PR #2423, opened 2024-01-02
Refactor conflict detection code and pass fileStatus to Conflict Checker

prakharjain09:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Refactor conflict detection code and pass fileStatus to Conflict Checker

## How was this patch tested?

Existing UTs

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
@@ -39,7 +39,7 @@ import org.apache.spark.sql.delta.metering.DeltaLogging
 import org.apache.spark.sql.delta.schema.{SchemaMergingUtils, SchemaUtils}
 import org.apache.spark.sql.delta.sources.DeltaSQLConf
 import org.apache.spark.sql.delta.stats._
-import org.apache.hadoop.fs.Path
+import org.apache.hadoop.fs.{FileStatus, Path}
 
 import org.apache.spark.SparkException
 import org.apache.spark.sql.{AnalysisException, Column, DataFrame, SparkSession}
@@ -1772,8 +1772,16 @@ trait OptimisticTransactionImpl extends TransactionalWrite
         tags = Map(TAG_LOG_STORE_CLASS -> deltaLog.store.getClass.getName)) {
 
     DeltaTableV2.withEnrichedUnsupportedTableException(catalogTable) {
+      val fileStatuses = getConflictingVersions(checkVersion)
+      val nextAttemptVersion = checkVersion + fileStatuses.size
 
-      val nextAttemptVersion = getNextAttemptVersion(checkVersion)
+      // validate that information about conflicting winning commit files is continuous and in the
+      // right order.
+      val expected = (checkVersion until nextAttemptVersion)
+      val found = fileStatuses.map(deltaVersion)
+      val mismatch = expected.zip(found).dropWhile{ case (v1, v2) => v1 == v2 }.take(10)
+      assert(mismatch.isEmpty,
+        s"Expected ${mismatch.map(_._1).mkString(",")} but got ${mismatch.map(_._2).mkString(",")}")
 
       val logPrefixStr = s"[attempt $attemptNumber]"
       val txnDetailsLogStr = {
@@ -1792,10 +1800,12 @@ trait OptimisticTransactionImpl extends TransactionalWrite
         s"[$checkVersion, $nextAttemptVersion) with current txn having $txnDetailsLogStr")
 
       var updatedCurrentTransactionInfo = currentTransactionInfo
-      (checkVersion until nextAttemptVersion).foreach { otherCommitVersion =>
+      (checkVersion until nextAttemptVersion)
+        .zip(fileStatuses)
+        .foreach { case (otherCommitVersion, otherCommitFileStatus) =>
         updatedCurrentTransactionInfo = checkForConflictsAgainstVersion(
           updatedCurrentTransactionInfo,
-          otherCommitVersion,
+          otherCommitFileStatus,
           commitIsolationLevel)
         logInfo(s"$logPrefixStr No conflicts in version $otherCommitVersion, " +
           s"${clock.getTimeMillis() - commitAttemptStartTime} ms since start")
@@ -1810,13 +1820,13 @@ trait OptimisticTransactionImpl extends TransactionalWrite
 
   protected def checkForConflictsAgainstVersion(
       currentTransactionInfo: CurrentTransactionInfo,
-      otherCommitVersion: Long,
+      otherCommitFileStatus: FileStatus,
       commitIsolationLevel: IsolationLevel): CurrentTransactionInfo = {
 
     val conflictChecker = new ConflictChecker(
       spark,
       currentTransactionInfo,
-      otherCommitVersion,
+      otherCommitFileStatus,
       commitIsolationLevel)
     conflictChecker.checkConflicts()
   }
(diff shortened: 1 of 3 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
```diff
@@ -1811,24 +1820,27 @@ trait OptimisticTransactionImpl extends TransactionalWrite
...
 
-  /** Returns the next attempt version given the last attempted version */
-  protected def getNextAttemptVersion(previousAttemptVersion: Long): Long = {
+  /** Returns the conflicting commit information */
+  protected def getConflictingVersions(previousAttemptVersion: Long): Seq[FileStatus] = {
+    assert(previousAttemptVersion == preCommitLogSegment.version + 1)
     val latestSnapshot = deltaLog.update()
     preCommitLogSegment = latestSnapshot.logSegment
-    latestSnapshot.version + 1
+    deltaLog.listFrom(previousAttemptVersion).filter(isDeltaFile).toSeq
```
ryan-johnson-databricks, line 1843: Doesn't this add extra listings to the conflict resolution path? 

The `deltaLog.update()` call just barely did the ~same listing two lines earlier; is there some way to reuse that work instead of repeating it?

prakharjain09: Good catch. Tried to reuse the same listing.

spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
```diff
@@ -628,6 +638,41 @@ trait SnapshotManagement { self: DeltaLog =>
     timestamp => cutoffOpt.forall(timestamp < _)
   }
 
+  /**
+   * Get the newest logSegment, using the previous logSegment as a hint. This is faster than
+   * doing a full update, but it won't work on table overwrites that overwrite the delta log.
```
ryan-johnson-databricks, line 643:
```suggestion
   * doing a full update, but it won't work if the table's log directory was replaced.
```

spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
```diff
@@ -628,6 +638,41 @@ trait SnapshotManagement { self: DeltaLog =>
...
+      versionToLoad = None,
+      files = Some(allFiles),
+      validateLogSegmentWithoutCompactedDeltas = false,
+      lastCheckpointInfo = lastCheckpointInfo,
+      oldCheckpointProviderOpt = Some(oldLogSegment.checkpointProvider)
+    ).getOrElse(oldLogSegment)
+    val fileStatusesOfConflictingCommits = newFiles.filter {
+      case DeltaFile(_, v) if v <= newLogSegment.version => true
+      case _ => false
+    }
```
ryan-johnson-databricks, line 672: nit
```suggestion
    val fileStatusesOfConflictingCommits = newFiles.collect {
      case DeltaFile(f, v) if v <= newLogSegment.version => f
    }
```

spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
```diff
@@ -286,7 +288,15 @@ trait SnapshotManagement { self: DeltaLog =>
...
       // i.e. some commit deltas are missing (although compacted-deltas are present).
-      validateDeltaVersions(deltasAfterCheckpoint, newCheckpointVersion, versionToLoad)
+      // We should not do this validation when we want to update postCommitSnapshot. In that
+      // specific flow, we just list from the committed version and reuse existing pre-commit
+      // logsegment together with listing result to create the new snapshot (Refer to
+      // [[SnapshotManagement.getUpdatedLogSegment]]). That codepath explicitly needs to disable
+      // validation as the validation is expected to fail in that as we don't have all delta file
+      // info when minor compactions are used in pre-commit snapshot.
+      if (validateLogSegmentWithoutCompactedDeltas) {
+        validateDeltaVersions(deltasAfterCheckpoint, newCheckpointVersion, versionToLoad)
```
ryan-johnson-databricks, line 298: I don't understand why the check needs to be(come) conditional? [validateDeltaVersions](https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala#L191-L217) doesn't seem to care whether every version has a commit .json file available? It seems to just verify that the list of commits and/or minor compactions covers all needed versions (in order) after the checkpoint. The existing log segment should trivially pass that check (because we checked when constructing it), and it seems like we still want to validate the new files in relation to old ones?

Could you elaborate on how/why "validation is expected to fail" in that case?

spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
ryan-johnson-databricks: Or
```suggestion
    val fileStatusesOfConflictingCommits = newFiles.filter {
      case DeltaFile(_, v) => v <= newLogSegment.version
    }
```

spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
```diff
@@ -628,6 +638,41 @@ trait SnapshotManagement { self: DeltaLog =>
...
+      // An empty listing likely implies a list-after-write inconsistency or that somebody clobbered
+      // the Delta log.
+      return (oldLogSegment, Seq.empty)
+    }
+    val allFiles = (
+      oldLogSegment.checkpointProvider.topLevelFiles ++
+        oldLogSegment.deltas ++
+        newFiles
+      ).toArray
+    val lastCheckpointInfo: Option[LastCheckpointInfo] = None
```
ryan-johnson-databricks, line 661: aside: I never know whether it's more scala-tastic to do
```suggestion
    val lastCheckpointInfo = Option.empty[LastCheckpointInfo]
```
?

(similar quandary to `Seq.empty` vs. `Nil` on L654 above)

spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
```diff
@@ -1772,8 +1772,16 @@ trait OptimisticTransactionImpl extends TransactionalWrite
...
+      val nextAttemptVersion = checkVersion + fileStatuses.size
 
-      val nextAttemptVersion = getNextAttemptVersion(checkVersion)
+      // validate that information about conflicting winning commit files is continuous and in the
+      // right order.
+      val expected = (checkVersion until nextAttemptVersion)
+      val found = fileStatuses.map(deltaVersion)
+      assert(expected == found,
+        s"Missing commit(s): ${(expected.toSet -- found.toSet).toSeq.sorted}, " +
+          s"Extra commit(s): ${(found.toSet -- expected.toSet).toSeq.sorted}")
```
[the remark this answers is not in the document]
ryan-johnson-databricks: Also, should this be an assertion or some other kind of exception?

spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
ryan-johnson-databricks: Update: `deltasAfterCheckpoint` _only_ contains deltas (see L255-256 above, which separates checkpoints, deltas, and minor compactions). Normal listing returns both deltas and minor compactions for the same commits, but log segments only include one or the other. Thus, the check would wrongly report gaps in the version sequence if the existing log segment included any minor compactions.

Downside is, by disabling the check we risk missing any problems in the newly listed deltas after the log segment. But I guess it's only an extra sanity check, so maybe not worth tearing up the code to keep track of the separation point between existing vs. new files?

spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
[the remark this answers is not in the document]
prakharjain09: Most of such invariants today are `assertions` in our code.

spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
prakharjain09: The 2nd implementation fails with a MatchError since `newFiles` could have other kind of files also?

spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
prakharjain09: Yes - keeping the check for getUpdatedLogSegment will need us to maintain the separation point - which doesn't seem to be worth the complexity.

spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
ryan-johnson-databricks: > The 2nd implementation fails

good point!

ryan-johnson-databricks (approved):
LGTM

PR #2429, opened 2024-01-03
[Spark] Throw exception when CREATE OR REPLACE is run on existing catalog tables with missing delta_log

dhruvarya-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Currently, CREATE OR REPLACE does throw any exception when it is run on an existing table where the backing delta_log directory has been manually deleted. Even though we don't throw any exceptions during CREATE OR REPLACE, other queries don't work on these broken tables.

This PR handles this scenario explicitly and throws an exception with the message:

```
[DELTA_METADATA_ABSENT_EXISTING_CATALOG_TABLE] The table `spark_catalog`.`default`.`delta_tbl` already exists in the catalog but no metadata could be found
for the table at the path file:/tmp/spark-5103f24e-a80a-4cc0-b1a7-86926fe8ff4f/subdir/_delta_log.
Did you manually delete files from the _delta_log directory? If so, dropping the
table by running `DROP TABLE `spark_catalog`.`default`.`delta_tbl`` and then recreating it by running the current command again
should enable you to recreate it.
```

## How was this patch tested?

New unit test.

Introduces a new error message when a user runs CREATE OR REPLACE on a catalog table for which the backing delta_log is missing:

```
[DELTA_METADATA_ABSENT_EXISTING_CATALOG_TABLE] The table `spark_catalog`.`default`.`delta_tbl` already exists in the catalog but no metadata could be found
for the table at the path file:/tmp/spark-5103f24e-a80a-4cc0-b1a7-86926fe8ff4f/subdir/_delta_log.
Did you manually delete files from the _delta_log directory? If so, dropping the
table by running `DROP TABLE `spark_catalog`.`default`.`delta_tbl`` and then recreating it by running the current command again
should enable you to recreate it.
```

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaErrors.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaErrors.scala
@@ -1600,6 +1600,12 @@ trait DeltaErrorsBase
       messageParameters = Array(DeltaSQLConf.DELTA_COMMIT_VALIDATION_ENABLED.key))
   }
 
+  def metadataAbsentForExistingCatalogTable(tableName: String, tablePath: String): Throwable = {
+    new DeltaIllegalStateException(
+      errorClass = "DELTA_METADATA_ABSENT_EXISTING_CATALOG_TABLE",
+      messageParameters = Array(tableName, tablePath, tableName))
+  }
+
   def updateSchemaMismatchExpression(from: StructType, to: StructType): Throwable = {
     new DeltaAnalysisException(
       errorClass = "DELTA_UPDATE_SCHEMA_MISMATCH_EXPRESSION",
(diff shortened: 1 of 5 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/DeltaErrors.scala
```diff
@@ -1599,6 +1599,12 @@ trait DeltaErrorsBase
       messageParameters = Array(DeltaSQLConf.DELTA_COMMIT_VALIDATION_ENABLED.key))
   }
 
+  def metadataAbsentForExistingCatalogTable(tableName: String, tablePath: String): Throwable = {
```
prakharjain09, line 1602: Add a test in `DeltaErrorsSuite` for this error.

dhruvarya-db: Done

prakharjain09 (approved):

PR #2433, opened 2024-01-04
[Kernel][Data skipping] Initial end-to-end support for data skipping for a small set expressions (AND, =, <, <=, >, >=)

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Part of #2229.

Adds initial end-to-end support for data skipping for a subset of expressions. Generates a data skipping filter based on the rules in `DataSkippingReader` in delta-spark.

## How was this patch tested?

Adds tests to `ScanSuite`. This includes copying of relevant tests from `DataSkiippingDeltaTests` as well as additional tests added just for kernel.

Note: currently tests tables are written using goldenTables. The utilities to write the test tables in ScanSuite have already been added and the table writing will be moved out of goldenTables (TODO).


kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -190,6 +206,56 @@ public void close() throws IOException {
         };
     }
 
+    private Optional<Predicate> getDataSkippingPredicate() {
+        Optional<Predicate> dataPredicate = getDataFilters();
```
scottsand-db, line 210: Something like this would be cleaner

```java
return getDataFilters().map(dataPredicate ->
    DataSkippingUtils.constructDataFilters(dataPredicate, metadata.getSchema())
)
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -190,6 +206,56 @@ public void close() throws IOException {
...
+        if (!dataPredicate.isPresent()) { // no data filter
+            return Optional.empty();
+        }
+        return DataSkippingUtils.constructDataFilters(dataPredicate.get(), metadata.getSchema());
+    }
+
+    private CloseableIterator<FilteredColumnarBatch> applyDataSkipping(
+        TableClient tableClient,
+        CloseableIterator<FilteredColumnarBatch> scanFileIter,
+        Predicate dataFilter) {
```
scottsand-db, line 220: nit: indent

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -190,6 +206,56 @@ public void close() throws IOException {
...
+        //    value of true means the stats say we must keep the file, and null means we could not
+        //    determine whether the file is safe to skip, because its stats were missing/null.
+        // 2. The coalesce(skip, true) converts null (= keep) to true
+        Predicate filterToEval = new Predicate(
+            "=",
+            new ScalarExpression(
+                "COALESCE",
+                Arrays.asList(dataFilter, Literal.ofBoolean(true))),
+            AlwaysTrue.ALWAYS_TRUE);
+        logger.info(String.format("totalFilter=%s", filterToEval));
```
scottsand-db, line 236: pretty sure you can just do

```java
logger.info("totalFilter={}", filterToEval);
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
```diff
@@ -0,0 +1,290 @@
...
+import io.delta.kernel.expressions.*;
+import io.delta.kernel.internal.actions.AddFile;
+import io.delta.kernel.types.*;
+import static io.delta.kernel.internal.InternalScanFileUtils.SCAN_FILE_SCHEMA;
+import static io.delta.kernel.internal.util.Preconditions.checkArgument;
+import static java.lang.String.format;
+
+public class DataSkippingUtils {
+
+    private static final int ADD_FILE_ORDINAL = SCAN_FILE_SCHEMA.indexOf("add");
```
scottsand-db, line 33: do we ever skip remove files?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
```diff
@@ -0,0 +1,290 @@
...
+    private static final int ADD_FILE_STATS_ORDINAL = AddFile.SCHEMA_WITH_STATS.indexOf("stats");
+
+    /**
+     * Given the table schema return the expected statistics schema.
+     */
+    public static StructType getStatsSchema(StructType dataSchema) {
+        return StatsSchemaHelper.getStatsSchema(dataSchema);
```
scottsand-db, line 40: this is a pass-through method. why wouldn't the caller just invoke `StatsSchemaHelper.getStatsSchema` directly?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    /**
+     * Given a {@code FilteredColumnarBatch} of scan files and the statistics schema to parse,
+     * return the parsed JSON stats from the scan files.
+     */
+    public static ColumnarBatch parseJsonStats(
+        TableClient tableClient, FilteredColumnarBatch scanFileBatch, StructType statsSchema) {
```
scottsand-db, line 48: nit: indentation, please apply elsewhere too

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
```diff
@@ -0,0 +1,290 @@
...
+    private static Predicate asPredicate(Expression expression) {
+        checkArgument(expression instanceof Predicate,
+            String.format("Expected predicate but got %s", expression));
+        return (Predicate) expression;
+    }
+
+    /**
+     * Utility method to return the left child of the binary input expression
+     */
+    static Expression getLeft(Expression expression) {
```
scottsand-db, line 271: Remind me why we don't just have a BinaryExpression class that supports a `.getLeft()` call?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/StatsSchemaHelper.java
```diff
@@ -0,0 +1,258 @@
...
+            add("string");
+        }
+    };
+
+    /**
+     * Returns true if the given data type is eligible for MIN/MAX data skipping.
+     */
+    private static boolean isSkippingEligibleDataType(DataType dataType) {
+        return skippingEligibleTypeNames.contains(dataType.toString()) ||
+            dataType instanceof DecimalType;
```
scottsand-db, line 52: why is decimal type a special case?

kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/internal/data/DefaultRowBasedColumnarBatch.java
```diff
@@ -83,6 +83,38 @@ public ColumnVector getColumnVector(int ordinal) {
         return columnVectors.get(ordinal).get();
     }
 
+    @Override
+    public ColumnarBatch withNewColumn(int ordinal, StructField columnSchema,
```
scottsand-db, line 87: nit: odd identation

scottsand-db (changes requested):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
tdas: +1

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -89,13 +92,26 @@ public CloseableIterator<FilteredColumnarBatch> getScanFiles(TableClient tableCl
             throw new IllegalStateException("Scan files are already fetched from this instance");
         }
         accessedScanFiles = true;
+
+        // Generate data skipping filter and decide if we should read the stats column
+        Optional<Predicate> dataFilter = getDataSkippingPredicate();
```
tdas, line 97: dataFilter -> dataSkippingFilter

also... why alternate between "filter" and "predicate"?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -190,6 +206,56 @@ public void close() throws IOException {
...
+        // 1. The predicate produces false for any file whose stats prove we can safely skip it. A
+        //    value of true means the stats say we must keep the file, and null means we could not
+        //    determine whether the file is safe to skip, because its stats were missing/null.
+        // 2. The coalesce(skip, true) converts null (= keep) to true
+        Predicate filterToEval = new Predicate(
+            "=",
+            new ScalarExpression(
+                "COALESCE",
+                Arrays.asList(dataFilter, Literal.ofBoolean(true))),
+            AlwaysTrue.ALWAYS_TRUE);
```
tdas, line 235: why do you need an additional ` = true`? wont coalesce return boolean? or is that we dont have expressions return values in our logical expressions...thus making just coalesce ambiguous?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
```diff
@@ -0,0 +1,290 @@
...
+     * Given a {@code FilteredColumnarBatch} of scan files and the statistics schema to parse,
+     * return the parsed JSON stats from the scan files.
+     */
+    public static ColumnarBatch parseJsonStats(
+        TableClient tableClient, FilteredColumnarBatch scanFileBatch, StructType statsSchema) {
+        ColumnVector statsVector = scanFileBatch.getData()
+            .getColumnVector(ADD_FILE_ORDINAL)
+            .getChild(ADD_FILE_STATS_ORDINAL);
+        return tableClient.getJsonHandler()
+            .parseJson(statsVector, statsSchema, scanFileBatch.getSelectionVector());
```
tdas, line 53: Not for this PR, but I feel like there is an opportunity for future optimizations here. just like we expressions evaluation has two steps - create the evaluator (with 1 time setup stuff), and then execute the evaluator (efficiently again and again)... the json parser could probably use a similar setup / execute break down.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
```diff
@@ -0,0 +1,290 @@
...
+     * both NULL -- which indicates that all values in the file are NULL. Fortunately, most of the
+     * operators we support data skipping for are NULL intolerant, and thus trivially satisfy this
+     * requirement because they never return TRUE for NULL inputs. The only NULL tolerant operator
+     * we support -- IS [NOT] NULL -- is specifically NULL aware.
+     *
+     * NOTE: The skipping predicate does *NOT* need to worry about missing stats columns (which also
+     * manifest as NULL). That case is handled separately by `verifyStatsForFilter` (which disables
+     * skipping for any file that lacks the needed stats columns).
+     */
+    public static Optional<Predicate> constructDataFilters(
```
tdas, line 143: isnt it better call it constructDataSkippingFilters? data filters is confusing because these are not filters on the table data. in fact you take the dataFilter as a parameter, so the returned filter cannot be a dataFilter as well .. right?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
```diff
@@ -0,0 +1,290 @@
...
+        Predicate dataFilter, StructType dataSchema) {
+        StatsSchemaHelper schemaHelper = new StatsSchemaHelper(dataSchema);
+        return constructDataFilters(dataFilter, schemaHelper);
+    }
+
+    public static Optional<Predicate> constructDataFilters(
```
tdas, line 149: why do both versions of this function need to be public?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/StatsSchemaHelper.java
```diff
@@ -0,0 +1,258 @@
...
+/**
+ * Provides information and utilities for statistics columns given a table schema. Specifically,
+ * it is used to:
+ * <ol>
+ *     <li>Get the expected statistics schema given a table schema</li>
+ *     <li>Check if a {@link Literal} or {@link Column} is skipping eligible</li>
+ *     <li>Get the statistics column for a given stat type and logical column</li>
+ * </ol>
+ */
+public class StatsSchemaHelper {
```
tdas, line 22: is this all new logic, or copied from some existing logic in delta-spark?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
vkorukanti: We need a Predicate which the coalesce is not.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
vkorukanti:
```suggestion
        logger.info(String.format("Data skipping filter = {}", filterToEval));
```

Also I am not sure if it is ok to print the filter in logs given it could contain literals which could be PII. Do we print them in delta-spark?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
vkorukanti: Main reason was to avoid the class bloat and keep the list of expression classes to a very minimal. We have an utility method in default handler, may be we should move that to kernel internal.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/StatsSchemaHelper.java
vkorukanti: it has a precision and scale :P

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/StatsSchemaHelper.java
```diff
@@ -0,0 +1,258 @@
...
+    private static Set<String> skippingEligibleTypeNames = new HashSet<String>() {
+        {
+            add("byte");
+            add("short");
+            add("integer");
+            add("long");
+            add("float");
+            add("double");
+            add("date");
+            add("timestamp");
```
vkorukanti, line 42: we just need to make sure that there aren't any timezone issues. As long as the JSON reader returns in UTC and the literal is in UTC should be fine.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/utils/ExpressionUtils.scala
```diff
@@ -0,0 +1,66 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.delta.kernel.defaults.utils
+
+import io.delta.kernel.expressions.{Column, Expression, Literal, Predicate}
+
+/** Useful helper functions for creating expressions in tests */
```
vkorukanti, line 20: we already have one here with the same name: io.delta.kernel.defaults.internal.expressions.ExpressionUtils.java

vkorukanti: Also I think this is better move into the API module (internal only)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/StatsSchemaHelper.java
scottsand-db: makes sense. this isn't obvious to a reader. this should have a comment

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
scottsand-db: hm. so `DataSkippingUtils` provides the expression get left / get right logic? Not, ExpressionUtils? This isn't cohesive.

See, this doesn't seem very findable. It's tough to search in the code for all helper/static/utility methods that take an Expression as input. ... espcially when some are in DataSkipppingUtils and some are in ExpressionUtils

scottsand-db: this seems like a very `c`, `struct` way of programming, not OOP. we should do a more OOP way. thoughts?

tdas (commented):
still reviewing.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
tdas: Got it.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
```diff
@@ -0,0 +1,290 @@
...
+                return Optional.empty();
+        }
+    }
+
+    //////////////////////////////////////////////////////////////////////////////////
+    // Useful helper functions for dealing with expressions
+    //////////////////////////////////////////////////////////////////////////////////
+
+    /** Construct the skipping predicate for a given comparator */
+    private static Predicate constructComparatorDataFilters(
```
tdas, line 209: same comment as about "DataFilter"

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
```diff
@@ -0,0 +1,290 @@
...
+            String comparator, Column leftCol, Literal rightLit, StatsSchemaHelper schemaHelper) {
+
+        switch (comparator.toUpperCase(Locale.ROOT)) {
+
+            // Match any file whose min/max range contains the requested point.
+            case "=":
+                return new And(
+                    new Predicate("<=", schemaHelper.getMinColumn(leftCol), rightLit),
```
tdas, line 217: this is a bit hard to read, could you please add the simplified SQL expression as comment.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
```diff
@@ -0,0 +1,290 @@
...
+        TableClient tableClient, FilteredColumnarBatch scanFileBatch, StructType statsSchema) {
+        ColumnVector statsVector = scanFileBatch.getData()
+            .getColumnVector(ADD_FILE_ORDINAL)
+            .getChild(ADD_FILE_STATS_ORDINAL);
+        return tableClient.getJsonHandler()
+            .parseJson(statsVector, statsSchema, scanFileBatch.getSelectionVector());
+    }
+
+    /**
+     * Returns a file skipping predicate expression, derived from the user query, which uses column
```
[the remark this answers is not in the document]
tdas: And can you clarify right in the start of the comment the semantics of that the predicate returned. What does True mean? What does False or null mean?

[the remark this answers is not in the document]
tdas: And can you clarify right in the start of the comment the semantics of that the predicate returned. What does True mean? What does False or null mean?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
tdas: the problem is we can use class/trait structure for this. That would have been the most OOP way. In absense of that, yeah, ExpressionUtils seems better.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/utils/ExpressionUtils.scala
tdas: +1 consolidate please.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/utils/ExpressionUtils.scala
```diff
@@ -0,0 +1,66 @@
...
+    new Predicate("NOT", pred)
+  }
+
+  def col(name: String): Column = new Column(name)
+
+  def nestedCol(name: String): Column = {
+    new Column(name.split("\\."))
+  }
+
+  /* ---------- NOT-YET SUPPORTED EXPRESSIONS ----------- */
```
tdas, line 53: why are these here if these are not yet supported?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
tdas: i guess not. we are skipping after log replay .. so only add files will be left. isnt it @allisonport-db

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/StatsSchemaHelper.java
```diff
@@ -0,0 +1,258 @@
...
+
+    //////////////////////////////////////////////////////////////////////////////////
+    // Public static fields and methods
+    //////////////////////////////////////////////////////////////////////////////////
+
+    /**
+     * Returns true if the given literal is skipping-eligible. Delta tracks min/max stats for a
+     * limited set of data types and only literals of those types are skipping eligible.
+     */
+    public static boolean isSkippingEligibleLiteral(Literal literal) {
```
tdas, line 105: why this single line function being called from only one location?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/StatsSchemaHelper.java
```diff
@@ -0,0 +1,258 @@
...
+        return isSkippingEligibleDataType(literal.getDataType());
+    }
+
+    /**
+     * Returns the expected statistics schema given a table schema.
+     */
+    public static StructType getStatsSchema(StructType dataSchema) {
```
tdas, line 112: why are some functions static and taking data schema and some function class methods? as in why is this not a class method?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/StatsSchemaHelper.java
```diff
@@ -0,0 +1,258 @@
...
+     */
+    private static boolean isSkippingEligibleDataType(DataType dataType) {
+        return skippingEligibleTypeNames.contains(dataType.toString()) ||
+            dataType instanceof DecimalType;
+    }
+
+    /**
+     * Given a data schema returns the expected schema for a min or max statistics column. This
+     * means 1) replace logical names with physical names 2) set nullable=true 3) only keep stats
+     * eligible fields (i.e. don't include fields with isSkippingEligibleDataType=false)
```
tdas, line 58: for this and the following method can you put the structure of the stats in some way? Maybe a pretty json. So that we can validate whether the code matches that. 

Also then this will become a great reference material for future.

tdas: and do these need to be public methods? i dont see it being called from anywhere outside this class

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -190,6 +206,56 @@ public void close() throws IOException {
...
+        return DataSkippingUtils.constructDataFilters(dataPredicate.get(), metadata.getSchema());
+    }
+
+    private CloseableIterator<FilteredColumnarBatch> applyDataSkipping(
+        TableClient tableClient,
+        CloseableIterator<FilteredColumnarBatch> scanFileIter,
+        Predicate dataFilter) {
+        // Get the stats schema
+        // TODO prune stats schema according to the data filter
+        StructType statsSchema = DataSkippingUtils.getStatsSchema(metadata.getSchema());
```
tdas, line 223: so we are converting the entire data schema into the stats schema, irrespective of which stats actually exist right? when we pass this to json parser, is the json parser creating large number of null-values column vectors for all the columns whose stats does not exist?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
tdas, line 235: @vkorukanti this code `AlwaysTrue.ALWAYS_TRUE` really looks odd. Can we replace this with `Literal.ALWAYS_TRUE` or `Literal.TRUE`?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ScanSuite.scala
```diff
@@ -15,28 +15,660 @@
...
-class ScanSuite extends AnyFunSuite with TestUtils {
+class ScanSuite extends AnyFunSuite with TestUtils with ExpressionUtils {
   import io.delta.kernel.defaults.ScanSuite._
 
-  test("conditionally read the statistics column based on the query filter") {
-    // For now we read the stats column whenever there is a data predicate, this test will need to
-    // be updated once we actually generate data filters and use the column statistics
+  // TODO no more golden tables, write the tables here!
+
+  // docs
```
tdas, line 44: what does hits and misses mean?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ScanSuite.scala
```diff
@@ -15,28 +15,660 @@
...
+          .withFilter(defaultTableClient, predicate)
+          .build())
+      assert(scanFiles.nonEmpty, s"Expected hit but got miss for $predicate")
+    }
+    misses.foreach { predicate =>
+      val scanFiles = collectScanFileRows(
+        snapshot.getScanBuilder(defaultTableClient)
+          .withFilter(defaultTableClient, predicate)
+          .build())
+      assert(scanFiles.isEmpty, s"Expected miss but got hit for $predicate")
```
tdas, line 59: can you print out the hit file stats for easier debugging?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
allisonport-db: I'm just going to remove this for now

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
allisonport-db: For now since we are skipping after log replay it's only add files (log replay only returns add files)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
allisonport-db: I can move these to `ExpressionUtils` in kernel-api and (and try to combine with the ones in kernel-defaults)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
allisonport-db: they don't I just missed changing the second to private

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/StatsSchemaHelper.java
allisonport-db: It will be used more once we add additional expressions to the skipping

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
allisonport-db: I copied names from delta-spark so the functions would match up... I'll go over and unify all the usages of terms instead

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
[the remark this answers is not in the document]
allisonport-db: Mostly verbatim. I've re-done the docs for this and the method below.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/StatsSchemaHelper.java
allisonport-db: Mostly new

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/StatsSchemaHelper.java
allisonport-db: There are no issues with UTC. But there is an issue since the json writer in spark truncates timestamps to milliseconds (instead of microseconds.) Delta-spark still does skipping by adding a millisecond to any MAX statistic but we need to support time addition for that. I've added this to the project plan and block min/max skipping on timestamp columns for now.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
vkorukanti: `AlwaysTrue` is a predicate and `Literal.TRUE` is not. Not sure if we can mix predicates and literals together.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
vkorukanti: +1. We should be able to get the columns used in the predicate.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/StatsSchemaHelper.java
```diff
@@ -0,0 +1,275 @@
...
+     */
+    private Map<Column, Tuple2<Column, DataType>> getLogicalToPhysicalColumnAndDataType(
+            StructType dataSchema) {
+        Map<Column, Tuple2<Column, DataType>> result = new HashMap<>();
+        for (StructField field : dataSchema.fields()) {
+            if (field.getDataType() instanceof StructType) {
+                Map<Column, Tuple2<Column, DataType>> nestedCols =
+                    getLogicalToPhysicalColumnAndDataType((StructType) field.getDataType());
+                for (Column childLogicalCol : nestedCols.keySet()) {
+                    Column childPhysicalCol = nestedCols.get(childLogicalCol)._1;
```
vkorukanti, line 238: one lookup instead of two `nestedCols.get(childLogicalCol)`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/StatsSchemaHelper.java
```diff
@@ -0,0 +1,275 @@
...
+                    DataType childColDataType = nestedCols.get(childLogicalCol)._2;
+                    result.put(
+                        getChildColumn(childLogicalCol, field.getName()),
+                        new Tuple2<>(
+                            getChildColumn(childPhysicalCol, getPhysicalName(field)),
+                            childColDataType
+                        ));
+                }
+            } else {
+                result.put(
```
vkorukanti, line 248: qn: do we collect stats for map keys or values?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/StatsSchemaHelper.java
```diff
@@ -0,0 +1,275 @@
...
+     * all leaf columns in the schema.
+     */
+    private Map<Column, Tuple2<Column, DataType>> getLogicalToPhysicalColumnAndDataType(
+            StructType dataSchema) {
+        Map<Column, Tuple2<Column, DataType>> result = new HashMap<>();
+        for (StructField field : dataSchema.fields()) {
+            if (field.getDataType() instanceof StructType) {
+                Map<Column, Tuple2<Column, DataType>> nestedCols =
+                    getLogicalToPhysicalColumnAndDataType((StructType) field.getDataType());
+                for (Column childLogicalCol : nestedCols.keySet()) {
```
vkorukanti, line 237: nit (not required in this PR): Have the `result` and current parent `name[]` as args and as soon as you visit a leaf level node, add it to the `result`. Avoids `getChildColumn` and multiple iterations of subfields.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
allisonport-db: Moved to ExpressionUtils

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/utils/ExpressionUtils.scala
allisonport-db: I've created `ExpressionUtils` in the kernel-api module and had already renamed this to `ExpressionTestUtils`. Not sure if we want to convolute that with all these methods--especially since I've included some unsupported expressions?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/utils/ExpressionUtils.scala
allisonport-db: Unsupported expressions just won't generate a data skipping filter. In tests I've included such expressions as copied over from the `DataSkippingDeltaTests` but adjusted them to expect no skipping. When we add support for those expressions these tests will fail and we'll adjust them to expect file skipping.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ScanSuite.scala
allisonport-db: Clarified in docs

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/StatsSchemaHelper.java
allisonport-db: I made the methods that don't use the instance fields or are only called once static. The class methods are all used more than once by `constructDataSkippingFilters` (given a specific data schema and the constructed instance fields.)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/StatsSchemaHelper.java
allisonport-db: Not sure which public method you're referring to?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
allisonport-db: We have no way of concretely knowing ahead of time which stats exist. But the follow-up PR I'm working on prunes this schema to only include any stats columns used in the generated skipping predicate.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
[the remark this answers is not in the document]
allisonport-db: I added an example of a data schema and stats to the top level docs.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/StatsSchemaHelper.java
allisonport-db: I added an example of a data schema and stats to the top level docs.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/StatsSchemaHelper.java
allisonport-db: We collect null_count stats (but not min/max). So we can generate skipping filters for expressions like `isNull` but not for others.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ScanSuite.scala
```diff
@@ -15,29 +15,955 @@
...
+        expNumPartitions = 1,
+        expNumFiles = 1) // 1 files with key = 'a'
+
+
+      checkResults(
+        predicate = equals(col("key"), ofString("b")),
+        expNumPartitions = 1,
+        expNumFiles = 1) // 1 files with key = 'b'
+
+      // TODO shouldn't partition filters on unsupported expressions just not prune instead of fail?
```
allisonport-db, line 715: @vkorukanti what do you think? It seems like it should just be pushed back to the connector as remainingPredicate

vkorukanti: Returning as the remaining predicate makes sense. Let me know, I can make a PR for this. We may need to split the partition predicate (if combined with AND) to figure out which parts we can use using the expression handler.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
```diff
@@ -0,0 +1,270 @@
...
+import io.delta.kernel.data.FilteredColumnarBatch;
+import io.delta.kernel.expressions.*;
+import io.delta.kernel.types.*;
+import io.delta.kernel.internal.actions.AddFile;
+import static io.delta.kernel.internal.util.ExpressionUtils.*;
+import static io.delta.kernel.internal.InternalScanFileUtils.SCAN_FILE_SCHEMA;
+
+public class DataSkippingUtils {
+
+    private static final int ADD_FILE_ORDINAL = SCAN_FILE_SCHEMA.indexOf("add");
```
scottsand-db, line 33:
```
private static final int ADD_FILE_ORDINAL = SCAN_FILE_SCHEMA.indexOf("add");
```

is already defined in `kernel/kernel-api/src/main/java/io/delta/kernel/internal/InternalScanFileUtils.java`. Why have 2 definitions of the same thing? Shouldn't we just make the InternalScanFileUtils one `public` instead of `private`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
scottsand-db: sorry, I still don't like this

You can just import `StatsSchemaHelper.getStatsSchema`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
scottsand-db: if we already know the exact issue # that will be fixing/improving this, can we include that next to the `TODO`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
```diff
@@ -0,0 +1,270 @@
...
+            case ">=":
+                return new Predicate(">=", schemaHelper.getMaxColumn(leftCol), rightLit);
+
+            default:
+                throw new IllegalArgumentException(
+                    String.format("Unsupported comparator expression %s", comparator));
+        }
+    }
+
+    private static Map<String, String> reverseComparators = new HashMap<String, String>(){
```
scottsand-db, line 253: `final`, too, since we don't want this to be re-assigned

scottsand-db: and then, of course, `REVERSE_COMPARATORS` to stick to java8 convention for final variables

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
```diff
@@ -0,0 +1,270 @@
...
+import io.delta.kernel.expressions.*;
+import io.delta.kernel.types.*;
+import io.delta.kernel.internal.actions.AddFile;
+import static io.delta.kernel.internal.util.ExpressionUtils.*;
+import static io.delta.kernel.internal.InternalScanFileUtils.SCAN_FILE_SCHEMA;
+
+public class DataSkippingUtils {
+
+    private static final int ADD_FILE_ORDINAL = SCAN_FILE_SCHEMA.indexOf("add");
+    private static final int ADD_FILE_STATS_ORDINAL = AddFile.SCHEMA_WITH_STATS.indexOf("stats");
```
scottsand-db, line 34: and I would just add `ADD_FILE_STATS_ORDINAL` to `InternalScanFileUtils`, so that all the ordinal/access stuff is there

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/StatsSchemaHelper.java
```diff
@@ -0,0 +1,304 @@
...
+            logicalToPhysicalColumnAndDataType.entrySet().stream()
+                .collect(
+                    Collectors.toMap(
+                        Map.Entry::getKey,
+                        e -> e.getValue()._2 // map to just the data type
+                    )
+                );
+    }
+
+    /** Given a logical column returns corresponding the MIN column in the statistic schema */
```
scottsand-db, line 130: `Given a logical column returns corresponding the MIN column` I don't get this sentence btw.
`returns corresponding` is an odd phrasing

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/StatsSchemaHelper.java
```diff
@@ -0,0 +1,304 @@
...
+     */
+    public boolean isSkippingEligibleNullCountColumn(Column column) {
+        return logicalToPhysicalColumn.containsKey(column);
+    }
+
+    //////////////////////////////////////////////////////////////////////////////////
+    // Private static fields and methods
+    //////////////////////////////////////////////////////////////////////////////////
+
+    private static String NUM_RECORDS = "numRecords";
```
scottsand-db, line 169: i'd give (comment) some context as to what these variables are used for. All I know is that they are a string. what do they represent semantically? it's not like whenever I want to log "numRecords" to the console, I should use the NUM_RECORDS constant instard, right? so, clearly, they have some other intended use or meaning

scottsand-db, line 169: `final`, too. we don't want this to get accidentally re-assigned

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/StatsSchemaHelper.java
```diff
@@ -0,0 +1,304 @@
...
+    private static String MIN = "minValues";
+    private static String MAX = "maxValues";
+    private static String NULL_COUNT = "nullCount";
+
+    private static Set<String> skippingEligibleTypeNames = new HashSet<String>() {
```
scottsand-db, line 174: final, etc

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/StatsSchemaHelper.java
```diff
@@ -0,0 +1,304 @@
...
+    private static Column getChildColumn(Column column, String parentName) {
+        return new Column(prependArray(column.getNames(), parentName));
+    }
+
+    /**
+     * Given an array {@code names} and a string element {@code preElem} return a new array with
+     * {@code preElem} inserted at the beginning
+     */
+    private static String[] prependArray(String[] arr, String preElem) {
+        String[] newNames = new String[arr.length + 1];
```
scottsand-db, line 299: how often will this be used? is this super common? it seems like we have chosen a data structure (array) that isn't very good for this use case (prepending). a LinkedList would be better. What are the other use cases of this array inside of column? Or, why was an array [] chosen?

kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/internal/expressions/DefaultExpressionUtils.java
```diff
@@ -220,28 +220,6 @@ static void compareBinary(ColumnVector left, ColumnVector right, int[] result) {
...
-     */
-    static Expression getRight(Expression expression) {
-        List<Expression> children = expression.getChildren();
-        checkArgument(
-            children.size() == 2,
-            format("%s: expected two inputs, but got %s", expression, children.size()));
-        return children.get(1);
-    }
-
     static Expression childAt(Expression expression, int index) {
```
scottsand-db, line 223: why does this belong in `DefaultExpressionUtils` and not in `ExpressionUtils`?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/utils/ExpressionTestUtils.scala
```diff
@@ -0,0 +1,74 @@
...
+  def col(name: String): Column = new Column(name)
+
+  def nestedCol(name: String): Column = {
+    new Column(name.split("\\."))
+  }
+
+  /* ---------- NOT-YET SUPPORTED EXPRESSIONS ----------- */
+
+  def nullSafeEquals(e1: Expression, e2: Expression): Predicate = {
+    new Predicate("<=>", e1, e2)
```
scottsand-db, line 60: maybe comment this out, and throw an exception if they are invoked?

or, why write code that can't be used, instead of just comments?

scottsand-db (changes requested):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/StatsSchemaHelper.java
allisonport-db: Updated this let me know if it's more clear; otherwise I can try to be even more specific

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/StatsSchemaHelper.java
allisonport-db: Added a comment

kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/internal/expressions/DefaultExpressionUtils.java
allisonport-db: I can move it over. I left it since it's only used by the default module currently

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/utils/ExpressionTestUtils.scala
allisonport-db: They will throw an exception if they are tried to be evaluated.

See as to why they are in the tests https://github.com/delta-io/delta/pull/2433#discussion_r1443646582. For now, unsupported expressions should just be ignored in the data skipping filter generation (and will be returned to the connector as part of the remaining predicate to evaluate.)

allisonport-db: I can add a comment here explaining this reasoning

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/StatsSchemaHelper.java
allisonport-db: @vkorukanti what do you think? can we change this to be a List?
I think the `names` field is mostly manipulated (prepended or appended to) as part of the data skipping

vkorukanti: We can get rid of this if we follow the method I described in my other comment above about how to construction the leaf level columns. 

But I am also ok to make this a list. It is just that lists have extra baggage.

vkorukanti: Not sure if I understand, but what I am thinking is maintain the current root path and whenever we visit a child allocate n+1 array (system.cop can do it) and assign the child name at the `n` index. See the comment [here](https://github.com/delta-io/delta/pull/2433/files#r1443554254). In any case, I am ok to change it to List and if that becomes a problem we can convert it back to array.

vkorukanti (approved):
lgtm

PR #2454, opened 2024-01-09
[Spark] Reorder checks in ConflictChecker

cstavr:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Reorder checks in Delta Conflict Checker:
- Move the check `checkIfDomainMetadataConflict` and `checkForUpdatedApplicationTransactionIdsThatCurrentTxnDependsOn` earlier to fail the transaction as early as possible.
- Move Row Tracking reconciliation before file-level checks to ensure that files do no have duplicate row IDs.

## How was this patch tested?

Existing tests.

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
@@ -135,15 +135,22 @@ private[delta] class ConflictChecker(
    * the transaction as if it had started while reading the `winningCommitVersion`.
    */
   def checkConflicts(): CurrentTransactionInfo = {
+    // Protocol and metadata checks.
     checkProtocolCompatibility()
     checkNoMetadataUpdates()
+    checkIfDomainMetadataConflict()
+    checkForUpdatedApplicationTransactionIdsThatCurrentTxnDependsOn()
+
+    // Row Tracking reconciliation. We perform this before the file checks to ensure that
+    // no files have duplicate row IDs.
+    reassignOverlappingRowIds()
+    reassignRowCommitVersions()
+
+    // Data file checks.
     checkForAddedFilesThatShouldHaveBeenReadByCurrentTxn()
     checkForDeletedFilesAgainstCurrentTxnReadFiles()
     checkForDeletedFilesAgainstCurrentTxnDeletedFiles()
-    checkForUpdatedApplicationTransactionIdsThatCurrentTxnDependsOn()
-    reassignOverlappingRowIds()
-    reassignRowCommitVersions()
-    checkIfDomainMetadataConflict()
+
     logMetrics()
     currentTransactionInfo
   }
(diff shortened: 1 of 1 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
```diff
@@ -135,15 +135,22 @@ private[delta] class ConflictChecker(
...
    */
   def checkConflicts(): CurrentTransactionInfo = {
+    // Protocol and metadata checks.
     checkProtocolCompatibility()
     checkNoMetadataUpdates()
-    checkForAddedFilesThatShouldHaveBeenReadByCurrentTxn()
-    checkForDeletedFilesAgainstCurrentTxnReadFiles()
-    checkForDeletedFilesAgainstCurrentTxnDeletedFiles()
+    checkIfDomainMetadataConflict()
     checkForUpdatedApplicationTransactionIdsThatCurrentTxnDependsOn()
```
c27kwan, line 142: nit: Add a comment for this one, since it's neither a protocol nor metadata check

spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
```diff
@@ -135,15 +135,22 @@ private[delta] class ConflictChecker(
    * the transaction as if it had started while reading the `winningCommitVersion`.
    */
   def checkConflicts(): CurrentTransactionInfo = {
+    // Protocol and metadata checks.
```
c27kwan, line 138: nit: I think it'd be nice if we left the rationale for the ordering behind so the next developer isn't afraid to reorder as necessary.  something like "we should enforce specifications in the protocol first"

spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
```diff
@@ -135,15 +135,22 @@ private[delta] class ConflictChecker(
...
     checkProtocolCompatibility()
     checkNoMetadataUpdates()
-    checkForAddedFilesThatShouldHaveBeenReadByCurrentTxn()
-    checkForDeletedFilesAgainstCurrentTxnReadFiles()
-    checkForDeletedFilesAgainstCurrentTxnDeletedFiles()
+    checkIfDomainMetadataConflict()
     checkForUpdatedApplicationTransactionIdsThatCurrentTxnDependsOn()
+
+    // Row Tracking reconciliation. We perform this before the file checks to ensure that
+    // no files have duplicate row IDs.
```
c27kwan, line 145: (and enforce the delta protocol/ align the files actions with the delta protocol/ avoid interacting with files that don't comply with the protocol)

c27kwan (approved):
lgtm!

spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
cstavr: done.

spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
cstavr: done.

spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
cstavr: done.

Issue #2297, opened 2023-11-15
[Feature Request] Enhance Iceberg compatibility in Delta Uniform (maps, list, etc)


lzlfred:
#### Which Delta project/connector is this regarding?

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

### Overview

The Delta UniForm announcement in Delta 3.0 attracted a lot of interest and feedback. The community found additional incompatibilities between Delta and Iceberg that we would like to resolve in the next Delta release.

### Further details
The IcebergCompatV1 table feature was released to ensure that a Delta table is compatible with Iceberg when generating Iceberg metadata. The current IcebergCompatV1 table feature has one major restriction: it cannot be enabled when a Delta table has List and Map. This is a major friction for users.
That aside, because different Iceberg readers have different support and behaviors, the following issues may exist and affect query results:

* Delta Uniform tables does not write partition values in parquet files; While OSS Spark and Iceberg is able to read those, some other Iceberg readers (eg, BigQuery, Athena) may rely on materialized partition values in parquet data files and show NULL as partition values in query results. 
* Spark writes timestamp as int96 to parquet as default, while Iceberg spec defines timestamp as int64. OSS Spark and Iceberg can read int96 as timestamp, while some other Iceberg readers (eg, Athena) will throw error on reading parquet containing int96. 

Thus this issue requests for enhanced Iceberg compatibility in Delta Uniform to address above pain points. 

### Willingness to contribute

The Delta Lake Community encourages new feature contributions. Would you or another member of your organization be willing to contribute an implementation of this feature?

- [ ] Yes. I can contribute this feature independently.
- [x] Yes. I would be willing to contribute this feature with guidance from the Delta Lake community.
- [ ] No. I cannot contribute this feature at this time.

PR #2459, opened 2024-01-09
[Spark] Support List and Map columns in Uniform

LukasRupprecht:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR adds support for List and Map columns in Uniform. To support these types, Delta column mapping needs to write additional field IDs to the parquet schema. List columns require one additional field ID for the 'element' subfield and Map columns require two additional field IDs for the 'key' and 'value' subfields inside the parquet file. These nested field IDs are added to the table schema during the generation of the IDs and physical names for column mapping. They are added to the parquet schema through a new class, DeltaParquetWriteSupport, that hooks into Spark's parquet write path and rewrites the parquet schema based on the additional field IDs.

This PR is part of https://github.com/delta-io/delta/issues/2297.

## How was this patch tested?

Unit tests will be added soon in a separate PR.

No


lzlfred (approved):
very excited to see Map and List support for Uniform Iceberg !

PR #2473, opened 2024-01-10
[Flink] Allow Flink connector to get snapshots via Kernel rather than standalone

nicklan:

- [ ] Spark
- [ ] Standalone
- [X] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Allow the Flink connector to get a snapshot via the kernel rather than standalone. We do so by having a delegator class that can use kernel whenever possible, but falls back to standalone when absolutely needed.

The option `io.delta.flink.kernel.enabled` (defined as `io.delta.flink.internal.DeltaFlinkHadoopConf.DELTA_KERNEL_ENABLED`) can be set to enable this feature.

## How was this patch tested?

Unit tests

No


connectors/flink/src/main/scala/io/delta/flink/internal/KernelSnapshotDelegator.scala
```diff
@@ -0,0 +1,99 @@
...
+class KernelSnapshotDelegator(
+    kernelSnapshot: SnapshotImplKernel,
+    hadoopConf: Configuration,
+    path: Path,
+    override val version: Long,
+    kernelDeltaLog: KernelDeltaLogDelegator,
+    standaloneDeltaLog: DeltaLogImpl)
+  extends SnapshotImpl(hadoopConf, path, -1, LogSegment.empty(path), -1, standaloneDeltaLog, -1) {
+
+  val snapshotWrapper = new KernelSnapshotWrapper(kernelSnapshot)
```
scottsand-db, line 45: `kernelSnapshotWrapper`

scottsand-db: id also add a short comment explaining/reminding us what a KernelSnapshotWrapper is, i.e. what is different about it from a SnapshotImplKernel

connectors/flink/src/main/scala/io/delta/flink/internal/KernelSnapshotDelegator.scala
```diff
@@ -0,0 +1,99 @@
...
+
+  override lazy val metadataScala: Metadata = {
+    val metadata = snapshotWrapper.getMetadata()
+    ConversionUtils.convertMetadataJ(metadata)
+  }
+
+  // provide a path to use the faster txn lookup in kernel
+  def getLatestTransactionVersion(id: String): Option[Long] = {
+    val versionJOpt = kernelSnapshot.getLatestTransactionVersion(id)
+    if (versionJOpt.isPresent) {
```
scottsand-db, line 64: you can't just return `kernelSnapshot.getLatestTransactionVersion(id).asScala`? look for other usages of this in the code base? should be easy to convert java optional to scala option

scottsand-db (changes requested):

connectors/flink/src/main/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitter.java
```diff
@@ -126,8 +126,12 @@ public DeltaGlobalCommitter(
         this.conf = conf;
         this.rowType = rowType;
         this.mergeSchema = mergeSchema;
-        this.deltaLog = DeltaLog.forTable(conf,
-            new org.apache.hadoop.fs.Path(basePath.toUri()));
+        if (conf.getBoolean("io.delta.flink.usekernel", false)) {
```
scottsand-db, line 129: `io.delta.flink.useKernel` probably? that's most consistent with our other flag names. See `StandaloneHadoopConf`

scottsand-db: should probably also make a `DeltaFlinkHadoopConf` *internal* class to document this flag, it's semantics, what it means, etc etc

connectors/flink/src/main/java/io/delta/flink/source/internal/enumerator/supplier/KernelSnapshotWrapper.java
```diff
@@ -0,0 +1,162 @@
...
+import io.delta.standalone.data.CloseableIterator;
+import io.delta.standalone.data.RowRecord;
+import io.delta.standalone.expressions.Expression;
+
+
+/**
+ * Wrap a {@link io.delta.kernel.Snapshot} such that it conforms to the the {@link
+ * io.delta.standalone.Snapshot} interface.
+ *
+ * NB: Currently only supports the getMetadata and getVersion methods. All others throw exceptions.
```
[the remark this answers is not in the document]
nicklan: https://en.wikipedia.org/wiki/Nota_bene

connectors/flink/src/main/scala/io/delta/flink/internal/KernelSnapshotDelegator.scala
nicklan: I don't think so. https://www.scala-lang.org/api/2.13.6/scala/jdk/OptionConverters$.html doesn't exist until scala 2.13

connectors/flink/src/main/java/io/delta/flink/source/internal/enumerator/supplier/KernelSnapshotWrapper.java
```diff
@@ -0,0 +1,162 @@
...
+
+        // Convert over the schema StructType
+        List<io.delta.kernel.types.StructField> kernelFields = kernelMetadata.getSchema().fields();
+        io.delta.standalone.types.StructField[] fields =
+            new io.delta.standalone.types.StructField[kernelFields.size()];
+        int index = 0;
+        for (io.delta.kernel.types.StructField kernelField: kernelFields) {
+            io.delta.standalone.types.FieldMetadata.Builder metadataBuilder =
+                io.delta.standalone.types.FieldMetadata.builder();
+            // TODO: Don't use Object and toString here
```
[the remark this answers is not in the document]
nicklan: fixed via tricky reflection

scottsand-db (approved):

PR #2474, opened 2024-01-10
[Kernel][Data skipping] Prune the statistics schema based on the columns used in the generated data skipping filter

### LINKED ISSUE #2458: [Kernel][Feature Request] Prune the stats schema based on the columns used in the generated data skipping filter

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

### Overview

(Part of #2229)

#2433 adds initial support for end-to-end data skipping. As a follow-up we should prune the stats read schema to only include the fields that are used by the generated filter.

### Motivation

Lower the parsing/reading overhead. If we only use a small subset of the stats columns (very likely) we shouldn't read all of them.

### Willingness to contribute

- [X] Yes. I can contribute this feature independently.

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

Closes  #2458

Prune the statistic schema based on the columns used in the generated data skipping filter. Collect the statistic columns used when constructing the data skipping filter and use it to prune the statistics schema before JSON parsing.

## How was this patch tested?

Adds tests


kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
```diff
@@ -260,4 +311,54 @@ private static Predicate reverseComparatorFilter(Predicate predicate) {
...
+    /**
+     * Prunes the given schema according to the provided set of top-level columns to keep. The
+     * columns in {@code schema} are taken to be nested under {@code parentPath}.
+     *
+     * @param schema schema to prune
+     * @param columnsToKeep set of top-level columns to keep
+     * @param parentPath parent path of the fields in {@code schema} relative to the top-level
+     *                   schema
+     */
+    private static StructType pruneSchema(
```
allisonport-db, line 324: I reconsidered changing `Column` to take a List instead of Array here. But I think this doesn't make much of a difference considering we will need to create _new_ lists every time we create a column anyways (or each time we recurse into a nested field).

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingPredicate.java
```diff
@@ -0,0 +1,44 @@
...
+ * by the expression.
+ */
+public class DataSkippingPredicate {
+
+    private final Predicate expression;
+    private final Set<Column> referencedCols;
+
+    DataSkippingPredicate(Predicate expression, Set<Column> referencedCols) {
+        this.expression = expression;
+        this.referencedCols = referencedCols;
```
scottsand-db, line 34: `Collections.unmodifiableSet`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingPredicate.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public Predicate getExpression() {
```
scottsand-db, line 37: why not `getPredicate`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingPredicate.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Set;
+
+import io.delta.kernel.expressions.Column;
+import io.delta.kernel.expressions.Predicate;
+
+/**
+ * Used to hold a pair of {@link Predicate} expression and {@link Set<Column>} of columns referenced
+ * by the expression.
+ */
+public class DataSkippingPredicate {
```
scottsand-db, line 27: why doesn't `Predicate` have a list of its own referenced columns?

something seems off that `DataSkippingPredicate` has-a `Predicate` but not is-a `Predicate`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingPredicate.java
```diff
@@ -0,0 +1,44 @@
...
+
+    private final Predicate expression;
+    private final Set<Column> referencedCols;
```
scottsand-db, line 30: i'd document what a reference column is, and what its relationship (coupling) to expression is

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
```diff
@@ -43,6 +43,13 @@ public static ColumnarBatch parseJsonStats(
             .parseJson(statsVector, statsSchema, scanFileBatch.getSelectionVector());
     }
 
+    /**
+     * Prunes the given schema to only include the referenced columns.
+     */
+    public static StructType pruneStatsSchema(StructType schema, Set<Column> referencedCols) {
```
scottsand-db, line 49: prune schema says `Prunes the given schema according to the provided set of top-level columns to keep`.

so do these `referencedCols` need to be top-level?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
scottsand-db, line 324: this method `pruneSchema` seems to accomplish the goal of the PR: `Prune the statistics schema based on the columns used in the generated data skipping filter`

What are all of the `DataSkippingPredicate` changes needed for?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
```diff
@@ -260,4 +311,54 @@ private static Predicate reverseComparatorFilter(Predicate predicate) {
...
+     *
+     * @param schema schema to prune
+     * @param columnsToKeep set of top-level columns to keep
+     * @param parentPath parent path of the fields in {@code schema} relative to the top-level
+     *                   schema
+     */
+    private static StructType pruneSchema(
+            StructType schema, Set<Column> columnsToKeep, String[] parentPath) {
+        List<StructField> prunedFields = new ArrayList<>();
+        for (StructField field: schema.fields()) {
```
scottsand-db, line 327: nit: `StructField field : schema.fields()` (spacing)

scottsand-db (changes requested):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -291,4 +299,11 @@ private Set<String> loadPartitionColNames() {
         }
         return Collections.unmodifiableSet(partitionColumnNames);
     }
+
+    private StructType loadDataSchema() {
+        return new StructType(metadata.getSchema().fields().stream()
+            .filter(field ->
+                !partitionColumnNames.contains(field.getName().toLowerCase(Locale.ENGLISH)))
```
scottsand-db, line 306: why english? isn't it usally `Locale.ROOT`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -291,4 +299,11 @@ private Set<String> loadPartitionColNames() {
         }
         return Collections.unmodifiableSet(partitionColumnNames);
     }
+
+    private StructType loadDataSchema() {
```
scottsand-db, line 303: not really `loading` anything. `getDataSchema()` is better. but anyways, this probably belongs in the `Metadata` class, no? it already has `getSchema` and `getPartitionCols` ... we could easily add `getDataSchema` etc there too?

so it should just be `metadata.getDataSchema()`

scottsand-db: and this is what delta-spark does, too

scottsand-db (changes requested):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingPredicate.java
allisonport-db: Switched it; was just copying `DataSkippingReader` in spark

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingPredicate.java
allisonport-db: Added

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
allisonport-db: Clarified

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
allisonport-db: Was just copying the above method switched them both

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
scottsand-db: Unfortunately, now I'm even more confused.

- `pruneSchema` says: `Prunes the given schema according to the provided set of top-level columns to keep.`
- `pruneStatsSchema` says that the `referencesCols` are the `leaf columns`

so the first asks for top-level columns, and the latter gives it leaf columns?

scottsand-db (changes requested):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingPredicate.java
allisonport-db: `Predicate` is part of our public expression framework and doesn't store referenced columns. We could rename `DataSkippingPredicate` to something else but I don't really see how it's semantically "not a `Predicate`". It's a predicate with its list of referenced columns... seems like it could be viewed as either having a predicate or being a predicate

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
allisonport-db: The changes gather the list of referenced columns while constructing the predicate. This is how delta-spark does it; although I guess if we wanted to it could be separate step

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
allisonport-db: `pruneStatsSchema` takes in `schema` and a set of leaf columns (for nested columns these should be the full column path)

`pruneSchema` recurses --> so the input `schema` here is nested under `parentPath`, but the `columnsToKeep` are not
- By "top-level leaf columns" I mean that the column paths are relative to the root schema and not to the current level of recursion

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
allisonport-db: This relies on `loadPartitionColNames` which loads the partition columns from `ArrayValue`; I can move both over and make them lazy

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingPredicate.java
scottsand-db: > but I don't really see how it's semantically "not a Predicate".

It doesn't extend Predicate, that's why. So it doesn't have a "is-a" relationship with `Predicate`. Since it's composed of a Predicate, it has a "has-a" relationship

Why doesn't it just extend `Predicate`? Every `DataSkippingPredicate` is a `Predicate`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
scottsand-db: Okay, I understand now (I think). I think we need to clarify this in the comments, then.

1.
```
pruneStatsSchema: Prunes the given schema to include only the referenced leaf columns. If a leaf column is nested, then it must be referenced using the full column path, e.g. `C_0.C_1.C_Leaf`
```

2. I would also rename `referencedCols` to `referencedLeafCols`

3. 
```
pruneSchema: Prunes the given schema according to include only the referenced leaf columns. These leaf columns (they can be nested, of course) are relative to the root schema, not to the current level of recursion. The `leafColumnsToKeep` doesn't change with each recursive call.

e.g.: <give an example, showing 2 layers of recursion>
```

4. if columnsToKeep doesn't change, why not mark it as final and indicate this?

5. rename to `leafColumnsToKeep`

scottsand-db: 6. For `pruneSchema`, Seems like `parentPath` is related to `schema`, yet they are not next to each other in the param list? Let's put related parameters together. I would do: `final leafColumnsToKeep, schema, parentPath`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
scottsand-db: Yup! This is what delta-spark does. Much more cohesive.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
allisonport-db: Updated; I think marking `leafColumnsToKeep` as final doesn't really mean much toward it being unchanged between recursive calls (and I think you could technically modify the set anyways). Mentioned it in the docs.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
```diff
@@ -315,23 +316,45 @@ private static Predicate reverseComparatorFilter(Predicate predicate) {
...
+     *          |   |--level_4_col: int
+     *          |--level_3_col: int
+     *          </pre>
+     *     </li>
+     *     <li>parentPath=["level1_struct", "level2_struct"]</li>
+     * </ol>
      *
      * @param schema schema to prune
-     * @param columnsToKeep set of top-level leaf columns to keep
+     * @param leafColumnsToKeep set of top-level leaf columns to keep
```
scottsand-db, line 345: I think `top-level leaf columns` is still confusing. I would just call it `leaf columns, relative to table schema root` or something along those lines

Really, since these are _exactly_ the same as what is passed to `pruneStatsSchema`, the same description should work and apply for both

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
```diff
@@ -342,9 +342,8 @@ private static Predicate reverseComparatorFilter(Predicate predicate) {
      * </ol>
      *
      * @param schema schema to prune
```
scottsand-db, line 344: nit (sry) update param docs order to match actual params, now that they've been rearranged

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingPredicate.java
allisonport-db: Ok I did this in the latest commit LMK if that looks better to you. Alternatively we could rename `DataSkippingPredicate` to something that is more reflective of it being a pair of predicate + referenced columns.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
```diff
@@ -364,11 +408,14 @@ private static Optional<Predicate> constructNotDataSkippingFilters(
                         schemaHelper.isSkippingEligibleLiteral(rightLit)) {
                         // Match any file whose min/max range contains anything other than the
                         // rejected point.
-                        // For example a = 1 --> minValue.a <= 1 AND 1 <= maxValue.a
+                        // For example a != 1 --> minValue.a < 1 OR maxValue.a > 1
```
scottsand-db, line 411: why did you change the comment for the case `=` to be for `!=`?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ScanSuite.scala
```diff
@@ -1497,6 +1497,47 @@ class ScanSuite extends AnyFunSuite with TestUtils with ExpressionTestUtils with
...
+      greaterThan(col("as_int"), ofInt(0)) -> Set(nestedCol("maxValues.as_int")),
+      greaterThanOrEqual(col("as_int"), ofInt(0)) -> Set(nestedCol("maxValues.as_int")),
+      lessThanOrEqual(col("as_int"), ofInt(0)) -> Set(nestedCol("minValues.as_int")),
+      new And(
+        lessThan(col("as_int"), ofInt(0)),
+        greaterThan(col("as_long"), ofInt(0))
+      ) -> Set(nestedCol("minValues.as_int"), nestedCol("maxValues.as_long"))
+    ).foreach { case (predicate, expectedCols) =>
+      val tableClient = tableClientVerifyJsonParseSchema(verifySchema(expectedCols))
+      collectScanFileRows(
```
scottsand-db, line 1533: Are you doing anything with the result of this read?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingPredicate.java
scottsand-db: `DataSkippingPredicate extends Predicate` LGTM

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
allisonport-db: It was wrong before this fixes it. This is in the constructNotDataSkippingFilter so it's NOT(=) --> !=

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ScanSuite.scala
allisonport-db: No just checking that the overridden tableClient doesn't throw an error. and collectScanFileRows is just so the iterators --> seq

scottsand-db (approved):

PR #2483, opened 2024-01-11
Remove disclaimer about the protocol being a draft

dhruvarya-db:
## Description

The protocol still has an old (~4 year old) disclaimer towards the top claiming that it is an in-progress draft. This PR removes that line.

## How was this patch tested?

No tests needed.


tdas (approved):
HAHAHAHA. Thank you for catching this. This should have removed LONG AGO.

PR #2499, opened 2024-01-12
[3.1][Kernel] Downgrade SLF4J version to 1.x to avoid compatibility issues with connectors

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Downgrades SLF4J version to 1.x. Version 2.x includes additional APIs that are unsupported in lower versions. This avoids using those APIs.

## How was this patch tested?

 Ran some unit tests and checked the output logs locally.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -246,13 +246,13 @@ lazy val kernelApi = (project in file("kernel/kernel-api"))
     Test / javaOptions ++= Seq("-ea"),
     libraryDependencies ++= Seq(
       "org.roaringbitmap" % "RoaringBitmap" % "0.9.25",
-      "org.slf4j" % "slf4j-api" % "2.0.9",
+      "org.slf4j" % "slf4j-api" % "1.7.36",
 
       "com.fasterxml.jackson.core" % "jackson-databind" % "2.13.5" % "test",
       "org.scalatest" %% "scalatest" % scalaTestVersion % "test",
       "junit" % "junit" % "4.13" % "test",
       "com.novocode" % "junit-interface" % "0.11" % "test",
-      "org.slf4j" % "slf4j-log4j12" % "2.0.9" % "test"
+      "org.slf4j" % "slf4j-log4j12" % "1.7.36" % "test"
     ),
     javaCheckstyleSettings("kernel/dev/checkstyle.xml"),
     // Unidoc settings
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -246,13 +246,13 @@ lazy val kernelApi = (project in file("kernel/kernel-api"))
     Test / javaOptions ++= Seq("-ea"),
     libraryDependencies ++= Seq(
       "org.roaringbitmap" % "RoaringBitmap" % "0.9.25",
-      "org.slf4j" % "slf4j-api" % "2.0.9",
+      "org.slf4j" % "slf4j-api" % "1.7.36",
```
allisonport-db, line 249: Chose based on "older stable version" here https://www.slf4j.org/download.html

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -473,24 +471,22 @@ protected Optional<LogSegment> getLogSegmentForVersion(
...
+        logDebug(() ->
             String.format(
                 "deltasAfterCheckpoint: %s",
                 Arrays.toString(deltasAfterCheckpoint.stream().map(
-                    x -> new Path(x.getPath()).getName()).toArray())
-            )
-        ).log();
+                    x -> new Path(x.getPath()).getName()).toArray())));
 
-        // todo again naming confusing (specify after checkpoint?)
```
allisonport-db, line 484: I think I accidentally merged this TODO in a previous PR...

nicklan (approved):
lgtm!

vkorukanti (approved):
lgtm

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -580,4 +576,11 @@ private Optional<LogSegment> getLogSegmentWithMaxExclusiveCheckpointVersion(
         // TODO
         return Optional.empty();
     }
+
+    // TODO logger interface to support this across kernel-api module
+    private void logDebug(Supplier<String> message) {
```
scottsand-db, line 581: nice!

scottsand-db (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
tdas: +1 lets make this interface. its going to be internal interface right? so its okay if this is dont after 3.1.

tdas (approved):

PR #2409, opened 2023-12-27
#1478 Add support for syncing Delta table schema and properties to HMS

dhruvarya-db:
#### Which Delta project/connector is this regarding?

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Follow up for #1478.

This PR adds the class `UpdateCatalog` which is used as a post commit hook and during table creation for syncing table schema and properties to HMS. The new behaviour in Delta Lake is that when `catalog.update.enabled` = `true`, the schema and properties of the Delta table will be synced to HMS during creation/replacement and whenever an update to the table updates either of the schema or properties (asynchronously, in a post-commit hook).

Some new additions are:

1. New Config: `catalog.update.enabled` -> Setting this to true will enable schema sync.
2. New OSS Config: `catalog.update.threadPoolSize` -> Controls the size of the thread pool which is used to asynchronously update the catalog.
3. HMS-specific Quirk: Hive does not allow for the removal of the partition columns once it has been added. When sending the table schema to HMS in `UpdateCatalog`, we always send the partition columns as regular data columns to ensure that we can remove these columns on schema update. However, for tables that have been converted from parquet to Delta, the partition columns would have already been set in HMS. Since OSS Spark always appends the partition columns to the end of the schema when it finds them in Hive, for converted tables with partitions, the column order returned by Hive will be incorrect.

## How was this patch tested?

Adds multiple tests in a new suite `DeltaUpdateCatalogSuite`.

Co-authored-by: dhruvarya-db <dhruv.arya@databricks.com>
Co-authored-by: schatterjee6 <Sirsha_Chatterjee@intuit.com>

PR #2505, opened 2024-01-13
[Doc] Add doc describing table schema/properties sync to HMS

dhruvarya-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [X] Other (Doc)

## Description

https://github.com/delta-io/delta/pull/2409 added table metadata sync (to HMS) to Delta Lake. This PR adds some documentation about this feature.

Screenshot 2024-01-12 at 5 39 35 PM

### DIFF
--- a/docs/source/delta-batch.md
+++ b/docs/source/delta-batch.md
@@ -1196,6 +1196,13 @@ You can also set `delta.`-prefixed properties during the first commit to a Delta
 
 See also the [_](/table-properties.md).
 
+## Syncing table schema and properties to the Hive metastore
+
+You can enable asynchronous syncing of table schema and properties to the metastore by setting `spark.databricks.delta.catalog.update.enabled` to `true`. Whenever the Delta client detects that either of these two were changed due to an update, it will sync the changes to the metastore. The config `spark.databricks.delta.catalog.update.threadPoolSize` can be optionally used to control the size of the threadpool that is used for these asynchronous updates.
+
+.. note::
+  - Since partition columns are immutable in Hive metastore, Delta stores them as normal data columns to allow for schema changes that involve partition columns. However, for partitioned tables that were converted into Delta, the initial partition columns will always appear towards end of the returned schema due to this Hive metastore quirk.
+
 <a id="table-metadata"></a>
 
 ## Table metadata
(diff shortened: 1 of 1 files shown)


docs/source/delta-batch.md
```diff
@@ -1196,6 +1196,13 @@ You can also set `delta.`-prefixed properties during the first commit to a Delta
 
 See also the [_](/table-properties.md).
 
+## Syncing table schema and properties to the Hive metastore
+
+You can enable asynchronous syncing of table schema and properties to the metastore by setting `spark.databricks.delta.catalog.update.enabled` to `true`. Whenever the Delta client detects that either of these two were changed due to an update, it will sync the changes to the metastore. The config `spark.databricks.delta.catalog.update.threadPoolSize` can be optionally used to control the size of the threadpool that is used for these asynchronous updates.
```
prakharjain09, line 1201:
```suggestion
You can enable asynchronous syncing of table schema and properties to the metastore by setting `spark.databricks.delta.catalog.update.enabled` to `true`. Whenever the Delta client detects that either of these two were changed due to an update, it will sync the changes to the metastore.```

`spark.databricks.delta.catalog.update.threadPoolSize` is an internal conf - lets avoid talking about it in docs. Also for all practical purposes, we don't expect users to play with it.
This is an

Comment (prakharjain09) [inline docs/source/delta-batch.md:1201]: In the documentation, we should also talk about how the schema would be available in HMS table properties with an example.

```
If Schema is small, It will be available as following in HMS table properties:
{
  'spark.sql.sources.schema': '{"name":"col1","type":"string","nullable":true, "metadata":{}},{"name":"col2","type":"string","nullable":true,"metadata":{}}'
}

If Schema is large, the schema will be broken down into multiple parts. Appending them together should give the correct schema. E.g.
{
  'spark.sql.sources.schema.numParts': '4',
  'spark.sql.sources.schema.part.1': '{"name":"col1","type":"string","nullable":tr',
  'spark.sql.sources.schema.part.2': 'ue, "metadata":{}},{"name":"co',
  'spark.sql.sources.schema.part.3': 'l2","type":"string","nullable":true,"meta',
  'spark.sql.sources.schema.part.4': 'data":{}}'
 }

```

prakharjain09 (approved):

vkorukanti (approved):
lgtm

zhangm365:
But how to configure HMS uris to synchronize metadata to HMS storage?

Issue #2515, opened 2024-01-16
[BUG] Avoid redudant jobs in buildRowIndexSetsForFilesMatchingCondition


andreaschat-db:
#### Which Delta project/connector is this regarding?

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

### Describe the problem
In `buildRowIndexSetsForFilesMatchingCondition` we join the results DataFrame with the candidate file list to fetch the Deletion Vector Descriptors. Currently, there is a sanity check there to verify we do not eliminate any rows due to path encoding issues. The sanity check compares before and after counts. The counts produce 2 extra jobs which can cause a significant performance overhead. For example, in the case of merge each job could be a full shuffle join between the source and the target. We need to eliminate the extra jobs while maintaining the sanity check.

#### Steps to reproduce
N/A.

### Environment information

* Delta Lake version:
* Spark version:
* Scala version:

### Willingness to contribute

The Delta Lake Community encourages bug fix contributions. Would you or another member of your organization be willing to contribute a fix for this bug to the Delta Lake code base?

- [x] Yes. I can contribute a fix for this bug independently.
- [ ] Yes. I would be willing to contribute a fix for this bug with guidance from the Delta Lake community.
- [ ] No. I cannot contribute a bug fix at this time.

foss-contributor:
@andreaschat-db, thank you for your contribution! This issue has been inactive for over 18 months and will be closed in one week. To keep it open, please leave a comment on this issue or contact us in the `#contributors` channel on [Delta Lake Slack](https://go.delta.io/slack). We appreciate your involvement in Delta Lake!

PR #2115, opened 2023-09-27
[Spark] Harmonize path-based Delta table resolution

ryan-johnson-databricks:
#### Which Delta project/connector is this regarding?

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Historically, path-based table resolution logic has been scattered all over Delta's code base. We can build on recent code improvements to start harmonizing and centralizing that logic in `DeltaTableV2`, which already plays a central role in all table resolution. Future changes can continue building on this base.

## How was this patch tested?

Existing unit tests.

## Does this PR introduce _any_ user-facing changes?

No.

PR #2516, opened 2024-01-16
[#2515][Spark] Avoid redudant jobs in buildRowIndexSetsForFilesMatchingCondition

andreaschat-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Avoid redudant jobs in `buildRowIndexSetsForFilesMatchingCondition`.
In buildRowIndexSetsForFilesMatchingCondition we join the results DataFrame with the candidate file list to fetch the Deletion Vector Descriptors. Currently, there is a sanity check there to verify we do not eliminate any rows due to path encoding issues. The sanity check compares before and after counts. The counts produce 2 extra jobs which can cause a significant performance overhead. For example, in the case of merge each job could be a full shuffle join between the source and the target. This PR eliminates the extra jobs while maintaining the sanity check.

Resolves https://github.com/delta-io/delta/issues/2515.

## How was this patch tested?
Added a new test in MergeIntoDVsSuite.

No.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/commands/DMLWithDeletionVectorsHelper.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/commands/DMLWithDeletionVectorsHelper.scala
@@ -391,14 +391,12 @@ object DeletionVectorBitmapGenerator {
       val filePathToDVDf = sparkSession.createDataset(filePathToDV)
 
       val joinExpr = filePathToDVDf("path") === matchedRowsDf(FILE_NAME_COL)
-      val joinedDf = matchedRowsDf.join(filePathToDVDf, joinExpr)
-      assert(joinedDf.count() == matchedRowsDf.count(),
-        s"""
-           |The joined DataFrame should contain the same number of entries as the original
-           |DataFrame. It is likely that _metadata.file_path is not encoded by Spark as expected.
-           |Joined DataFrame count: ${joinedDf.count()}
-           |matchedRowsDf count: ${matchedRowsDf.count()}
-           |""".stripMargin)
+      // Perform leftOuter join to make sure we do not eliminate any rows because of path
+      // encoding issues. If there is such an issue we will detect it during the aggregation
+      // of the bitmaps.
+      val joinedDf = matchedRowsDf.join(filePathToDVDf, joinExpr, "leftOuter")
+        .drop(FILE_NAME_COL)
+        .withColumnRenamed("path", FILE_NAME_COL)
       joinedDf
     } else {
       // When the table has no DVs, just add a column to indicate that the existing dv is null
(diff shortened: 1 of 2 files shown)


vkorukanti (approved):
lgtm, thanks for fixing this.

spark/src/main/scala/org/apache/spark/sql/delta/commands/DMLWithDeletionVectorsHelper.scala
```diff
@@ -391,14 +391,12 @@ object DeletionVectorBitmapGenerator {
...
-        s"""
-           |The joined DataFrame should contain the same number of entries as the original
-           |DataFrame. It is likely that _metadata.file_path is not encoded by Spark as expected.
-           |Joined DataFrame count: ${joinedDf.count()}
-           |matchedRowsDf count: ${matchedRowsDf.count()}
-           |""".stripMargin)
+      // Perform leftOuter join to make sure we do not eliminate any rows because of path
+      // encoding issues. If there is such an issue we will detect it during the aggregation
+      // of the bitmaps.
+      val joinedDf = matchedRowsDf.join(filePathToDVDf, joinExpr, "leftOuter")
```
felipepessoto, line 397: The sanity wasn't created only because of Spark 3.4.0, and bug was fixed in 3.4.1/3.5.0? Can't we just remove the assert?

felipepessoto:
The "Resolves https://github.com/delta-io/delta/pull/2115" is pointing to the incorrect issue.

spark/src/test/scala/org/apache/spark/sql/delta/MergeIntoDVsSuite.scala
```diff
@@ -143,6 +151,52 @@ class MergeIntoDVsSuite extends MergeIntoDVsTests {
         numDeletionVectorsRemoved = 1,
         numDeletionVectorsUpdated = 0)
     }
+
+    test(s"Verify error is produced when paths are not joined correctly") {
```
felipepessoto, line 155: Test is inside another test

spark/src/main/scala/org/apache/spark/sql/delta/commands/DMLWithDeletionVectorsHelper.scala
andreaschat-db: Thanks for the review @felipepessoto. We fixed related issues related to the sanity check in Spark 3.5. In any case, the sanity check is very cheap now and it is good to have a safeguard in case we come across any issues in the future.

felipepessoto: Oh ok, the join is supposed to always match, so we are not returning more rows than before.

felipepessoto (approved):

spark/src/main/scala/org/apache/spark/sql/delta/commands/DMLWithDeletionVectorsHelper.scala
andreaschat-db: Exactly :).

PR #2536, opened 2024-01-18
[#2515][3.1][Spark] Avoid redudant jobs in buildRowIndexSetsForFilesMatchingCondition 

andreaschat-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Avoid redudant jobs in buildRowIndexSetsForFilesMatchingCondition.
In buildRowIndexSetsForFilesMatchingCondition we join the results DataFrame with the candidate file list to fetch the Deletion Vector Descriptors. Currently, there is a sanity check there to verify we do not eliminate any rows due to path encoding issues. The sanity check compares before and after counts. The counts produce 2 extra jobs which can cause a significant performance overhead. For example, in the case of merge each job could be a full shuffle join between the source and the target. This PR eliminates the extra jobs while maintaining the sanity check.

This is a cherry pick of https://github.com/delta-io/delta/pull/2516.
## How was this patch tested?
Added a new test in MergeIntoDVsSuite.

No.


tdas (approved):
Approving this quickly since this is exact cherry pick of the master PR https://github.com/delta-io/delta/pull/2516

PR #2557, opened 2024-01-22
[Spark] Add Writer Protocol check in Vacuum Command

prakharjain09:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
This PR adds Write PROTOCOL check to the Vacuum Command since it makes changes to delta directory.

## How was this patch tested?

UTs

No

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/delta/DeltaProtocolVersionSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/DeltaProtocolVersionSuite.scala
@@ -445,7 +445,7 @@ trait DeltaProtocolVersionSuiteBase extends QueryTest
         overwrite = false,
         log.newDeltaHadoopConf())
 
-      // Both vacuum and vacuum dry run works as expected
+      // Both vacuum and vacuum dry run fails as expected
       vacuumCommandsToTry.foreach { command =>
         intercept[DeltaUnsupportedTableFeatureException] {
           spark.sql(command).collect()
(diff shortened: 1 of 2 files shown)


spark/src/test/scala/org/apache/spark/sql/delta/DeltaProtocolVersionSuite.scala
```diff
@@ -422,6 +422,38 @@ trait DeltaProtocolVersionSuiteBase extends QueryTest
...
+      val newProtocol = Protocol(
+        TABLE_FEATURES_MIN_READER_VERSION,
+        TABLE_FEATURES_MIN_WRITER_VERSION).withWriterFeatures(Seq("newUnsupportedWriterFeature"))
+      log.store.write(
+        deltaFile(log.logPath, snapshot.version + 1),
+        Iterator(Metadata().json, newProtocol.json),
+        overwrite = false,
+        log.newDeltaHadoopConf())
+
+      // Both vacuum and vacuum dry run works as expected
```
ryan-johnson-databricks, line 448:
```suggestion
      // Both vacuum and vacuum dry run fail as expected
```

ryan-johnson-databricks (approved):
LGTM. One wording nit in a code comment.

tdas (approved):
Ohh this is a very good check to do. I think this has always been a major risky gap that vacuum did not protocol version, which completely blocked making deeper changes in the protocol related to vacuum. 

I think this should be backported to 3.1 branch. Let's get this out as quickly as possible.

PR #2563, opened 2024-01-23
[Spark] Fix SQLMetric usage

davintjong-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Fix usage of SQLMetric not to use positive `initValue`, and to register these values with the AccumulatorContext as intended. Only in `WriteIntoDeltaLike.scala`.

Before, we were incorrectly using a metric name as the metric type, where we want to set that on registration rather than construction.

## How was this patch tested?

N/A

No.


sabir-akhadov (approved):
lgtm
The PR description is a bit too general, could mention it's for WriteIntoDeltaLike only.

PR #2566, opened 2024-01-24
[Spark] Add messageParameters to DeltaAnalysisException

larsk-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Add the correct `messageParameters` mapping to `DeltaAnalysisException` (like we have for `DeltaIllegalArgumentException` for example), so that Spark's `checkError` actually tests them, rather than testing against an empty `Map`.

## How was this patch tested?

Test-only PR.

No.


fred-db (approved):
LGTM, thanks for putting in the work and adding this! :)

PR #2571, opened 2024-01-25
[Spark] Use SparkPath in DV code path

xupefei:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR refactors the DV code path to use the `SparkPath` helper and clarifies requirements of whether a path string should or should not be escaped.

This PR also touches on some test cases as a preparation for testing tables with special characters -- which relies on a new Spark release and thus not shipped with this PR.

## How was this patch tested?

Existing tests.

No.


spark/src/main/scala/org/apache/spark/sql/delta/actions/DeletionVectorDescriptor.scala
```diff
@@ -132,7 +133,9 @@ case class DeletionVectorDescriptor(
     storageType match {
       case UUID_DV_MARKER =>
         val absolutePath = this.absolutePath(tableLocation)
-        this.copy(storageType = PATH_DV_MARKER, pathOrInlineDv = absolutePath.toString)
+        this.copy(
+          storageType = PATH_DV_MARKER,
+          pathOrInlineDv = SparkPath.fromPath(absolutePath).urlEncoded)
```
c27kwan, line 138: IIUC, we escape the path here instead of absolutePath because it's using the data in pathOrInlineDV. The tables with DVs already in use do not correctly encode the paths so we must work around this issue by doing the encoding on load, where we convert the Path to string. 

If this is the case, can we
* Make this very clear in a comment why it's encoded here.
* Add a test with an "existing" table that is storing DVs with incorrectly encoded paths and making sure that this solution works.

c27kwan: Now that you mention it, shouldn't this decision also make it to the protocol? It sounds like otherwise we'd have unreadable tables

larsk-db: The protocol states that absolute DVs paths behave the same as the `path` field in Add/RemoveFile actions. So we should go with option 1 then.
Anyway, there's no code that uses absolute paths, yet, so this change shouldn't break anything.

spark/src/test/scala/org/apache/spark/sql/delta/deletionvectors/DeletionVectorsSuite.scala
```diff
@@ -791,4 +810,23 @@ object DeletionVectorsSuite {
...
+  // the `withTableHavingSpecialCharInDVPath` method to create a copy of this table with paths
+  // filled.
+  // Content of this table is range(0, 10) with all even numbers deleted.
+  val tableHavingSpecialCharInDVTemplate = "src/test/resources/delta/table-with-dv-special-char"
+
+  def writeTableHavingSpecialCharInDVPath(path: File, pathIsEncoded: Boolean): Unit = {
+    FileUtils.copyDirectory(new File(tableHavingSpecialCharInDVTemplate), path)
+    val fullPath = new File(
+      path,
+      if (pathIsEncoded) "folder%26with%25special%20char" else "folder&with%special char")
```
c27kwan, line 824: nit: Can this be a named variable?

spark/src/test/scala/org/apache/spark/sql/delta/deletionvectors/DeletionVectorsSuite.scala
```diff
@@ -791,4 +810,23 @@ object DeletionVectorsSuite {
   val table5Sum = 21975159654L
   val table5CountByValues = (0 to 20).map(_ -> 99900000L).toMap + (21 -> 47436174L)
   val table5SumByValues = (0 to 20).map(v => v -> v * 99900000L).toMap + (21 -> 21 * 47436174L)
+
+  // A template table with special characters in DV path. To use this table, one must use
+  // the `withTableHavingSpecialCharInDVPath` method to create a copy of this table with paths
+  // filled.
+  // Content of this table is range(0, 10) with all even numbers deleted.
+  val tableHavingSpecialCharInDVTemplate = "src/test/resources/delta/table-with-dv-special-char"
```
c27kwan, line 818: This can be a val defined within writeTableHavingSpecialCharInDVPath.

spark/src/test/scala/org/apache/spark/sql/delta/deletionvectors/DeletionVectorsSuite.scala
```diff
@@ -791,4 +810,23 @@ object DeletionVectorsSuite {
...
+    FileUtils.copyDirectory(new File(tableHavingSpecialCharInDVTemplate), path)
+    val fullPath = new File(
+      path,
+      if (pathIsEncoded) "folder%26with%25special%20char" else "folder&with%special char")
+      .getCanonicalPath
+    val logJson = new File(path, "_delta_log/00000000000000000000.json")
+    val logJsonContent = FileUtils.readFileToString(logJson, "UTF-8")
+    val newLogJsonContent = logJsonContent.replace(
+      "{{FOLDER_WITH_SPECIAL_CHAR}}", fullPath)
+    FileUtils.write(logJson, newLogJsonContent, "UTF-8")
```
c27kwan, line 830: Since we are manually manipulating the JSON, can we read back the final JSON and validate that it correctly formed?

spark/src/test/scala/org/apache/spark/sql/delta/deletionvectors/DeletionVectorsSuite.scala
```diff
@@ -791,4 +810,23 @@ object DeletionVectorsSuite {
...
   val table5CountByValues = (0 to 20).map(_ -> 99900000L).toMap + (21 -> 47436174L)
   val table5SumByValues = (0 to 20).map(v => v -> v * 99900000L).toMap + (21 -> 21 * 47436174L)
+
+  // A template table with special characters in DV path. To use this table, one must use
+  // the `withTableHavingSpecialCharInDVPath` method to create a copy of this table with paths
+  // filled.
+  // Content of this table is range(0, 10) with all even numbers deleted.
+  val tableHavingSpecialCharInDVTemplate = "src/test/resources/delta/table-with-dv-special-char"
+
+  def writeTableHavingSpecialCharInDVPath(path: File, pathIsEncoded: Boolean): Unit = {
```
c27kwan, line 820: nit:
path -> tableDir
pathIsEncoded -> pathIsStoredEncoded

spark/src/test/resources/delta/table-with-dv-special-char/_delta_log/00000000000000000000.json
```diff
@@ -0,0 +1,4 @@
...
+{"metaData":{"id":"testId","format":{"provider":"parquet","options":{}},"schemaString":"{\"type\":\"struct\",\"fields\":[{\"name\":\"id\",\"type\":\"long\",\"nullable\":true,\"metadata\":{}}]}","partitionColumns":[],"configuration":{"delta.enableDeletionVectors":"true"},"createdTime":1708950819168}}
+{"protocol":{"minReaderVersion":3,"minWriterVersion":7,"readerFeatures":["deletionVectors"],"writerFeatures":["deletionVectors"]}}
+{"add":{"path":"part-00000-8d24f407-08d3-49ab-9d1c-f7f6c129e882-c000.snappy.parquet","partitionValues":{},"size":539,"modificationTime":1708950820783,"dataChange":true,"stats":"{\"numRecords\":10,\"minValues\":{\"id\":0},\"maxValues\":{\"id\":9},\"nullCount\":{\"id\":0},\"tightBounds\":false}","deletionVector":{"storageType":"p","pathOrInlineDv":"file:{{FOLDER_WITH_SPECIAL_CHAR}}/deletion_vector_67bc892e-2979-4760-a78c-856aba806564.bin","offset":1,"sizeInBytes":42,"cardinality":5}}}
```
c27kwan, line 4: If it's not too much work, maybe we should add a test for storageType "u" , where the parent directory is the one that needs special handling. Or is that handled by the other tests?

xupefei: Yes, that case will be handled by the next PR, in which we add special characters to `withTempDir`-generated folders.

xupefei:
I'll create a new PR since this one is becoming messy.

PR #2575, opened 2024-01-26
[Spark] Override Jackson string length limits

olaky:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Override the Jackson string length limit to allow reading and writing large json objects to and from the Delta Log.

## How was this patch tested?

Added a new test suite

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/util/JsonUtils.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/util/JsonUtils.scala
@@ -30,6 +30,8 @@ object JsonUtils {
     _mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
     _mapper.registerModule(DefaultScalaModule)
 
+    // We do not want to limit the length of JSON strings in the Delta log or table data. Also note
+    // that not having a limit was the default behavior before Jackson 2.15.
     val streamReadConstraints = StreamReadConstraints
       .builder()
       .maxStringLength(Int.MaxValue)
--- a/spark/src/main/scala/org/apache/spark/sql/delta/util/JsonUtils.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/util/JsonUtils.scala
@@ -17,6 +17,7 @@
 package org.apache.spark.sql.delta.util
 
 import com.fasterxml.jackson.annotation.JsonInclude.Include
+import com.fasterxml.jackson.core.StreamReadConstraints
 import com.fasterxml.jackson.databind.{DeserializationFeature, ObjectMapper}
 import com.fasterxml.jackson.module.scala.{DefaultScalaModule, ScalaObjectMapper}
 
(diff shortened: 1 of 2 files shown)


spark/src/test/scala/org/apache/spark/sql/delta/util/JsonUtilsSuite.scala
```diff
@@ -0,0 +1,73 @@
...
+    )
+    for ((obj, json) <- testCases) {
+      assert(JsonUtils.toJson(obj) == json)
+      assert(JsonUtils.fromJson[DataSize](json) == obj)
+    }
+  }
+
+  test("Serialize and de-serialize commit info with large message") {
+    val operationStringSize = StreamReadConstraints.DEFAULT_MAX_STRING_LEN * 10
+    assert(operationStringSize > StreamReadConstraints.DEFAULT_MAX_STRING_LEN)
```
c27kwan, line 51: Is this a bit redundant, since the variable is defined based on this constant?

spark/src/main/scala/org/apache/spark/sql/delta/util/JsonUtils.scala
```diff
@@ -28,6 +29,13 @@ object JsonUtils {
     _mapper.setSerializationInclusion(Include.NON_ABSENT)
     _mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
     _mapper.registerModule(DefaultScalaModule)
+
```
c27kwan, line 32: Maybe add a quick comment about us overriding the default. (e.g. why int.max and not long.maxvalue). Was this the old default value?

spark/src/test/scala/org/apache/spark/sql/delta/util/JsonUtilsSuite.scala
```diff
@@ -0,0 +1,73 @@
...
+import org.apache.spark.sql.test.SharedSparkSession
+
+class JsonUtilsSuite
+  extends QueryTest
+  with SharedSparkSession {
+
+  test("DataSize json serialization") {
+    val testCases = Seq(
+      DataSize() -> """{}""",
+      DataSize(bytesCompressed = Some(816L)) -> """{"bytesCompressed":816}""",
```
c27kwan, line 37: Can you add a bit more variance in the data by adding negative long and 0L?

c27kwan: the negative long is maybe more appropriate for the other test, since this one is data size, which shouldn't be negative.

spark/src/test/scala/org/apache/spark/sql/delta/util/JsonUtilsSuite.scala
c27kwan: Did you mean to assert operationStringSize < Int.MaxValue (which can also be its own constant like Delta_JSON_COMMIT_MAX_String_LEN

olaky: What this is supposed to check is that when the value of DEFAULT_MAX_STRING_LEN changes, the *10 does not lead to an overflow because for this test to be relevant, operationStringSize has to be bigger than the default max (because we want to validate that we have a custom default)

spark/src/test/scala/org/apache/spark/sql/delta/util/JsonUtilsSuite.scala
olaky: I added 0 and negative. While neagtive sizes are not valid, serialization should still handle them

spark/src/main/scala/org/apache/spark/sql/delta/util/JsonUtils.scala
olaky: I added a comment. Since the parameter is of type int, I think commenting why it is not long.maxvalue is very redundant

c27kwan (approved):
lgtm! Thanks for fixing

PR #2583, opened 2024-01-30
[Spark] Add config to make post-commit hooks throw on error

cstavr:
- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Currently most post-commit hooks, except manifest generation, suppress errors, i.e. they are logged but do not throw during the command that triggered them. Although this makes sense for production, it can hide issues in testing. This commit adds a config to make post-commit hooks always throw an exception.

## How was this patch tested?

Existing tests provide sufficient coverage.

No


larsk-db (approved):
LGTM

PR #2584, opened 2024-01-31
[Spark] Emit event for compaction errors

cstavr:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Emit Delta event when auto compaction fails.

## How was this patch tested?

Log-only change.

No


larsk-db (approved):
LGTM

Issue #2532, opened 2024-01-17
[Protocol Change Request] Improving Time Travel using In-Commit Timestamps


dhruvarya-db:
### Overview

This feature request is about changing Delta commit timestamps to improve time travel.

### Motivation

Delta currently relies on the file modification time to identify the timestamp of a commit. This timestamp is used for time travel queries, log cleanup, and staleness checks. However, file modification time is not a very reliable way of getting a timestamp — this can easily change when the files are copied/moved to another directory (e.g. for disaster recovery purposes) or when any manual fixes are performed to the Delta log. In such cases, time travel on the delta table breaks as of today. The possibility of non-monotonic file timestamps also adds lots of code complexity in Delta as we try to handle it heuristically in the best possible way.

### Further details

We propose a new Writer feature that will require clients to generate a timestamp just before performing a commit and store it in the commit itself.

Compliant writers will ensure that the timestamp stored in Commit X+1 is always greater than Commit X. To be able to ensure this, the client will need to perform conflict detection for these timestamps.
1. The writer will write this timestamp in the CommitInfo action. Furthermore, the writer will always write CommitInfo as the first action in a commit.
2. Clients that understand these new timestamps will now read the commit file to get the actual timestamp. These timestamps will now be used for time travel queries and by other operations that use timestamps.

The detail proposal and the required protocol changes are sketched out [in this doc](https://docs.google.com/document/d/1SqMcvrt5eeWsKH5Y4hzR1r1VeuDSrft5zoBmPuIbuv8/edit?usp=sharing).
### Willingness to contribute

The Delta Lake Community encourages new feature contributions. Would you or another member of your organization be willing to contribute an implementation of this feature?

- [X] Yes. I can contribute this feature independently.
- [ ] Yes. I would be willing to contribute this feature with guidance from the Delta Lake community.
- [ ] No. I cannot contribute this feature at this time.

dhruvarya-db:
This is being released as a preview feature in 3.2 (https://github.com/delta-io/delta/pull/2962). The feature will be generally available by 4.0.

allisonport-db:
This is being released in 3.3.0 (in RC stage right now)

PR #2596, opened 2024-02-02
[Spark] Introduce InCommitTimestamp feature and write monotonically increasing timestamps in CommitInfo

dhruvarya-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Follow-up for https://github.com/delta-io/delta/issues/2532.

Adds a new writer feature called `inCommitTimestamp`. When this feature is enabled, the writer will make sure that it writes `commitTimestamp` in CommitInfo which contains a monotonically increasing timestamp.

This PR is an initial implementation, it does not handle timestamp retrieval efficiently. It does not try to populate the inCommitTimestamp in Snapshot even in places where it is already available, instead Snapshot has to perform an IO to read the timestamp.

## How was this patch tested?

Added a new suite called `InCommitTimestampSuite`.

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
@@ -165,6 +165,7 @@ private[delta] class ConflictChecker(
     checkForAddedFilesThatShouldHaveBeenReadByCurrentTxn()
     checkForDeletedFilesAgainstCurrentTxnReadFiles()
     checkForDeletedFilesAgainstCurrentTxnDeletedFiles()
+    resolveTimestampOrderingConflicts()
 
     logMetrics()
     currentTransactionInfo
@@ -552,6 +553,53 @@ private[delta] class ConflictChecker(
     }
 
     currentTransactionInfo = currentTransactionInfo.copy(actions = newActions)
+  }
+
+  /**
+   * Adjust the current transaction's commit timestamp to account for the winning
+   * transaction's commit timestamp. If this transaction newly enabled ICT, also update
+   * the table properties to reflect the adjusted enablement version and timestamp.
+   */
+  private def resolveTimestampOrderingConflicts(): Unit = {
+    if (!DeltaConfigs.IN_COMMIT_TIMESTAMPS_ENABLED.fromMetaData(currentTransactionInfo.metadata)) {
+      return
+    }
+
+    val winningCommitTimestamp =
+      if (InCommitTimestampUtils.didCurrentTransactionEnableICT(
+              currentTransactionInfo.metadata, currentTransactionInfo.readSnapshot)) {
+        // Since the current transaction enabled inCommitTimestamps, we should use the file
+        // timestamp from the winning transaction as its commit timestamp.
+        winningCommitFileStatus.getModificationTime
+    } else {
+      // Get the inCommitTimestamp from the winning transaction.
+      CommitInfo.getRequiredInCommitTimestamp(
+        winningCommitSummary.commitInfo, winningCommitVersion.toString)
+    }
+    val currentTransactionTimestamp = CommitInfo.getRequiredInCommitTimestamp(
+      currentTransactionInfo.commitInfo, "NEW_COMMIT")
+    // getRequiredInCommitTimestamp will throw an exception if commitInfo is None.
+    val currentTransactionCommitInfo = currentTransactionInfo.commitInfo.get
+    val updatedCommitTimestamp = Math.max(currentTransactionTimestamp, winningCommitTimestamp + 1)
+    val updatedCommitInfo =
+      currentTransactionCommitInfo.copy(inCommitTimestamp = Some(updatedCommitTimestamp))
+    currentTransactionInfo = currentTransactionInfo.copy(commitInfo = Some(updatedCommitInfo))
+    val nextAvailableVersion = winningCommitVersion + 1L
+    val updatedMetadata =
+      InCommitTimestampUtils.getUpdatedMetadataWithICTEnablementInfo(
+        updatedCommitTimestamp,
+        currentTransactionInfo.readSnapshot,
+        currentTransactionInfo.metadata,
+        nextAvailableVersion)
+    updatedMetadata.foreach { updatedMetadata =>
+      currentTransactionInfo = currentTransactionInfo.copy(
+        metadata = updatedMetadata,
+        actions = currentTransactionInfo.actions.map {
+          case _: Metadata => updatedMetadata
+          case other => other
+        }
+      )
+    }
   }
 
   /** A helper function for pretty printing a specific partition directory. */
--- a/spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
@@ -40,6 +40,7 @@ import org.apache.spark.sql.delta.metering.DeltaLogging
 import org.apache.spark.sql.delta.schema.{SchemaMergingUtils, SchemaUtils}
 import org.apache.spark.sql.delta.sources.DeltaSQLConf
 import org.apache.spark.sql.delta.stats._
+import org.apache.spark.sql.util.ScalaExtensions._
 import org.apache.hadoop.fs.{FileStatus, Path}
 
 import org.apache.spark.SparkException
@@ -296,7 +297,7 @@ trait OptimisticTransactionImpl extends TransactionalWrite
    * Tracks the start time since we started trying to write a particular commit.
    * Used for logging duration of retried transactions.
    */
-  protected var commitAttemptStartTime: Long = _
+  protected var commitAttemptStartTimeMillis: Long = _
 
   /**
    * Tracks actions within the transaction, will commit along with the passed-in actions in the
@@ -370,6 +371,18 @@ trait OptimisticTransactionImpl extends TransactionalWrite
   // The commit store of a table shouldn't change. If it is changed by a concurrent commit, then it
   // will be detected as a conflict and the transaction will anyway fail.
   private[delta] val preCommitCommitStoreOpt: Option[CommitStore] = snapshot.commitStoreOpt
+
+  /**
+   * Generates a timestamp which is greater than the commit timestamp
+   * of the last snapshot. Note that this is only needed when the
+   * feature `inCommitTimestamps` is enabled.
+   */
+  protected[delta] def generateInCommitTimestampForFirstCommitAttempt(
+      currentTimestamp: Long): Option[Long] =
+    Option.when(DeltaConfigs.IN_COMMIT_TIMESTAMPS_ENABLED.fromMetaData(metadata)) {
+      val lastCommitTimestamp = snapshot.timestamp
+      math.max(currentTimestamp, lastCommitTimestamp + 1)
+    }
 
   /** The end to end execution time of this transaction. */
   def txnExecutionTimeMs: Option[Long] = if (commitEndNano == -1) {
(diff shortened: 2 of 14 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala
```diff
@@ -1171,6 +1194,20 @@ object CommitInfo {
     Some(s"Apache-Spark/${org.apache.spark.SPARK_VERSION} Delta-Lake/${io.delta.VERSION}")
   }
 
+  /**
+   * Returns the `inCommitTimestamp` of the given `commitInfoOpt` if it is defined.
+   * Throws an exception if either `commitInfoOpt` is not defined or `inCommitTimestamp` is not
+   * defined in `commitInfoOpt`.
```
ryan-johnson-databricks, line 1200:
```suggestion
   * Throws an exception if `commitInfoOpt` is empty or contains an empty `inCommitTimestamp`.
```

ryan-johnson-databricks (commented):
It seems like this PR needs to implement the version/timestamp enablement table properties as well. Even if they're not used yet, we can unit test their validity, including proper handling of physical conflicts when newly enabling the feature.

spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
```diff
@@ -547,6 +548,48 @@ private[delta] class ConflictChecker(
...
+
+    val currentTransactionCommitInfo = currentTransactionInfo.commitInfo.getOrElse {
+      throw new IllegalStateException("CommitInfo must be present in the current transaction" +
+        s" as it has ${InCommitTimestampTableFeature.name} feature enabled.")
+    }
+    val currentTransactionTimestamp = currentTransactionCommitInfo.inCommitTimestamp.getOrElse {
+      throw new IllegalStateException(
+        "CommitInfo.commitTimestamp must be present in the current transaction" +
+          s" as it has ${InCommitTimestampTableFeature.name} feature enabled.")
+    }
```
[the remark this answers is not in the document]
dhruvarya-db: The way I was thinking about it was: 
1. Winning Commit is missing timestamp: It was written by a non-compliant writer --> This is potentially a user error.
2. Current txn is missing timestamp: This is worse in the sense that it should never actually happen. It is inconsistent behaviour by the current writer.

spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
```diff
@@ -547,6 +548,48 @@ private[delta] class ConflictChecker(
     currentTransactionInfo = currentTransactionInfo.copy(actions = newActions)
   }
 
+  /** Ensure that commitInfo.commitTimestamp is monotonically increasing. */
+  private def resolveTimestampOrderingConflicts(): Unit = {
+    val isICTCurrentlyEnabled =
+      DeltaConfigs.IN_COMMIT_TIMESTAMPS_ENABLED.fromMetaData(currentTransactionInfo.metadata)
+    if (!isICTCurrentlyEnabled) {
```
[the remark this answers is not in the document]
dhruvarya-db: This was intentional so that I don't need another comment around L577 explaining that if we are at that point, then we already know that ICT is enabled in the current txn. It can be removed if you find it unnecessary from that perspective.

spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
```diff
@@ -547,6 +548,48 @@ private[delta] class ConflictChecker(
...
+    // the current transaction would have already been aborted since ICT-enablement is a
+    // metadata change operation (see [[checkNoMetadataUpdates]]).
+    val didCurrentTransactionEnableICT = isICTCurrentlyEnabled && !wasICTEnabledInReadSnapshot
+    val winningCommitTimestamp = if (didCurrentTransactionEnableICT) {
+        // Since the current transaction enabled inCommitTimestamps, we should use the file
+        // timestamp from the winning transaction as its commit timestamp.
+        winningCommitFileStatus.getModificationTime
+    } else {
+      // Get the inCommitTimestamp from the winning transaction.
+      CommitInfo.getRequiredInCommitTimestamp(winningCommitSummary.commitInfo, winningCommitVersion)
```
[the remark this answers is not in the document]
dhruvarya-db: > Also, why do we care at all about the read transaction here? Right now, none of the logic actually cares who enabled ICT -- we just need to extract the correct timestamp from the winning transaction, so we can adjust ours.

That is true, we just need to know whether the winning transaction has ICT enabled and get the right timestamp accordingly. However, the way to do that will effectively be the same as checking whether the current transaction enabled ICT.

spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
```diff
@@ -547,6 +548,36 @@ private[delta] class ConflictChecker(
...
+   * transaction's commit timestamp. If this transaction newly enabled ICT, also update
+   * the table properties to reflect the adjusted enablement version and timestamp.
+   */
+  private def resolveTimestampOrderingConflicts(): Unit = {
+    if (!DeltaConfigs.IN_COMMIT_TIMESTAMPS_ENABLED.fromMetaData(currentTransactionInfo.metadata)) {
+      return
+    }
+
+    val winningCommitTimestamp =
+      if (InCommitTimestampUtils.didCurrentTransactionEnableICT(currentTransactionInfo)) {
```
ryan-johnson-databricks, line 562: I don't see a definition for `InCommitTimestampUtils`... perhaps a file was not added to the PR?

dhruvarya-db: Yes, just added it.

spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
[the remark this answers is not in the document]
ryan-johnson-databricks: In this case, the whole function is gated on ICT being enabled in the current transaction. 
We shouldn't need a lot of reminders in the function body?

Meanwhile, the danger of redundant control flow is high complexity: every check we make implies a branch point in the logic, and somebody has to think about what happens on both sides of the branch. If one side is impossible, we risk wasting a lot of effort reasoning about (and even coding up) the dead case.

spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
```diff
@@ -1648,6 +1663,8 @@ trait OptimisticTransactionImpl extends TransactionalWrite
       for (attemptNumber <- 0 to maxRetryAttempts) {
         try {
           val postCommitSnapshot = if (attemptNumber == 0) {
+            updatedCurrentTransactionInfo = InCommitTimestampUtils
+              .updateInCommitTimestampEnablementInfo(updatedCurrentTransactionInfo, commitVersion)
```
ryan-johnson-databricks, line 1667: I don't love how the logic to update ICT info is scattered in so many parts of the code. It would be really nice if we could find a simpler way.

For example -- can we move this inside `doCommit`? It has access to everything we need, and then we wouldn't need to worry about duplicating that code inside the conflict checker, because the retry path also calls `doCommit` after resolving conflicts. The commit info update is anyway far separated from this code, so we wouldn't give up any locality.

Downside is, `doCommit` doesn't currently transform the actions, so it's an unexpected place for such logic. But this spot is no better in that regard.

ryan-johnson-databricks: It would be tempting to move the enablement check into `updateMetadataInternal`, where other table feature enablement checks already live, but it won't have the proposed commit timestamp yet.

spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
```diff
@@ -1082,15 +1095,16 @@ trait OptimisticTransactionImpl extends TransactionalWrite
...
-        op.jsonEncodedValues,
-        Map.empty,
-        Some(readVersion).filter(_ >= 0),
-        Option(isolationLevelToUse.toString),
-        Some(isBlindAppend),
-        getOperationMetrics(op),
-        getUserMetadata(op),
+        time = clock.getTimeMillis(),
+        operation = op.name,
+        inCommitTimestamp = generateInCommitTimestampForFirstCommitAttempt(),
```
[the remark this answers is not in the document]
ryan-johnson-databricks: Also, a third clock call at L1134 below.
```scala
commitAttemptStartTime = clock.getTimeMillis()
```
Seems helpful for the attempt start time to match the proposed ICT? 
Maybe just move this code above L1097 and then everything can use the `commitAttemptStartTime`?

spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
```diff
@@ -547,6 +548,36 @@ private[delta] class ConflictChecker(
...
+        // Since the current transaction enabled inCommitTimestamps, we should use the file
+        // timestamp from the winning transaction as its commit timestamp.
+        winningCommitFileStatus.getModificationTime
+    } else {
+      // Get the inCommitTimestamp from the winning transaction.
+      CommitInfo.getRequiredInCommitTimestamp(winningCommitSummary.commitInfo, winningCommitVersion)
```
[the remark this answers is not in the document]
dhruvarya-db: I have refactored the code a bit and removed `InCommitTimestampUtils .getRequiredCommitInfoAndCommitTimestamp`

spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
```diff
@@ -1081,28 +1095,44 @@ trait OptimisticTransactionImpl extends TransactionalWrite
...
+        isolationLevel = Option(isolationLevelToUse.toString),
+        isBlindAppend = Some(isBlindAppend),
+        operationMetrics = getOperationMetrics(op),
+        userMetadata = getUserMetadata(op),
         tags = if (tags.nonEmpty) Some(tags) else None,
         txnId = Some(txnId))
 
+      val firstAttemptVersion = getFirstAttemptVersion
+      val updatedMetadataOpt = commitInfo.inCommitTimestamp.flatMap { inCommitTimestamp =>
+        InCommitTimestampUtils.getUpdatedMetadataWithICTEnablementInfo(
```
dhruvarya-db, line 1115: The part where we update the enablement tracking properties has been moved to this place from `doCommitIteratively`.

spark/src/test/scala/org/apache/spark/sql/delta/InCommitTimestampSuite.scala
```diff
@@ -0,0 +1,343 @@
...
+          DeltaConfigs.IN_COMMIT_TIMESTAMP_ENABLEMENT_VERSION.fromMetaData(ver3Snapshot.metadata)
+        // Make sure that this transaction resulted in a conflict.
+        assert(filterUsageRecords(usageRecords, "delta.commit.retry").length == 1)
+        assert(observedEnablementTimestamp.get == getInCommitTimestamp(deltaLog, version = 3))
+        assert(observedEnablementVersion.get == 3)
+      }
+    }
+  }
+
+  test("commitLarge should correctly set the enablement tracking properties") {
```
dhruvarya-db, line 312: Added a new test for `commitLarge`

spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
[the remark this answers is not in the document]
dhruvarya-db: Updated all such callsites so that they use the same value.

spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
```diff
@@ -1177,8 +1191,9 @@ trait OptimisticTransactionImpl extends TransactionalWrite
     try {
       val tags = Map.empty[String, String]
       val commitInfo = CommitInfo(
-        time = clock.getTimeMillis(),
+        clock.getTimeMillis(),
         operation = op.name,
+        generateInCommitTimestampForFirstCommitAttempt(),
```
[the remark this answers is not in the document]
dhruvarya-db: >  I'm pretty surecommitLarge needs the ability to enable ICT as well

Yes, I completely missed this. I have updated this and added a new test for commitLarge.

spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
dhruvarya-db: Moved this to one level up in the stack trace to `commitImpl`.

spark/src/main/scala/org/apache/spark/sql/delta/Snapshot.scala
```diff
@@ -93,6 +95,26 @@ class Snapshot(
...
+   * When InCommitTimestampTableFeature is enabled, the timestamp
+   * is retrieved from the CommitInfo of the latest commit which
+   * can result in an IO operation.
+   */
+  def timestamp: Long = {
+    if (DeltaConfigs.IN_COMMIT_TIMESTAMPS_ENABLED.fromMetaData(metadata)) {
+      inCommitTimestampOpt.getOrElse {
+        val commitInfoOpt = DeltaHistoryManager.getCommitInfoOpt(
+          deltaLog.store, deltaLog.logPath, version, deltaLog.newDeltaHadoopConf())
+        CommitInfo.getRequiredInCommitTimestamp(commitInfoOpt, version)
```
prakharjain09, line 111: This method can throw an exception when the commitInfo doesn't have inCommitTimestamp or when commitInfo itself is not present?

In such a scenario, snapshot.timestamp will fail. Is this intentional?

dhruvarya-db: This is intentional. Such a thing happening is indicative of a writer not following the protocol. In my opinion, we should surface this issue as soon as we see it to prevent further breakage.

ryan-johnson-databricks (approved):
LGTM. There are several annoying almost-redundancies in the code, but there are just enough differences mixed in that I couldn't find any good way to factor them out better.

prakharjain09 (approved):

PR #2597, opened 2024-02-02
Add tests for schema evolution in MERGE with assignments qualified with target name

johanl-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Add a set of tests to capture the current behavior of schema evolution in MERGE when assignments are qualified with the target table name or alias, in particular when that qualifier conflicts with an existing column name, typ.:

```
-- target: key int, t struct<a: int> 
MERGE INTO target t
USING source s
ON t.key = s.key
WHEN MATCHED THEN UPDATE SET t.value = s.value
```

## How was this patch tested?
N/A, test-only change

N/A

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/delta/MergeIntoSchemaEvolutionSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/MergeIntoSchemaEvolutionSuite.scala
@@ -88,7 +88,7 @@ trait MergeIntoSchemaEvolutionMixin {
       }
     }
 
-    test(s"schema evolution - $name - enabled") {
+    test(s"schema evolution - $name") {
       withSQLConf((confs :+ (DeltaSQLConf.DELTA_SCHEMA_AUTO_MIGRATE.key, "true")): _*) {
         append(targetData, partitionCols)
         withTempView("source") {
@@ -242,7 +242,7 @@ trait MergeIntoSchemaEvolutionBaseTests {
   /**
    * Helper method similar to [[testEvolution()]] but without aliasing the target and source tables
    * as 't' and 's'. Used to check that attribute resolution works correctly with schema evolution
-   * when using column name qualified with the actual table name: `target.column`.
+   * when using column name qualified with the actual table name: `table_name.column`.
    */
   def testEvolutionWithoutTableAliases(name: String)(
       targetData: => DataFrame,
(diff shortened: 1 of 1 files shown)


spark/src/test/scala/org/apache/spark/sql/delta/MergeIntoSchemaEvolutionSuite.scala
```diff
@@ -951,6 +1012,168 @@ trait MergeIntoSchemaEvolutionBaseTests {
...
+      .add("t", new StructType()
+        .add("a", IntegerType)),
+    sourceSchema = new StructType()
+      .add("a", IntegerType)
+      .add("b", IntegerType)
+      .add("t", new StructType()
+        .add("a", IntegerType)
+        .add("b", IntegerType)),
+    cond = "1 = 1",
+    clauses = update("t.t.b = s.t.b") :: Nil,
```
olaky, line 1128: What happens if we mix in schemas here? t could also be the name of a schema, potentially even a catalog

spark/src/test/scala/org/apache/spark/sql/delta/MergeIntoSchemaEvolutionSuite.scala
```diff
@@ -951,6 +1012,168 @@ trait MergeIntoSchemaEvolutionBaseTests {
...
+    sourceData = Seq((1, 1, "extra1"), (2, 2, "extra2")).toDF("key", "value", "extra"),
+    clauses = update(set = "s.value = s.value") :: Nil,
+    // Assigning to the source is just wrong and should fail.
+    expected = ((0, 0) +: (3, 30) +: (1, 1) +: Nil)
+      .toDF("key", "value"),
+    expectErrorWithoutEvolutionContains = "cannot resolve s.value in UPDATE clause")
+
+  testNestedStructsEvolution("nested field assignment qualified with source alias")(
+    target = Seq("""{ "a": 1, "t": { "a": 2 } }"""),
+    // target = Seq("""{ "a": 1 }"""),
```
olaky, line 1027: debugging output? (also below)

johanl-db: Right, I removed it.

spark/src/test/scala/org/apache/spark/sql/delta/MergeIntoSchemaEvolutionSuite.scala
johanl-db: It would have little added value I think, the resolution / disambiguation when using <table>, <schema>.<table> or <catalog>.<schema>.<table> is the same (they all three resolve to a single table alias) so that covering just the <table> case vs <table>.<column> already covers most of the ground for merge

PR #2599, opened 2024-02-02
[Protocol Change] Add the In-Commit Timestamps spec change proposal

dhruvarya-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [X] Other (Protocol)

## Description

Adds the proposal for spec change In-Commit Timestamps (see https://github.com/delta-io/delta/issues/2532) to the RFC folder.

## How was this patch tested?

Spec-only change.

No

### DIFF
--- a/protocol_rfcs/in-commit-timestamps.md
+++ b/protocol_rfcs/in-commit-timestamps.md
@@ -25,7 +25,7 @@ When In-Commit Timestamp are enabled, writers are required to include a commitIn
 Field Name | Data Type | Description
 -|-|-
 _commit_version|`Long`| The table version containing the change. This can be derived from the name of the Delta log file that contains actions.
-_commit_timestamp|`Timestamp`| The timestamp associated when the commit was created. ~~This can be derived from the file modification time of the Delta log file that contains actions.~~ <ins> Depending on whether [In-Commit Timestamps](#in-commit-timestamps) are enabled, this is either the file modification time or the `inCommitTimestamp` stored in the `CommitInfo` action of the Delta log file with the version `__commit_version`.</ins>
+_commit_timestamp|`Timestamp`| The timestamp associated when the commit was created. ~~This can be derived from the file modification time of the Delta log file that contains actions.~~ <ins> Depending on whether [In-Commit Timestamps](#in-commit-timestamps) are enabled, this is either the file modification time or the `inCommitTimestamp` stored in the `commitInfo` action of the Delta log file with the version `__commit_version`.</ins>
 
 # In-Commit Timestamps
 > ***New Section after the [Clustered Table](#clustered-table) section***
@@ -42,7 +42,7 @@ Enablement:
 When In-Commit Timestamps is enabled, then:
 1. Writers must write the `commitInfo` (see [Commit Provenance Information](#commit-provenance-information)) action in the commit.
 2. The `commitInfo` action must be the first action in the commit.
-3. The `commitInfo` action must include a field named `inCommitTimestamp`, of type `long` (see [Primitive Types](#primitive-types)), which represents the Unix time in milliseconds when the commit is considered to have succeeded. It is the larger of two values:
+3. The `commitInfo` action must include a field named `inCommitTimestamp`, of type `long` (see [Primitive Types](#primitive-types)), which represents the time (in milliseconds since the Unix epoch) when the commit is considered to have succeeded. It is the larger of two values:
    - The Unix wall clock time at which the writer attempted the commit
    - One millisecond later than the previous commit's `inCommitTimestamp`
 4. If the table has commits from a period when this feature was not enabled, provenance information around when this feature was enabled must be tracked in table properties:
--- a/protocol_rfcs/in-commit-timestamps.md
+++ b/protocol_rfcs/in-commit-timestamps.md
@@ -42,8 +42,8 @@ Enablement:
 When In-Commit Timestamps is enabled, then:
 1. Writers must write the `commitInfo` (see [Commit Provenance Information](#commit-provenance-information)) action in the commit.
 2. The `commitInfo` action must be the first action in the commit.
-3. The `commitInfo` action must include a field named `inCommitTimestamp`, of type `long` (see [Primitive Types](#primitive-types)), which represents the UTC time in milliseconds when the commit is considered to have succeeded. It is the larger of two values:
-   - The UTC wall clock time at which the writer attempted the commit
+3. The `commitInfo` action must include a field named `inCommitTimestamp`, of type `long` (see [Primitive Types](#primitive-types)), which represents the Unix time in milliseconds when the commit is considered to have succeeded. It is the larger of two values:
+   - The Unix wall clock time at which the writer attempted the commit
    - One millisecond later than the previous commit's `inCommitTimestamp`
 4. If the table has commits from a period when this feature was not enabled, provenance information around when this feature was enabled must be tracked in table properties:
    - The property `delta.inCommitTimestampEnablementVersion` must be used to track the version of the table when this feature was enabled.
--- a/protocol_rfcs/in-commit-timestamps.md
+++ b/protocol_rfcs/in-commit-timestamps.md
@@ -22,13 +22,13 @@ When In-Commit Timestamp are enabled, writers are required to include a commitIn
 ...
 3. Change data readers should return the following extra columns:
 
-    Field Name | Data Type | Description
-    -|-|-
-    _commit_version|`Long`| The table version containing the change. This can be derived from the name of the Delta log file that contains actions.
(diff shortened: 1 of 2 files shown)


protocol_rfcs/in-commit-timestamps.md
```diff
@@ -0,0 +1,49 @@
...
+   - The UTC wall clock time at which the writer attempted the commit
+   - One millisecond later than the previous commit's `inCommitTimestamp`
+4. If the table has commits from a period when this feature was not enabled, provenance information around when this feature was enabled must be tracked in table properties:
+   - The property `delta.inCommitTimestampEnablementVersion` must be used to track the version of the table when this feature was enabled.
+   - The property `delta.inCommitTimestampEnablementTimestamp` must be the same as the `inCommitTimestamp` of the commit when this feature was enabled.
+5. The `inCommitTimestamp` of the commit that enables this feature must be greater than the file modification time of the immediately preceding commit.
+
+## Recommendations for Readers of Tables with In-Commit Timestamps
+
+For tables with In-Commit timestamps enabled, readers should use the `inCommitTimestamp` as the commit timestamp for operations like time travel.
```
ryan-johnson-databricks, line 43: Should we mention DESCRIBE HISTORY? It's technically a spark-specific command, but it seems reasonable that other clients would want to provide something similar if they don't already.

protocol_rfcs/in-commit-timestamps.md
```diff
@@ -0,0 +1,49 @@
...
+If a table has commits from a period before In-Commit timestamps were enabled, the table property `delta.inCommitTimestampEnablementVersion` would be set and can be used to identify commits that don't have `inCommitTimestamp`.
+To correctly determine the commit timestamp for these tables, readers can use the following rules:
+1. For commits with version >= `delta.inCommitTimestampEnablementVersion`, readers should use the `inCommitTimestamp` field of the `commitInfo` action.
+2. For commits with version < `delta.inCommitTimestampEnablementVersion`, readers should use the file modification timestamp.
```
ryan-johnson-databricks, line 47: This seems to be talking about version-based time travel, which isn't really related. If the client is trying to do timestamp-based time travel, they should first check whether the timestamp is before the `delta.inCommitTimestampEnablementTimestamp`; if so, timestamp resolution should only consider commits before `delta.inCommitTimestampEnablementVersion`, using file modification timestamps. Otherwise, timestamp resolution should only consider commits starting from `delta.inCommitTimestampEnablementVersion`, using in-commit timestamps.

protocol_rfcs/in-commit-timestamps.md
```diff
@@ -0,0 +1,63 @@
...
+### Commit Provenance Information
+> ***Change to existing section***
+
+A delta file can optionally contain additional provenance information about what higher-level operation was being performed as well as who executed it.
+
+Implementations are free to store any valid JSON [object literal](https://www.w3schools.com/js/js_json_objects.asp) as the `commitInfo` action <ins>unless some table feature (e.g. [In-Commit Timestamps](#in-commit-timestamps)) imposes additional requirements on the data</ins>.
+
+When In-Commit Timestamp are enabled, writers are required to include a commitInfo action with every commit, which must include the `inCommitTimestamp` field.
+
+#### Reader Requirements for AddCDCFile
```
dhruvarya-db, line 19: Added new changes to this section.

protocol_rfcs/in-commit-timestamps.md
```diff
@@ -0,0 +1,49 @@
...
+- The table must be on Writer Version 7.
+- The feature `inCommitTimestamps` must exist in the table `protocol`'s `writerFeatures`.
+- The table property `delta.enableInCommitTimestamps` must be set to `true`.
+
+## Writer Requirements for In-Commit Timestamps
+
+When In-Commit Timestamps is enabled, then:
+1. Writers must write the `commitInfo` (see [Commit Provenance Information](#commit-provenance-information)) action in the commit.
+2. The `commitInfo` action must be the first action in the commits.
+3. The `commitInfo` action must include a field named `inCommitTimestamp`, of type `timestamp` (see [Primitive Types](#primitive-types)), which represents the UTC time when the commit is considered to have succeeded. It is the larger of two values:
```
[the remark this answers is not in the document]
dhruvarya-db: Looks like CDC readers are expected to return a `Timestamp` (https://github.com/delta-io/delta/pull/2599/files#diff-ff95f55525ec15c41f84b2215cd0f45fbdf4a88b2e4c4ef126ed3bea0b900597R28) but I agree, we don't really need µs resolution so a `Long` should suffice.

protocol_rfcs/in-commit-timestamps.md
```diff
@@ -0,0 +1,49 @@
...
+   - The UTC wall clock time at which the writer attempted the commit
+   - One millisecond later than the previous commit's `inCommitTimestamp`
+4. If the table has commits from a period when this feature was not enabled, provenance information around when this feature was enabled must be tracked in table properties:
+   - The property `delta.inCommitTimestampEnablementVersion` must be used to track the version of the table when this feature was enabled.
+   - The property `delta.inCommitTimestampEnablementTimestamp` must be the same as the `inCommitTimestamp` of the commit when this feature was enabled.
```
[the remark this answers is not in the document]
dhruvarya-db: > Yes?

Yes, this is exactly correct. Should I just add this table to the spec?

> And it is an error for a commitInfo to not provide the in-commit timestamp, if it lies within the ICT window those properties identify?

Yes.

> Should we just leave the properties in place permanently, for simpler checking? Or are they bulky enough it's helpful to garbage collect them at some point? If we did GC them, we'd probably want to do it only when metadata is anyway changing for some other reason, since table property changes conflict with all concurrent transactions.

Would it be okay to just leave this up to the writer? The spec, as written, allows for both possibilities.

protocol_rfcs/in-commit-timestamps.md
dhruvarya-db: Added this as a separate section: https://github.com/delta-io/delta/pull/2599/files#r1480345698. Could combine these two if that looks better.

protocol_rfcs/in-commit-timestamps.md
```diff
@@ -42,8 +42,8 @@ Enablement:
 When In-Commit Timestamps is enabled, then:
 1. Writers must write the `commitInfo` (see [Commit Provenance Information](#commit-provenance-information)) action in the commit.
 2. The `commitInfo` action must be the first action in the commit.
-3. The `commitInfo` action must include a field named `inCommitTimestamp`, of type `long` (see [Primitive Types](#primitive-types)), which represents the UTC time in milliseconds when the commit is considered to have succeeded. It is the larger of two values:
-   - The UTC wall clock time at which the writer attempted the commit
+3. The `commitInfo` action must include a field named `inCommitTimestamp`, of type `long` (see [Primitive Types](#primitive-types)), which represents the Unix time in milliseconds when the commit is considered to have succeeded. It is the larger of two values:
+   - The Unix wall clock time at which the writer attempted the commit
```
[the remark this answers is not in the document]
dhruvarya-db: Replaced both instances with "in milliseconds since the Unix epoch"

protocol_rfcs/in-commit-timestamps.md
dhruvarya-db: Added a references to `DESCRIBE HISTORY`

protocol_rfcs/in-commit-timestamps.md
```diff
@@ -0,0 +1,63 @@
...
+
+# In-Commit Timestamps
+> ***New Section after the [Clustered Table](#clustered-table) section***
+
+The In-Commit Timestamps writer feature strongly associates a monotonically increasing timestamp with each commit by storing it in the commit's metadata.
+
+Enablement:
+- The table must be on Writer Version 7.
+- The feature `inCommitTimestamps` must exist in the table `protocol`'s `writerFeatures`.
+- The table property `delta.enableInCommitTimestamps` must be set to `true`.
```
ryan-johnson-databricks, line 38: What purpose would this table property serve? There isn't much of an enable/disable story, is there? No metadata or data to rewrite, no one-way doors, etc?

CC @bart-samwel since he's thought a _lot_ about properties/upgrade/downgrade issues.

ryan-johnson-databricks (approved):
Looks good to me except the one question about table property. But we can probably iterate on that as the prototype takes shape and we have more information to work from.

ryan-johnson-databricks (approved):
Oh, but we do need to add a proposed entry for this to the RFC index.md!

dhruvarya-db:
> Oh, but we do need to add a proposed entry for this to the RFC index.md!

Done

protocol_rfcs/in-commit-timestamps.md
dhruvarya-db: >  No metadata or data to rewrite

We might want to remove `delta.inCommitTimestampEnablementTimestamp/Version` from the metadata on disablement. Though it is technically okay even if we do not.

scottsand-db (approved):

protocol_rfcs/in-commit-timestamps.md
ryan-johnson-databricks: I think we would definitely want to remove them, unless we're willing to track multiple ranges of commits where ICT was enabled. But that can easily be done in the same commit that removes the feature from protocol.

PR #2613, opened 2024-02-07
[Kernel] Refactor SnapshotManager to take logPath and dataPath as part of the constructor

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Follow-up to https://github.com/delta-io/delta/pull/2607#discussion_r1480314605

## How was this patch tested?

Refactor PR existing tests suffice

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
@@ -38,20 +38,26 @@ public class TableImpl implements Table {
         } catch (IOException io) {
             throw new RuntimeException(io);
         }
-        final Path dataPath = new Path(resolvedPath);
-        final Path logPath = new Path(dataPath, "_delta_log");
-
-        return new TableImpl(logPath, dataPath);
+        return new TableImpl(resolvedPath);
     }
 
     private final SnapshotManager snapshotManager;
+    private final String tablePath;
 
-    public TableImpl(Path logPath, Path dataPath) {
+    public TableImpl(String tablePath) {
+        this.tablePath = tablePath;
+        final Path dataPath = new Path(tablePath);
+        final Path logPath = new Path(dataPath, "_delta_log");
         this.snapshotManager = new SnapshotManager(logPath, dataPath);
     }
 
     @Override
     public Snapshot getLatestSnapshot(TableClient tableClient) throws TableNotFoundException {
         return snapshotManager.buildLatestSnapshot(tableClient);
+    }
+
+    @Override
+    public String getPath() {
+        return tablePath;
     }
 }
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
@@ -44,18 +44,14 @@ public class TableImpl implements Table {
         return new TableImpl(logPath, dataPath);
     }
 
-    private final Path logPath;
-    private final Path dataPath;
     private final SnapshotManager snapshotManager;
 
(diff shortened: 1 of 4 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -237,7 +239,7 @@ protected final Optional<List<FileStatus>> listDeltaAndCheckpointFiles(
      * Load the Snapshot for this Delta table at initialization. This method uses the
      * `lastCheckpoint` file as a hint on where to start listing the transaction log directory.
      */
-    private SnapshotImpl getSnapshotAtInit(TableClient tableClient, Path logPath, Path dataPath)
```
allisonport-db, line 240: I thought this internal method made sense to remove from arg list. I'm good with whatever on the rest (would be fine to remove from any and all instance methods).

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
```diff
@@ -44,18 +44,14 @@ public static Table forPath(TableClient tableClient, String path)
         return new TableImpl(logPath, dataPath);
     }
 
-    private final Path logPath;
```
scottsand-db, line 47: Now that we are doing this refactor ... we should also create a `Table::getPath` API. It's silly that we can create a table instance that represents a delta table at a particular path, but cannot ask it what the path is.

I don't _think_ we need a `getLogPath` API yet?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
scottsand-db: We should remove all data path and log path params from _instance_ methods part of this class. They are all for the same table, same data path, same log path

scottsand-db (changes requested):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
allisonport-db: Added getPath API

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
allisonport-db: Updated but seems weird to not have `path` as an argument to methods like `listFrom`. Maybe these should be static methods

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
```diff
@@ -38,24 +38,26 @@ public static Table forPath(TableClient tableClient, String path)
...
-    private final Path dataPath;
     private final SnapshotManager snapshotManager;
+    private final String tablePath;
 
-    public TableImpl(Path logPath, Path dataPath) {
-        this.logPath = logPath;
-        this.dataPath = dataPath;
-        this.snapshotManager = new SnapshotManager();
+    public TableImpl(String tablePath) {
```
scottsand-db, line 47: Why change the input from a `Path` to a String?

allisonport-db: Pretty sure we plan to remove `Path` completely (cc @vkorukanti) so wanted to avoid using it in a new public API

scottsand-db (approved):

Issue #2598, opened 2024-02-02
[Protocol Change Request] Delta Coordinated Commits


prakharjain09:
#### Which Delta project/connector is this regarding?

This will be a PROTOCOL change.

### Overview

Today’s Delta commit protocol relies on the filesystem to provide commit atomicity. This writer-only table feature request is to allow an external commit-coordinator to decide whether a commit succeeded instead of relying on filesystem atomicity, similar to Iceberg's [metastore tables](https://iceberg.apache.org/spec/#metastore-tables). The filesystem remains the source of truth about the actual content of commits, and (unlike Iceberg metastore tables) filesystem-based readers can still access the table directly. This allows us to deal with various limitations of delta mentioned below in Motivation section.

### Motivation

Today’s Delta commit protocol relies on the filesystem to provide commit atomicity, both for durability and mutual exclusion purposes. This is problematic for several reasons:
1. No reliable way to involve a catalog with the commit
1.1. Impossible to keep the catalog and the table state in sync, and the catalog cannot reject commit attempts it wouldn’t like, because it doesn’t even know about them until after they become durable (and visible to readers).
1.2. No clear path to transactions that could span multiple tables and/or involve catalog updates, because filesystem commits cannot be made conditionally or atomically.
2. No way to tie commit ownership to a table
2.1. In general, Delta tables have no way to advertise that they are managed by catalog or LogStore X (at endpoint Y)
2.2. Today a table could be configured to use different log stores in different clusters. Each of the LogStore implementations tries to implement put-if-absent on their own. Since these implementations are not aware of each other, this leads to lost commits and table corruption.
2.3. The logStore setting today is cluster-wide, so you can't safely mix tables with different logStores.
3. No way to do commits over multiple tables atomically
3.1 We need a way where we can commit to multiple tables at once, allowing us to support multi-table transactions.

### Further details

The detail proposal and the required protocol changes are sketched out [in this doc](https://docs.google.com/document/d/1ZzNwKWrMVrNHc9qfhm7CZEn-AXcnD7NnosL0j6l5Iko/edit?usp=sharing).

**At high level**: We propose a new Delta writer table feature, `coordinatedCommits`. The table feature includes the following capabilities:
1. Conforming Delta clients will refuse to attempt filesystem-based commits against a table that enables `coordinateCommits`. 
2. A table that uses `coordinated commits` can include metadata (dictated by the table’s commit coordinator) that would-be writers can use to correctly perform a commit.
[...]

PR #2614, opened 2024-02-08
[Protocol Change] SPEC change proposal for managed-commits

prakharjain09:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [x] Other (PROTOCOL)

## Description

Adds the proposal for spec change for Managed-Commits (see https://github.com/delta-io/delta/issues/2598) to the RFC folder.

## How was this patch tested?

NA

NA

### DIFF
--- a/protocol_rfcs/managed-commits.md
+++ b/protocol_rfcs/managed-commits.md
@@ -152,7 +152,7 @@ interface CommitStore {
    *
    * @return a list of `Commit` which are tracked by commit-owner.
    * 
-   */	
+   */ 
     def getCommits(startVersion: Long): Seq[Commit]
 
   /**
@@ -297,7 +297,6 @@ E.g.
 }
 ```
 
-
 ## Writer Requirements for Managed Commits
 
 When supported and active:
@@ -305,6 +304,7 @@ When supported and active:
 - Writer must do commits via the commit-owner's commit protocol and it must not write the backfilled commit files directly.
 - Writer must not write checkpoints or log compaction files for the commits that are not backfilled.
 - Metadata cleanup must always preserve the newest k >= 1 backfilled commits.
+
 ## Reader Requirements for Managed Commits
 Managed commits is a writer feature. So it doesn't put any restrictions on the reader.
 
--- a/protocol_rfcs/managed-commits.md
+++ b/protocol_rfcs/managed-commits.md
@@ -161,7 +161,7 @@ Even after a commit succeeds, Delta clients can only discover the commit through
 have no way to determine which file in `_delta_log/_commits` directory corresponds to the actual commit `v`.
 
 The commit-owner is responsible to implement an API (defined by the Delta client) that Delta clients can use to retrieve information about un-backfilled commits maintained
-by the commit-owner. Delta clients who are unaware of the commit-owner (or unwilling to talk to it), will fall back to working only with backfilled commits and may encounter stale reads.
+by the commit-owner. Delta clients who are unaware of the commit-owner (or unwilling to talk to it), may not see recent un-backfilled commits and thus may encounter stale reads.
 
 ## Sample Commit Owner API
 
@@ -299,11 +299,9 @@ Suppose the commit-owner is tracking:
 
 Delta clients have two choices to read such tables:
 1. Any Delta client can read such table by listing the `_delta_log` directory and reading the delta/checkpoint/log-compaction files.
-   They cannot use the un-backfilled commits in `_delta_log/_commits` directory as they are not aware of the external commit-owner. Because of this, they may see a stale snapshot
-   of the table if the recent commits are not backfilled.
+   Without contacting the commit owner, they cannot access recent un-backfilled commits in the `_delta_log/_commits` directory, and may construct a stale snapshot.
     - In the above example, such delta implementation will see version 7 as the latest snapshot.
-2. A client that understands managed commits can supplement the file listing result by contacting the commit-owner for information about un-backfilled commits, in order to get the most
-   recent snapshot of the table.
+2. A client can guarantee freshness by additionally requesting the set of recent un-backfilled commits from the commit-owner.
     - In the above example, a delta implementation could get information about versions 0 through 7 from `_delta_log` directory and get information about un-backfilled commits (v8, v9) from the commit-owner.
 
 ## Maintenance operations on managed-commit tables
--- a/protocol_rfcs/managed-commits.md
+++ b/protocol_rfcs/managed-commits.md
@@ -109,9 +109,9 @@ A commit file is a [delta file](#delta-log-entries) that contains the actions wh
 There are two types of commit files:
 1. **Un-backfilled commit files**: These reside in the `_delta_log/_commits` directory.
     - The filename must follow the pattern: `<version>.<uuid>.json`. Here the `uuid` is a random UUID that is generated for each commit and `version` is the version `v` which is being committed, zero-padded to 20 digits.
-    - The commit-owner must track these files until they are backfilled to the `_delta_log` directory.
     - Mere existence of these files does not mean that the file is a _valid_ commit. It might correspond to a failed or in-progress commit.
       The commit-owner is the source of truth around which un-backfilled commits are valid.
+    - The commit-owner must track these files until they are backfilled to the `_delta_log` directory.
 
 2. **Backfilled commit files**: These reside in the `_delta_log` directory.
     - The filename must follow the pattern: `<version>.json`. Here the `version` is the version `v` which is being committed, zero-padded to 20 digits.
--- a/protocol_rfcs/managed-commits.md
+++ b/protocol_rfcs/managed-commits.md
@@ -12,7 +12,7 @@ the filesystem (s3, abfs etc). This allows us to deal with various limitations o
 2. No way to tie commit ownership to a table.
     - In general, Delta tables have no way to advertise that they are managed by catalog or LogStore X (at endpoint Y)
     - There is no central entity that needs to be contacted in order to commit to the table. So if the underlying file system is missing _putIfAbsent_ semantics, then there is no way to ensure that a commit is atomic, which could lead
-       to lost writes when concurrent writers are writing to the table.
+      to lost writes when concurrent writers are writing to the table.
 
 --------
 
@@ -131,33 +131,33 @@ The following is an example of a possible commit-owner API which some Java-based
 ```java
 
 interface CommitStore {
-  /**
-   * Commits the given set of `actions` to the given commit `version`.
-   *
-   * @param version The version we want to commit.
-   * @param actions Actions that need to be committed.
-   *
-   * returns CommitResponse which has details around the new committed delta file.
-   */
-   def commit(
-     version: Long,
-     actions: Iterator[String]): CommitResponse
+    /**
+     * Commits the given set of `actions` to the given commit `version`.
+     *
+     * @param version The version we want to commit.
+     * @param actions Actions that need to be committed.
+     *
+     * returns CommitResponse which has details around the new committed delta file.
+     */
+    def commit(
+        version: Long,
+        actions: Iterator[String]): CommitResponse
 
-  /**
-   * API to get the un-backfilled commits for the table represented by the given `tablePath` after
-   * the given `startVersion`. The returned commits are contiguous and in ascending version order.
-   * Note that the first version returned by this API may not be equal to the `startVersion`. This
-   * happens when few versions starting from `startVersion` are already backfilled and so
-   * CommitStore may have stopped tracking them.
-   *
-   * @return a list of `Commit` which are tracked by commit-owner.
-   * 
-   */ 
+    /**
+     * API to get the un-backfilled commits for the table represented by the given `tablePath` after
+     * the given `startVersion`. The returned commits are contiguous and in ascending version order.
+     * Note that the first version returned by this API may not be equal to the `startVersion`. This
+     * happens when few versions starting from `startVersion` are already backfilled and so
+     * CommitStore may have stopped tracking them.
+     *
+     * @return a list of `Commit` which are tracked by commit-owner.
+     *
+     */
     def getCommits(startVersion: Long): Seq[Commit]
 
-  /**
-   * API to ask the commit-owner to backfill all commits <= given `version`.
-   */
+    /**
+     * API to ask the commit-owner to backfill all commits <= given `version`.
+     */
     def backfillToVersion(version: Long): Unit
 }
 ```
@@ -291,7 +291,7 @@ E.g.
     "configuration":{
       "appendOnly": "true",
       "delta.managedCommitOwnerName": "commit-owner-1",
-      "delta.managedCommitOwnerConf": ""
+      "delta.managedCommitOwnerConf": "{\"batchSize\":\"10\",\"endpoint\":\"http://sample-url.com/commit\"}"
     }
   }
 }
--- a/protocol_rfcs/managed-commits.md
+++ b/protocol_rfcs/managed-commits.md
@@ -71,7 +71,23 @@ An action changes one aspect of the table's state, for example, adding or removi
    startVersion <= `cutOffCheckpoint`'s version.
     - <ins>Also delete all the [un-backfilled commit files](#commit-files) having version <= `cutOffCheckpoint`'s version from the `_delta_log/_commits` directory.</ins>
 
+### Checkpoints
+> ***Change to [existing section](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#checkpoints)***
 
+Checkpoints are also stored in the `_delta_log` directory, and can be created at any time, for any committed version of the table.
+For performance reasons, readers should prefer to use the newest complete checkpoint possible.
+
+<ins>**Note:** If [managed commits](#managed-commits) table feature is enabled on the table, a checkpoint can be created only for commit
+versions which are backfilled. Refer to [maintenance operations on managed-commit tables](#maintenance-operations-on-managed-commit-tables) section
+for more details</ins>
+
+### Log Compaction Files
+> ***Change to [existing section](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#log-compaction-files)***
+
+<ins>**Note:** If [managed commits](#managed-commits) table feature is enabled on the table, a log compaction file for commit
+range `[x, y]` i.e. `x.y.compacted.json` can be created only when commit `y` is already backfilled i.e. `_delta_log/<y>.json` must exist.
+Refer to [maintenance operations on managed-commit tables](#maintenance-operations-on-managed-commit-tables) section
+for more details.</ins>
 
 > ***The next set of sections will be added to the existing spec just before [Iceberg Compatibility V1](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#iceberg-compatibility-v1) section***
 
--- a/protocol_rfcs/managed-commits.md
+++ b/protocol_rfcs/managed-commits.md
@@ -16,9 +16,11 @@ the filesystem (s3, abfs etc). This allows us to deal with various limitations o
 
 --------
 
-> ***Change to existing section***
 
 ### Delta Log Entries
+
+> ***Change to [existing section](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#delta-log-entries)***
+
 Delta files are stored as JSON in a directory at the root of the table named `_delta_log`, and together with checkpoints make up the log of all changes that have occurred to a table. They are the unit of atomicity for a table.
 
 **Note:** If [managed commits](#managed-commits) table feature is enabled on the table, recently committed delta files may reside in the `_delta_log/_commits` directory. Delta clients have to contact
@@ -32,7 +34,7 @@ For example:
 ./_delta_log/00000000000000000000.json
 ```
 
-The delta files in the `_delta_log/_commits` directory on the other hand have a UUID embedded into them and follow the pattern <version>.<uuid>.json where the version corresponds to the next attempt version zero-padded to 20 digits.
+The delta files in the `_delta_log/_commits` directory have a UUID embedded into them and follow the pattern `<version>.<uuid>.json`, where the version corresponds to the next attempt version zero-padded to 20 digits.
 
 For example:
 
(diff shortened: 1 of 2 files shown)


tdas:
Can you update the descriptoin to match the github issue template.

protocol_rfcs/managed-commits.md
```diff
@@ -0,0 +1,315 @@
...
+
+## Maintenance operations on managed-commit tables
+
+Maintenance operations such as [checkpoints](#checkpoints-1), [log compaction](#log-compaction-files), [metadata cleanup](#metadata-cleanup) can be performed only on the backfilled part of a managed-commit table.
+
+- A [checkpoint](#checkpoints-1) for a version `v` can be created only if the commit `v` is already backfilled i.e. `_delta_log/<v>.json` exists.
+- A [log compaction](#log-compaction-files) can be done only on the backfilled part of the table. The log compaction file `<x>.<y>.compacted.json` can be created only if commit `y` is already backfilled i.e. `_delta_log/<y>.json` exists.
+- A [Metadata cleanup](#metadata-cleanup) must always preserve the newest k >= 1 backfilled commits. In other words, [metadata cleanup](#metadata-cleanup) must always choose a backfilled `cutOffCheckpoint`.
+    - This allows the table to remain discoverable by the filesystem-based delta readers even after metadata cleanup deletes everything before commit version `cutOffCheckpoint`.
```
ryan-johnson-databricks, line 270: This seems bulky/redundant... maybe
```suggestion
[Checkpoints](#checkpoints-1) and [log compaction files](#log-compaction-files) can only be created for commits in the `_delta_log` directory. In other words, in order to checkpoint version `v` or produce a compacted log file for commit range x <= v <= y, `_delta_log/<v>.json` must exist. Otherwise, filesystem-based readers who encountered the seemingly-extra files might think the table metadata was corrupted.

[Metadata cleanup](#metadata-cleanup) must not delete the newest commit in the `_delta_log` directory. In other words, if the cutoff checkpoint version is `v`, then `_delta_log/<v>.json` must exist. Otherwise, the table would appear deleted/empty to filesystem-based readers.
```

ryan-johnson-databricks: Actually: I'm not sure we need to say much (anything?) about metadata cleanup, because the newest checkpoint always corresponds to a backfilled commit, and we never delete the newest checkpoint? 

Also -- The current [metadata cleanup](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#metadata-cleanup) spec doesn't say so, but it's implied that metadata cleanup can only proceed if a checkpoint can be found. Maybe we should say that directly?

protocol_rfcs/managed-commits.md
```diff
@@ -0,0 +1,315 @@
...
+
+## Managed Commit Enablement
+
+The managed-commit feature is supported and active when:
+- The table must be on Writer Version 7.
+- The table has a `protocol` action with `writerFeatures` containing the feature `managedCommits`.
+- The table has a metadata property `delta.managedCommitOwnerName` in the [change-metadata](#change-metadata)'s configuration.
+- The table may have a metadata property `delta.managedCommitOwnerConf` in the [change-metadata](#change-metadata)'s configuration. The value of this property is a json-coded string-to-string map.
```
ryan-johnson-databricks, line 278: Should we do some namespacing here, with e.g. `delta.managedCommits.commitOwner` and `delta.managedCommits.commitOwnerConf`?

protocol_rfcs/managed-commits.md
```diff
@@ -0,0 +1,315 @@
...
+{
+  "metaData":{
+    "id":"af23c9d7-fff1-4a5a-a2c8-55c59bd782aa",
+    "format":{"provider":"parquet","options":{}},
+    "schemaString":"...",
+    "partitionColumns":[],
+    "configuration":{
+      "appendOnly": "true",
+      "delta.managedCommitOwnerName": "commit-owner-1",
+      "delta.managedCommitOwnerConf": "{\"batchSize\":\"10\",\"endpoint\":\"http://sample-url.com/commit\"}"
```
ryan-johnson-databricks, line 294:
```suggestion
      "delta.managedCommitOwnerConf": "{\"endpoint\":\"http://sample-url.com/commit\", \"authenticationMode\":\"oauth2\"}"
```
(no idea what "batchSize" might mean, and probably not worth the confusion just for an example)

ryan-johnson-databricks: Also, somewhere we should say that the commit owner conf should never include secrets such as auth tokens or credentials, because any reader with access to the table's storage location can see them.

protocol_rfcs/managed-commits.md
```diff
@@ -0,0 +1,315 @@
...
+    }
+  }
+}
+```
+
+## Writer Requirements for Managed Commits
+
+When supported and active:
+- Require that `inCommitTimestamp` table feature must be supported and active.
```
ryan-johnson-databricks, line 303:
```suggestion
- The `inCommitTimestamp` table feature must also be supported and active.
```
(we should eventually link to that feature's section of the spec -- but I realize that it's not yet in the spec to link to)

protocol_rfcs/managed-commits.md
```diff
@@ -0,0 +1,315 @@
...
+- Writer must do commits via the commit-owner's commit protocol and it must not write the backfilled commit files directly.
```
ryan-johnson-databricks, line 304:
```suggestion
- Writer must follow the commit-owner's commit protocol and must not perform filesystem-based commits.
```

protocol_rfcs/managed-commits.md
```diff
@@ -0,0 +1,315 @@
...
+- Writer must not write checkpoints or log compaction files for the commits that are not backfilled.
```
ryan-johnson-databricks, line 305:
```suggestion
- Writer must only create checkpoints or log compaction files for commits in the `_delta_log` directory.
```

protocol_rfcs/managed-commits.md
```diff
@@ -0,0 +1,315 @@
...
+
+--------
+
+> ***Change to existing section***
+
+### Delta Log Entries
+Delta files are stored as JSON in a directory at the root of the table named `_delta_log`, and together with checkpoints make up the log of all changes that have occurred to a table. They are the unit of atomicity for a table.
+
+**Note:** If [managed commits](#managed-commits) table feature is enabled on the table, recently committed delta files may reside in the `_delta_log/_commits` directory. Delta clients have to contact
+the corresponding commit-owner of the table in order to find the information about the [un-backfilled commits](#commit-backfills).
```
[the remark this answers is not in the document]
prakharjain09: Yes it was creating confusion as we are trying to simulate a diff. So now this just reflects final state of this section. i.e. how it looks now.

protocol_rfcs/managed-commits.md
```diff
@@ -0,0 +1,315 @@
...
+```
+
+The delta files in `_delta_log/_commits` directory may not be committed and should not be trusted. The [commit-owner](#commit-owner) is the source of truth
+for the information around which one of these are valid. Refer to [managed commits](#managed-commits) for more details.
+
+Delta files use new-line delimited JSON format, where every action is stored as a single line JSON document.
+A delta file, corresponding to version `n`, contains an atomic set of [_actions_](#Actions) that should be applied to the previous table state, corresponding to `n-1`, in order to construct the `n`th snapshot of the table.
+An action changes one aspect of the table's state, for example, adding or removing a file.
+
+> ***New Section***
```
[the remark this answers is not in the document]
prakharjain09: Adding reference:

> ***The next set of sections will be added to the existing spec just before [Iceberg Compatibility V1](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#iceberg-compatibility-v1) section***

protocol_rfcs/managed-commits.md
prakharjain09: Added this in `Managed Commit Enablement` section where we describe the table property.

protocol_rfcs/managed-commits.md
prakharjain09: Makes sense. Adding following line to metadata cleanup:

2. Identify the newest checkpoint that is not newer than the `cutOffCommit`. A checkpoint at the `cutOffCommit` is ideal, but an older one will do. Lets call it `cutOffCheckpoint`.
   We need to preserve the `cutOffCheckpoint` and all commits after it, because we need them to enable
   time travel for commits between `cutOffCheckpoint` and the next available checkpoint.
   - **If no checkpoint can be found, do not proceed with metadata cleanup as there is nothing to cleanup.**

Added Metadata Cleanup section also in the RFC.

protocol_rfcs/managed-commits.md
```diff
@@ -0,0 +1,315 @@
...
+understand `managed-commits`. Backfill also allows the commit-owner to reduce the number of commits it must track internally.
+
+Backfill must be sequential. In other words, a commit-owner must ensure that backfill of commit `v-1` is complete before initiating backfill of commit 'v'.
+
+`commit-owner`s are encouraged to backfill the commits frequently. This has several advantages:
+1. Filesystem-based Delta implementations may only understand backfilled commits, and frequent backfill allows them to access the most recent table snapshots.
+2. Frequent backfilling minimizes the chance of data loss in case the `commit-owner` loses its state due to some failover/issue.
+3. Some maintenance operations (such as checkpoints, log compaction, and metadata cleanup) can be performed only on the backfilled part of the table. Refer to the [Maintenance operations on managed-commit tables](#maintenance-operations-on-managed-commit-tables) section for more details.
+
+The commit-owner may also expose an API to backfill the commits. This will allow clients to ask the commit-owner to backfill the commits if needed in order to do some maintenance operations.
```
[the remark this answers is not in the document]
prakharjain09: Rephrased this to:

> The commit-owner also needs to expose an API to backfill the commits. This will allow clients to ask the commit-owner to backfill the commits if needed in order to do some maintenance operations.

protocol_rfcs/managed-commits.md
```diff
@@ -0,0 +1,315 @@
...
+     * API to get the un-backfilled commits for the table represented by the given `tablePath` after
+     * the given `startVersion`. The returned commits are contiguous and in ascending version order.
+     * Note that the first version returned by this API may not be equal to the `startVersion`. This
+     * happens when few versions starting from `startVersion` are already backfilled and so
+     * CommitStore may have stopped tracking them.
+     *
+     * @return a list of `Commit` which are tracked by commit-owner.
+     *
+     */
+    def getCommits(startVersion: Long): Seq[Commit]
```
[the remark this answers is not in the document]
prakharjain09: Yes agreed that it is better to put endVersion in the example API definition. done.

protocol_rfcs/managed-commits.md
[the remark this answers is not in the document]
ryan-johnson-databricks: That makes it hard to review for _changes_ tho... for example, the old text said something about commit files being the "unit of atomicity" for the table, and that's completely gone in this version. But nothing actually identifies that change. 

I did notice that mixing lots of <ins>insertions</ins> and ~deletions~ in the same text is hard to read, and it will often be better to just put full before/after versions, e.g. 

~Delta files are the unit of atomicity for a table, and are named using the next available version number, zero-padded to 20 digits.~

<ins>The delta files in `_delta_log` directory are named using the next available version number, zero-padded to 20 digits.</ins>

But any time we're reviewing markdown changes, it's important to View File, to see how it actually renders instead of trying to guess from the raw text:
image

protocol_rfcs/managed-commits.md
```diff
@@ -0,0 +1,315 @@
...
+the filesystem (s3, abfs etc). This allows us to deal with various limitations of delta:
+
+1. No reliable way for the table's owner to participate in commits.
+    - The table's owner (such as a catalog) cannot reliably stay in sync with the table state, nor reject commit attempts it wouldn’t like, because it doesn’t even know about writes until they are already durable (and visible to readers).
+    - No clear path to transactions that could span multiple tables and/or involve catalog updates, because filesystem commits cannot be made conditionally or atomically.
+2. No way to tie commit ownership to a table.
+    - In general, Delta tables have no way to advertise that they are managed by catalog or LogStore X (at endpoint Y)
+    - There is no central entity that needs to be contacted in order to commit to the table. So if the underlying file system is missing _putIfAbsent_ semantics, then there is no way to ensure that a commit is atomic, which could lead
+      to lost writes when concurrent writers are writing to the table.
```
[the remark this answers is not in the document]
ryan-johnson-databricks: (should we update the wording to reflect that?)

protocol_rfcs/managed-commits.md
```diff
@@ -43,19 +45,44 @@ For example:
...
+   time travel for commits between `cutOffCheckpoint` and the next available checkpoint.
+   - **If no checkpoint can be found, do not proceed with metadata cleanup as there is nothing to cleanup.**
+3. Delete all [delta log entries](#delta-log-entries and [checkpoint files](#checkpoints) before the
+   `cutOffCheckpoint` checkpoint. Also delete all the [log compaction files](#log-compaction-files) having
+   startVersion <= `cutOffCheckpoint`'s version.
+4. Now read all the available [checkpoints](#checkpoints-1) in the _delta_log directory and identify
+   the corresponding [sidecar files](#sidecar-files). These sidecar files need to be protected.
+5. List all the files in `_delta_log/_sidecars` directory, preserve files that are less than a day
+   old (as of midnight UTC), to not break in-progress checkpoints. Also preserve the referenced sidecar files
+   identified in Step-4 above. Delete everything else.
```
ryan-johnson-databricks, line 76: It's probably enough to just show the bullet that changed, e.g.

2. Identify the newest checkpoint that is not newer than the `cutOffCommit`. A checkpoint at the `cutOffCommit` is ideal, but an older one will do. Lets call it `cutOffCheckpoint`.
   We need to preserve the `cutOffCheckpoint` and all commits after it, because we need them to enable
   time travel for commits between `cutOffCheckpoint` and the next available checkpoint.
   - <ins>If no `cutOffCheckpoint` can be found, do not proceed with metadata cleanup as there is nothing to cleanup.</ins>

protocol_rfcs/managed-commits.md
```diff
@@ -144,16 +171,21 @@ interface CommitStore {
         actions: Iterator[String]): CommitResponse
 
     /**
-     * API to get the un-backfilled commits for the table represented by the given `tablePath` after
-     * the given `startVersion`. The returned commits are contiguous and in ascending version order.
+     * API to get the un-backfilled commits for the table represented by the given `tablePath` where
+     * `startVersion` <= version <= endVersion.
+     * If endVersion is -1, then it means that we want to get all the commits starting from `startVersion`
```
ryan-johnson-databricks, line 176: Option[Long] seems better than magic values?

protocol_rfcs/managed-commits.md
```diff
@@ -262,23 +294,21 @@ Delta clients have two choices to read such tables:
...
+- The table has a metadata property `delta.managedCommits.commitOwner` in the [change-metadata](#change-metadata)'s configuration.
+- The table may have a metadata property `delta.managedCommits.commitOwnerConf` in the [change-metadata](#change-metadata)'s configuration. The value of this property is a json-coded string-to-string map.
     - A commit-owner can store additional information (e.g. configuration information such as service endpoints) in this field, for use by the commit-owner client (it is opaque to the Delta client).
+    - This field should never include secrets such as auth tokens or credentials, because any reader with access to the table's storage location can see them.
 
-Note that a table is in invalid state if the change-metadata contains the `delta.managedCommitOwnerName` property but the table does not have the `managedCommits` feature in the `protocol` action or vice versa.
+Note that a table is in invalid state if the change-metadata contains the `delta.managedCommits.commitOwner` property but the table does not have the `managedCommits` feature in the `protocol` action or vice versa.
```
ryan-johnson-databricks, line 311:
```suggestion
Note that a table is in invalid state if the change-metadata contains the `delta.managedCommits.commitOwner` property but the table does not have the `managedCommits` feature in the `protocol` action (or vice versa).
```

protocol_rfcs/managed-commits.md
```diff
@@ -0,0 +1,345 @@
+# Managed Commits
+**Associated Github issue for discussions: https://github.com/delta-io/delta/issues/2598**
+
+This RFC proposes a new table feature `managedCommits` which changes the way Delta Lake manages commits.
```
ryan-johnson-databricks, line 4:
```suggestion
This RFC proposes a new table feature `managedCommits` which changes the way Delta Lake performs commits.
```

protocol_rfcs/managed-commits.md
```diff
@@ -0,0 +1,315 @@
...
+
+Today’s Delta commit protocol relies on the filesystem to provide commit atomicity. This feature request is to allow delta tables which gets commit atomicity using an external commit-owner and not
```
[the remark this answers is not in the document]
ryan-johnson-databricks: No change in latest PR version?

protocol_rfcs/managed-commits.md
```diff
@@ -0,0 +1,345 @@
...
+the filesystem (s3, abfs etc). This allows us to deal with various limitations of delta:
```
ryan-johnson-databricks, line 7: nit: s/delta/Delta/g for cases that refer to the table format (as opposed to "delta files" which are commit .json)

protocol_rfcs/managed-commits.md
```diff
@@ -0,0 +1,345 @@
...
+    def backfillToVersion(version: Long): Unit
+}
+```
+
+## Commit Backfills
+Backfilling is the process of copying the un-backfilled commits i.e. `_delta_log/_commits/<version>.<uuid>.json` to `_delta_log/<version>.json`.
+With the help of backfilling, the [delta files](#delta-log-entries) are visible even to the filesystem based Delta clients that do not
+understand `managed-commits`. Backfill also allows the commit-owner to reduce the number of commits it must track internally.
+
+Backfill must be sequential. In other words, a commit-owner must ensure that backfill of commit `v-1` is complete before initiating backfill of commit 'v'.
```
ryan-johnson-databricks, line 202:
```suggestion
Backfill must be sequential. In other words, a commit-owner must ensure that backfill of commit `v-1` is complete before initiating backfill of commit `v`.
```

protocol_rfcs/managed-commits.md
```diff
@@ -0,0 +1,345 @@
...
+   They cannot use the un-backfilled commits in `_delta_log/_commits` directory as they are not aware of the external commit-owner. Because of this, they may see a stale snapshot
+   of the table if the recent commits are not backfilled.
+    - In the above example, such delta implementation will see version 7 as the latest snapshot.
+2. A client that understands managed commits can supplement the file listing result by contacting the commit-owner for information about un-backfilled commits, in order to get the most
+   recent snapshot of the table.
+    - In the above example, a delta implementation could get information about versions 0 through 7 from _delta_log directory and get information about un-backfilled commits (v8, v9) from the commit-owner.
+
+## Maintenance operations on managed-commit tables
+
+[Checkpoints](#checkpoints-1) and [log compaction files](#log-compaction-files) can only be created for commits in the `_delta_log` directory. In other words, in order to
```
ryan-johnson-databricks, line 297: Should we also add a sentence about this to each of those sections as well? it's kind of an important detail

protocol_rfcs/managed-commits.md
```diff
@@ -0,0 +1,345 @@
...
+{
+  "metaData":{
+    "id":"af23c9d7-fff1-4a5a-a2c8-55c59bd782aa",
+    "format":{"provider":"parquet","options":{}},
+    "schemaString":"...",
+    "partitionColumns":[],
+    "configuration":{
+      "appendOnly": "true",
+      "delta.managedCommits.commitOwner": "commit-owner-1",
+       "delta.managedCommits.commitOwnerConf": "{\"endpoint\":\"http://sample-url.com/commit\", \"authenticationMode\":\"oauth2\"}"
```
ryan-johnson-databricks, line 324: This line is longer than the viewbox github renders for the snippet. Maybe we should manually break the line?
```suggestion
       "delta.managedCommits.commitOwnerConf": 
           "{\"endpoint\":\"http://sample-url.com/commit\", \"authenticationMode\":\"oauth2\"}"
```

protocol_rfcs/managed-commits.md
```diff
@@ -0,0 +1,345 @@
...
+    }
+  }
+}
+```
+
+## Writer Requirements for Managed Commits
+
+When supported and active:
+- The `inCommitTimestamp` table feature must also be supported and active.
```
ryan-johnson-databricks, line 333: I don't think this is explained anywhere... probably worth updating the backfill section to point out that ICT is needed because the backfilled commit will otherwise have a different timestamp than the original?

ryan-johnson-databricks (approved):
Looks good, just a few nits.

protocol_rfcs/managed-commits.md
prakharjain09: Isn't Option more scalaish? I kept it -1 to more reflect it like a pseudo code.
No strong opinion - I can change it to Option if thats more intutive.

protocol_rfcs/managed-commits.md
prakharjain09: Adding following text in `Commit Backfills` section:

Since commit backfills may happen at a later point in time, so the `file modification timestamp` of the backfilled file might be very different than the time of actual commit. For this reason, the `managed-commit` feature depends on another writer feature called [in-commit-timestamps](#TODO-Put-Relevant-Link) to make the commit timestamps more reliable. Refer to [Writer Requirements for Managed Commits](#writer-requirements-for-managed-commits) section for more details.

protocol_rfcs/managed-commits.md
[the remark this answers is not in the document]
prakharjain09: fixed.

protocol_rfcs/managed-commits.md
```diff
@@ -0,0 +1,315 @@
...
+## Commit Owner
+
+A commit-owner is an external entity which manages the commits on a delta table. It could be backed by a database, a file system, or any other storage system. Each commit-owner has its own spec around how they should be contacted and how they do a commit.
+
+## Commit Files
+
+A commit file is a [delta file](#delta-log-entries) that contains the actions that are committed / need to be committed.
+
+There are two types of commit files:
+1. **Un-backfilled commit files**: These reside in the `_delta_log/_commits` directory.
```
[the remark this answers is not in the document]
prakharjain09: The backfilled and un-backfilled terminology seems very explicit and clear as compared to recent commit -- although it is more technical. For now, keeping it to "[un]backfill". Will revisit this when converting RFC to Protocol.

protocol_rfcs/managed-commits.md
```diff
@@ -0,0 +1,315 @@
...
+
+At a high level, the `commit-owner` needs to provide:
+- API to atomically commit a version `x` with given set of `actions`. This is explained in detail in the [commit protocol](#commit-protocol) section.
+- API to retrieve information about the recent commits on the table. This is explained in detail in the [getting un-backfilled commits from commit-owner](#getting-un-backfilled-commits-from-commit-owner) section.
+
+### Commit Protocol
+
+When a `commit-owner` receives a request to commit version `v`, it must first verify that the previous version `v-1` already exists, and that version `v` does not yet exist. It then has following choices to publish the commit:
+1. Write the actions to an 'un-backfilled' [commit file](#commit-files) in the `_delta_log/_commits` directory, and **atomically** record that the new file now corresponds to version `v`.
+2. Atomically write a backfilled [commit file](#commit-files) in the `_delta_log` directory. Note that the commit will be considered to have succeeded as soon as the file becomes visible to
```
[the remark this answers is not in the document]
prakharjain09: `Write the delta file to the _delta_log/_commits directory` , `Atomically write the delta file to the _delta_log directory` implicitly assumes that the file inside _commits directory must have UUID while files inside `_delta_log` won't have UUID. This is true but it might be better to explicitly mention this by using backfill/un-backfilled commit words?

protocol_rfcs/managed-commits.md
prakharjain09: done.

PR #2615, opened 2024-02-08
Support map and arrays in ALTER TABLE CHANGE COLUMN

johanl-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This change addresses an issue where trying to change the key/value of a map or element of an array with ALTER TABLE CHANGE COLUMN would succeed while doing nothing.
 
In addition, a proper error is now thrown when trying to add or drop the key or value of a map or element of an array.

## How was this patch tested?

- Added tests to `DeltaAlterTableTests` to cover changing maps and arrays in ALTER TABLE CHANGE COLUMN.
- Added tests to `SchemaUtilsSuite` and `DeltaDropColumnSuite` to cover the updated error when trying to add/drop map key/value or array element.

Changing the type of the key or value of a map or of the elements of an array now fails if the type change isn't supported (= anything except setting the same type or moving between char, varchar, string):
```
CREATE TABLE table (m map<int, int>) USING DELTA;
ALTER TABLE table CHANGE COLUMN m.key key long;
-- Fails with DELTA_UNSUPPORTED_ALTER_TABLE_CHANGE_COL, previously succeeded while applying no change.
```
Similarly, adding a comment now also fails.

The error when trying to add or drop a map key/value or array element field is updated:
```
CREATE TABLE table (m map<int, int>) USING DELTA;
ALTER TABLE table ADD COLUMN m.key long;
-- Now fails with DELTA_ADD_COLUMN_PARENT_NOT_STRUCT instead of IllegalArgumentException: Don't know where to add the column m.key"
```

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/commands/alterDeltaTableCommands.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/commands/alterDeltaTableCommands.scala
@@ -581,7 +581,7 @@ case class AlterTableChangeColumnDeltaCommand(
           a.copy(elementType =
             SchemaUtils.changeDataType(a.elementType, newColumn.dataType, resolver))
 
-        case (_, other, _) => other
+        case (_, other @ (_: StructType | _: ArrayType | _: MapType), _) => other
       }
 
       // update `partitionColumns` if the changed column is a partition column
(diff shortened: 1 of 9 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/commands/alterDeltaTableCommands.scala
```diff
@@ -561,11 +561,27 @@ case class AlterTableChangeColumnDeltaCommand(
...
+          verifyMapArrayChange(sparkSession, originalField, resolver, txn)
+          m.copy(valueType = SchemaUtils.changeDataType(m.valueType, newColumn.dataType, resolver))
+
+        case (`columnPath`, a: ArrayType, _) if columnName == "element" =>
+          val originalField = StructField(columnName, a.elementType, nullable = a.containsNull)
+          verifyMapArrayChange(sparkSession, originalField, resolver, txn)
+          a.copy(elementType =
+            SchemaUtils.changeDataType(a.elementType, newColumn.dataType, resolver))
+
+        case (_, other, _) => other
```
sabir-akhadov, line 584: should we be more strict here and only match on StructType, MapType and ArrayType?

johanl-db: Yes, I updated it to restrict to struct, map, arrays which are all expected at this point

sabir-akhadov:
nit: typo in description, Similarly, adding a comment now also comment fails.

sabir-akhadov (approved):
lgtm

PR #2617, opened 2024-02-08
[2481] Column mapping removal check for invalid column names

sabir-akhadov:
Add new column mapping removal command basic class hidden behind a feature flag.
Add a check for invalid column names before applying the command.

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## How was this patch tested?
new tests

No


spark/src/main/resources/error/delta-error-classes.json
```diff
@@ -1130,6 +1130,15 @@
     ],
     "sqlState" : "42K05"
   },
+  "DELTA_INVALID_CHARACTERS_IN_COLUMN_NAMES_UPON_SETTING_COLUMN_MAPPING_TO_NONE" : {
```
tomvanbussel, line 1133: Nit: Let's maybe use a slightly error class.

tomvanbussel (approved):
One small nit, LGTM otherwise

spark/src/main/resources/error/delta-error-classes.json
sabir-akhadov: Is `DELTA_INVALID_COLUMN_NAMES_WHEN_REMOVING_COLUMN_MAPPING` better?

tomvanbussel: SGTM

PR #2627, opened 2024-02-09
[Spark] Add managed-commit table feature and CommitStore interface

prakharjain09:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Add Managed Commit table feature and Commit Store interface.

Related issue: https://github.com/delta-io/delta/issues/2598
## How was this patch tested?

UTs

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/managedcommit/CommitStore.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/managedcommit/CommitStore.scala
@@ -1,0 +1,153 @@
+/*
+ * Copyright (2021) The Delta Lake Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.sql.delta.managedcommit
+
+import scala.collection.mutable
+
+import org.apache.spark.sql.delta.{DeltaConfigs, ManagedCommitTableFeature, SerializableFileStatus, SnapshotDescriptor}
+import org.apache.spark.sql.delta.actions.{Action, CommitInfo, Metadata, Protocol}
+import org.apache.spark.sql.delta.storage.LogStore
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.{FileStatus, FileSystem, Path}
+
+import org.apache.spark.internal.Logging
+
+
+/** Representation of a commit file */
+case class Commit(
+  version: Long,
+  serializableFileStatus: SerializableFileStatus,
+  commitTimestamp: Long)
+
+
+/**
+ * Exception raised by [[CommitStore.commit]] method.
+ *  | retryable | conflict  | meaning                                                         |
+ *  |   no      |   no      | something bad happened (e.g. auth failure)                      |
+ *  |   no      |   yes     | permanent transaction conflict (e.g. multi-table commit failed) |
+ *  |   yes     |   no      | transient error (e.g. network hiccup)                           |
+ *  |   yes     |   yes     | physical conflict (allowed to rebase and retry)                 |
+ */
+class CommitFailedException(
+    retryable: Boolean, conflict: Boolean, message: String) extends Exception(message)
+
+/** Response container for [[CommitStore.commit]] API */
+case class CommitResponse(commit: Commit)
+
+/** A container class to inform the CommitStore about any changes in Protocol/Metadata */
+case class UpdatedActions(
+  commitInfo: CommitInfo,
+  newMetadata: Option[Metadata],
+  newProtocol: Option[Protocol])
+
+/**
+ * CommitStore is responsible for managing commits for a managed-commit delta table.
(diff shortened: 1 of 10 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/managedcommit/CommitStore.scala
```diff
@@ -0,0 +1,153 @@
...
+    hadoopConf: Configuration,
+    tablePath: Path,
+    commitVersion: Long,
+    actions: Iterator[String],
+    updatedActions: UpdatedActions): CommitResponse
+
+  /**
+   * API to get the un-backfilled commits for the table represented by the given `tablePath` after
+   * the given `startVersion`. The returned commits are contiguous and in ascending version order.
+   * version.
```
ryan-johnson-databricks, line 82:
```suggestion
   * API to get the un-backfilled commits for the table represented by the given `tablePath`. Commits
   * older than `startVersion`, or newer than `endVersion` (if given), are ignored. The returned
   * commits are contiguous and in ascending version order.
```

spark/src/main/scala/org/apache/spark/sql/delta/util/FileNames.scala
```diff
@@ -18,24 +18,34 @@ package org.apache.spark.sql.delta.util
...
+  private val checksumFilePattern = checksumFileRegex.pattern
+  private val checkpointFilePattern = checkpointFileRegex.pattern
 
   /** Returns the delta (json format) path for a given delta file. */
   def deltaFile(path: Path, version: Long): Path = new Path(path, f"$version%020d.json")
 
+  /** Returns the delta (json format) path for a given delta file. */
+  def uuidDeltaFile(logPath: Path, version: Long): Path = {
+    val basePath = commitDirPath(logPath)
+    val uuid = UUID.randomUUID.toString
```
ryan-johnson-databricks, line 45:
```suggestion
    val uuid = UUID.randomUUID
```
(`$uuid` will `toString` automatically)

spark/src/main/scala/org/apache/spark/sql/delta/util/FileNames.scala
```diff
@@ -158,7 +168,14 @@ object FileNames {
...
     def unapply(path: Path): Option[(Path, Long)] = {
-      deltaFileRegex.unapplySeq(path.getName).map(path -> _.head.toLong)
+      val parentDirName = path.getParent.getName
+      parentDirName match {
+        case COMMIT_SUBDIR =>
+          // If parent is _commits dir, then match against uuid commit file.
+          uuidDeltaFileRegex.unapplySeq(path.getName).map(path -> _.head.toLong)
+        case _ =>
+          deltaFileRegex.unapplySeq(path.getName).map(path -> _.head.toLong)
+      }
```
ryan-johnson-databricks, line 178:
```suggestion
      // If parent is _commits dir, then match against uuid commit file.
      val regex = if (parentDirName == COMMIT_SUBDIR) uuidDeltaFileRegex else deltaFileRegex
      regex.unapplySeq(path.getName).map(path -> _.head.toLong)
```

spark/src/test/scala/org/apache/spark/sql/delta/S3SingleDriverLogStoreSuite.scala
```diff
@@ -112,7 +114,9 @@ trait S3SingleDriverLogStoreSuiteBase extends LogStoreSuiteBase {
   }
 
   test("cache works correctly when writing an initial log version") {
-    withTempDir { dir =>
+    withTempDir { rootDir =>
+      val dir = new File(rootDir, "_delta_log")
+      dir.mkdir()
```
ryan-johnson-databricks, line 119: Yikes... this test was creating Delta metadata files directly in the table dir?

ryan-johnson-databricks (approved):
LGTM. The RFC needs to merge before we merge code tho.

spark/src/main/scala/org/apache/spark/sql/delta/managedcommit/CommitStore.scala
```diff
@@ -0,0 +1,153 @@
...
+  def build(conf: Map[String, String]): CommitStore
+}
+
+/** Factory to get the correct CommitStore for a table */
+object CommitStoreProvider {
+  // mapping from different commit-owner names to the corresponding [[CommitStoreBuilder]]s.
+  private val nameToBuilderMapping = mutable.Map.empty[String, CommitStoreBuilder]
+
+  /** Returns a [[CommitStore]] for the given `name` and `conf` */
+  def getCommitStore(name: String, conf: Map[String, String]): CommitStore = {
```
ryan-johnson-databricks, line 111: We should probably add `synchronized` to this and other methods that access the build map? 
(only the "clear" method has it right now)

ryan-johnson-databricks (approved):
Looks good, pending two last items to address.

spark/src/main/scala/org/apache/spark/sql/delta/managedcommit/CommitStore.scala
prakharjain09: Adding `syncronized` to registerBuilder. Do you see a need to syncronize on reader? i.e. getCommitStore method?

ryan-johnson-databricks: The reader definitely needs `synchronized` as well. It's not physically safe to read a normal `Map` that could be modified concurrently by other threads, because internal data structures could be(come) inconsistent during the read.

ryan-johnson-databricks: (granted, we don't expect any concurrency in the common case... but `synchronized` is ~free in single-thread env and saves us from obscure errors if multi-thread races ever happen)

Issue #1905, opened 2023-07-12
Inline deletion vector


sclmn:
Hi,

When taking a look at the protocol, there is an example for an inline deletion vector:
"wi5b=000010000siXQKl0rr91000f55c8Xg0@@D72lkbi5=-{L"

When looking at it, the deletion vector doesn't seem to have the outlined magic number (1681511376 vs the protocol which states it should be 1681511377) and has a different format. I'm wondering if the deletion vector is valid and if so how it can be decoded?

Thanks!

scottsand-db:
cc @schuermannator

foss-contributor:
@sclmn, thank you for your contribution! This issue has been inactive for over 18 months and will be closed in one week. To keep it open, please leave a comment on this issue or contact us in the `#contributors` channel on [Delta Lake Slack](https://go.delta.io/slack). We appreciate your involvement in Delta Lake!

felipepessoto:
Inline example is still using invalid value

PR #2628, opened 2024-02-12
Update PROTOCOL to replace incorrect Deletion Vector example

alexr-bq:
- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [x] Other (Spec Doc)

## Description

The current example inlined Deletion Vector in PROTOCOL.md does not work. When trying to read a table with this deletion vector in spark, this is the error:

24/02/12 09:34:59 WARN TaskSetManager: Lost task 2.0 in stage 11.0 (TID 159) (10.0.0.160 executor driver): TaskKilled (Stage cancelled: Job aborted due to stage failure: Task 0 in stage 11.0 failed 1 times, most recent failure: Lost task 0.0 in stage 11.0 (TID 157) (10.0.0.160 executor driver): java.io.IOException: Unexpected RoaringBitmapArray magic number -791463580

This PR replaces it with an inlined deletion vector that does work.

## How was this patch tested?

Tested loading the example DV in spark and in some rust code using the roaring crate.

Documentation.

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -883,7 +883,7 @@ Assuming that this DV is stored relative to an `s3://mytable/` directory, the ab
 ```json
 {
   "storageType" : "i",
-  "pathOrInlineDv" : "^Bg9^0rr910000000000iXQKl0rr91000935c8X    g000621onSa",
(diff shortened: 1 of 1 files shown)


felipepessoto:
Related: #1905, #2050

PR #2408, opened 2023-12-26
Update AWS SDK Java 1.x to AWS SDK Java 2.x dependency

Aakansha99:
#### Which Delta project/connector is this regarding?

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [x] Storage-s3-dynamodb
- [ ] Other (fill in here)

## Description

The AWS SDK for Java 1.x is being deprecated will enter maintenance mode. To address the package deprecation, we’ll need to upgrade AWS SDK Java 1.x in delta to AWS SDK Java 2.x.

Delta is undergoing an upgrade from AWS SDK Java version 1.x to version 2.x, aiming to leverage the enhanced features and improvements introduced in the latest SDK. The transition involves significant changes in the codebase, including the adoption of a builder pattern for constructing requests, a more functional programming approach utilizing lambda expressions, adjustments in response handling, and a new exception handling hierarchy (SdkServiceException). These changes contribute to improved code readability, flexibility, and compatibility with the latest AWS SDK practices.

## How was this patch tested?

Testing:
1. Unit test - SUCCEEDED
2. Manual test :

#### Step 1: Download the Required JAR file and Execute Spark SQL
```
aws s3 cp s3://jainaakz-emr-blr-dev/storage-s3-dynamodb/delta-storage-s3-dynamodb-3.1.0-SNAPSHOT.jar .

spark-sql --packages io.delta:delta-spark_2.12:3.0.0 --conf spark.delta.logStore.s3a.impl=io.delta.storage.S3DynamoDBLogStore --conf spark.io.delta.storage.S3DynamoDBLogStore.ddb.tableName=delta_log1 --conf spark.io.delta.storage.S3DynamoDBLogStore.ddb.region=us-east-1 --conf spark.sql.extensions=io.delta.sql.DeltaSparkSessionExtension --conf spark.sql.catalog.spark_catalog=org.apache.spark.sql.delta.catalog.DeltaCatalog --jars /home/hadoop/delta-storage-s3-dynamodb-3.1.0-SNAPSHOT.jar
```

#### Step-2: Create a Delta Table and Insert Initial Data

```

spark-sql (default)> CREATE TABLE oss_delta_test_db.my_delta_table11 (
                   >   id INT,
                   >   value INT
                   > ) USING delta;

spark-sql (default)> INSERT INTO oss_delta_test_db.my_delta_table11
                   > VALUES
                   >   (1, 100),
                   >   (2, 200),
                   >   (3, 300),
[...]

PR #2644, opened 2024-02-16
Extend S3DynamoDbLogStore to allow for use with localstack

rtyler:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This extends #2408 to allow for subclasses to override `getClient()`, but also allows `AWS_ENDPOINT_URL` overrides for the DynamoDb client. We require this in delta-rs so that we can run integration tests of S3DynamoDbLogStore with a localstack container

## How was this patch tested?

:shrug: I mean I ran some `sbt` targets :smile:


build.sbt
```diff
@@ -329,7 +330,7 @@ lazy val storageS3DynamoDB = (project in file("storage-s3-dynamodb"))
     // Test / publishArtifact := true,
 
     libraryDependencies ++= Seq(
-      "com.amazonaws" % "aws-java-sdk" % "1.12.262" % "provided",
+      "software.amazon.awssdk" % "dynamodb" % "2.18.40",
```
scottsand-db, line 333: why this change?

scottsand-db (changes requested):
Few questiosn about the SDK change.

Next, if the goal is "We require this in delta-rs so that we can run integration tests of S3DynamoDbLogStore with a localstack container" ... can't you test this with delta-rs? `build/sbt storageS3DynamoDB/publishM2` and test it with delta-rs?

rtyler:
> Few questiosn about the SDK change.

The SDK changes are from #2408 which might better address the question. Aside from the [v1 of the SDK being deprecated this year](https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-java-v1-x-on-december-31-2025/), there are newer APIs which only exist in version 2 that I require.

> Next, if the goal is "We require this in delta-rs so that we can run integration tests of S3DynamoDbLogStore with a localstack container" ... can't you test this with delta-rs? `build/sbt storageS3DynamoDB/publishM2` and test it with delta-rs?

v1 of the SDK doesn't allow for clean overriding of the DynamoDB endpoint URL the way that v2 does. I don't know if you're suggesting that we fork Delta/Spark just to do integration testing in our CI for delta-rs, or suggesting a one-time integration test here.  :confused: 

Anywho, I see this change (in some form, I'm dispassionate about my PR specifically) would also allow Delta/Spark to actually run [integration tests like this](https://github.com/delta-io/delta/blob/98fac5728f8efb297d3e68c1c6301871bfc93095/storage-s3-dynamodb/integration_tests/dynamodb_logstore.py#L185) in CI, since right now I don't _believe_ there is any automated acceptance/integration testing from change to change of `S3DynamoDbLogStore`

stantonk:
➕ to @rtyler 's PR here, since AWS's Java SDK doesn't support *any* of the other mechanisms for overriding the AWS endpoint (environment variables, system properties, etc. are all *still* not supported :( https://docs.aws.amazon.com/sdkref/latest/guide/feature-ss-endpoints.html), the only way to do integration tests with something like localstack is to allow overriding the `getClient()` method here.

stantonk:
@scottsand-db @rtyler FWIW, this PR has finally been approved which should add support for overriding the endpoint_url via environment variables! https://github.com/aws/aws-sdk-java-v2/pull/5562

PR #2624, opened 2024-02-09
[Protocol Change Request] Type Widening table feature

johanl-db:
### Description of the protocol change

This protocol change is part of the proposed Type Widening table feature, see feature request: https://github.com/delta-io/delta/issues/2622

Protocol RFC issue: https://github.com/delta-io/delta/issues/2623

The type widening table feature covers changing the type of existing columns or nested fields in a Delta table without having to rewrite the data.

### Willingness to contribute

The Delta Lake Community encourages protocol innovations. Would you or another member of your organization be willing to contribute this feature to the Delta Lake code base?

- [x] Yes. I can contribute.
- [ ] Yes. I would be willing to contribute with guidance from the Delta Lake community.
- [ ] No. I cannot contribute at this time.

Issue #2622, opened 2024-02-09
[Feature Request] Widen the type of existing columns or fields without rewriting the table


johanl-db:
#### Which Delta project/connector is this regarding?

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

### Overview
Delta tables currently lack the ability to change the type of a column or nested field after the table was created. Changing type then currently requires copying the while table over, whether that's by actually creating a copy of the table or doing the copy to the new type in place using e.g. column mapping.

This feature request targets specifically widening type changes. In the case of widening change, we are guaranteed that all values present in files that were written before the type change can be promoted to the new, wider type without the risk of overflow or precision loss.
In particular, the following type changes can be supported:
- byte -> short -> int -> long
- float -> double
- decimal precision/scale increase as long as the precision increases by at least as much as the scale to avoid loss of precision/overflow
- date -> timestamp_ntz - dates don't have a timezone and can only be promoted to timestamp without timezone unambiguously.

### Motivation
The type of a column or field is mostly fixed once the table has been created: we only allow setting a column or field to nullable.
The type of a column can become too narrow to store the required values in the lifetime of a table, for example:
- IDs stored in an integer column exceed 31 bits that the type can hold, and the column type needs to be extended to Long.
- A decimal column was initially created with a given precision and new data with a higher precision needs to be ingested.

The only way to do this today is to manually rewrite the table to add a new column with the type wanted and copy the data to the new column. This can be expensive for large tables that must be rewritten and will conflict with every concurrent operation.

### Further details
Currently writing a design doc, I'll share it when the draft is ready.

### Willingness to contribute

The Delta Lake Community encourages new feature contributions. Would you or another member of your organization be willing to contribute an implementation of this feature?

- [x] Yes. I can contribute this feature independently.
- [ ] Yes. I would be willing to contribute this feature with guidance from the Delta Lake community.
- [ ] No. I cannot contribute this feature at this time.

johanl-db:
Type widening is now fully supported in Delta Spark 4.0, the RFC was merged in the main protocol: https://github.com/delta-io/delta/commit/181226a267afe66028d18075ca8e951041e722aa

PR #2645, opened 2024-02-16
Type Widening in ALTER TABLE CHANGE COLUMN

johanl-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
This change introduces the `typeWidening` delta table feature, allowing to widen the type of existing columns and fields in a delta table using the `ALTER TABLE CHANGE COLUMN TYPE` or `ALTER TABLE REPLACE COLUMNS` commands.

The table feature is introduced as `typeWidening-dev` during implementation and is available in testing only.

For now, only byte -> short -> int are supported. Other changes will require support in the Spark parquet reader that will be introduced in Spark 4.0

## How was this patch tested?
A new test suite `DeltaTypeWideningSuite` is created, containing:
- `DeltaTypeWideningAlterTableTests`: Covers applying supported and unsupported type changes on partitioned columns, non-partitioned columns and nested fields
- `DeltaTypeWideningTableFeatureTests`: Covers adding the `typeWidening` table feature

The type widening table feature will introduce the following changes:
- Adding the `typeWidening` via a table property:
```
ALTER TABLE t SET TBLPROPERTIES (‘delta.enableTypeWidening' = true)
```
- Apply a widening type change:
```
ALTER TABLE t CHANGE COLUMN int_col TYPE long
```

Note: both ALTER TABLE commands reuse the existing syntax for setting a table property and applying a type change, no new SQL syntax is being introduced by this feature.


sabir-akhadov (approved):
lgtm

spark/src/main/scala/org/apache/spark/sql/delta/TypeWidening.scala
```diff
@@ -0,0 +1,59 @@
...
+   * Returns whether Type Widening is enabled on this table version. Checks that Type Widening is
+   * supported, which is a pre-requisite for enabling Type Widening, throws an error if
+   * not. When Type Widening is enabled, the type of existing columns or fields can be widened
+   * using ALTER TABLE CHANGE COLUMN.
+   */
+  def isEnabled(protocol: Protocol, metadata: Metadata): Boolean = {
+    val isEnabled = DeltaConfigs.ENABLE_TYPE_WIDENING.fromMetaData(metadata)
+    if (isEnabled && !isSupported(protocol)) {
+      throw new IllegalStateException(
+        s"Table property '${DeltaConfigs.ENABLE_TYPE_WIDENING.key}' is " +
```
bart-samwel, line 41: I guess this should be using the error framework?

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
```diff
@@ -0,0 +1,371 @@
...
+trait DeltaTypeWideningAlterTableTests extends QueryErrorsBase {
+  self: QueryTest with ParquetTest with DeltaTypeWideningTestMixin with DeltaDMLTestUtils =>
+
+  import testImplicits._
+
+  /**
+   * Represents the input of a type change test.
+   * @param fromType         The original type of the column 'value' in the test table.
+   * @param toType           The type to use when changing the type of column 'value'
+   * @param initialValues    The initial values to insert in column 'value after table creation,
```
bart-samwel, line 69:
```suggestion
   * @param initialValues    The initial values to insert in column 'value' after table creation,
```

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
```diff
@@ -0,0 +1,371 @@
...
+    def expectedResult: DataFrame =
+      initialValuesDF.union(additionalValuesDF).select($"value".cast(toType))
+  }
+
+  // Type changes that are supported by all Parquet readers. Byte, Short, Int are all stored as
+  // INT32 in parquet so these changes are guaranteed to be supported.
+  private val supportedTestCases: Seq[TypeEvolutionTestCase] = Seq(
+    TypeEvolutionTestCase(ByteType, ShortType,
+      Seq("1", "2", Byte.MinValue.toString),
+      Seq("4", "5", Int.MaxValue.toString)),
```
bart-samwel, line 94: `Short`?

The "after" cast would just return NULL, I guess. How about setting config ANSI_ENABLED for this test, to avoid using values that don't actually fit?

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
```diff
@@ -0,0 +1,371 @@
...
+      additionalValues.toDF("value").select($"value".cast(toType))
+
+    def expectedResult: DataFrame =
+      initialValuesDF.union(additionalValuesDF).select($"value".cast(toType))
+  }
+
+  // Type changes that are supported by all Parquet readers. Byte, Short, Int are all stored as
+  // INT32 in parquet so these changes are guaranteed to be supported.
+  private val supportedTestCases: Seq[TypeEvolutionTestCase] = Seq(
+    TypeEvolutionTestCase(ByteType, ShortType,
```
bart-samwel, line 92: Do we want to include some explicit tests for NULLs as well? Of course they should trivially widen, but their presence could interfere with widening of other values, e.g. they could cause incorrect widening of surrounding values due to mismatches in counting / not counting nulls in some counters or something like that.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
```diff
@@ -0,0 +1,371 @@
...
+      Seq("1", "2", Byte.MinValue.toString),
+      Seq("4", "5", Int.MaxValue.toString)),
+    TypeEvolutionTestCase(ByteType, IntegerType,
+      Seq("1", "2", Byte.MinValue.toString),
```
bart-samwel, line 96: Shall we do both min and max? And maybe include -1 as standard as well, for an additional sign extension test.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
```diff
@@ -0,0 +1,371 @@
...
+      Seq("4", "5", Int.MaxValue.toString)),
+    TypeEvolutionTestCase(ShortType, IntegerType,
+      Seq("1", "2", Byte.MinValue.toString),
```
bart-samwel, line 99:
```suggestion
      Seq("1", "2", Short.MinValue.toString),
```

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
```diff
@@ -0,0 +1,371 @@
...
+      withAllParquetReaders {
+        checkAnswer(readDeltaTable(tempPath).select("value"), testCase.expectedResult)
+      }
+    }
+  }
+
+  // Test type changes that aren't supported.
+  private val unsupportedNonTestCases: Seq[TypeEvolutionTestCase] = Seq(
+    TypeEvolutionTestCase(IntegerType, ByteType,
+      Seq("1", "2", Int.MinValue.toString)),
```
bart-samwel, line 133: It looks like the optionality of the second sequence parameter in TypeEvolutionTestCase is only to support the negative case. It might make sense to turn this into a separate "negative case" class, because the optional parameter allowed a missing second sequence parameter above.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
```diff
@@ -0,0 +1,371 @@
...
+        "named_struct('a', cast(a as byte)) as s",
+        "map(cast(b as byte), cast(c as short)) as m",
+        "array(cast(d as short)) as a"))
+
+    assert(readDeltaTable(tempPath).schema === new StructType()
+      .add("s", new StructType().add("a", ByteType))
+      .add("m", MapType(ByteType, ShortType))
+      .add("a", ArrayType(ShortType)))
+
+    sql(s"ALTER TABLE delta.`$tempPath` CHANGE COLUMN s.a TYPE short")
```
bart-samwel, line 214: Can you also change the type of a STRUCT column as a whole, e.g. CHANGE COLUMN s TYPE STRUCT<...> etc.? We should make sure that it either doesn't work, or we support it correctly if that induces type widenings.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
```diff
@@ -0,0 +1,371 @@
...
+      readDeltaTable(tempPath),
+      Seq((1, 2, 3, 4), (5, 6, 7, 8))
+        .toDF("a", "b", "c", "d")
+        .selectExpr("named_struct('a', cast(a as short)) as s", "map(b, c) as m", "array(d) as a"))
+  }
+
+  test("type widening using ALTER TABLE REPLACE COLUMNS") {
+    append(Seq(1, 2).toDF("value").select($"value".cast(ShortType)))
+    assert(readDeltaTable(tempPath).schema === new StructType().add("value", ShortType))
+    sql(s"ALTER TABLE delta.`$tempPath` REPLACE COLUMNS (value INT)")
```
bart-samwel, line 238: Same here -- what if we're replacing an entire STRUCT column and widening one of the fields inside the struct? Or a MAP or an ARRAY?

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
```diff
@@ -0,0 +1,371 @@
...
+    append(Seq(3, 4).toDF("value"))
+    checkAnswer(readDeltaTable(tempPath), Seq(Row(1), Row(2), Row(3), Row(4)))
+  }
+
+  test("row group skipping Short -> Int") {
+    withSQLConf(
+      SQLConf.FILES_MAX_PARTITION_BYTES.key -> 1024.toString) {
+      append((0 until 1024).toDF("value").select($"value".cast(ShortType)))
+      sql(s"ALTER TABLE delta.`$tempPath` CHANGE COLUMN value TYPE INT")
+      append((Short.MinValue + 1 until Short.MaxValue + 2048).toDF("value"))
```
bart-samwel, line 250: Why not - 1 so that we overstep the minimum bound as well?

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
```diff
@@ -0,0 +1,371 @@
...
+    sql(s"ALTER TABLE delta.`$tempPath` REPLACE COLUMNS (value INT)")
+    assert(readDeltaTable(tempPath).schema === new StructType().add("value", IntegerType))
+    checkAnswer(readDeltaTable(tempPath), Seq(Row(1), Row(2)))
+    append(Seq(3, 4).toDF("value"))
+    checkAnswer(readDeltaTable(tempPath), Seq(Row(1), Row(2), Row(3), Row(4)))
+  }
+
+  test("row group skipping Short -> Int") {
+    withSQLConf(
+      SQLConf.FILES_MAX_PARTITION_BYTES.key -> 1024.toString) {
```
bart-samwel, line 247: How do we verify that this actually leads to row group skipping? FWIW it looks like this should lead to regular min/max data skipping on the Delta log. The max partition bytes only affects how we read chunks, i.e., it'll split a parquet file of >1024 bytes over multiple tasks (1024 bytes each), but if the file is still a single row group, then only one of the tasks will actually read it (the task that gets the "middle" byte of the row group), and the other files will simply ignore the row group. And with the current way of writing, each file will have only one row group, so the row group skipping will be equal to the Delta-level file skipping, so it should never skip, really.

bart-samwel (commented):
Nice! I left a bunch of test feedback.

spark/src/main/scala/org/apache/spark/sql/delta/TypeWidening.scala
johanl-db: This should never happen unless there's a bug in the implementation, so I wouldn't give it an error class. We typically wouldn't want to document that error as a user-facing error

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
johanl-db: You can't use `CHANGE COLUMN` on anything but a leaf field, I added a test to cover it (I believe it's already covered in `AlterTableTests` but it makes it to cover it here)

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
johanl-db: Yes, I added nulls

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
johanl-db: Good catch, switched to running with ANSI enabled

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
johanl-db: 👍 I added more values to test

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
johanl-db: Good idea, I split the test case classes in two

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
johanl-db: Yes, that's actually a pretty important test case since replacing top-level columns is the only way you can change nested type with REPLACE COLUMNS. I added a couple of tests for this

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
johanl-db: This test doesn't make a lot of sense here anymore, I added it back when I found the overflow issue in parquet row group skipping and kept it around but it's redundant with the test I added when I fixed the issue in Spark: 
https://github.com/databricks/runtime/blob/a9d6a9d6191264ced1f0658ff6675b0f30e8e77f/sql/core/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetQuerySuite.scala#L1216

I'm removing it, I don't think it's worth the effort to get it to accurately cover an issue that the Spark test already covers a lot better

spark/src/main/scala/org/apache/spark/sql/delta/TableFeature.scala
```diff
@@ -625,6 +626,18 @@ object ManagedCommitTableFeature
   }
 }
 
+object TypeWideningTableFeature extends ReaderWriterFeature(name = "typeWidening-dev")
```
tdas, line 629: since we are using -dev right now and its not ready for users using it. is it behind the isTesting flag?

tdas (approved):
LGTM.

spark/src/main/scala/org/apache/spark/sql/delta/TableFeature.scala
johanl-db: Yes, it is: https://github.com/databricks/runtime/pull/83562/files#diff-7b78502e6bca8772b4197ce6652fdfa72c4211c15b49e3b03ddfef7199cfcf0aR353

PR #2059, opened 2023-09-15
Enable Row Tracking outside of testing

johanl-db:
#### Which Delta project/connector is this regarding?

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Support for writing fresh row IDs / row commit versions was introduced in the following PRs:
- https://github.com/delta-io/delta/pull/1723
- https://github.com/delta-io/delta/pull/1781
- https://github.com/delta-io/delta/pull/1896

This is sufficient to enable row tracking on a table and write to a table that has row tracking enabled but not to actually read row IDs / row commit versions back, which will require changes in Spark that will be released with Spark 3.5 (see e.g. https://github.com/apache/spark/pull/40677 and https://github.com/apache/spark/pull/40545)

Using row tracking is currently only allowed in testing, this change allows enabling row tracking outside of testing so that the upcoming Delta 3.0 release includes support for writing to tables with row tracking enabled, making Delta writers future-proof.

## How was this patch tested?
Tests have already been added in previous changes, this only flips the switch to let users enabled Row Tracking outside of tests

## Does this PR introduce _any_ user-facing changes?
Users are now able to enable Row Tracking when creating a delta table:
```
CREATE TABLE tbl(a int) USING DELTA TBLPROPERTIES ('delta.enableRowTracking' = 'true')
```

PR #2647, opened 2024-02-16
[Spark][TEST-ONLY] Add tests to check the behaviour of CONVERT TO DELTA with row IDs.

longvu-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Add tests to check the behaviour of the different combinations of CONVERT TO DELTA with row IDs.
We want to make sure Row Tracking is well tested before [Enabling Row Tracking outside of Testing](https://github.com/delta-io/delta/pull/2059).

## How was this patch tested?
New test suite added.

No.


johanl-db (approved):
LGTM!

PR #2650, opened 2024-02-17
[Spark] Implement incremental clustering using ZCUBE approach

dabao521:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Implement incremental Liquid clustering according to the deisgn [doc](https://docs.google.com/document/d/1FWR3odjOw4v4-hjFy_hVaNdxHVs4WuK1asfB6M6XEMw/edit?usp=sharing).

This implementation uses ZCube based approach to achieve incremental clustering. When a Zcube size is big enough, the zcube is sealed and the next clustering won't re-cluster those files and so less write amplification.

Key changes
Each clustered file is tagged with ZCUBE_ID to track which ZCUBE it belongs to and the id is generated using UUID. Also anther tag ZCUBE_ZORDER_BY is used to track the clustering columns. Each clustered file has the clsuteringProvider populated with liquid.

## How was this patch tested?

new unit tests.
No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala
@@ -225,6 +225,15 @@ class OptimizeExecutor(
     optimizeContext: DeltaOptimizeContext)
   extends DeltaCommand with SQLMetricsReporting with Serializable {
 
+  /**
+   * In which mode the Optimize command is running. There are three valid modes:
+   * 1. Compaction
+   * 2. ZOrder
+   * 3. Clustering
+   */
+  private val optimizeStrategy =
+    OptimizeTableStrategy(sparkSession, txn.snapshot, optimizeContext, zOrderByColumns)
+
   /** Timestamp to use in [[FileAction]] */
   private val operationTimestamp = new SystemClock().getTimeMillis()
 
@@ -239,15 +248,6 @@ class OptimizeExecutor(
       ClusteringColumnInfo.extractLogicalNames(txn.snapshot)
     } else {
       Nil
-    }
-  }
-
-  private lazy val curve: String = {
-    if (zOrderByColumns.nonEmpty) {
-      "zorder"
-    } else {
-      assert(isClusteredTable)
-      "hilbert"
     }
   }
 
@@ -269,7 +269,7 @@ class OptimizeExecutor(
       }
       val partitionsToCompact = filesToProcess.groupBy(_.partitionValues).toSeq
 
-      val jobs = groupFilesIntoBins(partitionsToCompact, maxFileSize)
+      val jobs = groupFilesIntoBins(partitionsToCompact)
 
       val maxThreads =
         sparkSession.sessionState.conf.getConf(DeltaSQLConf.DELTA_OPTIMIZE_MAX_THREADS)
@@ -320,18 +320,7 @@ class OptimizeExecutor(
           numDeletionVectorRowsRemoved = removedDVs.map(_.cardinality).sum))
       }
 
-      if (isMultiDimClustering) {
-        val inputFileStats =
-          ZOrderFileStats(removedFiles.size, removedFiles.map(_.size.getOrElse(0L)).sum)
-        optimizeStats.zOrderStats = Some(ZOrderStats(
-          strategyName = "all", // means process all files in a partition
-          inputCubeFiles = ZOrderFileStats(0, 0),
-          inputOtherFiles = inputFileStats,
-          inputNumCubes = 0,
-          mergedFiles = inputFileStats,
-          // There will one z-cube for each partition
-          numOutputCubes = optimizeStats.numPartitionsOptimized))
-      }
+      optimizeStrategy.updateOptimizeStats(optimizeStats, removedFiles, jobs)
 
       return Seq(Row(txn.deltaLog.dataPath.toString, optimizeStats.toOptimizeMetrics))
     }
@@ -365,13 +354,13 @@ class OptimizeExecutor(
    *
    * @param partitionsToCompact List of files to compact group by partition.
    *                            Partition is defined by the partition values (partCol -> partValue)
-   * @param maxTargetFileSize Max size (in bytes) of the compaction output file.
    * @return Sequence of bins. Each bin contains one or more files from the same
    *         partition and targeted for one output file.
    */
   private def groupFilesIntoBins(
-      partitionsToCompact: Seq[(Map[String, String], Seq[AddFile])],
-      maxTargetFileSize: Long): Seq[(Map[String, String], Seq[AddFile])] = {
+      partitionsToCompact: Seq[(Map[String, String], Seq[AddFile])])
+  : Seq[(Map[String, String], Seq[AddFile])] = {
+    val maxBinSize = optimizeStrategy.maxBinSize
     partitionsToCompact.flatMap {
       case (partition, files) =>
         val bins = new ArrayBuffer[Seq[AddFile]]()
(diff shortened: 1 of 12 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala
```diff
@@ -231,6 +231,17 @@ class OptimizeExecutor(
...
 
+  /**
+   * In which mode the Optimize command is running. There are three valid modes:
+   * 1. Compaction
+   * 2. ZOrder
+   * 3. Clustering
+   */
+  private val optimizeTableMode = OptimizeTableStrategy.getMode(txn.snapshot, zOrderByColumns)
+
+  private val optimizeStrategy = OptimizeTableStrategy(
```
zedtang, line 242: can you add a comment on what this does? similar to optimizeTableMode above

spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableStrategy.scala
```diff
@@ -0,0 +1,256 @@
...
+      CompactionOptimizeTableStrategy(sparkSession, optimizeContext)
+    case other => throw new UnsupportedOperationException(s"Unsupported mode $other")
+  }
+
+  def getMode(snapshot: Snapshot, zOrderBy: Seq[String]): OptimizeTableMode.Value = {
+    val isClusteredTable = ClusteredTableUtils.isSupported(snapshot.protocol)
+    val isZOrderBy = zOrderBy.nonEmpty
+    if (isClusteredTable) {
+      assert(!isZOrderBy)
+      OptimizeTableMode.CLUSTERING
```
zedtang, line 108: I think if clustering columns are empty, it should fall under compaction right?

spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala
```diff
@@ -231,6 +231,17 @@ class OptimizeExecutor(
...
+    optimizeTableMode, sparkSession, txn.snapshot, optimizeContext, zOrderByColumns)
+
   /** Timestamp to use in [[FileAction]] */
   private val operationTimestamp = new SystemClock().getTimeMillis()
```
zedtang, line 247: I think we can remove `isClusteredTable` and `isMultiDimClustering`?

spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableStrategy.scala
dabao521: I think we can do this as a followup given this chage has been large.

spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala
dabao521: `isMultiDimClustering ` is still needed for places such as creating `repartitionDF` in `runOptimizeBinJob`. I started hiding them as well in the strategy class but that made the strategy class much more complex since it has more denepency from the Optimize table command's context.

The same goes to `isClusteredTable `.

spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableStrategy.scala
```diff
@@ -0,0 +1,267 @@
...
+  /**
+   * Utility method to prepare files in a partition for optimization.
+   *
+   * By default it sorts files on the size for the binpack.
+   *
+   * @return Prepared files for the subsequent optimization.
+   */
+  def prepareFilesPerPartition(inputFiles: Seq[AddFile]): Seq[AddFile] = inputFiles.sortBy(_.size)
+
+  def optimizeTableMode: OptimizeTableMode.Value
```
imback82, line 56: for completeness, let's add a comment

spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableStrategy.scala
```diff
@@ -0,0 +1,267 @@
...
+   */
+  def tagAddFile(file: AddFile, binInfo: OptimizeTableStrategy.BinInfo): AddFile =
+    file.copy(dataChange = false)
+
+  /**
+   * Utility to update additional metrics after optimization.
+   *
+   * @param optimizeStats The input stats to update on.
+   * @param removedFiles Removed files.
+   * @param bins Bin-packed file groups from [[groupFilesIntoBins]].
```
imback82, line 80: it's weird to reference a specific caller from the interface. can you just describe what's here (e.g., what are the key/value pairs, etc.)

spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableStrategy.scala
```diff
@@ -0,0 +1,267 @@
...
+      inputOtherFiles = inputFileStats,
+      inputNumCubes = 0,
+      mergedFiles = inputFileStats,
+      // There will one z-cube for each partition
+      numOutputCubes = optimizeStats.numPartitionsOptimized))
+  }
+}
+
+/** Implements clustering strategy for clustered tables */
+case class ClusteringOptimizeTableStrategy(
```
imback82, line 181: Can we simply do `CompactionStrategy`, `ZOrderStrategy` and `ClusteringStrategy`?

spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableStrategy.scala
```diff
@@ -0,0 +1,267 @@
...
+  }
+
+  // Upon a new ZCube, allocate a [[ZCubeInfo]] with a new ZCUBE ID.
+  override def initNewBin: OptimizeTableStrategy.BinInfo = {
+    OptimizeTableStrategy.ZCubeBinInfo(ZCubeInfo(clusteringColumns))
+  }
+
+  override def tagAddFile(file: AddFile, binInfo: OptimizeTableStrategy.BinInfo): AddFile = {
+    val taggedFile = super.tagAddFile(file, binInfo)
+    val zCubeInfo = binInfo.asInstanceOf[OptimizeTableStrategy.ZCubeBinInfo].zCubeInfo
```
imback82, line 214: btw, can `initNewBin` instantiate the value internally, so that the user doesn't need to pass the `binInfo` here?

spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableStrategy.scala
```diff
@@ -0,0 +1,267 @@
...
+  def optimizeTableMode: OptimizeTableMode.Value
+
+  /**
+   * The clustering algorithm to be used by either by ZORDER or Liquid CLUSTERING.
+   * An error is thrown for COMPACTION.
+   */
+  def curve: String
+
+  /**
+   * Prepare a new Bin and returns its initialized [[BinInfo]].
```
imback82, line 65: Can you add a comment when this needs to be called?

spark/src/main/scala/org/apache/spark/sql/delta/skipping/clustering/ZCube.scala
```diff
@@ -0,0 +1,175 @@
...
+    }
+  }
+
+  /**
+   * Filter out files belonging to single ZCube.
+   *
+   * @param files - Files sorted by ZCubeId, unoptimized files first.
+   */
+  def filterOutSingleZCubes(files: Iterator[AddFileWithNumRecords])
+  : Iterator[AddFileWithNumRecords] = {
```
imback82, line 116: nit: is this the right indentation?

imback82 (commented):
LGTM! (few nits / questions)

spark/src/main/scala/org/apache/spark/sql/delta/zorder/ZCubeInfo.scala
```diff
@@ -0,0 +1,117 @@
...
+   * Automatically generates a unique zCubeID.
+   */
+  def apply(zOrderBy: Seq[String]): ZCubeInfo = {
+    val zCubeID = UUID.randomUUID.toString
+    ZCubeInfo(zCubeID, zOrderBy)
+  }
+
+  private val ZCUBE_ID_KEY = AddFile.tag(ZCUBE_ID)
+  private val ZORDER_BY_KEY = AddFile.tag(ZCUBE_ZORDER_BY)
+  private[sql] val ZORDER_CURVE = AddFile.tag(ZCUBE_ZORDER_CURVE)
```
imback82, line 59: is this used somewhere else?

imback82 (approved):
LGTM!

spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala
mwc360, line 1: @dabao521 @imback82 @zedtang - Can you guys share why Z-Cubes are the grain in which Liquid Clustering is incremental. This appears to work very different and much less efficient compared to in Databricks. In Databricks, it appears that the autotune target file size (starting at 256m) is used to determine when a file stops being a candidate for clustering and this results in far less data being reclustered over time.

In OSS, write amplification is excessive because the same data will be reclustered until the z-cube is 100GB. For many write scenarios it makes it appear like OSS Delta doesn’t support incremental clustering.

Thx!

PR #2656, opened 2024-02-19
[Spark][TEST-ONLY] Add missing Row ID High Watermark test

longvu-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Add missing Row IDs High Watermark test.

## How was this patch tested?
Add a new UT, no new features.

No

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/delta/rowid/RowIdSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/rowid/RowIdSuite.scala
@@ -193,6 +193,23 @@ class RowIdSuite extends QueryTest
     }
   }
 
+  test("Check missing High Watermark for newly created empty table") {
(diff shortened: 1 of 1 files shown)


spark/src/test/scala/org/apache/spark/sql/delta/rowid/RowIdSuite.scala
```diff
@@ -193,6 +193,23 @@ class RowIdSuite extends QueryTest
     }
   }
 
+  test("Check missing High Watermark") {
```
tomvanbussel, line 196: Can we give this a more descriptive name? Right now it's not clear what it checks.

tomvanbussel (approved):
Nice, this will get us closer to being able to turn on row tracking outside of tests! One small comment about the test name, but it LGTM otherwise.

PR #2676, opened 2024-02-23
[Delta Uniform] overwrite source column field id for Iceberg PartitionField with field id assigned by Delta

lzlfred:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

context: Delta and Iceberg traverse schema and assigns field id in different way. Delta uniform currently use a extra Iceberg txn to overwrite the schema in iceberg table with wrong field ids reassigned by Iceberg.

However, if the source field id for partition columns is different in that schema overwrite txn, Iceberg always expect field id for partition columns to be the same and does not have logic to reconcile that, and then Iceberg will fail the overwrite txn. This PR adds the logic to adopt new source column field id to Iceberg PartitionField if changed, so the overwrite txn can go through and set the correct source column field id for PartitionFields.

## How was this patch tested?

manual test; unit test will come soon.

### DIFF
--- a/icebergShaded/iceberg_src_patches/0005-iceberg-takes-updated-source-column-field-id.patch
+++ b/icebergShaded/iceberg_src_patches/0005-iceberg-takes-updated-source-column-field-id.patch
@@ -1,0 +1,38 @@
+Iceberg PartitionField takes source column field id from latest schema if changed
+
+--- a/core/src/main/java/org/apache/iceberg/TableMetadata.java
++++ b/core/src/main/java/org/apache/iceberg/TableMetadata.java
+@@ -664,13 +664,21 @@ public class TableMetadata implements Serializable {
+     return new Builder(this).upgradeFormatVersion(newFormatVersion).build();
+   }
+ 
+-  private static PartitionSpec updateSpecSchema(Schema schema, PartitionSpec partitionSpec) {
++  private static PartitionSpec updateSpecSchema(Schema schema, Schema currSchema, PartitionSpec partitionSpec) {
+     PartitionSpec.Builder specBuilder =
+         PartitionSpec.builderFor(schema).withSpecId(partitionSpec.specId());
+ 
+-    // add all the fields to the builder. IDs should not change.
++    // add all the fields to the builder. IDs may change so it looks up the source field id by
++    // name from schema to update
+     for (PartitionField field : partitionSpec.fields()) {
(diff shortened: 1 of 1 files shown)


icebergShaded/iceberg_src_patches/0005-iceberg-takes-updated-source-column-field-id.patch
```diff
@@ -0,0 +1,38 @@
+Iceberg PartitionField takes source column field id from latest schema if changed
+
+--- a/core/src/main/java/org/apache/iceberg/TableMetadata.java
++++ b/core/src/main/java/org/apache/iceberg/TableMetadata.java
+@@ -664,13 +664,21 @@ public class TableMetadata implements Serializable {
+     return new Builder(this).upgradeFormatVersion(newFormatVersion).build();
+   }
+ 
+-  private static PartitionSpec updateSpecSchema(Schema schema, PartitionSpec partitionSpec) {
++  private static PartitionSpec updateSpecSchema(Schema schema, Schema currSchema, PartitionSpec partitionSpec) {
```
LukasRupprecht, line 10:
```suggestion
+  private static PartitionSpec updateSpecSchema(Schema newSchema, Schema currSchema, PartitionSpec partitionSpec) {
```

icebergShaded/iceberg_src_patches/0005-iceberg-takes-updated-source-column-field-id.patch
```diff
@@ -0,0 +1,38 @@
...
+         PartitionSpec.builderFor(schema).withSpecId(partitionSpec.specId());
+ 
+-    // add all the fields to the builder. IDs should not change.
++    // add all the fields to the builder. IDs may change so it looks up the source field id by
++    // name from schema to update
+     for (PartitionField field : partitionSpec.fields()) {
+-      specBuilder.add(field.sourceId(), field.fieldId(), field.name(), field.transform());
++      String partFieldSourceName = currSchema.findField(field.sourceId()).name();
++      int partFieldSourceInt;
++      if (schema.findField(partFieldSourceName) == null) {
```
LukasRupprecht, line 21: Make `schema.findField(partFieldSourceName)` a val and replace here and in line 24 below.

LukasRupprecht (approved):
2 nits, otherwise LGTM!

tdas (approved):
is this fixing a bug in uniform? then should this be backported to 3.1 branch?

PR #2677, opened 2024-02-23
[Spark][Test-only] Check that RESTORE doesn't downgrade protocol when restoring to a version where row tracking is not enabled, unless the flag is set.

longvu-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Check that RESTORE doesn't downgrade protocol when restoring to a version where row tracking is not enabled, unless the flag is set.

## How was this patch tested?
New UT.

No.


fred-db (approved):
LGTM! :)

PR #2687, opened 2024-02-26
[Spark] Fix adding wide bounds to files with incomplete stats

ala:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

When a DML command marks rows in a file as deleted, we have to update the file stats from tight bounds to wide bounds. The implementation expected that the file stats will include `nullCount`, `minValue` and `maxValue` fields. In some scenarios these fields might be missing, for example:

- When the `delta.dataSkippingNumIndexedCols` table property was set to 0
- When the table contains only BINARY columns (we only collect `nullCount` for this data type)

In such scenarios, the DML command fails.

This patch makes changes to `StatisticsCollection.updateStatsToWideBounds()` to ensure we don't try to access stats fields that are not already present.

## How was this patch tested?

New tests in `TightBoundsSuite`.

No.


cstavr:
LGTM!

cstavr (approved):

PR #2689, opened 2024-02-26
[Spark] Limit number of retries for non-conflict retryable commit failures

sumeet-db:
- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

We add limited retries (default = 10) for retryable CommitFailedExceptions which are non-conflict. The number of default retries for retryable conflict CommitFailedExceptions is set to 10M right now.

## How was this patch tested?

UT to make sure the change is working as intended

No


prakharjain09 (approved):

tdas (approved):
Seems like an obvious improvement in hindsight.

PR #2696, opened 2024-02-28
[Spark] Allow specifying custom prefixes for temp folders used by tests

xupefei:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR introduces a new trait, `DeltaSQLTestUtils`, to be used by all tests as a replacement for Apache Spark's `SQLTestUtils.` The new trait provides an ability for each test to specify a custom name prefix for temporary folders, so we can test the behavior of Delta Lake when a table's path contains special characters.

## How was this patch tested?

Not needed.

No.


xupefei:
@c27kwan I created this PR according to your suggestion. Could you review it?

spark/src/test/scala/org/apache/spark/sql/delta/test/DeltaSQLTestUtils.scala
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright (2021) The Delta Lake Project Authors.
```
c27kwan, line 2: DYK if we're supposed to bump the year on new files or if we expect 2021 every time? cc: @scottsand-db ?

spark/src/test/scala/org/apache/spark/sql/delta/test/DeltaSQLTestUtils.scala
```diff
@@ -0,0 +1,48 @@
...
+  /**
+   * Generates a temporary path with a prefix without creating the actual file/directory,
+   * and passes it to `f`. If a file/directory is created by `f`, it will be deleted
+   * after `f` returns.
+   *
+   * This method is copied over from [[SQLTestUtils]] of Apache Spark.
+   */
+  def withTempDir(prefix: String)(f: File => Unit): Unit = {
+    val path = Utils.createTempDir(namePrefix = prefix)
+    path.delete()
```
c27kwan, line 34: What is the purpose of this `.delete()` call? Did we meant to do it here or part of the `finally` clause? I don't see `delete` in the Spark implementation. https://github.com/apache/spark/blob/master/sql/core/src/test/scala/org/apache/spark/sql/test/SQLTestUtils.scala#L78

xupefei: I copied from here: https://src.dev.databricks.com/databricks/runtime@master/-/blob/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/plans/SQLHelper.scala?L41:5

Apparently there are multiple implementations... alright now I'll copy from SQLTestUtils.scala instead.

spark/src/test/scala/org/apache/spark/sql/delta/test/DeltaSQLTestUtils.scala
```diff
@@ -0,0 +1,47 @@
...
+import java.io.File
+
+import org.apache.spark.sql.test.SQLTestUtils
+import org.apache.spark.util.Utils
+
+trait DeltaSQLTestUtils extends SQLTestUtils {
+  /**
+   * Generates a temporary path with a prefix without creating the actual file/directory,
+   * and passes it to `f`. If a file/directory is created by `f`, it will be deleted
+   * after `f` returns.
```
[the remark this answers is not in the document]
xupefei: It's not stale but incorrect :)
Now this method will create the actual folder.

PR #2708, opened 2024-03-01
Record type widening metadata

johanl-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

It introduces metadata to record information about type changes that were applied using `ALTER TABLE`. This metadata is stored in table schema, as specified in https://github.com/delta-io/delta/pull/2624/files#diff-114dec1ec600a6305fe7117bed7acb46e94180cdb1b8da63b47b12d6c40760b9R28

For example, changing a top-level column `a` from `int` to `long` will update the schema to include metadata:
```
{
    "name" : "a",
    "type" : "long",
    "nullable" : true,
    "metadata" : { 
      "delta.typeChanges": [
        {
          "tableVersion": 1,
          "fromType": "integer",
          "toType": "long"
        },
        {
          "tableVersion": 5,
          "fromType": "integer",
          "toType": "long"
        }
      ]
    }
  }
```

## How was this patch tested?
- A new test suite `DeltaTypeWideningMetadataSuite` is created to cover methods handling type widening metadata.
- Tests covering adding metadata to the schema when running `ALTER TABLE CHANGE COLUMN` are added to `DeltaTypeWideningSuite`

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMetadata.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMetadata.scala
@@ -126,6 +126,9 @@ private[delta] object TypeWideningMetadata {
 
     SchemaMergingUtils.transformColumns(schema, oldSchema) {
       case (_, newField, Some(oldField), _) =>
+        // Record the version the transaction will attempt to use in the type change metadata. If
+        // there's a conflict with another transaction, the version in the metadata will be updated
+        // during conflict resolution. See [[ConflictChecker.updateTypeWideningMetadata()]].
         val typeChanges =
           collectTypeChanges(oldField.dataType, newField.dataType, txn.getFirstAttemptVersion)
         TypeWideningMetadata(typeChanges).appendToField(newField)
--- a/spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMetadata.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMetadata.scala
@@ -1,0 +1,192 @@
+/*
+ * Copyright (2021) The Delta Lake Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMetadata.scala
```diff
@@ -0,0 +1,192 @@
...
+      oldSchema: StructType): StructType = {
+
+    if (!TypeWidening.isEnabled(txn.protocol, txn.metadata)) return schema
+
+    if (DataType.equalsIgnoreNullability(schema, oldSchema)) return schema
+
+    SchemaMergingUtils.transformColumns(schema, oldSchema) {
+      case (_, newField, Some(oldField), _) =>
+        val typeChanges =
+          collectTypeChanges(oldField.dataType, newField.dataType, txn.getFirstAttemptVersion)
```
sabir-akhadov, line 130: Is the getFirstAttemptVersion ever updated in the case of concurrent transactions on the table? Maybe add a comment?

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningMetadataSuite.scala
```diff
@@ -0,0 +1,505 @@
...
+
+  /**
+   * Short-hand to build the metadata for a type change to cut down on repetition.
+   */
+  private def typeChangeMetadata(
+      version: Long,
+      fromType: String,
+      toType: String,
+      path: String = ""): Metadata = {
+    val builder = new MetadataBuilder()
```
sabir-akhadov, line 61: maybe use TypeChange.toMetadata directly?

johanl-db: The issue is that it could hide bugs in TypeChange.toMetadata. e.g. we wouldn't see if fromType gets serialized to INT instead of integer. That's why I pass all the args as plain strings and also explicitly write out the metadata keys instead of reusing the constants in TypeChangeMetadata: these should never change, if they do we want tests to fail.

spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMetadata.scala
johanl-db: I added a comment

PR #2709, opened 2024-03-01
Disentangle schema evolution in MERGE INTO command

johanl-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

### Context
Schema evolution in the MERGE INTO command relies on manually updating the plan of the target table and replace the original output attributes to match the schema after evolution.
Manually updating the target plan is unnecessarily complex, error-prone and ultimately wrong: the new attributes don't match the actual target schema that was used to resolve all expressions. This worked so far for schema evolution that only adds new fields or columns that are then implicitly filled with `null`s on read but breaks when introducing type evolution due to type mismatches that can't be reconciled.

### Changes
The target plan isn't manually updated to support schema evolution in the MERGE INTO command anymore. Instead, the original target output is used and we rely on the different expressions used to write out the data to produce an output that supports schema evolution. E.p.:
- For target columns that are assigned to by an UPDATE action: `generateUpdateExpressions` already does the heavy lifting, it is simply updated to use the actual target output instead of trying to reference the evolved schema. This allows simplifying a bit the method.
- For target columns that aren't assigned to by an UPDATE action (either columns without an assignment or with a DELETE action and the deleted row must be written to the CDC output): no-op expressions are generated by casting the target output attributes to their corresponding type after evolution in `getTargetOutputCols`.

The manual handling in `buildTargetPlanWithIndex` and `replaceFileIndex` that isn't needed anymore is removed. This in turns allow splitting `replaceFileIndex` in two methods with separate concerns: `replaceFileIndex` (already exists) and `dropColumns` to manually remove columns from the target plan.

## How was this patch tested?
This is extensively covered by existing MERGE tests, in particular:
- `MergeIntoSchemaEvolutionTests`
- `MergeCDCSuite` - `schema evolution with non-nullable schema` that covers handling non nullable fields that turn nullable due to the outer join in MERGE.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/UpdateExpressionsSupport.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/UpdateExpressionsSupport.scala
@@ -196,8 +196,8 @@ trait UpdateExpressionsSupport extends SQLConfHelper with AnalysisHelper {
   }
 
   /**
-   * Given a list of target-column expressions and a set of update operations, generate a list
-   * of update expressions, which are aligned with given target-column expressions.
+   * Given a target schema and a set of update operations, generate a list of update expressions,
+   * which are aligned with the given schema.
    *
    * For update operations to nested struct fields, this method recursively walks down schema tree
    * and apply the update expressions along the way.
@@ -210,16 +210,20 @@ trait UpdateExpressionsSupport extends SQLConfHelper with AnalysisHelper {
    *
    * this method works as follows:
    *
-   * generateUpdateExpressions(targetCols=[a,z], updateOps=[(a.b, 1), (a.c, 2), (z, 3)])
-   *   generateUpdateExpressions(targetCols=[b,c,d], updateOps=[(b, 1),(c, 2)], pathPrefix=["a"])
+   * generateUpdateExpressions(
+   *   targetSchema=[a,z], originalExprs=[a, z], updateOps=[(a.b, 1), (a.c, 2), (z, 3)])
+   *   generateUpdateExpressions(
+   *     targetSchema=[b,c,d], originalExprs=[b, c, d], updateOps=[(b, 1),(c, 2)], pathPrefix=["a"])
    *     end-of-recursion
    *   -> returns (1, 2, d)
    * -> return ((1, 2, d), 3)
    *
-   * @param targetCols a list of expressions to read named columns; these named columns can be
-   *                   either the top-level attributes of a table, or the nested fields of a
-   *                   StructType column.
+   * @param targetSchema schema to follow to generate update expressions. Due to schema evolution,
+   *                     it may contain additional columns or fields not present in the original
+   *                     table schema.
    * @param updateOps a set of update operations.
+   * @param defaultExprs the expressions to use when no update operation is provided for a column
+   *                      or field. This is typically the output from the base table.
    * @param pathPrefix the path from root to the current (nested) column. Only used for printing out
    *                   full column path in error messages.
    * @param allowStructEvolution Whether to allow structs to evolve. When this is false (default),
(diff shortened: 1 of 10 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/UpdateExpressionsSupport.scala
```diff
@@ -201,22 +201,26 @@ trait UpdateExpressionsSupport extends SQLConfHelper with AnalysisHelper {
...
-   *   targetSchema=[a,z], originalExprs=[a, z], updateOps=[(a.b, 1), (a.c, 2), (z, 3)])
+   *   targetSchema=[s1,s2,z], defaultExprs=[s1,s2, z], updateOps=[(s1.a, 1), (s1.b, 2), (z, 3)])
+   *   -> generates expression for s1 - build recursively from child assignments
    *   generateUpdateExpressions(
-   *     targetSchema=[b,c,d], originalExprs=[b, c, d], updateOps=[(b, 1),(c, 2)], pathPrefix=["a"])
+   *     targetSchema=[a,b,c], defaultExprs=[a, b, c], updateOps=[(a, 1),(b, 2)], pathPrefix=["s1"])
    *     end-of-recursion
-   *   -> returns (1, 2, d)
-   * -> return ((1, 2, d), 3)
+   *   -> returns (1, 2, a.d)
```
bart-samwel, line 219:
```suggestion
   *   -> returns (1, 2, a.c)
```

spark/src/main/scala/org/apache/spark/sql/delta/UpdateExpressionsSupport.scala
```diff
@@ -201,22 +201,26 @@ trait UpdateExpressionsSupport extends SQLConfHelper with AnalysisHelper {
...
+   *   -> generates expression for s2 - no child assignment and no update expression: use
+   *      default expression `s2`
+   *   -> generates expression for z - use available update expression `3`
+   * -> returns ((1, 2, a.d), s2, 3)
```
bart-samwel, line 223:
```suggestion
   * -> returns ((1, 2, a.c), s2, 3)
```

spark/src/main/scala/org/apache/spark/sql/delta/UpdateExpressionsSupport.scala
```diff
@@ -210,16 +210,20 @@ trait UpdateExpressionsSupport extends SQLConfHelper with AnalysisHelper {
    *
    * this method works as follows:
    *
-   * generateUpdateExpressions(targetCols=[a,z], updateOps=[(a.b, 1), (a.c, 2), (z, 3)])
-   *   generateUpdateExpressions(targetCols=[b,c,d], updateOps=[(b, 1),(c, 2)], pathPrefix=["a"])
+   * generateUpdateExpressions(
```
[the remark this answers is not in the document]
bart-samwel: Very nice!

spark/src/main/scala/org/apache/spark/sql/delta/commands/MergeIntoCommandBase.scala
```diff
@@ -89,6 +89,40 @@ trait MergeIntoCommandBase extends LeafRunnableCommand
     }
   }
 
+  /**
+   * Output expressions from the target adjusted after applying schema evolution. There are two
```
bart-samwel, line 93: This could be even clearer. "Output expressions from the target" can be interpreted in multiple ways. IIUIC this produces expressions to convert from a pre-evolution target row to the post-evolution target row, for use to copy rows unmodified, as well as for default expressions for columns that are not modified updated rows. Is that right?

bart-samwel (commented):
Nice! A couple more points, but this is coming together nicely.

spark/src/main/scala/org/apache/spark/sql/delta/commands/MergeIntoCommandBase.scala
johanl-db: Correct, that's much better, I rephrased that comment

bart-samwel (approved):
All right, let's go! Looks much better. :)

PR #2715, opened 2024-03-04
Refactor iceberg to delta partition value conversion

lzlfred:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

refactor  iceberg to delta partition value conversion to make code more modular. No function change. 

## How was this patch tested?

existing unit test.


iceberg/src/main/scala/org/apache/spark/sql/delta/IcebergFileManifest.scala
```diff
@@ -171,4 +155,37 @@ class IcebergFileManifest(
   }
 
   override def close(): Unit = fileSparkResults.map(_.unpersist())
+
+  def convertIcebergPartitionToPartitionValues(partition: StructLike):
+      Option[Map[String, String]] = {
+    if (spark.sessionState.conf
```
mingdai-db, line 161: Can we move the conf check out of this help function

mingdai-db (approved):
LGTM

PR #2716, opened 2024-03-04
[Spark] Add LatestTableVersion to CommitStore GetCommits API response

sumeet-db:
- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
With this change, we require CommitStores to maintain and return the latestTableVersion in the getCommits API. The motivation here is for readers to get more context from CommitStore when all commits are backfilled and the Commit Response is empty.

Clients can use latestTableVersion to determine the version the table is when all commits are backfilled and the commits list is empty in the getCommits response.

## How was this patch tested?
Unit tests

No


LukasRupprecht (approved):
Overall LGTM. Could you add a bit more details around what the clients can/would do with the additional context of latestCommitVersion that will now be returned?

sumeet-db:
> Overall LGTM. Could you add a bit more details around what the clients can/would do with the additional context of latestCommitVersion that will now be returned?

Added.

PR #2719, opened 2024-03-05
[Spark] Deletion Vector descriptor `path` type should URI-escape special characters

xupefei:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR fixes an issue where the `path` type of `DeletionVectorDescriptor` must escape special characters in the DV path. A special character may exist if the DV is referring to an existing file from another table that contains such a character in its path.

## How was this patch tested?

New tests.

No.


xupefei:
Hello @larsk-db, could you review this PR?

larsk-db (approved):
LGTM, thanks.

scottsand-db (approved):

PR #2720, opened 2024-03-05
Drop Type Widening Table Feature

johanl-db:
## Description

It adds the ability to remove the type widening table feature by running the `ALTER TABLE DROP FEATURE` command.
Before dropping the table feature, traces of it are removed from the current version of the table:
- Files that were written before the latest type change and thus contain types that differ from the current table schema are rewritten using an internal `REORG TABLE` operation.
- Metadata in the table schema recording previous type changes is removed.

## How was this patch tested?
- A new set of tests are added to `DeltaTypeWideningSuite` to cover dropping the table feature with tables in various states: with/without files to rewrite or metadata to remove.

When the feature is available, this change enables the following user action:
- Drop the type widening table feature:
```
ALTER TABLE t DROP FEATURE typeWidening
```
This succeeds immediately if no version of the table contains traces of the table feature (= no type changes were applied in the available history of the table.
Otherwise, if the current version contains traces of the feature, these are removed: files are rewritten if needed and type widening metadata is removed from the table schema. Then, an error `DELTA_FEATURE_DROP_WAIT_FOR_RETENTION_PERIOD` is thrown, telling the user to retry once the retention period expires.

If only previous versions contain traces of the feature, no action is applied on the table, and an error `DELTA_FEATURE_DROP_HISTORICAL_VERSIONS_EXIST` is thrown, telling the user to retry once the retention period expires.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
@@ -559,7 +559,9 @@ private[delta] class ConflictChecker(
    *     to handle the row tracking feature being enabled by the winning transaction.
    */
   private def reassignRowCommitVersions(): Unit = {
-    if (!RowTracking.isSupported(currentTransactionInfo.protocol)) {
+    if (!RowTracking.isSupported(currentTransactionInfo.protocol) &&
+      // Type widening relies on default row commit versions to be set.
+      !TypeWidening.isSupported(currentTransactionInfo.protocol)) {
       return
     }
 
--- a/spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
@@ -161,6 +161,9 @@ private[delta] class ConflictChecker(
     reassignOverlappingRowIds()
     reassignRowCommitVersions()
 
+    // Update the table version in newly added type widening metadata.
+    updateTypeWideningMetadata()
+
     // Data file checks.
     checkForAddedFilesThatShouldHaveBeenReadByCurrentTxn()
     checkForDeletedFilesAgainstCurrentTxnReadFiles()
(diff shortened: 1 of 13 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
```diff
@@ -559,7 +559,9 @@ private[delta] class ConflictChecker(
    *     to handle the row tracking feature being enabled by the winning transaction.
    */
   private def reassignRowCommitVersions(): Unit = {
-    if (!RowTracking.isSupported(currentTransactionInfo.protocol)) {
+    if (!RowTracking.isSupported(currentTransactionInfo.protocol) &&
+      // Type widening relies on default row commit versions to be set.
+      !TypeWidening.isSupported(currentTransactionInfo.protocol)) {
```
sabir-akhadov, line 564: I don't understand this condition. If row tracking is not supported but type widening is, we will still reassign the default row commit version? How can we have a row commit version without row tracking?

spark/src/main/scala/org/apache/spark/sql/delta/PreDowngradeTableFeatureCommand.scala
```diff
@@ -126,3 +126,75 @@ case class V2CheckpointPreDowngradeCommand(table: DeltaTableV2)
...
+   * - Remove the type widening metadata attached to fields in the current table schema.
+   *
+   * @return Return true if files were rewritten or metadata was removed. False otherwise.
+   */
+  override def removeFeatureTracesIfNeeded(): Boolean = {
+    if (TypeWideningTableFeature.validateRemoval(table.initialSnapshot)) return false
+
+    val startTimeNs = System.nanoTime()
+    val properties = Seq(DeltaConfigs.ENABLE_TYPE_WIDENING.key)
+    AlterTableUnsetPropertiesDeltaCommand(table, properties, ifExists = true).run(table.spark)
```
sabir-akhadov, line 148: just checking my understanding: removing the table property doesn't mean the table will become unreadable if we don't succeed in rewriting all files?

spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
johanl-db: We only need a very limited subset of row tracking: that each AddFile action committed when the type widening table feature is supported contains a (correct) defaultRowCommitVersion value.

E.p. we don't require a backfill for older files, that other row tracking information is populated - base row id, high water mark - or that the defaultRowCommitVersion be accessible other than by directly reading it from AddFile action.

I updated the tests for dropping the table feature to run with both row tracking enabled and disabled.

spark/src/main/scala/org/apache/spark/sql/delta/PreDowngradeTableFeatureCommand.scala
johanl-db: The table is still readable after unsetting the property, it simply prevents further type changes to be applied.

sabir-akhadov:
| This PR includes changes from https://github.com/delta-io/delta/pull/2708 which isn't merged yet.
I believe it is now?

sabir-akhadov (approved):
lgtm

PR #2732, opened 2024-03-07
[CI] Disable running unidoc from the spark CI tests for now

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [X] Other (fill in here)

## Description

This removes compiling unidoc from the spark CI tests since we've seen a flaky failure on #2727.

We will follow up and create a separate job just for unidoc (which makes more sense anyways as it should run for changes in any of the projects.)

## How was this patch tested?

CI runs.


tdas (approved):
yeah, best to move it to a different github action.

PR #2733, opened 2024-03-08
An example fix of DeltaCatalog

mingdai-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)


tdas:
Sorry, what is this fix an example of?

tdas:
There arent any unit tests?

PR #2752, opened 2024-03-14
Support Java8 time objects

yhosny:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Added support for Java8 time objects when Java8Api is enabled. 

The following query fails when Java8Api is enabled:

```
INSERT INTO $tableName REPLACE
       where (DATE IN (DATE('2024-03-11'), DATE('2024-03-12')))
       VALUES ('1', DATE('2024-03-13'))
```

## How was this patch tested?

Unit Test.

The potential public surface area for this change is limited to only external DateType, TimestampType. By default, Spark converts DateType values to java.sql.Date, TimestampType -> java.sql.Timestamp but with the SQL conf enabled, Spark uses java.time.LocalDate as the external type for DateType and java.time.Instant for TimestampType.

Necessity of fix: If users use Spark 3.0 and enable Java 8, they will be unable to parse and use dates in Delta tables correctly, leading to matching errors.

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/delta/DeltaSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/DeltaSuite.scala
@@ -1607,6 +1607,20 @@ class DeltaSuite extends QueryTest
     }
   }
 
+  test("support Java8 time objects") {
+    withSQLConf(SQLConf.DATETIME_JAVA8API_ENABLED.key -> "true") {
+      val tableName = "my_table"
+      withTable(tableName) {
+        spark.sql(s"CREATE TABLE $tableName (id STRING, date DATE) USING DELTA;")
+        spark.sql(
+          s"""
+             |INSERT INTO $tableName REPLACE
+             |where (DATE IN (DATE('2024-03-11'), DATE('2024-03-13')))
+             |VALUES ('2', DATE('2024-03-13')), ('3', DATE('2024-03-11'))
+             |""".stripMargin)
+      }
+    }
+  }
 
   test("all operations with special characters in path") {
     withTempDir { dir =>
--- a/spark/src/test/scala/org/apache/spark/sql/delta/DeltaSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/DeltaSuite.scala
@@ -1630,7 +1630,7 @@ class DeltaSuite extends QueryTest
         spark.sql(
           s"""
              |INSERT INTO $tableName REPLACE
-             |where 
+             |where
              | (timestamp IN (TIMESTAMP('2022-12-22 15:50:00'), TIMESTAMP('2022-12-23 15:50:00')))
              | VALUES
              | ('2', TIMESTAMP('2022-12-22 15:50:00')), ('3', TIMESTAMP('2022-12-23 15:50:00'))
--- a/spark/src/test/scala/org/apache/spark/sql/delta/DeltaSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/DeltaSuite.scala
@@ -1607,7 +1607,7 @@ class DeltaSuite extends QueryTest
     }
   }
 
-  test("support Java8 time object") {
+  test("support Java8 API for DATE type") {
     withSQLConf(SQLConf.DATETIME_JAVA8API_ENABLED.key -> "true") {
       val tableName = "my_table"
       withTable(tableName) {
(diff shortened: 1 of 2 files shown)


spark/src/test/scala/org/apache/spark/sql/delta/DeltaSuite.scala
```diff
@@ -1607,6 +1607,20 @@ class DeltaSuite extends QueryTest
     }
   }
 
+  test("support Java8 time objects") {
+    withSQLConf(SQLConf.DATETIME_JAVA8API_ENABLED.key -> "true") {
+      val tableName = "my_table"
+      withTable(tableName) {
+        spark.sql(s"CREATE TABLE $tableName (id STRING, date DATE) USING DELTA;")
```
sandip-db, line 1614: Add a column with `TIMESTAMP` type to test LocalDateTime.
Please make sure that the issue repro with `TIMESTAMP` as well.

sandip-db (changes requested):

spark/src/test/scala/org/apache/spark/sql/delta/DeltaSuite.scala
yhosny: Good catch, tested with `TIMESTAMP` which uses `time.Instant` not `time.LocalDateTime`

yhosny: Also verified that the test does repro for timestamp.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaSuite.scala
```diff
@@ -1607,6 +1607,37 @@ class DeltaSuite extends QueryTest
...
+          s"""
+             |INSERT INTO $tableName REPLACE
+             |where (DATE IN (DATE('2024-03-11'), DATE('2024-03-13')))
+             |VALUES ('2', DATE('2024-03-13')), ('3', DATE('2024-03-11'))
+             |""".stripMargin)
+      }
+    }
+  }
+
+  test("support Java8 time object") {
```
sandip-db, line 1625:
```suggestion
  test("support Java8 API for TIMESTAMP type") {
```

spark/src/test/scala/org/apache/spark/sql/delta/DeltaSuite.scala
```diff
@@ -1607,6 +1607,37 @@ class DeltaSuite extends QueryTest
     }
   }
 
+  test("support Java8 time object") {
```
sandip-db, line 1610:
```suggestion
  test("support Java8 API for DATE type") {
```

sandip-db (approved):
test cannot have the same name

spark/src/test/scala/org/apache/spark/sql/delta/DeltaSuite.scala
```diff
@@ -1607,6 +1607,37 @@ class DeltaSuite extends QueryTest
...
+    withSQLConf(SQLConf.DATETIME_JAVA8API_ENABLED.key -> "true") {
```
tdas, line 1611: can you give me a reference to support for this in Spark code? I want to understand how this is tested in Spark.

tdas (changes requested):
I would like to see a bit more information about the Java8 APIs from the Spark stuff before expanding public API support. 

I dont agree with your description that this does not change public API completely. Its quite possible that this SQL conf will be enable in Spark by default. That would implicitly allow new SQL operations in Delta that would have failed earlier. So I do want to understand the public API surface area before I approve this.

tdas (approved):

yhosny:
> I would like to see a bit more information about the Java8 APIs from the Spark stuff before expanding public API support.
> 
> I dont agree with your description that this does not change public API completely. Its quite possible that this SQL conf will be enable in Spark by default. That would implicitly allow new SQL operations in Delta that would have failed earlier. So I do want to understand the public API surface area before I approve this.

Summarizing offline conversation: 

- The potential public-facing surface area of the fix: The potential public surface area for this change is limited to only one external DateType. By default, Spark converts DateType values to java.sql.Date, but with the SQL conf enabled, Spark uses java.time.LocalDate as the external type for DateType.
- The necessity of the fix: If users use Spark 3.0 and enable Java 8, they will be unable to parse and use dates in Delta tables correctly, leading to matching errors.
- The availability: This fix would be available from Spark 3.0 onwards.

Issue #1707, opened 2023-04-21
[Feature Request] Improve unit test execution time


felipepessoto:
### Overview

The unit tests are taking longer every new version. As a reference, the build in this PR, from a year ago took 61-77 minutes: https://github.com/delta-io/delta/pull/887

<img width="694" alt="image" src="https://user-images.githubusercontent.com/1336227/233565002-ee64a475-b5ca-4c2c-aa90-3d407a86a1cf.png">

### Motivation

I think we need to improve it before it becomes out of control.

### Further details

Parallel tests are disabled:

```
    // Don't execute in parallel since we can't have multiple Sparks in the same JVM
    Test / parallelExecution := false,
```

Do we have any alternatives?

### Willingness to contribute

The Delta Lake Community encourages new feature contributions. Would you or another member of your organization be willing to contribute an implementation of this feature?

- [ ] Yes. I can contribute this feature independently.
- [ ] Yes. I would be willing to contribute this feature with guidance from the Delta Lake community.
- [X] No. I cannot contribute this feature at this time.

felipepessoto:
Since I opened the issue, tests went from ~230m, to more than 360m (they are timing out)

felipepessoto:
@edmondop, @tdas, @scottsand-db, @allisonport-db (from #1249)

We could add tags (Group1, Group2....Group 10) to unit tests and change run-tests.py adding a testOnly argument, and a runPythonTests.

It would work independently of the infra being used. You could start several VMs/Agents each one calling run-tests.py with different tags. What you think? I can send a PR, but would like to confirm if somebody will be able to review it.

felipepessoto:
Now we have shards and TEST_PARALLELISM_COUNT=4

PR #2754, opened 2024-03-15
[DO_NOT_MERGE] Update spark_test.yaml to test larger runners

tdas:


felipepessoto:
I think we need to split some huge Suites taking 20-30m like MergeCDC and Hilbert Index, to improve parallelism.

tdas:
Yeah, I am trying to see easy parallelism vs cost of using larger runners.

felipepessoto:
I have being doing this. https://github.com/delta-io/delta/issues/1707#issuecomment-1636558990 

But it requires manually balancing the group of tags

PR #2755, opened 2024-03-15
[SPARK] Managed Commit support for cold and hot snapshot update

prakharjain09:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR adds support for cold and hot snapshot update for managed-commits.

## How was this patch tested?

UTs

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
@@ -59,14 +59,14 @@ trait SnapshotManagement { self: DeltaLog =>
 
   @volatile private[delta] var asyncUpdateTask: Future[Unit] = _
 
+  /** Use ReentrantLock to allow us to call `lockInterruptibly` */
+  protected val snapshotLock = new ReentrantLock()
+
   /**
    * Cached fileStatus for the latest CRC file seen in the deltaLog.
    */
   @volatile protected var lastSeenChecksumFileStatusOpt: Option[FileStatus] = None
   @volatile protected var currentSnapshot: CapturedSnapshot = getSnapshotAtInit
-
-  /** Use ReentrantLock to allow us to call `lockInterruptibly` */
-  protected val snapshotLock = new ReentrantLock()
 
   /**
    * Run `body` inside `snapshotLock` lock using `lockInterruptibly` so that the thread
@@ -79,20 +79,6 @@ trait SnapshotManagement { self: DeltaLog =>
     } finally {
       snapshotLock.unlock()
     }
-  }
-
-  /**
-   * Get the LogSegment that will help in computing the Snapshot of the table at DeltaLog
-   * initialization, or None if the directory was empty/missing.
-   *
-   * @param startingCheckpoint A checkpoint that we can start our listing from
-   */
-  protected def getLogSegmentFrom(
-      startingCheckpoint: Option[LastCheckpointInfo]): Option[LogSegment] = {
-    getLogSegmentForVersion(
-      versionToLoad = None,
-      lastCheckpointInfo = startingCheckpoint
-    )
   }
 
   /** Get an iterator of files in the _delta_log directory starting with the startVersion. */
@@ -231,11 +217,11 @@ trait SnapshotManagement { self: DeltaLog =>
    * @return Some LogSegment to build a Snapshot if files do exist after the given
    *         startCheckpoint. None, if the directory was missing or empty.
    */
-  protected def getLogSegmentForVersion(
+  protected def createLogSegment(
       versionToLoad: Option[Long] = None,
       oldCheckpointProviderOpt: Option[UninitializedCheckpointProvider] = None,
-      lastCheckpointInfo: Option[LastCheckpointInfo] = None,
-      commitStoreOpt: Option[CommitStore] = None): Option[LogSegment] = {
+      commitStoreOpt: Option[CommitStore] = None,
+      lastCheckpointInfo: Option[LastCheckpointInfo] = None): Option[LogSegment] = {
     // List based on the last known checkpoint version.
     // if that is -1, list from version 0L
     val lastCheckpointVersion = getCheckpointVersion(lastCheckpointInfo, oldCheckpointProviderOpt)
@@ -251,6 +237,12 @@ trait SnapshotManagement { self: DeltaLog =>
       oldCheckpointProviderOpt = oldCheckpointProviderOpt,
       lastCheckpointInfo = lastCheckpointInfo
     )
+  }
+
+  private def createLogSegment(previousSnapshot: Snapshot): Option[LogSegment] = {
+    createLogSegment(
+      oldCheckpointProviderOpt = Some(previousSnapshot.checkpointProvider),
+      commitStoreOpt = previousSnapshot.commitStoreOpt)
   }
 
   /**
@@ -324,7 +316,7 @@ trait SnapshotManagement { self: DeltaLog =>
           // deleting files. Either way, we can't safely continue.
           //
           // For now, we preserve existing behavior by returning Array.empty, which will trigger a
-          // recursive call to [[getLogSegmentForVersion]] below.
+          // recursive call to [[createLogSegment]] below.
           Array.empty[FileStatus]
         }
 
@@ -335,7 +327,7 @@ trait SnapshotManagement { self: DeltaLog =>
       } else if (newFiles.isEmpty) {
         // The directory may be deleted and recreated and we may have stale state in our DeltaLog
         // singleton, so try listing from the first version
-        return getLogSegmentForVersion(versionToLoad = versionToLoad)
+        return createLogSegment(versionToLoad = versionToLoad)
       }
       val (checkpoints, deltasAndCompactedDeltas) = newFiles.partition(isCheckpointFile)
       val (deltas, compactedDeltas) = deltasAndCompactedDeltas.partition(isDeltaFile)
@@ -498,30 +490,20 @@ trait SnapshotManagement { self: DeltaLog =>
    * file as a hint on where to start listing the transaction log directory. If the _delta_log
    * directory doesn't exist, this method will return an `InitialSnapshot`.
    */
-  protected def getSnapshotAtInit: CapturedSnapshot = {
+  protected def getSnapshotAtInit: CapturedSnapshot = withSnapshotLockInterruptibly {
     recordFrameProfile("Delta", "SnapshotManagement.getSnapshotAtInit") {
-      val currentTimestamp = clock.getTimeMillis()
+      val snapshotInitWallclockTime = clock.getTimeMillis()
       val lastCheckpointOpt = readLastCheckpointFile()
-      createSnapshotAtInitInternal(
-        initSegment = getLogSegmentFrom(lastCheckpointOpt),
-        timestamp = currentTimestamp
-      )
+      val initialSegmentForNewSnapshot = createLogSegment(
+        versionToLoad = None,
+        lastCheckpointInfo = lastCheckpointOpt)
+      val snapshot = getUpdatedSnapshot(
+        oldSnapshotOpt = None,
+        initialSegmentForNewSnapshot = initialSegmentForNewSnapshot,
+        initialCommitStore = None,
+        isAsync = false)
+      CapturedSnapshot(snapshot, snapshotInitWallclockTime)
     }
-  }
-
-  protected def createSnapshotAtInitInternal(
-      initSegment: Option[LogSegment],
-      timestamp: Long): CapturedSnapshot = {
-    val snapshot = initSegment.map { segment =>
-      val snapshot = createSnapshot(
-        initSegment = segment,
-        checksumOpt = None)
-      snapshot
-    }.getOrElse {
-      logInfo(s"Creating initial snapshot without metadata, because the directory is empty")
-      new InitialSnapshot(logPath, this)
-    }
-    CapturedSnapshot(snapshot, timestamp)
   }
 
   /**
@@ -696,7 +678,7 @@ trait SnapshotManagement { self: DeltaLog =>
      * Instead, just do a general update to the latest available version. The racing commits
      * can then use the version check short-circuit to avoid constructing a new snapshot.
      */
-    getLogSegmentForVersion(
+    createLogSegment(
       oldCheckpointProviderOpt = Some(oldCheckpointProvider),
       commitStoreOpt = commitStoreOpt
     ).getOrElse {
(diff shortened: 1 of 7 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
```diff
@@ -907,48 +898,102 @@ trait SnapshotManagement { self: DeltaLog =>
...
+  /** Creates a Snapshot for the given `segmentOpt` */
+  protected def getSnapshotForLogSegmentInternal(
+      previousSnapshotOpt: Option[Snapshot],
       segmentOpt: Option[LogSegment],
-      updateTimestamp: Long,
       isAsync: Boolean): Snapshot = {
     segmentOpt.map { segment =>
-      if (segment == previousSnapshot.logSegment) {
+      if (previousSnapshotOpt.exists(_.logSegment == segment)) {
         // If no changes were detected, just refresh the timestamp
```
ryan-johnson-databricks, line 957: Stale comment? The timestamp manipulation code here was deleted?

spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
```diff
@@ -907,48 +898,102 @@ trait SnapshotManagement { self: DeltaLog =>
...
     }.getOrElse {
-      logInfo(s"No delta log found for the Delta table at $logPath")
-      replaceSnapshot(new InitialSnapshot(logPath, this), updateTimestamp)
+      logInfo(s"Creating initial snapshot without metadata, because the directory is empty")
+      new InitialSnapshot(logPath, this)
     }
-    currentSnapshot.snapshot
   }
 
   /** Replace the given snapshot with the provided one. */
```
ryan-johnson-databricks, line 973: Stale comment, now that we added the keep-same-snapshot semantics?

spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
```diff
@@ -907,48 +898,102 @@ trait SnapshotManagement { self: DeltaLog =>
...
-    oldSnapshot.uncache()
+    if (currentSnapshot == null) {
+      // cold snapshot initialization
+      currentSnapshot = CapturedSnapshot(newSnapshot, updateTimestamp)
+      return newSnapshot
+    }
+    val CapturedSnapshot(oldSnapshot, oldTimestamp) = currentSnapshot
+    if (oldSnapshot eq newSnapshot) {
+      // Same snapshot as before, so just refresh the timestamp
+      val timestampToUse = math.max(updateTimestamp, oldTimestamp)
```
ryan-johnson-databricks, line 989: This is the guaranteed-fresh-as-of timestamp, right? So we want it to be as large as possible? If so, should we also include `newSnapshot.timestamp` in the max, in case the snapshot landed _after_ the snapshot update process began? (it's probably a pretty rare race condition, tho)

(I realize this is just existing logic moved from `installLogSegmentInternal`, so maybe we should address this in a separate PR, if we decide it's important to do so?)

spark/src/main/scala/org/apache/spark/sql/delta/managedcommit/InMemoryCommitStore.scala
```diff
@@ -126,6 +126,16 @@ class InMemoryCommitStore(val batchSize: Long) extends AbstractBatchBackfillingC
       versionsToRemove.foreach(tableData.commitsMap.remove)
     }
   }
+
+  def registerTable(
+      logPath: Path,
+      maxCommitVersion: Long): Unit = {
+    val newPerTableData = new PerTableData()
+    newPerTableData.maxCommitVersion = maxCommitVersion
```
ryan-johnson-databricks, line 134: nit: any reason this isn't a constructor arg?

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitSuite.scala
```diff
@@ -116,6 +123,405 @@ class ManagedCommitSuite
...
+    }
+  }
+
+  test("cold snapshot initialization - with and without initial commit store") {
+    CommitStoreProvider.registerBuilder(TrackingInMemoryCommitStoreBuilder(batchSize = 10))
+    withTempDir { tempDir =>
+      val tablePath = tempDir.getAbsolutePath
+      Seq(1).toDF.write.format("delta").mode("overwrite").save(tablePath) // version 0
+      Seq(2).toDF.write.format("delta").mode("overwrite").save(tablePath) // version 1
+      Seq(2).toDF.write.format("delta").mode("overwrite").save(tablePath) // version 2
```
ryan-johnson-databricks, line 153: Are these repeated overwrites somehow important, out of curiosity?

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitSuite.scala
```diff
@@ -116,6 +123,405 @@ class ManagedCommitSuite
...
+          dataPath = new Path(tablePath),
+          options = Map.empty,
+          initialCommitStoreOpt = Some(cs),
+          tableIdentifier = None)
+        val snapshot = log.unsafeVolatileSnapshot
+        assert(snapshot.version === 2)
+        assert(snapshot.commitStoreOpt.nonEmpty)
+      }
+      assert(filterUsageRecords(usageRecords2, "delta.readChecksum").size === 1)
+
```
ryan-johnson-databricks, line 183:
```suggestion
```

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitSuite.scala
```diff
@@ -116,6 +123,405 @@ class ManagedCommitSuite
...
+        assert(snapshot.commitStoreOpt.nonEmpty)
+      }
+      assert(filterUsageRecords(usageRecords1, "delta.readChecksum").size === 2)
+
+      // Case-2: Create deltaLog without initial commit store
+      DeltaLog.clearCache()
+      val usageRecords2 = Log4jUsageLogger.track {
+        val ownerName = spark.conf.get(MANAGED_COMMIT_OWNER_NAME.defaultTablePropertyKey)
+        val ownerConfStr = spark.conf.get(MANAGED_COMMIT_OWNER_CONF.defaultTablePropertyKey)
+        val ownerConf = JsonUtils.fromJson[Map[String, String]](ownerConfStr)
```
ryan-johnson-databricks, line 170: aside: does this also work?
```suggestion
        val ownerConf: Map[String, String] = JsonUtils.fromJson(ownerConfStr)
```
(just scala nerdery, doesn't really matter)

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitSuite.scala
```diff
@@ -116,6 +123,405 @@ class ManagedCommitSuite
...
+        val cs = CommitStoreProvider.getCommitStore(ownerName, ownerConf)
+        val log = DeltaLog.forTable(
+          spark,
+          dataPath = new Path(tablePath),
+          options = Map.empty,
+          initialCommitStoreOpt = Some(cs),
+          tableIdentifier = None)
+        val snapshot = log.unsafeVolatileSnapshot
```
ryan-johnson-databricks, line 178: nit: Is there no suitable `forTableWithSnapshot` overload available?
(not necessarily worth adding one just for this, but worth double checking)

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitSuite.scala
```diff
@@ -116,6 +123,405 @@ class ManagedCommitSuite
...
+  }
+
+  // Test commit owner changed on concurrent cluster
+  test("snapshot is updated recursively when FS table is converted to commit owner" +
+      " table on a concurrent cluster") {
+    val commitStore = new TrackingCommitStore(new InMemoryCommitStore(batchSize = 10))
+    val builder = TrackingInMemoryCommitStoreBuilder(batchSize = 10, Some(commitStore))
+    CommitStoreProvider.registerBuilder(builder)
+    val oldCommitOwnerValue = spark.conf.get(MANAGED_COMMIT_OWNER_NAME.defaultTablePropertyKey)
+    spark.conf.unset(MANAGED_COMMIT_OWNER_NAME.defaultTablePropertyKey)
```
ryan-johnson-databricks, line 194: aside: I guess `withSqlConf` can't handle this `unset` case?

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitSuite.scala
```diff
@@ -116,6 +123,405 @@ class ManagedCommitSuite
...
+        assert(snapshotV2.version === 2)
+        assert(snapshotV2.commitStoreOpt.isEmpty)
+        DeltaLog.clearCache()
+
+        // Add new commit to convert FS table to managed-commit table
+        val deltaLog2 = DeltaLog.forTable(spark, tablePath)
+        val oldMetadata = snapshotV2.metadata
+        val newMetadata = snapshotV2.metadata.copy(
+          configuration =
+            oldMetadata.configuration + (MANAGED_COMMIT_OWNER_NAME.key -> "tracking-in-memory"))
```
ryan-johnson-databricks, line 213: nit
```suggestion
        val commitOwner = (MANAGED_COMMIT_OWNER_NAME.key -> "tracking-in-memory")
        val newMetadata = oldMetadata.copy(configuration = oldMetadata.configuration + commitOwner)
```

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitSuite.scala
```diff
@@ -116,6 +123,405 @@ class ManagedCommitSuite
...
+      assert(log.getCapturedSnapshot().updateTimestamp == clock.getTimeMillis())
+    }
+  }
+
+  // This test has the following setup:
+  // Setup:
+  // There are two table owners: CS1 and CS2 both of which are pointing to same underlying
+  // in-memory implementation.
+  // 1. Make 3 commits on the table with CS1 as owner.
+  // 2. Modify the content of commit-2. Change the table owner from 1 to 2 as part of it.
```
ryan-johnson-databricks, line 256: AFAIK, the [RFC](https://github.com/delta-io/delta/blob/master/protocol_rfcs/managed-commits.md#converting-an-existing-filesystem-based-table-to-managed-commit-table) doesn't say anything about transferring a table between commit owners?

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitSuite.scala
```diff
@@ -116,6 +123,405 @@ class ManagedCommitSuite
...
+  // 3. Do cold read from table and confirm we can construct snapshot v2 automatically. This will
+  //    need recursive snapshot update and both CS1 and CS2 will be contacted one after the other.
+  // 4. Write commit 3/4 using new commit owner.
+  // 5. Read the table again and make sure right APIs are called:
+  //    a) If read query is run in scala, we do listing 2 times. So CS2.getCommits will be called
+  //       twice. We should not be contacting CS1 anymore.
```
ryan-johnson-databricks, line 262: I guess we expect backfill somewhere along the way? But when does that occur, and wouldn't commit-2 and its backfilled counterpart disagree after we modify?

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitSuite.scala
```diff
@@ -116,6 +123,405 @@ class ManagedCommitSuite
...
+        MANAGED_COMMIT_OWNER_NAME.defaultTablePropertyKey -> builder1.name) {
+      withTempDir { tempDir =>
+        val tablePath = tempDir.getAbsolutePath
+        // Step-1: Make 3 commits on the table with CS1 as owner.
+        Seq(0).toDF.write.format("delta").mode("append").save(tablePath) // version 0
+        Seq(1).toDF.write.format("delta").mode("append").save(tablePath) // version 1
+        Seq(2).toDF.write.format("delta").mode("append").save(tablePath) // version 2
+        DeltaLog.clearCache()
+        checkAnswer(sql(s"SELECT * FROM delta.`$tablePath`"), Seq(Row(0), Row(1), Row(2)))
+        def deltaLog(): DeltaLog = DeltaLog.forTable(spark, tablePath)
```
ryan-johnson-databricks, line 318: I only see one call to this method (L350 below) and also see code reusing `log`. Maybe this method isn't needed?

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitSuite.scala
ryan-johnson-databricks: Update: Actually, batch size 10 means we don't backfill anything, and instead rely on CS2 to know about commits performed by CS1, and also rely on warm start snapshot update to start directly from CS2?

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitSuite.scala
```diff
@@ -116,6 +123,405 @@ class ManagedCommitSuite
...
+  // 5. Do commit v3 on table. As part of this, change commit-owner to CS2. This will be recorded
+  //    in both commit stores CS1, CS2 as CS1 was managing the table when the commit happened.
+  // 6. Do commit v4 on table. This will be recorded in CS2 only as CS2 is the owner now.
+  // 7. Invoke deltaLog.update() on the old deltaLog object which is still pointing to v1.
+  //    - While doing this, we will inject failure in CS2 so that it fails twice when cs2.getCommits
+  //      is called.
+  //    - Because of this old delta log will firstly contact cs1 and get newer commits  i.e. v2/v3
+  //      and create a snapshot out of it. Then it will contact cs2 and fail. So deltaLog.update()
+  //      won't succeed and throw exception. But underlying DeltaLog object now has reference to v3.
+  //      The recorded timestamp for this must be commit timestamp and not clock timestamp.
```
ryan-johnson-databricks, line 403: How could `DeltaLog` instance have a reference to v3, when `getUpdatedSnapshot` only installs the new snapshot until after the commit-owner-changed loop exits?

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitSuite.scala
```diff
@@ -116,6 +123,405 @@ class ManagedCommitSuite
...
+          logPath: Path,
+          commitVersion: Long,
+          commitFile: FileStatus,
+          commitTimestamp: Long): CommitResponse = {
+        super.commitImpl(logStore, hadoopConf, logPath, commitVersion, commitFile, commitTimestamp)
+        underlyingCommitStore2
+          .commitImpl(logStore, hadoopConf, logPath, commitVersion, commitFile, commitTimestamp)
+      }
+
+      override protected[delta] def registerBackfill(
```
ryan-johnson-databricks, line 422: nit: we could simplify the override since this class is anyway scoped to this specific test case?
```suggestion
      override def registerBackfill(
```

ryan-johnson-databricks (commented):
Nice start!

I especially like that the logic backing `getSnapshotAtInit` and `updateInternal` has been unified. The redundancy was annoying.

spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
```diff
@@ -59,15 +59,15 @@ trait SnapshotManagement { self: DeltaLog =>
...
+  protected val snapshotLock = new ReentrantLock()
+
   /**
    * Cached fileStatus for the latest CRC file seen in the deltaLog.
    */
   @volatile protected var lastSeenChecksumFileStatusOpt: Option[FileStatus] = None
   @volatile protected var currentSnapshot: CapturedSnapshot = getSnapshotAtInit
 
-  /** Use ReentrantLock to allow us to call `lockInterruptibly` */
-  protected val snapshotLock = new ReentrantLock()
```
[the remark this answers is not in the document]
prakharjain09: This is used by `getSnapshotAtInit`. So we need to define it before `getSnapshotAtInit` is called in the constructor.

spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
prakharjain09: Makes sense - will do this in a followup PR.

spark/src/main/scala/org/apache/spark/sql/delta/managedcommit/InMemoryCommitStore.scala
prakharjain09: This is a class and not a case class. So constructor is not defined.

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitSuite.scala
prakharjain09: In this test case, they are not relevant. Removing these.

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitSuite.scala
prakharjain09: Yes `forTableWithSnapshot` is not available.

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitSuite.scala
prakharjain09: Yes exactly.

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitSuite.scala
prakharjain09: We should add a section to RFC around Transferring commit ownership.

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitSuite.scala
prakharjain09: > Actually, batch size 10 means we don't backfill anything, and instead rely on CS2 to know about commits performed by CS1, and also rely on warm start snapshot update to start directly from CS2

Yes - this is correct. Even if we add a constrain in RFC that commits must be backfilled on ownership change, still our delta-spark code could recursively update snapshots across different owners even when the commits where ownership changes are not backfilled. This test case is testing the same.

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitSuite.scala
prakharjain09: The description here is out-of-sync from the actual logic. Fixing this.

spark/src/main/scala/org/apache/spark/sql/delta/managedcommit/InMemoryCommitStore.scala
ryan-johnson-databricks:
```scala
class Foo(var x: Int = -1)
val f = new Foo(10)
f.x += 10
println(f.x)
```
Works just fine and prints `20`?

(case classes just define `apply` and `unapply` helpers automatically)

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitSuite.scala
ryan-johnson-databricks: > Even if we add a constrain in RFC that commits must be backfilled on ownership change, still our delta-spark code could recursively update snapshots across different owners even when the commits where ownership changes are not backfilled. This test case is testing the same.

Why would we need to validate that scenario, once the spec forbids it?

spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
```diff
@@ -907,48 +898,102 @@ trait SnapshotManagement { self: DeltaLog =>
...
+      isAsync: Boolean): Snapshot = {
+    var commitStoreUsed = initialCommitStore
+    var newSnapshot = getSnapshotForLogSegmentInternal(
+      oldSnapshotOpt,
+      initialSegmentForNewSnapshot,
+      isAsync
+    )
+    // If the commit store has changed, we need to recursively invoke updateSnapshot so that we
+    // could get the latest commits from the new commit store.
+    while (newSnapshot.version >= 0 && newSnapshot.commitStoreOpt != commitStoreUsed) {
```
ryan-johnson-databricks, line 935: What commit owner does the initial snapshot use? If it's the default commit owner, attempting an `update` would fail because the default owner doesn't know about this (not yet existing) table? Seems like we need to force commit owner to None for initial snapshot, and let commit 0 install a commit owner if it wishes?

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitSuite.scala
prakharjain09: The RFC already talks about FS -> MC and MC  -> FS. The same could be combined to change the owners.
i.e. MC1 -> FS -> MC2.

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitSuite.scala
prakharjain09: That scenario has been removed.

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitSuite.scala
ryan-johnson-databricks: After a bunch of discussion offline, it turns out direct ownership transfers are really messy, and hard to do securely, because they require two commit owners to coordinate in some ad-hoc way. At this point it looks better to keep the RFC spec as-is:
1. If the Delta client wants to propose an owner for a FS table, they must send a commit request to the proposed owner; if the owner agrees to take over the table, they arrange for a direct-backfilled commit that makes the owner change.
2. If the Delta client wants to remove the owner of a table (making it FS based), they must send one last commit request to the current owner; if the owner agrees to disown the table, they arrange for a direct-backfilled commit that makes the owner change.

The direct-backfilled commit may be done client-side by the owner's commit store, or server-side by the owner itself (that's up to the owner to decide).

ryan-johnson-databricks: If we are in agreement about the above, we should update this test (and possibly others as well) to follow spec?

spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
```diff
@@ -231,11 +217,11 @@ trait SnapshotManagement { self: DeltaLog =>
    * @return Some LogSegment to build a Snapshot if files do exist after the given
    *         startCheckpoint. None, if the directory was missing or empty.
    */
-  protected def getLogSegmentForVersion(
+  protected def createLogSegment(
```
ryan-johnson-databricks, line 220: aside: This is a welcome change... the old name never really did make much sense.

spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
```diff
@@ -253,6 +239,12 @@ trait SnapshotManagement { self: DeltaLog =>
     )
   }
 
+  private def createLogSegment(previousSnapshot: Snapshot): Option[LogSegment] = {
```
ryan-johnson-databricks, line 242: This is just a convenience overload to create a new log segment, using the previous snapshot as starting point, right? Otherwise nothing special?
(maybe worth a quick doc comment)

spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
```diff
@@ -498,30 +490,20 @@ trait SnapshotManagement { self: DeltaLog =>
    * file as a hint on where to start listing the transaction log directory. If the _delta_log
    * directory doesn't exist, this method will return an `InitialSnapshot`.
    */
-  protected def getSnapshotAtInit: CapturedSnapshot = {
+  protected def getSnapshotAtInit: CapturedSnapshot = withSnapshotLockInterruptibly {
```
ryan-johnson-databricks, line 493: To make sure I'm understanding the new flow correctly:
* At construction time, `getSnapshotAtInit`:
  * calls `createLogSegment` with last checkpoint (if any) as the starting point
  * passes the resulting log segment to `getUpdatedSnapshot` (which handles commit owner changes)
  * directly returns the resulting `CapturedSnapshot` so it can be assigned
* At update time, `updateInternal`:
  * calls `createLogSegment` with current snapshot as the starting point
  * passes the resulting log segment to `getUpdatedSnapshot` (which handles commit owner changes)
  * calls `installSnapshot` to update or replace the captured snapshot as needed.

Nice. Much cleaner than the previous flow, which was creating log segments in different ways for init vs. update paths. Good-bye, `createSnapshotAtInitInternal`!

ryan-johnson-databricks (approved):
Looks great. Just needs some unit tests cleanup and we should be good to merge!

spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
prakharjain09: nicely summarized.

tdas (approved):

PR #2764, opened 2024-03-19
Automatic type widening in MERGE 

johanl-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

It adds automatic type widening as part of schema evolution in MERGE INTO:
- During resolution of the `DeltaMergeInto` plan, when merging the target and source schema to compute the schema after evolution, we keep the wider source type when type widening is enabled on the table.
- When updating the table schema at the beginning of MERGE execution, metadata is added to the schema to record type changes.

## How was this patch tested?
- A new test suite `DeltaTypeWideningSchemaEvolutionSuite` is added to cover type evolution in MERGE

The table feature is available in testing only, there are no user-facing changes as of now.

When automatic schema evolution is enabled in MERGE and the source schema contains a type that is wider than the target schema:

With type widening disabled: the type in the target schema is not changed. the ingestion behavior follows the `storeAssignmentPolicy` configuration:
- LEGACY: source values that overflow the target type are stored as `null`
- ANSI: a runtime check is injected to fail on source values that overflow the target type.
- STRICT: the MERGE operation fails during analysis.

With type widening enabled: the type in the target schema is updated to the wider source type. The MERGE operation always succeeds:
```
-- target: key int, value short
-- source: key int, value int
MERGE INTO target
USING source
ON target.key = source.key
WHEN MATCHED THEN UPDATE SET *
```
After the MERGE operation, the target schema is `key int, value int`.


spark/src/main/scala/org/apache/spark/sql/delta/ResolveDeltaMergeInto.scala
```diff
@@ -278,14 +278,25 @@ object ResolveDeltaMergeInto {
         })
 
       val migrationSchema = filterSchema(source.schema, Seq.empty)
+      val allowTypeWidening = EliminateSubqueryAliases(target) match {
+        case DeltaFullTable(_, index) =>
+          TypeWidening.isEnabled(
+            index.snapshotAtAnalysis.protocol,
+            index.snapshotAtAnalysis.metadata
+          )
+        case o => throw DeltaErrors.notADeltaSourceException("MERGE", Some(o))
```
sabir-akhadov, line 287: nit: o -> other
Also, this case is probably unreachable code. But outside of that it's mixing too much different concerns, one is a table feature and the other verifying a target table is a Delta table.

spark/src/main/scala/org/apache/spark/sql/delta/TypeWidening.scala
```diff
@@ -62,6 +62,19 @@ object TypeWidening {
       case _ => false
     }
 
+  /**
+   * Returns whether the given type change is eligible for **automatic** widening. Only a subset of
```
sabir-akhadov, line 66: What is automatic widening? Is it a well established term?

spark/src/test/scala/org/apache/spark/sql/delta/DeltaErrorsSuite.scala
```diff
@@ -1432,7 +1432,10 @@ trait DeltaErrorsSuiteBase
       val e = intercept[DeltaAnalysisException] {
         val s1 = StructType(Seq(StructField("c0", IntegerType, true)))
         val s2 = StructType(Seq(StructField("c0", StringType, false)))
-        SchemaMergingUtils.mergeSchemas(s1, s2, false, false, Set("c0"))
+        SchemaMergingUtils.mergeSchemas(s1, s2,
+          allowImplicitConversions = false,
+          keepExistingType = false,
+          allowTypeWidening = false, Set("c0"))
```
sabir-akhadov, line 1438:
```suggestion
          allowTypeWidening = false,
          Set("c0")
        )
```

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningAutomaticSuite.scala
```diff
@@ -0,0 +1,303 @@
...
+import org.apache.spark.sql.internal.SQLConf
+import org.apache.spark.sql.internal.SQLConf.StoreAssignmentPolicy
+import org.apache.spark.sql.types._
+
+
+/**
+ * Suite covering widening columns and fields type as part of automatic schema evolution when the
+ * type widening table feature is supported.
+ */
+class DeltaTypeWideningAutomaticSuite
```
sabir-akhadov, line 34: Sorry to do this but it probably still needs to say ..AutomaticTypeEvolutionSuite. Otherwise it's not clear what an AutomaticSuite is.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
```diff
@@ -67,15 +67,22 @@ trait DeltaTypeWideningTestMixin extends SharedSparkSession {
   protected def enableTypeWidening(tablePath: String, enabled: Boolean = true): Unit =
     sql(s"ALTER TABLE delta.`$tablePath` " +
           s"SET TBLPROPERTIES('${DeltaConfigs.ENABLE_TYPE_WIDENING.key}' = '${enabled.toString}')")
+
+  protected def metadata(
```
sabir-akhadov, line 71: nit: method name too general. What is this metadata for?

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
```diff
@@ -219,6 +201,45 @@ trait DeltaTypeWideningAlterTableTests extends QueryErrorsBase {
...
       Seq(BigDecimal("-67.89"), BigDecimal("9" * (Decimal.MAX_LONG_DIGITS - 2) + ".99")))
   )
+}
+
+/**
+ * Trait collecting a subset of tests providing core coverage for type widening using ALTER TABLE
+ * CHANGE COLUMN TYPE.
+ */
+trait DeltaTypeWideningAlterTableTests extends QueryErrorsBase with DeltaTypeWideningTestCases {
+  self: QueryTest with ParquetTest with DeltaTypeWideningTestMixin with DeltaDMLTestUtils =>
```
sabir-akhadov, line 211: I don't understand all the usecases for this type of reverse inheritance, but could these just be normal mixins or there is a deeper reason?

sabir-akhadov (commented):
left some comments.

spark/src/main/scala/org/apache/spark/sql/delta/ResolveDeltaMergeInto.scala
johanl-db: Updated.
We have a similar pattern later on in `PreprocessTableMerge` and I don't know if there are corner cases that could actually hit it - e.g. some temp view with a filter. To be safe Im' throwing the same error as in `PreprocessTableMerge` so that the behavior doesn't change when schema evolution is enabled vs. disabled

spark/src/main/scala/org/apache/spark/sql/delta/TypeWidening.scala
johanl-db: I reworded to "type change during schema evolution"

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningAutomaticSuite.scala
johanl-db: 👍 I kept the prefix `DeltaTypeWidening` so that type widening test suites are grouped together

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
johanl-db: Updated + added a comment

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
johanl-db: I use it mainly as a way to document dependencies on classes/traits: putting in the self type dependencies to test helpers/classes that are typically shared with other tests and directly inheriting traits that are specific to this test trait.

There's no functional reason behind it, if that's hurts readability I'll stop doing it.

sabir-akhadov: Yeah, it makes sense, you are basically putting limitations on which suites/traits are allowed to inherit from this one. 
One more a less legit use case I've seen is us breaking up a large MergeInto class into several traits and each trait also using the methods of the "child" class. Basically just organizing the code better without having to pass around too much state explicitly in constructor parameters or method args.

spark/src/main/scala/org/apache/spark/sql/delta/ResolveDeltaMergeInto.scala
johanl-db: Actually changed matching to make it more straight to the point: find the Delta relation (which really should always be there) if we don't find it then consider that type widening is disabled, which is likely safer since something's obviously wrong already

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSchemaEvolutionSuite.scala
```diff
@@ -0,0 +1,318 @@
...
+        // trivial insert only merge to make it easier to validate results.
+        executeMerge(
+          tgt = s"delta.`$tempPath` t",
+          src = "source",
+          cond = "0 = 1",
+          clauses = insert("*"))
+
+        assert(readDeltaTable(tempPath).schema("value").dataType === testCase.toType)
+        checkAnswer(
+          readDeltaTable(tempPath).select("value").sort("value"),
```
sabir-akhadov, line 76: nit: is the sort required? checkAnswer will sort rows anyways.

sabir-akhadov (approved):
lgtm

spark/src/main/scala/org/apache/spark/sql/delta/ResolveDeltaMergeInto.scala
```diff
@@ -278,14 +278,22 @@ object ResolveDeltaMergeInto {
         })
 
       val migrationSchema = filterSchema(source.schema, Seq.empty)
+      val allowTypeWidening = target.exists {
+        case DeltaTable(fileIndex) =>
+          TypeWidening.isEnabled(fileIndex.protocol, fileIndex.metadata)
```
tomvanbussel, line 283: What will happen if type widening is disabled after this statement, but before the main transaction starts?

spark/src/main/scala/org/apache/spark/sql/delta/TypeWidening.scala
```diff
@@ -62,6 +62,19 @@ object TypeWidening {
...
+  /**
+   * Returns whether the given type change can be applied during schema evolution. Only a
+   * subset of supported type changes are considered for schema evolution.
+   */
+  def isTypeChangeSupportedForSchemaEvolution(fromType: AtomicType, toType: AtomicType): Boolean =
+    (fromType, toType) match {
+      case (from, to) if !isTypeChangeSupported(from, to) => false
+      case (from, to) if from == to => true
+      case (ByteType, ShortType) => true
+      case (ByteType | ShortType, IntegerType) => true
```
tomvanbussel, line 74: Nit: This can be replaced by
```scala
case (ByteType | ShortType, ShortType | IntegerType) => true`
```

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSchemaEvolutionSuite.scala
```diff
@@ -0,0 +1,318 @@
...
+            src = "source",
+            cond = "0 = 1",
+            clauses = insert("*"))
+          assert(readDeltaTable(tempPath).schema("value").dataType === testCase.fromType)
+        }
+      }
+    }
+  }
+
+  test("type widening isn't applied in merge when schema evolution is disabled") {
```
tomvanbussel, line 107: Do we also have a test for when the feature/table property is disabled but `DELTA_SCHEMA_AUTO_MIGRATE` is set to true?

tomvanbussel (approved):
Nice! This is going to be really powerful when combined with the already existing schema auto migration! I have two questions and a small nit, but otherwise it LGTM.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSchemaEvolutionSuite.scala
johanl-db: We do in MergeIntoSchemaEvolutionSuite, specifically tests covering storeAssignmentPolicy, including source long to target int.

I added a simple test with type widening disabled here, as a kind of smoke test when only this suite is run and not the full MergeIntoSchemaEvolutionSuite,

spark/src/main/scala/org/apache/spark/sql/delta/TypeWidening.scala
johanl-db: It's harder to read though, I'd rather keep this split

spark/src/main/scala/org/apache/spark/sql/delta/ResolveDeltaMergeInto.scala
johanl-db: The two calls to SchemaMergingUtils.mergeSchemas (here and in ImplicitMetadataOperation) will return a different schema: first one with the wider type, second one with the original type.

This works currently because the second call to mergeSchemas doesn't allow implicit casts and will fail but it's quite brittle. I added a check at the start of MERGE to properly fail when type widening is enabled/disabled concurrently

tdas (approved):
This is awesome.

PR #2766, opened 2024-03-19
[Spark] Current Date/Time resolution in constraints

andreaschat-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Heading towards the removal of codegenFallback from Date/Time expressions (https://github.com/apache/spark/pull/44261), delta constraints need to resolve current_datetime expressions during the analysis of the invariants. The proposed changes work for both Spark 3.5 and Spark master.

## How was this patch tested?

Existing tests. In particular, `CheckConstaintsSuite` covers constraints with `current_timestamp()` expressions. Added an extra test to cover `current_date()`.

No.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/constraints/DeltaInvariantCheckerExec.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/constraints/DeltaInvariantCheckerExec.scala
@@ -17,9 +17,11 @@
 package org.apache.spark.sql.delta.constraints
 
 import scala.collection.mutable
+
 import org.apache.spark.sql.delta.{DeltaErrors, DeltaIllegalStateException}
 import org.apache.spark.sql.delta.constraints.Constraints.{Check, NotNull}
 import org.apache.spark.sql.delta.schema.SchemaUtils
+
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.catalyst.InternalRow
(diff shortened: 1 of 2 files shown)


tdas (changes requested):
any unit tests?

spark/src/main/scala/org/apache/spark/sql/delta/constraints/DeltaInvariantCheckerExec.scala
```diff
@@ -71,9 +73,16 @@ case class DeltaInvariantCheckerExec(
...
       DeltaInvariantCheckerExec.buildInvariantChecks(child.output, constraints, session)
-    val boundRefs = invariantChecks.map(_.withBoundReferences(child.output))
+
+    // Resolve current_date()/current_time() expressions.
+    // We resolve currentTime for all invariants together to make sure we use the same timestamp.
+    val invariantsFakePlan = AnalysisHelper.FakeLogicalPlan(invariantChecks, Nil)
+    val newInvariantsPlan = optimizer.ComputeCurrentTime(invariantsFakePlan)
 
     child.execute().mapPartitionsInternal { rows =>
+      val boundRefs = newInvariantsPlan.expressions
```
cloud-fan, line 83: can we put the `boundRefs` variable out of the RDD `mapPartitionsInternal` closure as before?

andreaschat-db: I moved it there because I was having a serialization issue that only appeared in some Spark versions. Are there any issues for having the `boundRefs` there?

cloud-fan: referencing `child.output` in the RDD closure means serializing the entire `child` plan. We should at least do `val localOutput = child.output` before the `mapPartitionsInternal` call and reference this local variable in the closure.

andreaschat-db: Done thanks!

cloud-fan (approved):

tdas (approved):
This is great. Thank you for making this such that it will work with both spark master and 3.5. 

My only worry is that we are using `optimizer.ComputeCurrentTime` .. which is spark internal API, and there is a risk that this is not going to stable across spark versions.

PR #2777, opened 2024-03-20
[Spark] OPTIMIZE on clustered table with no clustering columns should run compaction

zedtang:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Currently, when running OPTIMIZE on a clustered table without any clustering columns(after ALTER TABLE CLUSTER BY NONE), it would fail with a long stack trace:
```
[info]   org.apache.spark.SparkException: Exception thrown in awaitResult:
[info]   at org.apache.spark.util.SparkThreadUtils$.awaitResult(SparkThreadUtils.scala:56)
[info]   at org.apache.spark.util.ThreadUtils$.awaitResult(ThreadUtils.scala:310)
[info]   at org.apache.spark.util.ThreadUtils$.parmap(ThreadUtils.scala:383)
[info]   at org.apache.spark.sql.delta.commands.OptimizeExecutor.$anonfun$optimize$1(OptimizeTableCommand.scala:276)
...
[info]   Cause: java.util.concurrent.ExecutionException: Boxed Error
[info]   at scala.concurrent.impl.Promise$.resolver(Promise.scala:87)
...
[info]   Cause: java.lang.AssertionError: assertion failed: Cannot cluster by zero columns!
[info]   at scala.Predef$.assert(Predef.scala:223)
[info]   at org.apache.spark.sql.delta.skipping.MultiDimClustering$.cluster(MultiDimClustering.scala:51)
[info]   at org.apache.spark.sql.delta.commands.OptimizeExecutor.runOptimizeBinJob(OptimizeTableCommand.scala:427)
[info]   at org.apache.spark.sql.delta.commands.OptimizeExecutor.$anonfun$optimize$6(OptimizeTableCommand.scala:277)
...
```

This change makes OPTIMIZE on a clustered table without any clustering columns run regular compaction, which is the desired behavior.
## How was this patch tested?

This change adds a new test to verify the correct behavior. The test would fail without the fix.

No.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala
@@ -239,7 +239,9 @@ class OptimizeExecutor(
 
   private val isClusteredTable = ClusteredTableUtils.isSupported(txn.snapshot.protocol)
 
-  private val isMultiDimClustering = isClusteredTable || zOrderByColumns.nonEmpty
+  private val isMultiDimClustering =
+    optimizeStrategy.isInstanceOf[ClusteringStrategy] ||
+    optimizeStrategy.isInstanceOf[ZOrderStrategy]
 
   private val clusteringColumns: Seq[String] = {
     if (zOrderByColumns.nonEmpty) {
--- a/spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableStrategy.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableStrategy.scala
@@ -118,8 +118,9 @@ object OptimizeTableStrategy {
 
   private def getMode(snapshot: Snapshot, zOrderBy: Seq[String]): OptimizeTableMode.Value = {
     val isClusteredTable = ClusteredTableUtils.isSupported(snapshot.protocol)
+    val hasClusteringColumns = ClusteringColumnInfo.extractLogicalNames(snapshot).nonEmpty
     val isZOrderBy = zOrderBy.nonEmpty
-    if (isClusteredTable) {
+    if (isClusteredTable && hasClusteringColumns) {
       assert(!isZOrderBy)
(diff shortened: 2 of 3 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala
```diff
@@ -239,7 +239,9 @@ class OptimizeExecutor(
 
   private val isClusteredTable = ClusteredTableUtils.isSupported(txn.snapshot.protocol)
 
-  private val isMultiDimClustering = isClusteredTable || zOrderByColumns.nonEmpty
+  private val isMultiDimClustering =
+    optimizeStrategy.isInstanceOf[ClusteringStrategy] ||
+    optimizeStrategy.isInstanceOf[ZOrderStrategy]
```
dabao521, line 244: Does this fix something or just a good refactoring?

dabao521 (approved):

spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala
zedtang: We need both this fix and the fix in OptimizeTableStrategy:
* isMultiDimClustering: controls whether we run `MultiDimClustering.cluster` or compaction ([pointer](https://github.com/delta-io/delta/blob/a11b92de2850dfbe439fad6ec7cca747be503b77/spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala#L423))
* Fix in getMode in OptimizeTableStrategy: to determine the correct mode

zedtang: Note that the previous logic here was wrong:
`isClusteredTable || zOrderByColumns.nonEmpty` should be
`(isClusteredTable && hasClusteringColumns) || zOrderByColumns.nonEmpty`

tdas (approved):

PR #2785, opened 2024-03-22
Automatic type widening in INSERT

johanl-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

It adds automatic type widening as part of schema evolution in INSERT. During resolution, when schema evolution and type widening are enabled, type differences between the input query and the target table are handled as follows:
- If the type difference qualifies for automatic type evolution: the input type is left as is, the data will be inserted with the new type and the table schema will be updated in `ImplicitMetadataOperation` (already implemented as part of MERGE support)
- If the type difference doesn't qualify for automatic type evolution: the current behavior is preserved: a cast is added from the input type to the existing target type.

## How was this patch tested?
- Tests are added to `DeltaTypeWideningAutomaticSuite` to cover type evolution in INSERT

When automatic schema evolution is enabled in INSERT and the source schema contains a type that is wider than the target schema:

With type widening disabled: the type in the target schema is not changed. A cast is added to the input to insert to match the expected target type.

With type widening enabled: the type in the target schema is updated to the wider source type. 
```
-- target: key int, value short
-- source: key int, value int
INSERT INTO target SELECT * FROM source
```
After the INSERT operation, the target schema is `key int, value int`.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaAnalysis.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaAnalysis.scala
@@ -66,7 +66,7 @@ import org.apache.spark.sql.execution.datasources.parquet.ParquetFileFormat
 import org.apache.spark.sql.execution.datasources.v2.DataSourceV2Relation
 import org.apache.spark.sql.execution.streaming.StreamingRelation
 import org.apache.spark.sql.internal.SQLConf
-import org.apache.spark.sql.types.{ArrayType, DataType, IntegerType, MapType, StructField, StructType}
+import org.apache.spark.sql.types._
 import org.apache.spark.sql.util.CaseInsensitiveStringMap
 
 /**
@@ -81,8 +81,8 @@ class DeltaAnalysis(session: SparkSession)
   override def apply(plan: LogicalPlan): LogicalPlan = plan.resolveOperatorsDown {
     // INSERT INTO by ordinal and df.insertInto()
     case a @ AppendDelta(r, d) if !a.isByName &&
-        needsSchemaAdjustmentByOrdinal(d.name(), a.query, r.schema) =>
-      val projection = resolveQueryColumnsByOrdinal(a.query, r.output, d.name())
+        needsSchemaAdjustmentByOrdinal(d, a.query, r.schema) =>
+      val projection = resolveQueryColumnsByOrdinal(a.query, r.output, d)
       if (projection != a.query) {
         a.copy(query = projection)
       } else {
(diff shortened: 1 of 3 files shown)


spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSchemaEvolutionSuite.scala
```diff
@@ -0,0 +1,810 @@
...
+      .add("a", ArrayType(IntegerType), nullable = true,
+        metadata = typeWideningMetadata(
+          version = 1,
+          from = ShortType,
+          to = IntegerType,
+          path = Seq("element")))
+  )
+
+
+    testInsertTypeEvolution("nested struct type evolution with field upcast")(
```
tomvanbussel, line 731: Nit: indentation is off

spark/src/main/scala/org/apache/spark/sql/delta/DeltaAnalysis.scala
```diff
@@ -890,47 +892,69 @@ class DeltaAnalysis(session: SparkSession)
...
       case (s, t) if s == t =>
         attr
       case (s: StructType, t: StructType) if s != t =>
-        addCastsToStructs(tblName, attr, s, t)
+        addCastsToStructs(tblName, attr, s, t, allowTypeWidening)
       case (ArrayType(s: StructType, sNull: Boolean), ArrayType(t: StructType, tNull: Boolean))
           if s != t && sNull == tNull =>
-        addCastsToArrayStructs(tblName, attr, s, t, sNull)
+        addCastsToArrayStructs(tblName, attr, s, t, sNull, allowTypeWidening)
+      case (s: AtomicType, t: AtomicType)
```
tomvanbussel, line 915: What about structs that can be evolved using type evolution? Can this case be handled without `addCastsToStructs`?

tomvanbussel (approved):
Nice! This is going to be really convenient! I have a question and a small nit, but otherwise it LGTM

spark/src/main/scala/org/apache/spark/sql/delta/DeltaAnalysis.scala
johanl-db: Yes, you could check that the struct only strictly contains widening type changes - no new fields or other type differences as that would require a cast.
In that case we don't currently modify the struct, just deconstruct it and reconstruct it as it was, which would land itself to being optimized away by an optimzier rule (not sure if that happens though).

This isn't going to happen often though, it means a type in that struct will actually be widened during this operation. I'd rather keep the code simpler and less error prone rather than trying to optimize for this case by comparing the input and output structs for new columns or non-widening type changes

tdas (approved):
oh this is amazing.

PR #2789, opened 2024-03-22
Refactor to Delta shallow clone

lzlfred:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

minor refactor to delta shallow clone to reorder and simplify logics. 

## How was this patch tested?

UT


mingdai-db (approved):

tdas (approved):
generally i dont like refactors by themselves, but this PR is small and it removes more line than it adds. So I am okay with this.

PR #2790, opened 2024-03-22
[BUG] Default columns is also a reader feature, the PROTOCOL doc is wrong.

mwc360:
The docs here correctly state that enabling Default Columns will prevent older versions without the Default Column feature from reading the table: https://docs.delta.io/3.1.0/delta-default-columns.html
The PROTOCOL doc conflicts with the above. I tested and confirmed that Default Columns is a table feature that requires the reader to support.

Signed-off-by: Miles Cole <m.w.c.360@gmail.com>

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Documentation needs updating

## How was this patch tested?
Just documentation

No

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -1601,7 +1601,7 @@ Feature | Name | Readers or Writers?
 [Column Invariants](#column-invariants) | `invariants` | Writers only
 [`CHECK` constraints](#check-constraints) | `checkConstraints` | Writers only
 [Generated Columns](#generated-columns) | `generatedColumns` | Writers only
-[Default Columns](#default-columns) | `allowColumnDefaults` | Writers only
+[Default Columns](#default-columns) | `allowColumnDefaults` | Readers and Writers
 [Change Data Feed](#add-cdc-file) | `changeDataFeed` | Writers only
 [Column Mapping](#column-mapping) | `columnMapping` | Readers and writers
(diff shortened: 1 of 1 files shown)


felipepessoto:
Previously the decision was to make it a write only feature. That changed? Or docs are incorrect?

https://github.com/delta-io/delta/pull/2240#issuecomment-1788109698

mwc360:
> Previously the decision was to make it a write only feature. That changed? Or docs are incorrect?
> 
> [#2240 (comment)](https://github.com/delta-io/delta/pull/2240#issuecomment-1788109698)
You're right, something with my testing was wrong. It does not block reader access for older delta versions,

The documentation [here ](https://docs.delta.io/3.1.0/delta-default-columns.html) incorrectly notes that "Enabling default column values for a table upgrades the Delta [table version](https://docs.delta.io/3.1.0/versioning.html) as a byproduct of enabling table features. This protocol upgrade is irreversible. **Tables with default column values enabled can only be read in Delta Lake 3.1 and above.**"

mwc360:
I'll cancel this PR and open one up for the doc that should be fixed. thanks!

mwc360:
New PR opened here: https://github.com/delta-io/delta/pull/2794

PR #2792, opened 2024-03-23
[WIP] First draft : Support least common type coalesce

### LINKED ISSUE #2417: [Kernel][Defaults] Support `COALESCE` for different data types using least-common-type resolution in `DefaultExpressionHandler`

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

### Overview

#2415 adds support for `COALESCE` for boolean type expressions and #2416 is to add support for all data types (needs to be finished before this issue). This issue is to support different `expr` data types if they can all be up-casted to one least-common-type.

### Further details

Need to update `DefaultExpressionEvaluator` to find a least-common-type if it exists and add implicit casts for each expression.

zzl-7:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description
Rough draft to support least common type resolution for coalesce

Resolves https://github.com/delta-io/delta/issues/2417

## How was this patch tested?
Unit test, with new test added


kernel/kernel-api/src/main/java/io/delta/kernel/expressions/ScalarExpression.java
```diff
@@ -73,4 +80,11 @@ public String getName() {
     public List<Expression> getChildren() {
         return children;
     }
+
+    public DataType getOutputType() {
+        if (this.outputType == null) {
```
zzl-7, line 85: This might not be the best idea, but we need to propagate outputType to avoid calculating leas common type multiple times, should we make all ScalarExpression have outputType?

kernel/kernel-api/src/main/java/io/delta/kernel/types/ArrayType.java
```diff
@@ -35,6 +36,11 @@ public ArrayType(DataType elementType, boolean containsNull) {
         this.containsNull = containsNull;
     }
 
+    @Override
+    public byte getPromotionGroup(DataType dataType) {
```
zzl-7, line 40: This is redundant code 
```
    @Override
    public byte getPromotionGroup(DataType dataType) {
        return typePromotionGroup;
    }
```
But need to be here because polymorphism does not play nicely with static variable, the alternative is to make `typePromotionGroup` an instance variable instead, neither option looks clean, perhaps code reviewers can point me a better alternative.

kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/internal/data/ColumnVectorConverter.java
```diff
@@ -0,0 +1,210 @@
...
+
+public class ColumnVectorConverter {
+    public static ColumnVector convertToOutputType(ColumnVector c, DataType outputType) {
+        DataType inputType = c.getDataType();
+        if (inputType == outputType) {
+            return c;
+        }
+        if (outputType instanceof BooleanType) {
+            // TODO: how to convert string to boolean?
+            // "true" "TRUE", "T" -> true ?
```
zzl-7, line 33: I couldn't find info on how to resolve string to boolean https://spark.apache.org/docs/latest/sql-ref-datatypes.html 
So I am making the least common type resolution throwing an error temporary.

kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/internal/data/ColumnVectorConverter.java
```diff
@@ -0,0 +1,210 @@
...
+            return convertToInteger(c);
+        } else if (outputType instanceof LongType) {
+            return convertToLong(c);
+        } else if (outputType instanceof DecimalType) {
+            return convertToDecimal(c);
+        } else if (outputType instanceof FloatType) {
+            return convertToFloat(c);
+        } else if (outputType instanceof DoubleType) {
+            return convertToDouble(c);
+        } else {
```
zzl-7, line 47: Its not clear to me if complex type can be recursive from here https://spark.apache.org/docs/latest/sql-ref-datatypes.html , I am guessing yes, but would like to confirm with code reviewer thats the case before handling complex least common type resolution for its element

> *** For a complex type, the precedence rule applies recursively to its component elements.

kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/internal/data/ColumnVectorConverter.java
```diff
@@ -0,0 +1,210 @@
...
+            String msg = String.format(
+                    "`%s` is not supported to convert to data type %s`",
+                    inputType, outputType);
+            throw new IllegalStateException(msg);
+        }
+    }
+
+    protected static ColumnVector convertToShort(ColumnVector c) {
```
zzl-7, line 55: There is too much boilerplate code here, every conversion have to handle all possible least common type conversion with the same destination outputType (ie. double have to handle byte, short, int, long, decimal, float, string) , I am not confident that this is the best way to handle this, next step is to play around with using Object[] to see if we can decrease boilerplate code.

kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/internal/types/LeastCommonTypeResolver.java
```diff
@@ -0,0 +1,284 @@
...
+            default:
+                String msg = String.format(
+                        "`%s` is not supported for least common type " +
+                                "resolution for number related type`",
+                        precedence);
+                throw new IllegalStateException(msg);
+        }
+    }
+    public static boolean isCompatible(DataType a, DataType b) {
+        if (a instanceof ByteType) {
```
zzl-7, line 218: Having a static immutable set also make sense, it seems like we can't use `set.of` operation due to java version, so I opt to enumerate it here instead of creating a set by adding possible value one by one.

kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/internal/data/ColumnVectorConverter.java
```diff
@@ -0,0 +1,210 @@
...
+            } else if (d instanceof LongType) {
+                values[i] = new BigDecimal(c.getLong(i));
+            } else {
+                String msg = String.format(
+                        "`%s` cannot convert to big decimal`",
+                        d);
+                throw new IllegalStateException(msg);
+            }
+        }
+        return new DefaultDecimalVector(
```
zzl-7, line 145: There is some choice here if original column have non default `int precision, int scale` should we honor that when we do the conversion?

foss-contributor:
@zzl-7, thank you for your contribution! This pull request has been inactive for over 18 months and will be closed in one week. To keep it open, please leave a comment on this pull request or contact us in the `#contributors` channel on [Delta Lake Slack](https://go.delta.io/slack). We appreciate your involvement in Delta Lake!

PR #2794, opened 2024-03-24
[BUG] Default columns is a writer table feature, not a reader feature.

mwc360:
The documentation incorrectly notes that "Enabling default column values for a table upgrades the Delta table version as a byproduct of enabling table features. This protocol upgrade is irreversible. Tables with default column values enabled can only be read in Delta Lake 3.1 and above."

Signed-off-by: Miles Cole [m.w.c.360@gmail.com](mailto:m.w.c.360@gmail.com)

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Default columns is a writer table feature, not a reader feature.

## How was this patch tested?
I confirmed that a table with the defaultColumns table feature was readable by delta lake 3.0

No

### DIFF
--- a/docs/source/delta-default-columns.md
+++ b/docs/source/delta-default-columns.md
@@ -12,7 +12,7 @@ This information is stored in the [StructField](https://github.com/delta-io/delt
 
 ## How to enable <Delta> default column values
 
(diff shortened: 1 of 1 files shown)


xupefei:
It looks good to me -- the current PROTOCOL doc is wrong. The feature is writer-only.

Also could you, by the way, improve the message? To something like
```
Enabling default column values will upgrade the Delta table [protocol version](versioning.md) to support [table features](#versioning). This protocol upgrade is irreversible.
```

dtenedor (approved):
LGTM, this was my mistake. Thanks for fixing it.

PR #2796, opened 2024-03-25
[Spark]Optimize Directory Creation in ensureLogDirectoryExists method

sumeet-db:
- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
This PR updates the ensureLogDirectoryExist method to optimistically handle directory creation by attempting to create the directory before checking its existence. This is efficient because we're assuming it's more likely that the directory doesn't exist and it saves an filesystem existence check in that case.

Cloud Hadoop implementations are expected to throw org.apache.hadoop.fs.FileAlreadyExistsException if the file already exists. e.g. [S3AFileSystem.java](https://github.com/apache/hadoop/blob/fc166d3aec7c95110a8cd4ef6ce1fbf4955107e5/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L3763)

Unix-based File Systems are expected to throw java.nio.file.FileAlreadyExistsException if the file already exists.

To cover all bases, including unforeseen file systems, it retains a final existence check for exceptions outside these specific cases, ensuring robustness without compromising functionality.

## How was this patch tested?

Existing UTs

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala
@@ -439,15 +439,28 @@ class DeltaLog private(
   def ensureLogDirectoryExist(): Unit = {
     val fs = logPath.getFileSystem(newDeltaHadoopConf())
     def createDirIfNotExists(path: Path): Unit = {
-      var success = false
-      try {
-        success = fs.exists(path) || fs.mkdirs(path)
+      // Optimistically attempt to create the directory first without checking its existence.
+      // This is efficient because we're assuming it's more likely that the directory doesn't
+      // exist and it saves an filesystem existence check in that case.
+      val (success, mkdirsIOExceptionOpt) = try {
+        // Return value of false should mean the directory already existed (not an error) but
+        // we will verify below because we're paranoid about buggy FileSystem implementations.
+        (fs.mkdirs(path), None)
       } catch {
+        // Only needed because buggy Hadoop FileSystem.mkdir wrongly throws on existing dir.
         case io: IOException =>
-          throw DeltaErrors.cannotCreateLogPathException(logPath.toString, io)
+          val dirExists =
+            try {
+              fs.getFileStatus(path).isDirectory
+            } catch {
+              case NonFatal(_) => false
+            }
+          (dirExists, Some(io))
       }
       if (!success) {
-        throw DeltaErrors.cannotCreateLogPathException(logPath.toString)
+        throw DeltaErrors.cannotCreateLogPathException(
+          logPath = logPath.toString,
+          cause = mkdirsIOExceptionOpt.orNull)
       }
     }
     createDirIfNotExists(FileNames.commitDirPath(logPath))
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala
@@ -447,7 +447,9 @@ class DeltaLog private(
         // we will verify below because we're paranoid about buggy FileSystem implementations.
         (fs.mkdirs(path), None)
       } catch {
-        // Only needed because buggy Hadoop FileSystem.mkdir wrongly throws on existing dir.
+        // A FileAlreadyExistsException is expected if a non-directory object exists but an explicit
+        // check is needed because buggy Hadoop FileSystem.mkdir wrongly throws the exception even
+        // on existing directories.
         case io: IOException =>
           val dirExists =
             try {
--- a/spark/src/main/scala/org/apache/spark/sql/delta/hooks/CheckpointHook.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/hooks/CheckpointHook.scala
@@ -33,8 +33,6 @@ object CheckpointHook extends PostCommitHook {
       committedActions: Seq[Action]): Unit = {
     if (!txn.needsCheckpoint) return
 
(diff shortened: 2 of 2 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala
```diff
@@ -439,16 +439,27 @@ class DeltaLog private(
...
+        // Optimistically attempt to create the directory first without checking its existence.
+        // This is efficient because we're assuming it's more likely that the directory doesn't
+        // exist and it saves an filesystem existence check in that case.
+        if (!fs.mkdirs(path) && !fs.exists(path)) {
+          throw DeltaErrors.cannotCreateLogPathException(logPath.toString)
+        }
       } catch {
+        // Cloud Hadoop File Systems implementations are expected to throw
+        // org.apache.hadoop.fs.FileAlreadyExistsException if the file already exists.
+        // e.g. S3AFileSystem.
```
scovich, line 452: Actually... Hadoop [FileSystem.mkdirs](https://hadoop.apache.org/docs/current/api/org/apache/hadoop/fs/FileSystem.html#mkdirs-org.apache.hadoop.fs.Path-) documentation _specifically_ states that the method should return `false` if the directory already exists, and that 
> Existence of the directory hierarchy is not an error.

So this is a flat-out bug in `S3AFileSystem`. We should file the bug report with them, and update this doc comment accordingly.

spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala
```diff
@@ -439,16 +439,27 @@ class DeltaLog private(
...
+        case _: org.apache.hadoop.fs.FileAlreadyExistsException => // Ignore this exception
+        // Unix-based File Systems might throw java.nio.file.FileAlreadyExistsException
+        // if the file already exists.
+        case _: java.nio.file.FileAlreadyExistsException => // Ignore this exception
```
scovich, line 456: These two catches are unsafe -- the docs for [Files.createDirectories](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#createDirectories-java.nio.file.Path-java.nio.file.attribute.FileAttribute...-) state that 
> an exception is not thrown if the directory could not be created because it already exists. 

... except for one special case that we definitely don't want to ignore:
>  `FileAlreadyExistsException` - if dir exists but is not a directory (optional specific exception)

Thus, we can only ignore the exception if we prove the existing path is a directory.

spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala
```diff
@@ -439,16 +439,27 @@ class DeltaLog private(
...
+        // Catching a generic Exception after specific ones ensures that no unexpected exception
+        // type goes unnoticed. However, if the directory exists, we ignore it too.
+        case _: Exception if fs.exists(path) => // Ignore this exception
```
scovich, line 459: Also unsafe, unless we prove the existing path is a directory.

spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala
scovich: I think the safest approach is to adjust the existing code to do something like this:
```scala
var success = try {
  // NOTE: Return value of false should mean the directory already existed (not an error) but
  // we will verify below because we're paranoid about buggy FileSystem implementations.
  fs.mkdirs(...)
} catch {
  // NOTE: Only needed because buggy Hadoop FileSystem.mkdir wrongly throws on existing dir.
  case _: IOException => false
}
if (!success) {
  success = try {
    fs.getFileStatus(...).isDir
  } catch {
    NonFatal(_) => false
  }
}
if (!success) {
  throw DeltaErrors.cannotCreateLogPathException(...)
}
```

sumeet-db: Added this + some handling for IOExceptions (which we already had)

spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala
```diff
@@ -439,15 +439,28 @@ class DeltaLog private(
...
-        success = fs.exists(path) || fs.mkdirs(path)
+      // Optimistically attempt to create the directory first without checking its existence.
+      // This is efficient because we're assuming it's more likely that the directory doesn't
+      // exist and it saves an filesystem existence check in that case.
+      val (success, mkdirsIOExceptionOpt) = try {
+        // Return value of false should mean the directory already existed (not an error) but
+        // we will verify below because we're paranoid about buggy FileSystem implementations.
+        (fs.mkdirs(path), None)
       } catch {
+        // Only needed because buggy Hadoop FileSystem.mkdir wrongly throws on existing dir.
```
scovich, line 450: This is tricky... if a not-directory object already exists with the same name, then a `FileAlreadyExistsException` can and should be thrown. The bug is that HDFS _also_ throws if the already-existing object is a proper directory (so we have to double check whether the exception was wrongly thrown).

scovich (approved):
Nice -- one optional code comment nit

spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala
sumeet-db: Updated the comment.

PR #2807, opened 2024-03-26
[Spark] Move VacuumProtocolCheck Table Feature from Dev/Testing to Prod

sumeet-db:
- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
We move the Table Feature from dev/test only to prod as part of this PR.

## How was this patch tested?
Existing UTs

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/TableFeature.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/TableFeature.scala
@@ -335,6 +335,7 @@ object TableFeature {
       IcebergCompatV1TableFeature,
       IcebergCompatV2TableFeature,
       DeletionVectorsTableFeature,
+      VacuumProtocolCheckTableFeature,
       V2CheckpointTableFeature)
     if (DeltaUtils.isTesting) {
       features ++= Set(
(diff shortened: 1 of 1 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/TableFeature.scala
```diff
@@ -335,6 +335,7 @@ object TableFeature {
       IcebergCompatV1TableFeature,
       IcebergCompatV2TableFeature,
       DeletionVectorsTableFeature,
+      VacuumProtocolCheckTableFeature,
```
prakharjain09, line 338: As part of this change, we should also finalize the vacuumProtocolCheck RFC and update the Delta spec.

sumeet-db: Added it in a separate PR https://github.com/delta-io/delta/pull/2808/files

prakharjain09 (approved):

spark/src/main/scala/org/apache/spark/sql/delta/TableFeature.scala
```diff
@@ -682,7 +682,7 @@ object InCommitTimestampTableFeature
  * feature exists
  */
 object VacuumProtocolCheckTableFeature
-  extends ReaderWriterFeature(name = "vacuumProtocolCheck-dev")
+  extends ReaderWriterFeature(name = "vacuumProtocolCheck")
```
tdas, line 685: isnt there a change needed in the tests to make sure that you dont have to enable some testing-only internal flag to use this feature?

sumeet-db: The change to ensure the feature is available without enabling a testing-only internal flag is included in the current file by moving it from testing-set to prod-set. 

```
if (DeltaUtils.isTesting) {
      features ++= Set(
        TestLegacyWriterFeature, ....
```

tdas: I see. thats how it works. good.

tdas (approved):

PR #2811, opened 2024-03-27
[Spark][TEST-ONLY] Add DDLTestUtils helper

c27kwan:
- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
In this PR, we add DDLTestUtils helper and ColumnSpec traits so that we have flexible interfaces for defining Delta tables. This will enable us to test both Scala and SQL interfaces easier. This PR introduces the helpers but this interface will not be used until later. 

We don't reuse the helper functions in GeneratedColumnSuite because they are less flexible. We can rewrite GeneratedColumnSuite to use DDLTestUtils later on.

## How was this patch tested?
Unused code. Introducing the interface only.

No.

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/delta/DDLTestUtils.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/DDLTestUtils.scala
@@ -1,0 +1,193 @@
+/*
+ * Copyright (2021) The Delta Lake Project Authors.
+ *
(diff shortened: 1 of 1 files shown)


c27kwan:
@larsk-db , can you review this PR? Thanks!

larsk-db (approved):
LGTM

PR #2813, opened 2024-03-27
[Spark] InCommitTimestamps: Use parallel radix search for ICT commit history

dhruvarya-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Makes DeltaHistoryManager.getActiveCommitAtTime aware of `inCommitTimestamps`. DeltaHistoryManager.getActiveCommitAtTime will now bisect the history of commits into non-ICT and ICT commits based on the ICT enablement version and only search the range of commits relevant to the timestamp. If the relevant commit range is composed of ICT commits, `getActiveCommitAtTime` will perform a parallelized search (see `getActiveCommitAtTimeFromICTRange` for details).

## How was this patch tested?

New tests in InCommitTimestampSuite.

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
@@ -21,19 +21,24 @@ import java.io.FileNotFoundException
 import java.sql.Timestamp
 
 import scala.collection.mutable
+import scala.concurrent.{ExecutionContext, ExecutionContextExecutorService, Future}
+import scala.concurrent.duration.Duration
 
 import org.apache.spark.sql.delta.actions.{Action, CommitInfo, CommitMarker, JobInfo, NotebookInfo}
 import org.apache.spark.sql.delta.metering.DeltaLogging
+import org.apache.spark.sql.delta.sources.DeltaSQLConf
 import org.apache.spark.sql.delta.storage.LogStore
 import org.apache.spark.sql.delta.util.{DateTimeUtils, DeltaCommitFileProvider, FileNames, TimestampFormatter}
 import org.apache.spark.sql.delta.util.FileNames._
+import org.apache.spark.sql.delta.util.threads.DeltaThreadPool
+import org.apache.spark.sql.util.ScalaExtensions._
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.{FileStatus, Path}
 
 import org.apache.spark.SparkEnv
 import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.internal.SQLConf
-import org.apache.spark.util.SerializableConfiguration
+import org.apache.spark.util.{SerializableConfiguration, ThreadUtils}
 
 /**
  * This class keeps tracks of the version of commits and their timestamps for a Delta table to
@@ -104,6 +109,25 @@ class DeltaHistoryManager(
   }
 
   /**
+   * Returns the latest commit that happened at or before `time` in the range `[start, end)`.
+   * All the commits in the range `[start, end)` are assumed to not have inCommitTimestamps.
+   * If no such commit exists, the earliest commit is returned.
+   */
+  def getCommitFromNonICTRange(start: Long, end: Long, time: Long): Commit = {
+    if (end - start > 2 * maxKeysPerList) {
+      parallelSearch(time, start, end)
+    } else {
+      val commits = getCommits(
+        deltaLog.store,
+        deltaLog.logPath,
+        start,
+        Some(end),
+        deltaLog.newDeltaHadoopConf())
+      lastCommitBeforeTimestamp(commits, time).getOrElse(commits.head)
+    }
+  }
+
+  /**
    * Returns the latest commit that happened at or before `time`.
    * @param timestamp The timestamp to search for
    * @param canReturnLastCommit Whether we can return the latest version of the table if the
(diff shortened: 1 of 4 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -303,6 +401,127 @@ object DeltaHistoryManager extends DeltaLogging {
...
+              .orElse {
+                getFirstCommitAndICTAfter(
+                  chunkStart,
+                  upperBoundExclusive = getChunkEnd(chunkStart),
+                  basePath,
+                  logStore,
+                  conf,
+                  commitFileProvider
+                )
+              }
```
scovich, line 502: We can use [Future.successful](https://www.scala-lang.org/api/2.12.6/scala/concurrent/Future$.html#successful[T](result:T):scala.concurrent.Future[T]) instead of burning a thread pool entry on chunk starts:
```suggestion
          if (chunkStart == curStartCommit.version) {
            Future.successful(curStartCommit)
          } else {
            threadPool.submit(spark) {
              getFirstCommitAndICTAfter(
                chunkStart,
                upperBoundExclusive = getChunkEnd(chunkStart),
                basePath,
                logStore,
                conf,
                commitFileProvider
              )
            }
          }
```

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -303,6 +401,127 @@ object DeltaHistoryManager extends DeltaLogging {
...
+          }
+        }
+      val knownTightestLowerBoundCommit =
+        chunkStartICTFutures.map(ThreadUtils.awaitResult(_, Duration.Inf))
+          .takeWhile(_.forall(_.timestamp <= searchTimestamp))
+          .flatten
+          .lastOption
+          .getOrElse {
+            return None
+          }
```
scovich, line 512: odd indentation...
```suggestion
      val knownTightestLowerBoundCommit = chunkStartICTFutures
        .map(ThreadUtils.awaitResult(_, Duration.Inf))
        .takeWhile(_.forall(_.timestamp <= searchTimestamp))
        .flatten
        .lastOption
        .getOrElse {
          return None
        }
```

spark/src/main/scala/org/apache/spark/sql/delta/util/InCommitTimestampUtils.scala
```diff
@@ -63,4 +63,21 @@ object InCommitTimestampUtils {
...
+      DeltaConfigs.IN_COMMIT_TIMESTAMP_ENABLEMENT_TIMESTAMP.fromMetaData(metadata)
+    val enablementVersionOpt =
+      DeltaConfigs.IN_COMMIT_TIMESTAMP_ENABLEMENT_VERSION.fromMetaData(metadata)
+    (enablementTimestampOpt, enablementVersionOpt) match {
+      case (Some(enablementTimestamp), Some(enablementVersion)) =>
+        Some(DeltaHistoryManager.Commit(enablementVersion, enablementTimestamp))
+      case (None, None) =>
+        None
+      case _ =>
+        // TODO: create a proper DeltaError message.
```
scovich, line 78: Is it really a user-facing error that needs a "proper" message?

spark/src/test/scala/org/apache/spark/sql/delta/InCommitTimestampSuite.scala
```diff
@@ -325,7 +328,8 @@ class InCommitTimestampSuite
       val ver0Snapshot = deltaLog.snapshot
       assert(DeltaConfigs.IN_COMMIT_TIMESTAMPS_ENABLED.fromMetaData(ver0Snapshot.metadata))
       // Disable ICT in version 1.
-      spark.sql(s"ALTER TABLE delta.`${tempDir.getAbsolutePath}`" +
+      spark.sql(
+        s"ALTER TABLE delta.`${tempDir.getAbsolutePath}`" +
```
scovich, line 332: nit: Missing space between the two string segments? But I guess the SQL parser didn't care?
```suggestion
        s"ALTER TABLE delta.`${tempDir.getAbsolutePath}` " +
```

spark/src/test/scala/org/apache/spark/sql/delta/InCommitTimestampSuite.scala
```diff
@@ -441,5 +445,401 @@ class InCommitTimestampSuite
...
     }
   }
+
+  test("DeltaHistoryManager.getActiveCommitAtTimeFromICTRange") {
+    withTempDir { tempDir =>
+      spark.range(10).write.format("delta").save(tempDir.getAbsolutePath)
+      val startTime = System.currentTimeMillis()
+      val clock = new ManualClock(startTime)
+      // Ensure that a cached version of deltaLog without the ManualClock is not used.
+      DeltaLog.clearCache()
```
scovich, line 455: That seems... bad... should the clock be part of the cache key??

spark/src/test/scala/org/apache/spark/sql/delta/InCommitTimestampSuite.scala
```diff
@@ -441,5 +445,401 @@ class InCommitTimestampSuite
...
+      val deltaLog =
+        DeltaLog.forTable(spark, new Path(tempDir.getCanonicalPath), clock)
+      val commitTimeDelta = 10
+      val numberAdditionalCommits = 25
+      assert(clock.getTimeMillis() == deltaLog.clock.getTimeMillis())
```
scovich, line 460: simpler to just do this?
```suggestion
      assert(clock eq deltaLog.clock)
```

scovich (approved):
Overall, looks good. A few nits.

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
dhruvarya-db: Right, thanks for catching that!

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -103,6 +108,25 @@ class DeltaHistoryManager(
...
+    if (end - start > 2 * maxKeysPerList) {
+      parallelSearch(time, start, end)
+    } else {
+      val commits = getCommits(
+        deltaLog.store,
+        deltaLog.logPath,
+        start,
+        Some(end),
+        deltaLog.newDeltaHadoopConf())
+      lastCommitBeforeTimestamp(commits, time).getOrElse(commits.head)
```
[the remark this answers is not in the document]
dhruvarya-db: I believe this is only possible in extremely rare scenarios and even in those cases, it will end up throwing some unrecoverable error downstream. This part of the code is mostly untouched: https://github.com/delta-io/delta/pull/2813/files#diff-98df89d4b0ce76abb5263d4fbb6f991660083a254dcdb9d922a4973a9937a4e2L139

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -118,28 +142,92 @@ class DeltaHistoryManager(
...
+        // start non-ICT search over [earliestVersion, ictEnablementVersion)
+        Some(getCommitFromNonICTRange(earliestVersion, end = ictEnablementCommit.version, time))
+      }
+    }
+
+    // We need to fetch the correct timestamp for the earliest commit if it was the result of the
+    // search.
+    // If commitOpt == ictEnablementCommit, we also need to validate the existence of the ICT
+    // enablement commit.
+    if (commitOpt.contains(placeholderEarliestCommit) || commitOpt.contains(ictEnablementCommit)) {
```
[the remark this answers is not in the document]
dhruvarya-db: Yes, the snapshot construction will fail at some step downstream.

spark/src/test/scala/org/apache/spark/sql/delta/InCommitTimestampSuite.scala
dhruvarya-db: Hmm, if the constructor, which takes a clock as an argument, is mostly used only for testing, it might not be worth maintaining the clock object in the key.

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -303,6 +401,127 @@ object DeltaHistoryManager extends DeltaLogging {
...
+      searchTimestamp: Long,
+      startCommit: Commit,
+      end: Long,
+      conf: Configuration,
+      basePath: Path,
+      logStore: LogStore,
+      numChunks: Long,
+      spark: SparkSession,
+      commitFileProvider: DeltaCommitFileProvider): Option[Commit] = {
+    require(startCommit.version < end, "start must be less than end")
```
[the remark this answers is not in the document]
dhruvarya-db: There is only one callsite for this function, which looks something like this:
```        
getActiveCommitAtTimeFromICTRange(
    time,
    searchWindowLowerBoundCommit,
    latestVersion + 1,
```
`latestVersion` = snapshot.version, and `searchWindowLowerBoundCommit` will always lie between 0 and snapshot.version.

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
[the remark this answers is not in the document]
scovich: Is it important to verify here that the enablement commit exists, then?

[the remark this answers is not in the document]
dhruvarya-db: It is only *required* when `canReturnEarliest` is `true`, so that we return some valid commit in that case. Otherwise, we can probably get away with not doing this at the cost of throwing a somewhat confusing error message instead of "proper" one that exists here: https://github.com/delta-io/delta/pull/2813/files/4a08b6e6b72a1c617bee0ca046849ecd790b7075#diff-98df89d4b0ce76abb5263d4fbb6f991660083a254dcdb9d922a4973a9937a4e2L149 . Since most clients are not trying to query for the earliest version of the table, I believe that this additional IO might be a reasonable penalty to pay.

PR #2816, opened 2024-03-28
[Spark] Add TransactionExecutionObserver

longvu-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

1. Add an observer field to `OptimisticTransaction` which can can be set via a thread-local to a custom instrumentation class. Instrumentation methods are invoked when creating a new transaction, for `prepareCommit`, for `doCommit`, and for completion of a transaction (failure or successful).
2. The default observer simply performs no-ops.
3. Added a testing `PhaseLocking` observer implementation that allows both observing, but also blocking the transaction's thread until unblocked by another thread. This allows fine control of how the transaction progresses, which is needed for some testing scenarios.

## How was this patch tested?
Added UTs.

No.


larsk-db (approved):
LGTM

PR #2818, opened 2024-03-28
[Spark] Update deltaLog.getHistory to serve [start, end] version range

sumeet-db:
- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Previously, `deltaLog.getHistory(start, endOpt)` served the `[start, endOpt]` range. However, this behavior was accidentally changed in #2799 to serve the `[start, endOpt)` range.

We revert it back to the original behavior and also update the function comment.

## How was this patch tested?

UTs

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
@@ -68,7 +68,7 @@ class DeltaHistoryManager(
   }
 
   /**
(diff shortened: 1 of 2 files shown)


felipepessoto:
Hi @sumeet-db, should we add a test to validate the behavior?

prakharjain09 (approved):

prakharjain09 (approved):
Lets also add a test to validate the behavior.

sumeet-db:
> Hi @sumeet-db, should we add a test to validate the behavior?

Added

PR #2820, opened 2024-03-29
only overwrite schema and field id for create table Iceberg conversion txn

lzlfred:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

The existing Uniform convertion txn has a logic to launch a second txn to set schema, after the first CREATE TABLE or REPLACE TABLE txn, to set the correct field ids because Iceberg may reassign those. This behavior has following flaws:

Firstly iceberg core does NOT reassign field id for REPLACE txn if the table already exists (in Uniform case it always does). So set schema for REPLACE TABLE is not necessary.

Secondly, Uniform uses the replace transaction when number of snapshots to convert exceeds threshold. The replace txn will set last Delta converted version as -1, which can be confusing or erroneous.

This PR fixes above flaws by NOT set schema for REPLACE txn.

## How was this patch tested?

Manually tested. Unit test will come in separate PR.


harperjiang (approved):
lgtm

PR #2834, opened 2024-04-01
[Spark] Make DeltaHistoryManager.getHistory aware of ICT

dhruvarya-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Return the ICT (In-Commit Timestamp) timestamp for ICT-enabled ranges in DeltaHistoryManager.getHistory when ICT is currently enabled.

## How was this patch tested?

New tests:
1. When all of history has ICT enabled.
2. When requested history is not an ICT range.
3. When requested range contains both ICT and non-ICT commits.

No visible differences for the user, the timestamp returned by `DESCRIBE HISTORY` will be different depending on whether ICT is currently enabled.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
@@ -135,8 +135,20 @@ class DeltaHistoryManager(
       start: Long,
       end: Option[Long] = None,
       snapshotOpt: Option[Snapshot] = None): Seq[DeltaHistory] = {
-    val snapshot = snapshotOpt.getOrElse(deltaLog.update())
-    val upperBoundExclusive = end.getOrElse(snapshot.version + 1)
+    val currentSnapshot = deltaLog.unsafeVolatileSnapshot
+    val (snapshot, upperBoundExclusive) = end match {
+        case Some(endExclusive) if currentSnapshot.version + 1 >= endExclusive =>
+          // Use the cache snapshot if it's fresh enough for the [start, endExclusive) query.
+          (currentSnapshot, math.min(currentSnapshot.version + 1, endExclusive))
+        case _ =>
+          // Either end doesn't exist or the currently cached snapshot isn't new enough to
+          // satisfy it.
+          val snapshot = snapshotOpt.getOrElse(deltaLog.update())
+          val endExclusive = end
+            .filter(_ <= snapshot.version + 1)
+            .getOrElse(snapshot.version + 1)
+          (snapshot, endExclusive)
+      }
     if (!DeltaConfigs.IN_COMMIT_TIMESTAMPS_ENABLED.fromMetaData(snapshot.metadata)) {
       getHistoryImpl(start, upperBoundExclusive, isRangeICT = false, snapshot)
     } else {
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
@@ -67,36 +67,32 @@ class DeltaHistoryManager(
    * chronological order.
    */
   def getHistory(limitOpt: Option[Int]): Seq[DeltaHistory] = {
+    val snapshot = deltaLog.update()
     val listStart = limitOpt.map { limit =>
-      math.max(deltaLog.update().version - limit + 1, 0)
+      math.max(snapshot.version - limit + 1, 0)
     }.getOrElse(getEarliestDeltaFile(deltaLog))
-    getHistory(listStart)
+    getHistory(listStart, snapshotOpt = Some(snapshot))
   }
 
   /**
-   * Get the commit information of the Delta table from commit `[start, end]`. If `end` is `None`,
-   * we return all commits from start to now.
+   * Get the commit information of the Delta table from commit `[start, end)`.
+   *
+   * @param isRangeICT Whether the commit range is ICT-enabled.
+   *                   If `true`, all commits in the range must have ICTs and
+   *                   the timestamp returned for each commit will be the ICT.
+   *                   If `false`, the file modification time will be used as the timestamp.
    */
-  def getHistory(
+  private[delta] def getHistoryImpl(
       start: Long,
-      endOpt: Option[Long] = None): Seq[DeltaHistory] = {
+      end: Long,
+      isRangeICT: Boolean,
+      snapshot: Snapshot): Seq[DeltaHistory] = {
     import org.apache.spark.sql.delta.implicits._
     val conf = getSerializableHadoopConf
     val logPath = deltaLog.logPath.toString
-    val currentSnapshot = deltaLog.unsafeVolatileSnapshot
-    val (snapshotForCommitFileProvider, end) = endOpt match {
-      case Some(end) if currentSnapshot.version >= end =>
-        // Use the cache snapshot if it's fresh enough for the [start, end] query.
-        (currentSnapshot, end)
-      case _ =>
-        // Either end doesn't exist or the currently cached snapshot isn't new enough to satisfy it.
-        val newSnapshot = deltaLog.update()
-        val endVersion = endOpt.getOrElse(newSnapshot.version).min(newSnapshot.version)
-        (newSnapshot, endVersion)
-    }
-    val commitFileProvider = DeltaCommitFileProvider(snapshotForCommitFileProvider)
+    val commitFileProvider = DeltaCommitFileProvider(snapshot)
     // We assume that commits are contiguous, therefore we try to load all of them in order
-    val info = spark.range(start, end + 1)
+    val info = spark.range(start, end)
       .mapPartitions { versions =>
         val logStore = LogStore(SparkEnv.get.conf, conf.value)
         val basePath = new Path(logPath)
@@ -104,9 +100,15 @@ class DeltaHistoryManager(
         versions.flatMap { commit =>
           try {
             val deltaFile = commitFileProvider.deltaFile(commit)
-            val ci = DeltaHistoryManager.getCommitInfo(logStore, deltaFile, conf.value)
-            val metadata = fs.getFileStatus(deltaFile)
-            Some(ci.withTimestamp(metadata.getModificationTime))
+            val ci = DeltaHistoryManager
+              .getCommitInfoOpt(logStore, deltaFile, conf.value)
+              .getOrElse(CommitInfo.empty(Some(commit)))
+            val timestamp = if (isRangeICT) {
+              CommitInfo.getRequiredInCommitTimestamp(Some(ci), commit.toString)
+            } else {
+              fs.getFileStatus(deltaFile).getModificationTime
+            }
+            Some(ci.withTimestamp(timestamp))
           } catch {
             case _: FileNotFoundException =>
               // We have a race-condition where files can be deleted while reading. It's fine to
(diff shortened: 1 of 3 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -115,8 +117,60 @@ class DeltaHistoryManager(
...
+  /**
+   * Get the commit information of the Delta table from commit `[start, end)`. If `end` is `None`,
+   * we return all commits from start to now.
+   * @param start The start of the commit range, inclusive.
+   * @param end The end of the commit range, exclusive.
+   */
+  def getHistory(
+      start: Long,
+      end: Option[Long] = None,
+      snapshotOpt: Option[Snapshot] = None): Seq[DeltaHistory] = {
```
sumeet-db, line 137: Remove `snapshotOpt` and rely on unsafeVolatileSnapshot instead. That will simplify the interface without any performance regressions.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaHistoryManagerSuite.scala
```diff
@@ -628,12 +628,12 @@ abstract class DeltaHistoryManagerBase extends DeltaTimeTravelTests
         assert(filterUsageRecords(usageRecords, "deltaLog.update").size === expectedLogUpdates)
       }
 
-      testGetHistory(start = 0, endOpt = Some(2), versions = Seq(2, 1, 0), expectedLogUpdates = 0)
-      testGetHistory(start = 1, endOpt = Some(1), versions = Seq(1), expectedLogUpdates = 0)
+      testGetHistory(start = 0, endOpt = Some(3), versions = Seq(2, 1, 0), expectedLogUpdates = 0)
```
sumeet-db, line 631: Undo these changes as well.

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -67,46 +67,48 @@ class DeltaHistoryManager(
    * chronological order.
    */
   def getHistory(limitOpt: Option[Int]): Seq[DeltaHistory] = {
+    val snapshot = deltaLog.update()
     val listStart = limitOpt.map { limit =>
-      math.max(deltaLog.update().version - limit + 1, 0)
+      math.max(snapshot.version - limit + 1, 0)
     }.getOrElse(getEarliestDeltaFile(deltaLog))
-    getHistory(listStart)
+    getHistory(listStart, snapshotOpt = Some(snapshot))
```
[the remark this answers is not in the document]
dhruvarya-db: That's a good idea, it will also keep the function signature stable. Done.

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -115,8 +117,60 @@ class DeltaHistoryManager(
...
+    val monotonizedCommits = if (isRangeICT) {
+      info.collect()
+    } else {
+      monotonizeCommitTimestamps(info.collect())
+    }
+    monotonizedCommits.reverse
+  }
+
+  /**
+   * Get the commit information of the Delta table from commit `[start, end)`. If `end` is `None`,
```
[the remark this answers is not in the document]
dhruvarya-db: Done.

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -67,46 +67,48 @@ class DeltaHistoryManager(
...
-   * we return all commits from start to now.
+   * Get the commit information of the Delta table from commit `[start, end)`.
+   *
+   * @param isRangeICT Whether the commit range is ICT-enabled.
+   *                   If `true`, all commits in the range must have ICTs and
+   *                   the timestamp returned for each commit will be the ICT.
+   *                   If `false`, the file modification time will be used as the timestamp.
    */
-  def getHistory(
+  private[delta] def getHistoryImpl(
```
sumeet-db, line 85: Can we make this `end` inclusive as well (and upperBoundExclusive -> upperBoundInclusive) for consistency across different sites in this file?

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -67,46 +67,48 @@ class DeltaHistoryManager(
...
       start: Long,
-      endOpt: Option[Long] = None): Seq[DeltaHistory] = {
+      end: Long,
+      isRangeICT: Boolean,
+      snapshot: Snapshot): Seq[DeltaHistory] = {
```
sumeet-db, line 89: Pass `commitFileProvider` here instead of snapshot?

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -115,8 +117,57 @@ class DeltaHistoryManager(
...
+   * Get the commit information of the Delta table from commit `[start, end]`. If `end` is `None`,
+   * we return all commits from start to now.
+   * @param start The start of the commit range, inclusive.
+   * @param end The end of the commit range, inclusive.
+   */
+  def getHistory(
+      start: Long,
+      end: Option[Long] = None): Seq[DeltaHistory] = {
+    val currentSnapshot = deltaLog.unsafeVolatileSnapshot
+    val (snapshot, upperBoundExclusive) = end match {
```
sumeet-db, line 138: `snapshot` -> `snapshotForCommitFileProvider`
`upperBoundExclusive` -> `resolvedEnd`

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -115,8 +117,57 @@ class DeltaHistoryManager(
...
+          val snapshot = deltaLog.update()
+          val endInclusive = end.getOrElse(snapshot.version).min(snapshot.version)
+          (snapshot, endInclusive + 1)
+      }
+    if (!DeltaConfigs.IN_COMMIT_TIMESTAMPS_ENABLED.fromMetaData(snapshot.metadata)) {
+      getHistoryImpl(start, upperBoundExclusive, isRangeICT = false, snapshot)
+    } else {
+      InCommitTimestampUtils.getValidatedICTEnablementInfo(snapshot.metadata) match {
+        case Some(Commit(ictEarliest, _)) =>
+          val nonICTCommits = if (ictEarliest > start) {
```
sumeet-db, line 154: I think relying on `getHistoryImpl` to return an empty Seq when start(inclusive) > end(inclusive) is a fair assumption. It can simplify the if-else's here.

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -115,8 +117,57 @@ class DeltaHistoryManager(
...
+          } else {
+            Seq.empty
+          }
+          val ictCommits = if (upperBoundExclusive > ictEarliest) {
+            getHistoryImpl(
+              math.max(ictEarliest, start), upperBoundExclusive, isRangeICT = true, snapshot)
+          } else {
+            Seq.empty
+          }
+          ictCommits ++ nonICTCommits
```
sumeet-db, line 166: The ordering here is important but it requires the reader to reason about the ordering of reverse and appending lists here to prove correctness. I'm not sure if there's a more intuitive way to order them here though.

We can add some comments:
```
// Merge the two sequences, ensuring ICT commits are listed first as they are more recent,
// followed by non-ICT commits, maintaining the reverse chronological order
```

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -67,46 +67,48 @@ class DeltaHistoryManager(
...
         val fs = basePath.getFileSystem(conf.value)
         versions.flatMap { commit =>
           try {
             val deltaFile = commitFileProvider.deltaFile(commit)
-            val ci = DeltaHistoryManager.getCommitInfo(logStore, deltaFile, conf.value)
-            val metadata = fs.getFileStatus(deltaFile)
-            Some(ci.withTimestamp(metadata.getModificationTime))
+            val ci = DeltaHistoryManager
+              .getCommitInfoOpt(logStore, deltaFile, conf.value)
+              .getOrElse(CommitInfo.empty(Some(commit)))
```
sumeet-db, line 105: Using an empty CommitInfo here will result in throwing [missingCommitTimestamp](https://github.com/delta-io/delta/blob/01c0ef91565d9a3e06fb9f62295f7c80fba09351/spark/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala#L1203) error here instead of `missingCommitInfo` error. I'm wondering if the latter would be better instead 🤔

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -67,46 +67,48 @@ class DeltaHistoryManager(
...
+   *                   If `true`, all commits in the range must have ICTs and
+   *                   the timestamp returned for each commit will be the ICT.
+   *                   If `false`, the file modification time will be used as the timestamp.
    */
-  def getHistory(
+  private[delta] def getHistoryImpl(
       start: Long,
-      endOpt: Option[Long] = None): Seq[DeltaHistory] = {
+      end: Long,
+      isRangeICT: Boolean,
```
sumeet-db, line 88: Nit: A simple `useInCommitTimestamps` name would also work?

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -67,46 +67,48 @@ class DeltaHistoryManager(
...
+      math.max(snapshot.version - limit + 1, 0)
     }.getOrElse(getEarliestDeltaFile(deltaLog))
-    getHistory(listStart)
+    getHistory(listStart, end = Some(snapshot.version))
   }
 
   /**
-   * Get the commit information of the Delta table from commit `[start, end]`. If `end` is `None`,
-   * we return all commits from start to now.
+   * Get the commit information of the Delta table from commit `[start, end)`.
```
sumeet-db, line 78: Nit: Mention `in reverse chronological order`

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -115,8 +117,57 @@ class DeltaHistoryManager(
...
+      info.collect()
+    } else {
+      monotonizeCommitTimestamps(info.collect())
+    }
+    monotonizedCommits.reverse
+  }
+
+  /**
+   * Get the commit information of the Delta table from commit `[start, end]`. If `end` is `None`,
+   * we return all commits from start to now.
```
sumeet-db, line 130: Nit: Mention `in reverse chronological order`

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
dhruvarya-db: Done

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
dhruvarya-db: `missingCommitInfo` is definitely the better error message in that case. Restructured the code to fix this.

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
dhruvarya-db: > upperBoundExclusive -> resolvedEnd
Done 

> snapshot -> snapshotForCommitFileProvider

Since the snapshot is used for determining the enablement status of ICT as well, I haven't changed the name.

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -115,8 +119,61 @@ class DeltaHistoryManager(
           }
         }.map(DeltaHistory.fromCommitInfo)
       }
-    // Spark should return the commits in increasing order as well
```
sumeet-db, line 118: Undo comment removal

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -115,8 +119,61 @@ class DeltaHistoryManager(
...
+   * Get the commit information of the Delta table from commit `[start, end]` in reverse
+   * chronological order. If `end` is `None`, we return all commits from start to now.
+   * @param start The start of the commit range, inclusive.
+   * @param end The end of the commit range, inclusive.
+   */
+  def getHistory(
+      start: Long,
+      end: Option[Long] = None): Seq[DeltaHistory] = {
+    val currentSnapshot = deltaLog.unsafeVolatileSnapshot
+    val (snapshot, resolvedEnd) = end match {
```
sumeet-db, line 141: `snapshot` -> `snapshotForCommitFileProvider` since the snapshot might not necessarily be useful for other usecases.

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -115,8 +119,61 @@ class DeltaHistoryManager(
...
+              useInCommitTimestamps = false,
+              commitFileProvider)
+          val ictCommits = getHistoryImpl(
+              math.max(ictEarliest, start),
+              resolvedEnd,
+              useInCommitTimestamps = true,
+              commitFileProvider)
+          // Merge the two sequences, ensuring ICT commits are listed first as they are more recent,
+          // followed by non-ICT commits, maintaining the reverse chronological order.
+          ictCommits ++ nonICTCommits
```
sumeet-db, line 172: I think it might be a problem that we don't monotonize timestamps across ICT and non-ICT commits?

dhruvarya-db: This will only happen for tables where the file modification timestamps of non-ICT commits have been changed after the enablement of ICT (since the first ICT > last non-ICT file modification time). I think this is an acceptable tradeoff. Eventually, these non-ICT commits will expire. It is preferable to ensure that ICT timestamps are unchanging.

sumeet-db (approved):

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
dhruvarya-db: Renamed to `snapshotNewerThanResolvedEnd`.

PR #2841, opened 2024-04-03
[Spark] Protocol version downgrade in the presence of table features

andreaschat-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR adds support for protocol versions downgrade when table features exist in the protocol. The downgraded protocol versions should be the minimum required to support all available table features. For example, `Protocol(3, 7, DeletionVectors, RowTracking)` can be downgraded to `Protocol(1, 7, RowTracking)` after removing the DV feature.

## How was this patch tested?
Added new UTs in DeltaProtocolVersionSuite. Furthermore, existing UTs cover a significant part of the functionality. These these are the following:

- Downgrade protocol version on table created with (3, 7).
- Downgrade protocol version on table created with (1, 7).
- Protocol version downgrade on a table with table features and added legacy feature.
- Protocol version is not downgraded when writer features exist.
- Protocol version is not downgraded when reader+writer features exist.
- Protocol version is not downgraded when multiple reader+writer features exist.

Yes. Dropping a table feature from a table with multiple features may now result to a Protocol versions downgrade. For example, `Protocol(3, 7, DeletionVectors, RowTracking)` can now be downgraded to `Protocol(1, 7, RowTracking)`.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/actions/TableFeatureSupport.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/actions/TableFeatureSupport.scala
@@ -351,7 +351,7 @@ trait TableFeatureSupport { this: Protocol =>
   def downgradeProtocolVersionsIfNeeded: Protocol = {
     if (nativeReaderAndWriterFeatures.nonEmpty) {
       val (minReaderVersion, minWriterVersion) =
-        TableFeatureProtocolUtils.minimumRequiredVersions(nativeReaderAndWriterFeatures)
+        TableFeatureProtocolUtils.minimumRequiredVersions(readerAndWriterFeatures)
       // It is guaranteed by the definitions of WriterFeature and ReaderFeature, that we cannot
       // end up with invalid protocol versions such as (3, 3). Nevertheless,
       // we double check it here.
--- a/spark/src/main/scala/org/apache/spark/sql/delta/actions/TableFeatureSupport.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/actions/TableFeatureSupport.scala
@@ -289,7 +289,7 @@ trait TableFeatureSupport { this: Protocol =>
     }
 
     // When `to` protocol contains table features, protocol versions may still
-    // get downgraded. However, the resulting protocol need to support at least writer features.
+    // get downgraded. However, the resulting protocol needs to support at least writer features.
     // Note, we check the minimum required versions only when table features exist in the protocol.
     // This because we do not always downgrade the versions of protocols with no table features.
     if (!to.supportsWriterFeatures ||
@@ -395,8 +395,8 @@ trait TableFeatureSupport { this: Protocol =>
       // end up with invalid protocol versions such as (3, 3). Nevertheless,
       // we double check it here.
       val newProtocol =
-      Protocol(minReaderVersion, minWriterVersion).withFeatures(readerAndWriterFeatures)
-      require(
+        Protocol(minReaderVersion, minWriterVersion).withFeatures(readerAndWriterFeatures)
+      assert(
         newProtocol.supportsWriterFeatures,
         s"Downgraded protocol should at least support writer features, but got $newProtocol.")
       return newProtocol
@@ -406,7 +406,7 @@ trait TableFeatureSupport { this: Protocol =>
       TableFeatureProtocolUtils.minimumRequiredVersions(readerAndWriterFeatures)
     val newProtocol = Protocol(minReaderVersion, minWriterVersion)
 
-    require(
+    assert(
       !newProtocol.supportsReaderFeatures && !newProtocol.supportsWriterFeatures,
       s"Downgraded protocol should not support table features, but got $newProtocol.")
 
--- a/spark/src/main/scala/org/apache/spark/sql/delta/actions/TableFeatureSupport.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/actions/TableFeatureSupport.scala
@@ -189,10 +189,11 @@ trait TableFeatureSupport { this: Protocol =>
     readerAndWriterFeatureNames.toSeq.flatMap(TableFeature.featureNameToFeature)
 
   /**
-   * A sequence of non-legacy [[TableFeature]]s.
+   * A sequence of native [[TableFeature]]s. This is derived by filtering out all explicitly
+   * supported legacy features.
    */
   @JsonIgnore
-  lazy val nonLegacyReaderAndWriterFeatures: Seq[TableFeature] =
+  lazy val nativeReaderAndWriterFeatures: Seq[TableFeature] =
     readerAndWriterFeatures.filterNot(_.isLegacyFeature)
 
   /**
@@ -256,7 +257,7 @@ trait TableFeatureSupport { this: Protocol =>
    *    downgrade is only supported with table features.
    *  - We can only remove one feature at a time.
    *  - The protocol versions may be downgraded when dropping a table feature.
-   *    There are main two cases:
+   *    There are two cases:
    *    a) If `to` protocol contains table features, the resulting versions should be the
    *       minimum required versions to support all table features.
    *    b) If `to` protocol only contains legacy features, the resulting versions
@@ -292,7 +293,7 @@ trait TableFeatureSupport { this: Protocol =>
     // Note, we check the minimum required versions only when table features exist in the protocol.
     // This because we do not always downgrade the versions of protocols with no table features.
     if (!to.supportsWriterFeatures ||
-      to.nonLegacyReaderAndWriterFeatures.nonEmpty && !toSupportsRequiredMinVersions) {
+      to.nativeReaderAndWriterFeatures.nonEmpty && !toSupportsRequiredMinVersions) {
       return false
     }
 
@@ -387,9 +388,9 @@ trait TableFeatureSupport { this: Protocol =>
    * features. After we remove the last native feature we downgrade the protocol to (1, 1).
    */
   def downgradeProtocolVersionsIfNeeded: Protocol = {
-    if (nonLegacyReaderAndWriterFeatures.nonEmpty) {
+    if (nativeReaderAndWriterFeatures.nonEmpty) {
       val (minReaderVersion, minWriterVersion) =
-        TableFeatureProtocolUtils.minimumRequiredVersions(nonLegacyReaderAndWriterFeatures)
+        TableFeatureProtocolUtils.minimumRequiredVersions(nativeReaderAndWriterFeatures)
       // It is guaranteed by the definitions of WriterFeature and ReaderFeature, that we cannot
       // end up with invalid protocol versions such as (3, 3). Nevertheless,
       // we double check it here.
--- a/spark/src/main/scala/org/apache/spark/sql/delta/actions/TableFeatureSupport.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/actions/TableFeatureSupport.scala
@@ -189,6 +189,13 @@ trait TableFeatureSupport { this: Protocol =>
     readerAndWriterFeatureNames.toSeq.flatMap(TableFeature.featureNameToFeature)
 
   /**
+   * A sequence of non-legacy [[TableFeature]]s.
+   */
+  @JsonIgnore
+  lazy val nonLegacyReaderAndWriterFeatures: Seq[TableFeature] =
+    readerAndWriterFeatures.filterNot(_.isLegacyFeature)
+
+  /**
    * Get all features that are implicitly supported by this protocol, for example, `Protocol(1,2)`
    * implicitly supports `appendOnly` and `invariants`. When this protocol is capable of requiring
    * writer features, no feature can be implicitly supported.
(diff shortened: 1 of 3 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/actions/TableFeatureSupport.scala
```diff
@@ -244,12 +251,16 @@ trait TableFeatureSupport { this: Protocol =>
...
    * includes the following:
+   *
    *  - The current protocol needs to support at least writer features. This is because protocol
    *    downgrade is only supported with table features.
-   *  - The protocol version can only be downgraded when there are no non-legacy table features.
    *  - We can only remove one feature at a time.
-   *  - When downgrading protocol versions, the resulting versions must support exactly the same
-   *    set of legacy features supported by the current protocol.
+   *  - The protocol versions may be downgraded when dropping a table feature.
+   *    There are main two cases:
```
xupefei, line 259:
```suggestion
   *    There are two cases:
```

spark/src/main/scala/org/apache/spark/sql/delta/actions/TableFeatureSupport.scala
```diff
@@ -258,24 +269,32 @@ trait TableFeatureSupport { this: Protocol =>
   def canDowngradeTo(to: Protocol, droppedFeatureName: String): Boolean = {
     if (!supportsWriterFeatures) return false
 
-    // When `to` protocol does not have any features version downgrades are possible. However,
+    val featureNames = readerAndWriterFeatureNames - droppedFeatureName
+    val minRequiredVersions = TableFeatureProtocolUtils.minimumRequiredVersions(
+      featureNames.flatMap(TableFeature.featureNameToFeature).toSeq)
+    val toSupportsRequiredMinVersions =
+      (to.minReaderVersion, to.minWriterVersion) == minRequiredVersions
```
xupefei, line 276: I wonder can we compare if
```
Protocol(minRequiredVersions._1, minRequiredVersions._2).implicitlyAndExplicitlySupportedFeatures ===
   to.implicitlyAndExplicitlySupportedFeatures
```
instead?
Only when they supports the same feature we could downgrade.

andreaschat-db: In the case of version downgrade with table features that would not work. The issue is that `Protocol(minRequiredVersions._1, minRequiredVersions._2).implicitlyAndExplicitlySupportedFeatures` would have an empty explicit features set. This is because `implicitlyAndExplicitlySupportedFeatures` generates only the implicit features while the the explicit feature have to be added, well, explicitly :D.

xupefei: I see!

xupefei (approved):
Lgtm!

spark/src/main/scala/org/apache/spark/sql/delta/actions/TableFeatureSupport.scala
```diff
@@ -368,7 +388,19 @@ trait TableFeatureSupport { this: Protocol =>
...
   def downgradeProtocolVersionsIfNeeded: Protocol = {
-    if (!readerAndWriterFeatures.forall(_.isLegacyFeature)) return this
+    if (nativeReaderAndWriterFeatures.nonEmpty) {
+      val (minReaderVersion, minWriterVersion) =
+        TableFeatureProtocolUtils.minimumRequiredVersions(nativeReaderAndWriterFeatures)
+      // It is guaranteed by the definitions of WriterFeature and ReaderFeature, that we cannot
+      // end up with invalid protocol versions such as (3, 3). Nevertheless,
+      // we double check it here.
+      val newProtocol =
+      Protocol(minReaderVersion, minWriterVersion).withFeatures(readerAndWriterFeatures)
```
larsk-db, line 398: indent

spark/src/main/scala/org/apache/spark/sql/delta/actions/TableFeatureSupport.scala
```diff
@@ -368,7 +388,19 @@ trait TableFeatureSupport { this: Protocol =>
...
+      require(
```
larsk-db, line 399: `assert`? This method takes no arguments, so you can hardly require something of them.

larsk-db (approved):
LGTM

felipepessoto:
Is the description correct? Reader v1 doesn’t support DV

larsk-db:
> Is the description correct? Reader v1 doesn’t support DV

I think the intent of "can be downgraded" was an implicit "after removing the DV feature".

spark/src/main/scala/org/apache/spark/sql/delta/actions/TableFeatureSupport.scala
```diff
@@ -258,24 +270,32 @@ trait TableFeatureSupport { this: Protocol =>
...
     }
 
-    // When `to` protocol contains table features we cannot downgrade the protocol version.
-    if (to.minReaderVersion != this.minReaderVersion) return false
-    if (to.minWriterVersion != this.minWriterVersion) return false
+    // When `to` protocol contains table features, protocol versions may still
+    // get downgraded. However, the resulting protocol needs to support at least writer features.
+    // Note, we check the minimum required versions only when table features exist in the protocol.
+    // This because we do not always downgrade the versions of protocols with no table features.
+    if (!to.supportsWriterFeatures ||
```
bart-samwel, line 295: Please use parenteses when combining `||` and `&&` -- I always forget the precedence order, and it's easy to get it wrong.

spark/src/main/scala/org/apache/spark/sql/delta/actions/TableFeatureSupport.scala
```diff
@@ -258,24 +270,32 @@ trait TableFeatureSupport { this: Protocol =>
   def canDowngradeTo(to: Protocol, droppedFeatureName: String): Boolean = {
     if (!supportsWriterFeatures) return false
```
bart-samwel, line 271:
```suggestion
    if (!supportsWriterFeatures) {
      // Feature dropping only works for tables that support Table Features.
      return false
    }
```

spark/src/main/scala/org/apache/spark/sql/delta/actions/TableFeatureSupport.scala
```diff
@@ -258,24 +270,32 @@ trait TableFeatureSupport { this: Protocol =>
...
+    val featureNames = readerAndWriterFeatureNames - droppedFeatureName
+    val minRequiredVersions = TableFeatureProtocolUtils.minimumRequiredVersions(
+      featureNames.flatMap(TableFeature.featureNameToFeature).toSeq)
+    val toSupportsRequiredMinVersions =
+      (to.minReaderVersion, to.minWriterVersion) == minRequiredVersions
+
+    // When `to` protocol does not have any features, versions might get downgraded. However,
     // the current protocol needs to contain one non-legacy feature. We also allow downgrade when
     // there are only legacy features. This is to accommodate the case when the user attempts to
     // remove a legacy feature in a table that only contains legacy features.
```
bart-samwel, line 282: This comment is complicated. It distinguishes between only-legacy-features and at-least-one-non-legacy-feature but the calculation is really simple: the table after downgrade needs to have the right set of features.
```suggestion
    // If the target protocol doesn't have any explicit table features and only uses
    // legacy protocol versions, then we can downgrade if the downgraded protocol
    // has the implicitly supported features that we need, i.e., the current protocol's
    // features minus the dropped feature.
```

spark/src/main/scala/org/apache/spark/sql/delta/actions/TableFeatureSupport.scala
```diff
@@ -258,24 +270,32 @@ trait TableFeatureSupport { this: Protocol =>
...
     if (to.readerAndWriterFeatureNames.isEmpty) {
-      val featureNames = readerAndWriterFeatureNames - droppedFeatureName
       val sameLegacyFeaturesSupported = featureNames == to.implicitlySupportedFeatures.map(_.name)
-      val minRequiredVersions = TableFeatureProtocolUtils.minimumRequiredVersions(
-        featureNames.flatMap(TableFeature.featureNameToFeature).toSeq)
 
       return sameLegacyFeaturesSupported &&
         (to.minReaderVersion, to.minWriterVersion) == minRequiredVersions &&
         readerAndWriterFeatures.filterNot(_.isLegacyFeature).size <= 1
```
bart-samwel, line 288: Isn't this last condition implied by the other conditions? I mean, you can't get to a legacy-features-only protocol by removing one feature unless you start with at most one non-legacy feature.

spark/src/main/scala/org/apache/spark/sql/delta/actions/TableFeatureSupport.scala
```diff
@@ -258,24 +270,32 @@ trait TableFeatureSupport { this: Protocol =>
...
     }
 
-    // When `to` protocol contains table features we cannot downgrade the protocol version.
-    if (to.minReaderVersion != this.minReaderVersion) return false
-    if (to.minWriterVersion != this.minWriterVersion) return false
+    // When `to` protocol contains table features, protocol versions may still
```
bart-samwel, line 291: I find this very hard to understand because we're handling so many cases separately. I think it's organically grown into this, but basically all we're trying to say is "the new protocol should be the old protocol minus the one feature that we're dropping". I think the entirety of this function can be simplified into something as stupid as:

```
(this.implicitlyAndExplicitlySupportedFeatures -- to.implicitlyAndExplicitlySupportedFeatures).toSeq == Seq(TableFeature.featureNameToFeature(droppedFeature))
```

I know that there's the point below that "Note, we check the minimum required versions only when table features exist in the protocol." But I think that's a choice that gets made at a different level, where we calculate the "to" protocol that we're testing against. But the basic test for "is this a valid downgrade" can be very simple.

andreaschat-db: Yes that is great. I will just turn it a bit around because like this it requires a symmetric difference.

spark/src/main/scala/org/apache/spark/sql/delta/actions/TableFeatureSupport.scala
```diff
@@ -368,13 +349,25 @@ trait TableFeatureSupport { this: Protocol =>
...
+      // It is guaranteed by the definitions of WriterFeature and ReaderFeature, that we cannot
+      // end up with invalid protocol versions such as (3, 3). Nevertheless,
+      // we double check it here.
+      val newProtocol =
+        Protocol(minReaderVersion, minWriterVersion).withFeatures(readerAndWriterFeatures)
+      assert(
+        newProtocol.supportsWriterFeatures,
+        s"Downgraded protocol should at least support writer features, but got $newProtocol.")
+      return newProtocol
+    }
```
bart-samwel, line 364: IIUIC `(minReaderVersion, minWriterVersion)` will end up being either (1, 7) or (3, 7) depending on whether the `nativeReaderAndWriterFeatures` contain any reader features. If the ColumnMapping feature is present, and a native writer feature, then this will return `(1, 7)`, but then the `newProtocol` below will use `withFeatures` to add back ColumnMapping. I think that would fail because `withFeatures` can't upgrade the legacy protocol version numbers.

Should this have been `readerAndWiterFeatures` in the third line? In that case we would have ended up with (2, 7). And then `withFeatures` works for adding ColumnMapping, because it accepts legacy features that are already implicitly supported.

Let's make sure we have a test for downgrading to (2, 7).

andreaschat-db: Yes. I thought that once a table is in the table features territory there should be no interaction with legacy features when it comes to protocol versions. But that is incorrect. Fixed and added a test.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaProtocolVersionSuite.scala
```diff
@@ -3427,6 +3427,23 @@ trait DeltaProtocolVersionSuiteBase extends QueryTest
...
   }
 
+  for (truncateHistory <- BOOLEAN_DOMAIN)
+  test(s"Protocol version downgrade with Table Features - include legacy reader features: " +
+    s"truncateHistory: ${truncateHistory}") {
+    val expectedFeatures =
+      Seq(DomainMetadataTableFeature, ChangeDataFeedTableFeature, ColumnMappingTableFeature)
+
+    // Although the protocol contains legacy features, the versions are only downgraded to
+    // table feature versions.
```
bart-samwel, line 3437: This comment is stale?

bart-samwel (approved):
One final nit. Let's ship it!

spark/src/test/scala/org/apache/spark/sql/delta/DeltaProtocolVersionSuite.scala
andreaschat-db: This is true in this PR but it won't be true when https://github.com/delta-io/delta/pull/2848 merges. Also, this test will have to be adapted once I rebase the latter PR. Anyway, I cleaned it up proactively.

PR #2848, opened 2024-04-04
[Spark] Improve Delta Protocol Transitions

andreaschat-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Currently, protocol transitions can be hard to manage. A few examples:
- It is hard to predict the output of certain operations.
- Once a legacy protocol transitions to a Table Features protocol it is quite hard to transition back to a legacy protocol.
- Adding a feature in a protocol and then removing it might lead to a different protocol.
- Adding an explicit feature to a legacy protocol always leads to a table features protocol although it might not be necessary.
- Dropping features from legacy protocols is not supported. As a result, the order the features are dropped matters.  
- Default protocol versions are ignored in some cases.
- Enabling table features by default results in feature loss in legacy protocols.
- CREATE TABLE ignores any legacy versions set if there is also a table feature in the definition.

This PR proposes several protocol transition improvements in order to simplify user journeys. The high level proposal is the following: 

Two protocol representations with singular operational semantics. This means that we have two ways to represent a protocol: a) The legacy representation and b) the table features representation. The latter representation is more powerful than the former, i.e the table features representation can represent all legacy protocols but the opposite is not true. This is followed by three simple rules:

1. All operations should be allowed to be performed on both protocol representations and should yield equivalent results.
2. The result should always be represented with the weaker form when possible.
3. Conversely, if the result of an operation on a legacy protocol cannot be represented with the legacy representation, use the Table Features representation.

**The PR introduces the following behavioural changes:**

1. Now all protocol operations are followed by denormalisation and then normalisation. Up to now, normalisation would only be performed after dropping a features.
2. Legacy features can now be dropped directly from a legacy protocol. The result is represented with table features if it cannot be represented with a legacy protocol.
3. Operations on table feature protocols now take into account the default versions. For example, enabling deletion vectors on table results to protocol `(3, 7, AppendOnly, Invariants, DeletionVectors)`.
5. Operations on table feature protocols now take into account any protocol versions set on the table. For example, creating a table with protocol `(1, 3)` and deletion vectors results to protocol `(3, 7, AppendOnly, Invariants, CheckConstraints, DeletionVectors)`.
6. It is not possible now to have a table features protocol without table features. For example, creating a table with `(3, 7)` and no table features is now normalised to `(1, 1)`.
7. Column Mapping can now be automatically enabled on legacy protocols when the mode is changed explicitly.

## How was this patch tested?
Added `DeltaProtocolTransitionsSuite`. Also modified existing tests in `DeltaProtocolVersionSuite`.

Yes.


xupefei:
One question. Consider the following command sequence:
```
CREATE TABLE x TBLPROPERTIES (
    delta.feature.RemovableReaderWriterFeature = 'supported',
    delta.feature.ChangeDataFeedTableFeature = 'supported')
```

```
ALTER TABLE x DROP FEATURE RemovableReaderWriterFeature
```

```
ALTER TABLE x SET TBLPROPERTIES (
  'delta.minReaderVersion' = 1,
  'delta.minWriterVersion' = 4) 
```

Should the table get `(1,4)` in the end? I think it should as we automatically add all legacy features in the 3rd command.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaProtocolVersionSuite.scala
```diff
@@ -3438,6 +3438,98 @@ trait DeltaProtocolVersionSuiteBase extends QueryTest
       expectedDowngradedProtocol = protocolWithReaderFeature(TestRemovableReaderWriterFeature))
   }
 
+  private def validVersions = Seq((1, 1), (1, 2), (1, 3), (1, 4), (2, 5), (2, 6), (1, 7), (3, 7))
+  private def invalidVersions = Seq((2, 2), (2, 3))
+  for ((readerVersion, writerVersion) <- validVersions ++ invalidVersions)
+  test(s"Legacy features are added when setting legacy versions: " +
```
xupefei, line 3444:
```suggestion
  test("Legacy features are added when setting legacy versions: " +
```

spark/src/test/scala/org/apache/spark/sql/delta/DeltaProtocolVersionSuite.scala
```diff
@@ -3438,6 +3438,98 @@ trait DeltaProtocolVersionSuiteBase extends QueryTest
...
+           |  'delta.minWriterVersion' = $writerVersion
+           |)""".stripMargin)
+
+      val expected = Protocol(readerVersion, writerVersion)
+        .implicitlySupportedFeatures + TestRemovableWriterFeature
+      assert(deltaLog.update().protocol.readerAndWriterFeatureNames === expected.map(_.name))
+    }
+  }
+
+  test(s"Setting protocol versions in a legacy table should not add any features") {
```
xupefei, line 3468:
```suggestion
  test("Setting protocol versions in a legacy table should not add any features") {
```

spark/src/test/scala/org/apache/spark/sql/delta/DeltaProtocolVersionSuite.scala
```diff
@@ -3438,6 +3438,98 @@ trait DeltaProtocolVersionSuiteBase extends QueryTest
...
+
+      // There should be no explicitly added legacy features.
+      assert(deltaLog.update().protocol == Protocol(1, 4))
+    }
+  }
+
+  for {
+    setLegacyVersions <- BOOLEAN_DOMAIN
+    tableFeatureToAdd <- Seq(TestRemovableWriterFeature, TestRemovableReaderWriterFeature)
+  } test(s"SOP for downgrading to legacy protocol versions for tables created with features. " +
```
xupefei, line 3494:
```suggestion
  } test("SOP for downgrading to legacy protocol versions for tables created with features. " +
```

andreaschat-db:
> One question. Consider the following command sequence:
> 
> ```
> CREATE TABLE x TBLPROPERTIES (
>     delta.feature.RemovableReaderWriterFeature = 'supported',
>     delta.feature.ChangeDataFeedTableFeature = 'supported')
> ```
> 
> ```
> ALTER TABLE x DROP FEATURE RemovableReaderWriterFeature
> ```
> 
> ```
> ALTER TABLE x SET TBLPROPERTIES (
>   'delta.minReaderVersion' = 1,
>   'delta.minWriterVersion' = 4) 
> ```
> 
> Should the table get `(1,4)` in the end? I think it should as we automatically add all legacy features in the 3rd command.

Very good point. Command 2 will result to `Protocol(3, 7, ChangeDataFeedTableFeature)`. Then, command 3 will result to `Protocol(3, 7, ChangeDataFeedTableFeature + rest features in (1, 4))`. Dropping one of the existing legacy features cannot result in (1, 4) because the legacy features list won't match exactly. To get out of this and downgrade to (1, 4) the user will have to add a feature to the table and then drop it... I could address this in a separate PR in the future since there is a way out of it and it is a rare case, i.e., downgrade the protocol versions of a table with NO table features but with table feature versions.

zsxwing:
Asking the user to set `delta.minReaderVersion` and `delta.minWriterVersion` is not user-friendly. This means users need to understand how to map a feature to `minReaderVersion/minWriterVersion`.

larsk-db (approved):
LGTM

felipepessoto:
For me it seems the existing behavior you described is better:

“For example, consider creating a table with Protocol(3, 7, RemovableReaderWriterFeature, ChangeDataFeed) and then dropping RemovableReaderWriterFeature. The resulting protocol will be Protocol(3, 7, ChangeDataFeed) instead of Protocol(1, 4)”

Protocol 3,7 besides being newer has less requirements. Meaning if you “downgrade” to 1,4 you are making your table less compatible to clients that doesn’t support all the features from writer v4, but supports v7 + CDC

spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
```diff
@@ -512,6 +512,7 @@ trait OptimisticTransactionImpl extends TransactionalWrite
       val newProtocolForLatestMetadata =
         Protocol(readerVersionAsTableProp, writerVersionAsTableProp)
       val proposedNewProtocol = protocolBeforeUpdate.merge(newProtocolForLatestMetadata)
+        .addImplicitLegacyFeatures(newProtocolForLatestMetadata)
```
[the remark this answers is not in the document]
andreaschat-db: It turns out we already added this capability at some point to `protocol.merge` so this call is not needed at all. However, `downgradeProtocolVersionsIfNeeded` should be called here instead of merge to avoid any surprise downgrades.

spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
```diff
@@ -511,7 +511,13 @@ trait OptimisticTransactionImpl extends TransactionalWrite
 
       val newProtocolForLatestMetadata =
         Protocol(readerVersionAsTableProp, writerVersionAsTableProp)
-      val proposedNewProtocol = protocolBeforeUpdate.merge(newProtocolForLatestMetadata)
+      // When the protocol versions are explicitly set on table features protocol we may
+      // downgrade to legacy protocol versions.
```
bart-samwel, line 515: I love a bit of color:
```suggestion
      // When the protocol versions are explicitly set on table features protocol we may
      // downgrade to legacy protocol versions. Legacy protocol versions can only be
      // used if a table supports *exactly* the set of features in that legacy protocol
      // version, with no "gaps". By merging in the protocol features from a particular
      // protocol version, we may end up with such a "gap-free" protocol. E.g. if a table
      // has only table feature "checkConstraints" (added by writer protocol version 3)
      // but not "invariants" and "appendOnly", then setting the minWriterVersion to
      // 2 or 3 will add "invariants" and "appendOnly", filling in the gaps for writer
      // protocol version 3, and then we can downgrade to version 3.  
```

spark/src/main/scala/org/apache/spark/sql/delta/actions/TableFeatureSupport.scala
```diff
@@ -281,12 +281,34 @@ trait TableFeatureSupport { this: Protocol =>
...
+   * we only have available the the start and end states of the Protocol, i.e. the intermediate
+   * state with all the required implicit features is not available. It is
+   * downgradeProtocolToLegacyVersionsIfNeeded which validates whether the intermediate state has
+   * all required legacy features are in the protocol before downgrading the protocol versions.
+   */
+  def canDowngradeToLegacy(to: Protocol): Boolean = {
+    require(supportsWriterFeatures)
+    require(!to.supportsWriterFeatures)
+
+    readerAndWriterFeatures.forall(_.isLegacyFeature) &&
```
bart-samwel, line 301: Now that I read this, I realize that it's actually an upgrade in some sense. :) It would be good to sprinkle that idea throughout the code in comments, so that people realize that it's not a real downgrade. 

```suggestion
    // Note that the downgrade can *add* legacy features. It's technically
    // a protocol *upgrade* then in terms of features but a *downgrade*
    // in terms of protocol versions.
    readerAndWriterFeatures.forall(_.isLegacyFeature) &&
```

spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
```diff
@@ -511,7 +511,13 @@ trait OptimisticTransactionImpl extends TransactionalWrite
...
+      val proposedNewProtocol = protocolBeforeUpdate
+        .merge(newProtocolForLatestMetadata)
+        .downgradeProtocolToLegacyVersionsIfNeeded(
```
bart-samwel, line 518: OK, so in this case we only downgrade to the explicitly specified protocol, NOT to _any_ legacy protocol. E.g. if the table contains the feature added by legacy protocol 6, but the caller asks to go down to protocol 5, then we won't downgrade -- you have to _exactly_ hit the target protocol version. Should we instead just go all the way and do a a downgrade to _any_ legacy version when people ask for a legacy version?

spark/src/main/scala/org/apache/spark/sql/delta/actions/TableFeatureSupport.scala
```diff
@@ -281,12 +281,34 @@ trait TableFeatureSupport { this: Protocol =>
...
+      readerAndWriterFeatureNames.subsetOf(to.implicitlySupportedFeatures.map(_.name))
+  }
+
   /**
    * True if this protocol can be upgraded or downgraded to the 'to' protocol.
    */
   def canTransitionTo(to: Protocol, op: Operation): Boolean = {
     op match {
       case drop: DeltaOperations.DropTableFeature => canDowngradeTo(to, drop.featureName)
+      case _ if supportsWriterFeatures && !to.supportsWriterFeatures => canDowngradeToLegacy(to)
```
bart-samwel, line 311:
```suggestion
      case _ if supportsWriterFeatures && !to.supportsWriterFeatures =>
        // This is called a "downgrade" but it can actually only be a "side-grade"
        // (equivalent features, different representation) or an upgrade (more
        // features, lower legacy protocol version).
        canDowngradeToLegacy(to)
```

spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
andreaschat-db: Done.

spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
```diff
@@ -512,7 +512,22 @@ trait OptimisticTransactionImpl extends TransactionalWrite
...
+      // protocol version, we may end up with such a "gap-free" protocol. E.g. if a table
+      // has only table feature "checkConstraints" (added by writer protocol version 3)
+      // but not "invariants" and "appendOnly", then setting the minWriterVersion to
+      // 2 or 3 will add "invariants" and "appendOnly", filling in the gaps for writer
+      // protocol version 3, and then we can downgrade to version 3.
+      val mergedProtocol = protocolBeforeUpdate.merge(newProtocolForLatestMetadata)
+      val proposedNewProtocol = if (newProtocolForLatestMetadata.supportsWriterFeatures) {
+        mergedProtocol
+      } else {
+        mergedProtocol.downgradeProtocolToLegacyVersionsIfNeeded
```
bart-samwel, line 529: Can we just do this unconditionally? I.e., whenever the protocol happens to be _perfectly_ representable by a legacy version, use the legacy protocol version? And if we have https://github.com/delta-io/delta/pull/2841/files, can this just use `downgradeProtocolVersionsIfNeeded`?

I think the goal should be simply _always use the "most legacy" protocol representation that can represent the feature set_. I don't think we'd call it an _invariant_ (because other implementations may not do this), but we have to try to do this normalization whenever we change the protocol.

spark/src/main/scala/org/apache/spark/sql/delta/actions/TableFeatureSupport.scala
```diff
@@ -281,12 +281,41 @@ trait TableFeatureSupport { this: Protocol =>
...
+      readerAndWriterFeatureNames.subsetOf(to.implicitlySupportedFeatures.map(_.name))
+  }
+
   /**
    * True if this protocol can be upgraded or downgraded to the 'to' protocol.
    */
   def canTransitionTo(to: Protocol, op: Operation): Boolean = {
     op match {
       case drop: DeltaOperations.DropTableFeature => canDowngradeTo(to, drop.featureName)
+      case _ if supportsWriterFeatures && !to.supportsWriterFeatures =>
```
bart-samwel, line 314: I think we can simplify this, similar to what we did in https://github.com/delta-io/delta/pull/2841. Basically, we can fold this into `canUpgradeTo`. The change in `OptimisticTransaction` already ensures that the right features have been added. Right now `canUpgradeTo` still looks at protocol versions explicitly, and refuses to do protocol version downgrades. I think all it would take is to remove the first two lines from `canUpgradeTo`, so that it simply looks at "is the new feature set a superset of the old feature set".

andreaschat-db: Done.

spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
andreaschat-db: > Can we just do this unconditionally? I.e., whenever the protocol happens to be perfectly representable by a legacy version, use the legacy protocol version?

This is how I originally did this. The issue was is it changes current behaviour. The result is that setting table features protocol versions to a legacy protocol won't "stick" when there are no table features. For example, currently, setting `(1, 7)` to a table with `(2, 2)` results to `Protocol(2, 7, {"appendOnly", "invariants"})`. With the proposed change it will go back to `(2, 2)` and therefore be a noop. On the other hand, I do not see why someone would want to convert a protocol to table features without adding a particular table feature. It just might be a bit confusing his operation to be noop. So, I ended up removing the condition.

> And if we have https://github.com/delta-io/delta/pull/2841/files, can this just use downgradeProtocolVersionsIfNeeded?

Moving the normalization to `downgradeProtocolVersionsIfNeeded` will result to downgrading protocol versions with "gaps" directly at DROP command without the extra version setting step. That would be a different user journey from the one proposed in the PR. However, I like the concept of normalization. I refactored `downgradeProtocolToLegacyVersionsIfNeeded` to clearly show there is a normalization process when we set protocol versions. Also, it is now clear this does not happen when we call the DROP FEATURE command.

spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
```diff
@@ -523,7 +523,20 @@ trait OptimisticTransactionImpl extends TransactionalWrite
...
+      // downgrade to legacy protocol versions. Legacy protocol versions can only be
+      // used if a table supports *exactly* the set of features in that legacy protocol
+      // version, with no "gaps". By merging in the protocol features from a particular
+      // protocol version, we may end up with such a "gap-free" protocol. E.g. if a table
+      // has only table feature "checkConstraints" (added by writer protocol version 3)
+      // but not "invariants" and "appendOnly", then setting the minWriterVersion to
+      // 2 or 3 will add "invariants" and "appendOnly", filling in the gaps for writer
+      // protocol version 3, and then we can downgrade to version 3.
+      val proposedNewProtocol = protocolBeforeUpdate
+        .merge(newProtocolForLatestMetadata)
```
bart-samwel, line 537: It seems to me that this line and the next line should do the same thing. Looking at `merge`, it will add all implicitly supported features in any of the merged protocols. So if `protocolBeforeUpdate` is (3, 7), and `newProtocolForLatestMetadata` is (2, 5), then it will actually add the implicitly supported features.

andreaschat-db: Moved legacy feature normalization in `protocol.merge`. This results into normalizing the protocol after operations such as `ALTER TABLE`, `CLONE`, `RTAS` etc.

xupefei (approved):
Looks good to me!

spark/src/test/scala/io/delta/tables/DeltaTableSuite.scala
```diff
@@ -575,12 +575,12 @@ class DeltaTableHadoopOptionsSuite extends QueryTest
...
       assert(log.update().protocol === Protocol(1, 7)
-        .merge(Protocol(1, 2)).withFeature(TestWriterFeature))
+        .withFeature(TestWriterFeature).merge(Protocol(1, 2)))
       table.addFeatureSupport(TestReaderWriterFeature.name)
       assert(
         log.update().protocol === Protocol(3, 7)
-          .merge(Protocol(1, 2))
-          .withFeatures(Seq(TestWriterFeature, TestReaderWriterFeature)))
+          .withFeatures(Seq(TestWriterFeature, TestReaderWriterFeature))
+          .merge(Protocol(1, 2)))
```
c27kwan, line 583: nit: it wasn't clear to me why this change was necessary. It seems a bit error-prone to write the expected result this way. Let's change it to be an explicit Protocol (without merging two protocol objects).

spark/src/test/scala/org/apache/spark/sql/delta/DeltaColumnMappingSuite.scala
```diff
@@ -1907,8 +1895,9 @@ class DeltaColumnMappingSuite extends QueryTest
         s"""CREATE TABLE $testTableName
            |USING DELTA
            |TBLPROPERTIES(
-           |'$minReaderKey' = '2',
-           |'$minWriterKey' = '7'
+           |'$minReaderKey' = '3',
+           |'$minWriterKey' = '7',
+           |'${DeltaConfigs.ENABLE_DELETION_VECTORS_CREATION.key}' = 'true'
```
c27kwan, line 1900: You shouldn't need this change as part of this PR,  right? I initially made the minReaderKey 2 to expose a bug: https://github.com/delta-io/delta/commit/a655bed59e436dd0e9500a02441e65bfa9f103d6

spark/src/main/scala/org/apache/spark/sql/delta/DeltaColumnMapping.scala
```diff
@@ -156,10 +153,6 @@ trait DeltaColumnMappingBase extends DeltaLogging {
             (newProtocol.supportsWriterFeatures && satisfiesReaderVersion)) {
           newProtocol = newProtocol.withFeature(ColumnMappingTableFeature)
         }
-
-        if (!satisfiesColumnMappingProtocol(newProtocol)) {
-          throw DeltaErrors.changeColumnMappingModeOnOldProtocol(oldProtocol)
-        }
```
c27kwan, line 162: If you're deleting this code, you can actually delete everything under the `else` here and simplify this whole method. It's a lot of throwaway work :D

spark/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala
```diff
@@ -254,7 +254,15 @@ object Protocol {
 
     val (readerVersion, writerVersion, enabledFeatures) =
       minProtocolComponentsFromMetadata(spark, metadata)
-    Protocol(readerVersion, writerVersion).withFeatures(enabledFeatures)
+    // New table protocols should always be denormalized and then normalized to convert the
+    // protocol to the weakest possible form. For example:
+    // 1) (3, 7, RowIDs) is normalized to (3, 7, RowIDs).
```
c27kwan, line 259: `RowId` -> `RowTracking`. Also "weakest possible form" is not clear to the casual reader. Can you define this and add an `i.e.` to the comment?

spark/src/test/scala/org/apache/spark/sql/delta/columnmapping/DropColumnMappingFeatureSuite.scala
```diff
@@ -82,17 +85,12 @@ class DropColumnMappingFeatureSuite extends RemoveColumnMappingSuiteUtils {
          |USING delta
          |TBLPROPERTIES ('${DeltaConfigs.COLUMN_MAPPING_MODE.key}' = 'name',
          |        '${DeltaConfigs.ENABLE_DELETION_VECTORS_CREATION.key}' = 'false',
-         |        'delta.minReaderVersion' = '1',
-         |        'delta.minWriterVersion' = '1')
+         |        'delta.minReaderVersion' = '3',
+         |        'delta.minWriterVersion' = '7')
```
c27kwan, line 89: Why did we need to update this test? the minReaderVersion and minWriterVersion are "suggestions" and are ignored anyways since we enable DVs which will make this table have (3,7) in the end anyways. Are we enforcing users' minReaderVersion and minWriterVersion after this change?

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTableFeatureSuite.scala
```diff
@@ -205,30 +207,13 @@ class DeltaTableFeatureSuite
...
+    assert(tfProtocol1.merge(Protocol(1, 2)) === Protocol(1, 2))
+    assert(tfProtocol2.merge(Protocol(2, 6)) === Protocol(2, 6))
   }
 
   test("protocol upgrade compatibility") {
     assert(Protocol(1, 1).canUpgradeTo(Protocol(1, 1)))
     assert(Protocol(1, 1).canUpgradeTo(Protocol(2, 1)))
-    assert(!Protocol(1, 2).canUpgradeTo(Protocol(1, 1)))
-    assert(!Protocol(2, 2).canUpgradeTo(Protocol(2, 1)))
     assert(
```
c27kwan, line 217: I'm a bit concerned that this needed to be removed in this PR. Intuitively, (1, 2) -> (1,1) is not an "upgrade" so as a developer I expect this to return `false`. Have we changed the definition of `canUpgradeTo` to mean `canTransitionTo`?

spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala
```diff
@@ -423,6 +423,14 @@ trait DeltaSQLConfBase {
...
       .createWithDefault(1)
 
+  val TABLE_FEATURES_TEST_FEATURES_ENABLED =
+    buildConf("tableFeatures.testFeatures.enabled")
+      .internal()
+      .doc("Controls whether test features are enabled in testing mode. " +
+        "This config is only used for testing purposes. ")
+      .booleanConf
+      .createWithDefault(true)
+
```
c27kwan, line 433: Do we still need this? I see we use this in TableFeature, but we never use it in tests AFAICT.

spark/src/main/scala/org/apache/spark/sql/delta/DeltaColumnMapping.scala
andreaschat-db: Right of course.

spark/src/main/scala/org/apache/spark/sql/delta/actions/TableFeatureSupport.scala
```diff
@@ -287,12 +277,13 @@ trait TableFeatureSupport { this: Protocol =>
...
 
-    if (mergedProtocol.supportsReaderFeatures || mergedProtocol.supportsWriterFeatures) {
-      mergedProtocol.withFeatures(mergedImplicitFeatures)
-    } else {
-      mergedProtocol
-    }
+    // The merged protocol is always normalized in order to represent the protocol
+    // with the weakest possible form. This enables backward compatibility.
+    // This is preceded by a denormalization step. This allows to fix invalid legacy Protocols.
+    // For example, (2, 3) is normalized to (1, 3).
```
larsk-db, line 285: Explain why this is invalid (not everyone knows all the features in every version by heart by now :P)

spark/src/main/scala/org/apache/spark/sql/delta/actions/TableFeatureSupport.scala
```diff
@@ -336,50 +326,50 @@ trait TableFeatureSupport { this: Protocol =>
...
-      // end up with invalid protocol versions such as (3, 3). Nevertheless,
-      // we double check it here.
-      val newProtocol =
-        Protocol(minReaderVersion, minWriterVersion).withFeatures(readerAndWriterFeatures)
-      assert(
-        newProtocol.supportsWriterFeatures,
-        s"Downgraded protocol should at least support writer features, but got $newProtocol.")
-      return newProtocol
-    }
+  def normalize: Protocol = {
```
larsk-db, line 339: Total naming nit, but "(de-)normalize" sounds to me like an in-place change. But this returns a new protocol, so I think it should be called "(de-)normalized", since it returns a "(de-)normalized protocol", no?

spark/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala
```diff
@@ -254,7 +254,17 @@ object Protocol {
...
+    // protocol to the weakest possible form. This means either converting a table features
+    // protocol to a legacy protocol or reducing the versions of a table features protocol.
+    // For example:
+    // 1) (3, 7, RowTracking) is normalized to (1, 7, RowTracking).
+    // 2) (3, 7, AppendOnly, Invariants) is normalized to (1, 2).
+    // 3) (2, 3) is normalized to (1, 3).
+    Protocol(readerVersion, writerVersion)
+      .withFeatures(enabledFeatures)
+      .denormalize
+      .normalize
```
larsk-db, line 267: Should we introduce a helper that does both?

spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala
larsk-db: Should we double guard this so by default it's also only enabled in testing mode?

spark/src/test/scala-spark-master/org/apache/spark/sql/delta/DeltaVariantSuite.scala
```diff
@@ -52,9 +52,11 @@ class DeltaVariantSuite
...
       sql("INSERT INTO tbl (SELECT 'foo', parse_json(cast(id + 99 as string)) FROM range(1))")
       assert(spark.table("tbl").selectExpr("v::int").head == Row(99))
-      assert(
-        getProtocolForTable("tbl") ==
-        VariantTypeTableFeature.minProtocolVersion.withFeature(VariantTypeTableFeature)
+      assert(getProtocolForTable("tbl") ==
+          VariantTypeTableFeature.minProtocolVersion.withFeatures(Seq(
+            AppendOnlyTableFeature,
+            InvariantsTableFeature,
+            VariantTypeTableFeature))
```
larsk-db, line 59: Do we really want to assert the exact protocol here? Or rather just that it supports `VariantTypeTableFeature`? Like these we'll have to update it every time the defaults change, no?

spark/src/test/scala/org/apache/spark/sql/delta/DeltaErrorsSuite.scala
```diff
@@ -25,6 +25,7 @@ import java.util.Locale
 import scala.sys.process.Process
 
 // scalastyle:off import.ordering.noEmptyLine
+// scalastyle:off line.size.limit
```
larsk-db, line 28: This seems unnecessary when only deleting stuff?

spark/src/test/scala/org/apache/spark/sql/delta/schema/InvariantEnforcementSuite.scala
```diff
@@ -407,14 +407,9 @@ class InvariantEnforcementSuite extends QueryTest
...
             ("delta.constraints.mychk" -> "valueA < valueB"))
-        assert(txn.protocol.minWriterVersion === writerVersion)
         txn.commit(Seq(newMetadata), DeltaOperations.ManualUpdate)
-        val upVersion = if (TableFeatureProtocolUtils.supportsWriterFeatures(writerVersion)) {
-          TableFeatureProtocolUtils.TABLE_FEATURES_MIN_WRITER_VERSION
-        } else {
-          CheckConstraintsTableFeature.minWriterVersion
-        }
-        assert(table.deltaLog.unsafeVolatileSnapshot.protocol.minWriterVersion === upVersion)
+        assert(table.deltaLog.unsafeVolatileSnapshot.protocol.minWriterVersion ===
```
larsk-db, line 411:
```suggestion
        assert(table.deltaLog.update().protocol.minWriterVersion ===
```
While we are at it^^

spark/src/test/scala/org/apache/spark/sql/delta/DeltaErrorsSuite.scala
larsk-db: Ah, but then, don't add it at the top, but just scope it where it's needed, please? Or is just sooooo many places that that would be crazy?

andreaschat-db: I am afraid there are 38 errors :(.

spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala
andreaschat-db: It turns out I need to wrap this with an ugly try block because there are cases where the `conf` is not initialized...

spark/src/test/scala/org/apache/spark/sql/delta/DeltaErrorsSuite.scala
larsk-db: That are all non-consecutive?

andreaschat-db: I am afraid they are not.

scottsand-db:
There are spark master test failures:

https://github.com/delta-io/delta/actions/runs/9891350186/job/27321586977?pr=2848

```
[info] CreateCheckpointSuite:
[info] - commits containing adds and removes, and no previous checkpoint
[info] - commits containing adds, and no previous checkpoint
[info] - commits containing adds and removes, and a previous checkpoint created using Spark (actions/perfile): 1000000
[info] - commits containing adds and removes, and a previous checkpoint created using Spark (actions/perfile): 3
[info] - commits containing adds, and a previous checkpoint created using Spark (actions/perfile): 1000000
[info] - commits containing adds, and a previous checkpoint created using Spark (actions/perfile): 3
[info] - commits with metadata updates
[info] - commits with protocol updates *** FAILED ***
[info]   == Results ==
[info]   
[info]   == Expected Answer - 1 ==
[info]   ([2,2])
[info]   
[info]   == Result - 1 ==
[info]   ([1,2]) (TestUtils.scala:386)
```

andreaschat-db:
> There are spark master test failures:
> 
> https://github.com/delta-io/delta/actions/runs/9891350186/job/27321586977?pr=2848
> 
> ```
> [info] CreateCheckpointSuite:
> [info] - commits containing adds and removes, and no previous checkpoint
> [info] - commits containing adds, and no previous checkpoint
> [info] - commits containing adds and removes, and a previous checkpoint created using Spark (actions/perfile): 1000000
> [info] - commits containing adds and removes, and a previous checkpoint created using Spark (actions/perfile): 3
> [info] - commits containing adds, and a previous checkpoint created using Spark (actions/perfile): 1000000
> [info] - commits containing adds, and a previous checkpoint created using Spark (actions/perfile): 3
> [info] - commits with metadata updates
> [info] - commits with protocol updates *** FAILED ***
> [info]   == Results ==
> [info]   
> [info]   == Expected Answer - 1 ==
> [info]   ([2,2])
> [info]   
> [info]   == Result - 1 ==
> [info]   ([1,2]) (TestUtils.scala:386)
> ```

The PR is not ready to be merged. The particular failure is fixed by https://github.com/delta-io/delta/pull/3356.

PR #2850, opened 2024-04-04
[Spark] Make `delta.dataSkippingStatsColumns` more lenient for nested columns

### LINKED ISSUE #2822: [Feature Request] Make `delta.dataSkippingStatsColumns` more lenient for nested columns

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

### Overview

Setting `delta.dataSkippingStatsColumns` to a struct column will fail if any column inside the struct does not support gathering stats (binary, arrays, maps, etc). This should be more lenient to just skip columns that aren't supported rather than throwing an exception.

### Motivation

Our use case is a single top level struct that has all the import fields we care about, and we want to gather stats on all of them. But this struct has a mix of column types, including arrays and maps, which means we currently can't use `delta.dataSkippingStatsColumns`. We also have different columns in this struct in various tables, so trying to include each individually supported field would also be extremely difficult. This is also inconsistent with `dataSkippingNumIndexedCols`, which allows for these unsupported types (and in fact still gathers null counts on them, even if it can't gather min/max stats).

While it may make sense to raise in exception if the direct column specified by `dataSkippingStatsColumns` is unsupported, it should be allowed to specify a struct that has a mix of types within it.

### Willingness to contribute

Kimahriman:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Resolves #2822 

Make `delta.dataSkippingStatsColumns` more lenient for nested columns by not throwing an exception if a nested column doesn't support gathering stats. This more closely matches the behavior of the `dataSkippingNumIndexedCols` which allows for unsupported types in those columns (and seems to still gather null counts for those unsupported types). This also allows more use cases where you might have a wide variety of types inside a top level struct, and you simply want to gather stats on whatever columns inside that struct you can.

I kept the duplicate column checking in place to avoid less changes, but I'm not sure how necessary that really is besides letting users know they are doing something dumb.

## How was this patch tested?

A couple tests were removed that were specifically testing for the now-allowed behavior, and a new test was added to verify the new behavior works.

Yes, specifying a struct with unsupported stats gathering types in `delta.dataSkippingStatsColumns` is now allowed instead of throwing an exception.


Kimahriman:
@kamcheungting-db since you added this originally

Kimahriman:
@longvu-db since you're reviewing my other PR 😊. This would be a great quality of life improvement for complex nested schemas

longvu-db:
@Kimahriman Will take a look!

spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala
```diff
@@ -458,15 +458,18 @@ object StatisticsCollection extends DeltaCommand {
    * @param name The name of the data skipping column for validating data type.
    * @param dataType The data type of the data skipping column.
    * @param columnPaths The column paths of all valid fields.
+   * @param inStruct Whether we are inside a struct already, in which case we don't care about type
```
longvu-db, line 461:
```suggestion
   * @param inStruct Whether we are inside a struct already, in which case we don't care about the type.
```

spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala
```diff
@@ -458,15 +458,18 @@ object StatisticsCollection extends DeltaCommand {
...
   private def validateDataSkippingType(
       name: String,
       dataType: DataType,
-      columnPaths: ArrayBuffer[String]): Unit = dataType match {
+      columnPaths: ArrayBuffer[String],
+      inStruct: Boolean = false): Unit = dataType match {
     case s: StructType =>
       s.foreach { field =>
-        validateDataSkippingType(name + "." + field.name, field.dataType, columnPaths)
+        validateDataSkippingType(name + "." + field.name, field.dataType, columnPaths, true)
```
longvu-db, line 470: `insideStructs = true`

spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala
```diff
@@ -458,15 +458,18 @@ object StatisticsCollection extends DeltaCommand {
...
    * @param dataType The data type of the data skipping column.
    * @param columnPaths The column paths of all valid fields.
+   * @param inStruct Whether we are inside a struct already, in which case we don't care about type
    */
   private def validateDataSkippingType(
       name: String,
       dataType: DataType,
-      columnPaths: ArrayBuffer[String]): Unit = dataType match {
+      columnPaths: ArrayBuffer[String],
+      inStruct: Boolean = false): Unit = dataType match {
```
longvu-db, line 467:
```suggestion
      insideStructs: Boolean = false): Unit = dataType match {
```

?

spark/src/test/scala/org/apache/spark/sql/delta/stats/StatsCollectionSuite.scala
```diff
@@ -499,8 +499,7 @@ class StatsCollectionSuite
     ("BINARY", "BinaryType"),
     ("BOOLEAN", "BooleanType"),
     ("ARRAY<TINYINT>", "ArrayType(ByteType,true)"),
```
longvu-db, line 501: Let's add a nested struct test (like double struct or something)

spark/src/test/scala/org/apache/spark/sql/delta/stats/DataSkippingDeltaTests.scala
```diff
@@ -599,6 +599,27 @@ trait DataSkippingDeltaTestsBase extends DeltaExcludedBySparkVersionTestMixinShi
     deltaStatsColNamesOpt = Some("b.c")
   )
 
+  testSkipping(
```
longvu-db, line 602: Should we consider double struct test?

spark/src/test/scala/org/apache/spark/sql/delta/stats/StatsCollectionSuite.scala
Kimahriman: This is just testing invalid types, not sure what I would add with structs here

spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala
Kimahriman: Renamed

spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala
Kimahriman: Updated

spark/src/test/scala/org/apache/spark/sql/delta/stats/DataSkippingDeltaTests.scala
Kimahriman: Added a double nested struct as well

spark/src/test/scala/org/apache/spark/sql/delta/stats/DataSkippingDeltaTests.scala
```diff
@@ -599,6 +599,31 @@ trait DataSkippingDeltaTestsBase extends DeltaExcludedBySparkVersionTestMixinShi
...
+        "c": [4, 5, 6],
+        "d": 7,
+        "e": 8,
+        "f": {
+          "g": 9
+        }
+      },
+      "i": 10
+    }""".replace("\n", ""),
+    hits = Seq(
```
longvu-db, line 616: Could we have hit tests for elements inside struct and the double struct as well?

Btw click on re-request review so that I get notified.

spark/src/test/scala/org/apache/spark/sql/delta/stats/StatsCollectionSuite.scala
longvu-db: STRUCT<STRUCT<a, b>, c> for example?

spark/src/test/scala/org/apache/spark/sql/delta/stats/StatsCollectionSuite.scala
```diff
@@ -499,14 +499,13 @@ class StatsCollectionSuite
...
-    ("MAP<DATE, INT>", "MapType(DateType,IntegerType,true)"),
-    ("STRUCT<c60:INT, c61:ARRAY<INT>>", "ArrayType(IntegerType,true)")
+    ("MAP<DATE, INT>", "MapType(DateType,IntegerType,true)")
```
longvu-db, line 502: Could we also look at other tests in StatsCollectionSuite and update them, if applicable, for our use-cases?

Examples:
https://github.com/delta-io/delta/blob/da5a5d29b6f00265439bc84020580080a2d4568f/spark/src/test/scala/org/apache/spark/sql/delta/stats/StatsCollectionSuite.scala#L611
https://github.com/delta-io/delta/blob/da5a5d29b6f00265439bc84020580080a2d4568f/spark/src/test/scala/org/apache/spark/sql/delta/stats/StatsCollectionSuite.scala#L611

longvu-db:
@Kimahriman Just out of curiosity, why are you making this change? Do you have some complex nested schemas that you want to skip data on?

Kimahriman:
> @Kimahriman Just out of curiosity, why are you making this change? Do you have some complex nested schemas that you want to skip data on?

Yes, we have structs with mixed arrays and primitive types. Currently I have to recursively count the fields to set the num index columns to the right value

spark/src/test/scala/org/apache/spark/sql/delta/stats/StatsCollectionSuite.scala
Kimahriman: That's a valid type, there's nothing to do with that in this test

longvu-db: @Kimahriman Right, sorry for the confusion

spark/src/test/scala/org/apache/spark/sql/delta/stats/StatsCollectionSuite.scala
Kimahriman: Added more valid types to the test

spark/src/test/scala/org/apache/spark/sql/delta/stats/DataSkippingDeltaTests.scala
Kimahriman: Added more hit tests

spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala
```diff
@@ -458,15 +458,20 @@ object StatisticsCollection extends DeltaCommand {
...
-      columnPaths: ArrayBuffer[String]): Unit = dataType match {
+      columnPaths: ArrayBuffer[String],
+      insideStruct: Boolean = false): Unit = dataType match {
     case s: StructType =>
       s.foreach { field =>
-        validateDataSkippingType(name + "." + field.name, field.dataType, columnPaths)
+        validateDataSkippingType(name + "." + field.name, field.dataType, columnPaths,
+          insideStruct = true)
       }
+    case _ if insideStruct => columnPaths.append(name)
```
longvu-db, line 474: When we are inside the struct, we keep appending columns regardless of the data type right?

So would the description of the "@param columnPaths" no longer correct? 

This seems outdated as well

https://github.com/delta-io/delta/blob/da5a5d29b6f00265439bc84020580080a2d4568f/spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala#L481

spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala
```diff
@@ -458,15 +458,20 @@ object StatisticsCollection extends DeltaCommand {
    * @param name The name of the data skipping column for validating data type.
    * @param dataType The data type of the data skipping column.
    * @param columnPaths The column paths of all valid fields.
+   * @param insideStruct Whether the datatype is inside a struct already, in which case we don't
```
longvu-db, line 461: Instead of saying "we don't clear", we can make it clearer what we mean by that I think.

spark/src/test/scala/org/apache/spark/sql/delta/stats/DataSkippingDeltaTests.scala
longvu-db: My understanding is that when we are inside struct, we gather stats for the valid datatypes inside the struct, but the invalid ones we still don't right? Would we want to test that we still don't skip the invalid datatypes inside the struct?

spark/src/test/scala/org/apache/spark/sql/delta/stats/StatsCollectionSuite.scala
```diff
@@ -608,7 +590,8 @@ class StatsCollectionSuite
 
   Seq(
     "BIGINT", "DATE", "DECIMAL(3, 2)", "DOUBLE", "FLOAT", "INT", "SMALLINT", "STRING",
-    "TIMESTAMP", "TIMESTAMP_NTZ", "TINYINT"
+    "TIMESTAMP", "TIMESTAMP_NTZ", "TINYINT", "STRUCT<c3: BIGINT>",
```
longvu-db, line 593: Looks like there are relevant tests in DataSkippingDeltaTests as well https://github.com/delta-io/delta/blob/da5a5d29b6f00265439bc84020580080a2d4568f/spark/src/test/scala/org/apache/spark/sql/delta/stats/DataSkippingDeltaTests.scala#L442

spark/src/test/scala/org/apache/spark/sql/delta/stats/DataSkippingDeltaTests.scala
longvu-db: Also, if we can check another skipping valid type beside integer, like timestamp, string, ... that would be great

spark/src/test/scala/org/apache/spark/sql/delta/stats/StatsCollectionSuite.scala
Kimahriman: Those tests are for specifying the number of indexed columns, not columns by name

spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala
Kimahriman: Changed, not sure if it's anymore clear

spark/src/test/scala/org/apache/spark/sql/delta/stats/DataSkippingDeltaTests.scala
Kimahriman: Technically null counts are collected, but they can't be used for skipping. We could add that to the test, but seems kind of out of scope because you can't do min/max things with arrays/maps anyway, so it'd have to be like an element contains or something

longvu-db: I think there are other non-skipping eligible datatypes than lists datatypes

Kimahriman: I'm not sure how to specify a Binary type in the inferred JSON. Is that the only non-complex type?

kamcheungting-db:
Hi Adam,

I intentionally throw error when the column doesn't support delta stats.
The reason behind is: if we don't throw error, user does;t know why the delta stat is not collected. It is hard for user to differentiate between unsupported data type with a bug.

kamcheungting-db:
If we want to swallow the supported error, there should be a way to tell user that the column is unsupported.

Kimahriman:
Happy to log a warning, but it's not actually unsupported. Null counts are still collected for all columns. If users are confused why their arrays and maps aren't collecting mins and maxed, they have bigger problems. This behavior matches how the number of indexed columns work.

kamcheungting-db:
> ###
> **[Kimahriman](/Kimahriman) ** commented [Aug 18, 2024](#issuecomment-2295377047) •

Warming sounds a good indicator.

Kimahriman:
> Warming sounds a good indicator.

Added a warning log, let me know if you have thoughts on exact verbiage, just copied the error text

spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala
```diff
@@ -458,16 +458,22 @@ object StatisticsCollection extends DeltaCommand {
...
+      columnPaths: ArrayBuffer[String],
+      insideStruct: Boolean = false): Unit = dataType match {
     case s: StructType =>
       s.foreach { field =>
-        validateDataSkippingType(name + "." + field.name, field.dataType, columnPaths)
+        validateDataSkippingType(name + "." + field.name, field.dataType, columnPaths,
+          insideStruct = true)
       }
     case SkippingEligibleDataType(_) => columnPaths.append(name)
+    case d if insideStruct =>
```
kamcheungting-db, line 474: should we also make the non-struct field have the same behavior?

kamcheungting-db (approved):

spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala
Kimahriman: Up to you. The idea was if you directly specify an in-eligible type, throw an exception because you did something wrong. If you specify a struct, just work with the sub fields that are supported

kamcheungting-db: Then, I am good about it.
Thank you

Kimahriman:
Gentle ping, would be great to get this merged for the next release

longvu-db:
@Kimahriman Hey Adam, sorry I totally forgot about this, was very busy these days, I will take another look today.

spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala
```diff
@@ -459,16 +459,22 @@ object StatisticsCollection extends DeltaCommand {
...
```
longvu-db, line 475: @Kimahriman 

Could we make this `d` something more descriptive? 

I'm not sure if I follow this part here, why are we appending to the `columnPaths` `name` while logging this message

`"Data skipping is not supported for column $name of type $dataType"`

Will the current description still holds true "The column paths of all valid fields."? If not, I think we should modify it accordingly.

We can put a comment here as well since I'm not really sure what is the full intent of this part here.

spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala
```diff
@@ -459,16 +459,22 @@ object StatisticsCollection extends DeltaCommand {
    * @param name The name of the data skipping column for validating data type.
    * @param dataType The data type of the data skipping column.
    * @param columnPaths The column paths of all valid fields.
+   * @param insideStruct Tracks if we are inside a struct, in which case all types are valid
```
longvu-db, line 462: "we are" -> "The data skipping column that we are validating is inside a struct"

longvu-db, line 462:
```suggestion
   * @param insideStruct Tracks if we are inside a struct, in which case all types are valid.
```

longvu-db: I think we should specify why "all types are valid" here, it would be confusing if an array is invalid but struct<array> is not.

The goal here is to collect stats as much as possible for only the **valid** types inside a struct if I'm not wrong.

spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala
Kimahriman: In this case "valid" means we don't throw an error, not we can actually collect stats on it. We still append to the `columnPaths` because that's purely used to see if the user provided the same column twice. The warning is so users can be aware that stats _won't_ be collected on that column, but we don't throw an error because you didn't specify that column directly, it was inside a struct the user specified.

Kimahriman: Also `d` isn't used so just changing to `_`

Kimahriman: We could not append to columnPaths in that case, it doesn't _really_ matter. It's an odd edge case for ensuring the user didn't specify the same struct column multiple times

spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala
Kimahriman: Added more to the param doc to try to clarify more

Kimahriman: Updated this whole param doc more

spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala
longvu-db: @Kimahriman What does "user-specified struct" mean? I am not sure if I follow "as the user didn't specify them directly", isn't the user that creates the table and the schema?

I think logWarning logs to the Spark Logs right? I'm not sure if it is easily readable in that case as Spark Logs are massive and the messages will be like needle in a haystack.

Could we please update the definition of the function `validateDataSkippingType`, what "valid" mean and `columnPaths` accordingly in that case? The code should be straightforward and easily to understand IMO since the data skipping codebase is already complicated, it took me quite some time to read the function and trying to understand what the intention here, and how it actually works. 

If we have to explain this in a code comment then chances are it is worth making the comment as clear as possible in the codebase as more people will stumble across this function.

Kimahriman: This function is purely for parsing and validating `DATA_SKIPPING_STATS_COLUMNS`, so "user-specified" in this case means user specified values in `DATA_SKIPPING_STATS_COLUMNS`. I can add to the function comment that this is all the function is for. A user can specify a struct column in this config in which case children fields will automatically be included for stats collection, which is the case this whole PR is about. Any larger confusion about how it works in general or why it works this way would have to go back to the original authors.

spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala
```diff
@@ -454,21 +454,32 @@ object StatisticsCollection extends DeltaCommand {
   }
 
   /**
-   * This method validates that the data type of data skipping column supports data skipping
-   * based on file statistics.
+   * This method validates that the data type of a data skipping column provided in
+   * DATA_SKIPPING_STATS_COLUMNS supports data skipping based on file statistics. If a struct
```
longvu-db, line 458:
```suggestion
   * [[DeltaConfigs.DATA_SKIPPING_STATS_COLUMNS]] supports data skipping based on file statistics. If a struct
```

spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala
```diff
@@ -454,21 +454,32 @@ object StatisticsCollection extends DeltaCommand {
...
+   * column is specified, all its children are considered valid because stats will be collected
```
longvu-db, line 459: @Kimahriman This comment that you added made things so much clearer.

spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala
```diff
@@ -454,21 +454,32 @@ object StatisticsCollection extends DeltaCommand {
...
+   * for any supported types, and unsupported types will be ignored.
```
longvu-db, line 460: 1. 
```suggestion
   * for any supported types. For unsupported types, stats will not be collected.
```
@Kimahriman 

2. Could you please also add the reasons why we consider all data types specifically for the **struct** case to be considered valid here? E.g. to collect stats whenever we can and to be more lenient, as in if a child is unsupported then we will not through an exception and we have the stats of the other valid children.

3. If you can also specify the definition (can be short) of "valid" then it would also be awesome, I'm aware we didn't add it before but it would make things clearer.

longvu-db (commented):
Thanks @Kimahriman  , last round of comments.

spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala
Kimahriman: Made another update to try to clarify further

longvu-db (approved):
Thanks @Kimahriman for making the code much more clearer, LGTM!

PR #2854, opened 2024-04-05
[Spark] Add ConflictCheckerRowIdSuite

longvu-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Add the `ConflictCheckerRowIdSuite`, to test the Row Tracking's Conflict Checker logic. We want to ensure that each row ID can only be assigned by one transaction. If we detect that a concurrent transaction also assigned row IDs, then we will assign different row IDs to the files added by the current transaction.

## How was this patch tested?
Add UTs.

No.


larsk-db (approved):
LGTM

PR #2857, opened 2024-04-08
[Spark] Identity Columns APIs in DeltaColumnBuilder

c27kwan:
- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
This PR is part of https://github.com/delta-io/delta/issues/1959
* We introduce `generateAlwaysAsIdentity` and `generatedByDefaultAsIdentity`APIs into DeltaColumnBuilder so that users can create Delta table with Identity column. 
* We guard the creation of identity column tables with a feature flag until development is complete.

## How was this patch tested?
New tests. 

Yes, we introduce `generateAlwaysAsIdentity` and `generatedByDefaultAsIdentity` interfaces to DeltaColumnBuilder for creating identity columns. 
**Interfaces**
```
def generatedAlwaysAsIdentity(): DeltaColumnBuilder
def generatedAlwaysAsIdentity(start: Long, step: Long): DeltaColumnBuilder
def generatedByDefaultAsIdentity(): DeltaColumnBuilder
def generatedByDefaultAsIdentity(start: Long, step: Long): DeltaColumnBuilder
```
When the `start` and the `step` parameters are not specified, they default to `1L`. `generatedByDefaultAsIdentity` allows users to insert values into the column while a column specified with`generatedAlwaysAsIdentity` can only ever have system generated values.

**Example Usage**
```
// Creates a Delta identity column.
io.delta.tables.DeltaTable.columnBuilder(spark, "id")
      .dataType(LongType)
      .generatedAlwaysAsIdentity()
// Which is equivalent to the call
io.delta.tables.DeltaTable.columnBuilder(spark, "id")
      .dataType(LongType)
      .generatedAlwaysAsIdentity(start = 1L, step = 1L)
```


c27kwan:
I split the table feature part into a separate PR, which will need to be merged first: https://github.com/delta-io/delta/pull/2859

spark/src/main/scala/org/apache/spark/sql/delta/TableFeature.scala
```diff
@@ -353,6 +353,8 @@ object TableFeature {
         TestFeatureWithDependency,
         TestFeatureWithTransitiveDependency,
         TestWriterFeatureWithTransitiveDependency,
+        // identity columns are under development and only available in testing.
```
larsk-db, line 356:
```suggestion
        // Identity columns are under development and only available in testing.
```

larsk-db (approved):
LGTM

spark/src/main/scala/org/apache/spark/sql/delta/TableFeature.scala
c27kwan: Addressed in https://github.com/delta-io/delta/pull/2859 :)

spark/src/main/scala/io/delta/tables/DeltaColumnBuilder.scala
```diff
@@ -96,6 +99,69 @@ class DeltaColumnBuilder private[tables](
...
+
+  /**
+   * :: Evolving ::
+   *
+   * Specify a column as an identity column that is always generated by the system, i.e., does not
+   * allow user specified values.
+   *
+   * @param start the start value of the identity column
+   * @param step the increment step of the identity column
+   * @since 3.2.0
```
felipepessoto, line 123: Wow, glad to see this PR so quickly. Are you planning to release it in 3.2.0?
Thanks.

c27kwan: It's a stretch goal at the moment, since the release for Delta 3.2 is quite soon. I should know by end of week whether I can realistically make 3.2. I plan to leave this PR up for at least one week anyways. :)

It's safe to start merging the code even if we can only make 4.0 since the DeltaSQLConf guards creating identity columns.

c27kwan: For context, this PR is on the critical path for identity columns implementation in Delta 3.2. We need an interface to specify identity columns. The SQL syntax can only be introduced in the next Spark release (Spark 4), which means we can only have it for Delta 4.0 earliest.

spark/src/main/scala/io/delta/tables/DeltaColumnBuilder.scala
```diff
@@ -96,6 +99,69 @@ class DeltaColumnBuilder private[tables](
...
+   *
+   * Specify a column as an identity column that is always generated by the system, i.e., does not
+   * allow user specified values using a start of 1 and a step of 1.
+   *
+   * @since 3.2.0
+   */
+  @Evolving
+  def generatedAlwaysAsIdentity(): DeltaColumnBuilder = {
+    generatedAlwaysAsIdentity(start = 1, step = 1)
+  }
```
ala, line 113: Could we maybe use constants for the default start and step? 
We could then use it in spark/src/test/scala/org/apache/spark/sql/delta/DDLTestUtils.scala instead of hard-coding the 1s.

spark/src/main/scala/io/delta/tables/DeltaColumnBuilder.scala
```diff
@@ -96,6 +99,69 @@ class DeltaColumnBuilder private[tables](
     this
   }
 
+  /**
+   * :: Evolving ::
+   *
+   * Specify a column as an identity column that is always generated by the system, i.e., does not
+   * allow user specified values using a start of 1 and a step of 1.
```
ala, line 106: Could we perhaps reword this a bit? "does not allow user specified values using a start of 1 and a step of 1" is a lot without punctuation.

spark/src/main/scala/io/delta/tables/DeltaColumnBuilder.scala
```diff
@@ -123,6 +189,26 @@ class DeltaColumnBuilder private[tables](
...
     }
+
+    if (identityAllowExplicitInsert.isDefined) {
+      if (generationExpr.nonEmpty) {
+        throw DeltaErrors.identityColumnWithGenerationExpression()
+      }
+
+      if (dataType != null && dataType != LongType) {
+        throw DeltaErrors.identityColumnDataTypeNotSupported()
+      }
```
ala, line 200: Could we perhaps parametrise the error with the unsupported type?

spark/src/main/scala/org/apache/spark/sql/delta/DeltaErrors.scala
```diff
@@ -2398,6 +2398,18 @@ trait DeltaErrorsBase
...
+  }
+
+  def identityColumnIllegalStep(): Throwable = {
+    new AnalysisException("Identity column step cannot be 0.")
+  }
+
+  def identityColumnDataTypeNotSupported(): Throwable = {
+    new AnalysisException("Identity column does not support this data type.")
+  }
+
```
ala, line 2412: Could we perhaps use error classes for these?

spark/src/main/scala/org/apache/spark/sql/delta/DeltaErrors.scala
```diff
@@ -2398,6 +2398,18 @@ trait DeltaErrorsBase
...
+  def identityColumnWithGenerationExpression(): Throwable = {
+    new AnalysisException("Identity column cannot be specified with a generated column expression.")
+  }
+
+  def identityColumnIllegalStep(): Throwable = {
+    new AnalysisException("Identity column step cannot be 0.")
+  }
+
+  def identityColumnDataTypeNotSupported(): Throwable = {
+    new AnalysisException("Identity column does not support this data type.")
```
ala, line 2410: It would be nice to tell the user what is the supported type and the unsupported type :)

spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala
```diff
@@ -1709,6 +1709,20 @@ trait DeltaSQLConfBase {
...
+  // IDENTITY COLUMN
+  ///////////////////
+
+  val DELTA_IDENTITY_COLUMN_ENABLED =
+    buildConf("identityColumn.enabled")
+      .internal()
+      .doc(
+        """
+          | The umbrella config to turn on/off the IDENTITY column support.
+          | If true, enable Delta IDENTITY column support.
```
ala, line 1722: Could we perhaps explain here what is the expected behaviour if this is disabled?
If a table already had identity column, is it unreadable, uwritable with this setting off?

spark/src/test/scala/org/apache/spark/sql/delta/DDLTestUtils.scala
```diff
@@ -73,8 +74,54 @@ case class TestColumnSpec(
...
+        col.generatedAlwaysAsIdentity(start = 1L, step = step)
+      case (GeneratedAsIdentityType.GeneratedAlways, None, None) => col.generatedAlwaysAsIdentity()
+      case (GeneratedAsIdentityType.GeneratedByDefault, Some(start), Some(step)) =>
+        col.generatedByDefaultAsIdentity(start, step)
+      case (GeneratedAsIdentityType.GeneratedByDefault, Some(start), None) =>
+        col.generatedByDefaultAsIdentity(start = start, step = 1L)
+      case (GeneratedAsIdentityType.GeneratedByDefault, None, Some(step)) =>
+        col.generatedByDefaultAsIdentity(start = 1L, step = step)
+      case _ => col.generatedByDefaultAsIdentity()
+    }
```
ala, line 114: Could we perhaps use the default values to simplify this?
Like `val start = startsWith.getOrElse(default_value)`?

spark/src/main/scala/io/delta/tables/DeltaColumnBuilder.scala
felipepessoto: Got it. Just for my knowledge, the required changes in Spark 4 are only to support SQL syntax, meaning Scala could be supported?

c27kwan: That's part of what I'm assessing this week in my prototype. To put the UDF for generating identity column values in  only the Delta repo requires adding workarounds so that it works on Spark 3.5. Once I have enough confidence about the idea, I'll post a PR.

Everything else doesn't seem to require Spark changes.

spark/src/main/scala/org/apache/spark/sql/delta/DeltaErrors.scala
c27kwan: I checked and it seems the other existing error message don't provide the list of supported types. It would be more consistent to not report the supported types.

ala: That's OK.

ala (approved):
LGTM.
Thank you!

spark/src/main/scala/io/delta/tables/DeltaColumnBuilder.scala
c27kwan: I'm not going to make Delta 3.2. I've discussed with @tdas offline and we're aiming for the next Delta version, Delta 3.3. The API doc comment in this PR has been adjusted accordingly.

vkorukanti (approved):

PR #2859, opened 2024-04-08
[Spark] Add IdentityColumnsTableFeature

c27kwan:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
This PR is part of https://github.com/delta-io/delta/issues/1959

In this PR, we introduce the IdentityColumnsTableFeature to test-only so that we can start developing with it.

Note, we do not add support to minWriterVersion 6 yet to properties.defaults.minWriterVersion because that will enable the table feature outside of testing.

## How was this patch tested?
Existing tests pass. 

No, this is a test-only change.


xupefei (approved):
Lgtm!

PR #2860, opened 2024-04-08
[Spark] Always create checkpoint when dropping reader+writer features

andreaschat-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

In the DROP FEATURE command we generate a checkpoint after cleaning reader+writer feature traces to allow metadata cleaning operations to truncate history before the cleanup. However, the checkpoint is not created if the cleanup operation does not do any work. This might cause an issue in some cases if the cleanup operation does not perform any work but there are historical traces that needed to be truncated. A scenario like this may occur if the user manually cleans the feature traces before invoking the DROP FEATURE command. This PR resolves this issue by relaxing the conditions the checkpoint is created. Note, this has as a side effect to create a checkpoint in both invocations of the DROP FEATURE command.

## How was this patch tested?

Added a new test in DeltaProtocolVersionSuite.

No.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/TableFeature.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/TableFeature.scala
@@ -751,7 +751,7 @@ private[sql] object TestRemovableWriterFeature
   }
 
   override def validateRemoval(snapshot: Snapshot): Boolean =
-    !snapshot.metadata.configuration.contains(TABLE_PROP_KEY)
+    !snapshot.metadata.configuration.get(TABLE_PROP_KEY).exists(_.toBoolean)
 
   override def preDowngradeCommand(table: DeltaTableV2): PreDowngradeTableFeatureCommand =
     TestWriterFeaturePreDowngradeCommand(table)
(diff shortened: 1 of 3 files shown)


larsk-db (approved):
LGTM

xupefei (approved):
Lgtm!

spark/src/main/scala/org/apache/spark/sql/delta/TableFeature.scala
```diff
@@ -751,7 +751,7 @@ private[sql] object TestRemovableWriterFeature
   }
 
   override def validateRemoval(snapshot: Snapshot): Boolean =
-    !snapshot.metadata.configuration.contains(TABLE_PROP_KEY)
+    !snapshot.metadata.configuration.get(TABLE_PROP_KEY).exists(_.toBoolean)
```
ala, line 754: Can you please add a comment explaining what is this doing?
`Make sure that the property is not enabled on the table.`
Or alternatively `snapshot.metadata.configuration.get(TABLE_PROP_KEY).forall(!_.toBoolean)` is a bit more readable.

ala (approved):

PR #2865, opened 2024-04-08
[Spark] Remove preview guard for clustered table

zedtang:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Now that incremental clustering, ALTER TABLE CLUSTER BY and DESC DETAIL are merged, we can remove the preview guard for clustered table, to prepare for OSS 3.2 release.
## How was this patch tested?

Existing tests.

Yes. User will no longer get the preview warning when they try out clustered table.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala
@@ -1667,13 +1667,6 @@ trait DeltaSQLConfBase {
   // Clustered Table
   //////////////////
 
-  val DELTA_CLUSTERING_TABLE_PREVIEW_ENABLED =
-    buildConf("clusteredTable.enableClusteringTablePreview")
-      .internal()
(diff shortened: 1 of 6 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala
```diff
@@ -1667,13 +1667,6 @@ trait DeltaSQLConfBase {
   // Clustered Table
   //////////////////
 
-  val DELTA_CLUSTERING_TABLE_PREVIEW_ENABLED =
```
ami7o, line 1670: Is it safe to remove this config? Is it only used in testing

zedtang: It's a config that warns user when using clustered table in Delta 3.1. They'll need to explicitly turn on this config to use clustered table. In tests it's ignored.

It's safe to remove since it's only used as guard for preview feature.

PR #2856, opened 2024-04-06
[Spark] Add read support for RowId

longvu-db:
#### Which Delta project/connector is this regarding?

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
1. Add the Analyzer Rule `GenerateRowIds` to generate default Row IDs.
2. Add the `row_id` field to the `_metadata` column for Delta tables, allowing us to read the `row_id` from the file metadata after it is stored.

## How was this patch tested?
Added UTs.

## Does this PR introduce _any_ user-facing changes?

No.

PR #1896, opened 2023-07-11
Assign materialized Row ID and Row commit version column names

johanl-db:
## Description
This change is part of implementing row tracking as specified in https://github.com/delta-io/delta/pull/1610 and https://github.com/delta-io/delta/commit/7272b04db76507a98e07e96dc88e2c1a54283693.
It covers assigning a column name for the materialized Row ID and Row commit version columns by setting them in the table metadata when creating or cloning a table.

## How was this patch tested?
- Add test suite `rowtracking.MaterializedColumnSuite` to cover assigning materialized column names in various table creation and clone scenarios.

PR #1781, opened 2023-05-22
Add Default Row Commit Version to AddFile and RemoveFile

tomvanbussel:
## Description

This PR implements part of the changes proposed in https://github.com/delta-io/delta/pull/1747. It adds the `defaultRowCommitVersion` field to `AddFile` and `RemoveFile`, and it makes sure that it's populated during commits and read during log replay. It **does not** handle any transaction conflicts yet.

## How was this patch tested?

Added a new test suite `DefaultRowCommitVersionSuite`.

## Does this PR introduce _any_ user-facing changes?

No

PR #1723, opened 2023-04-25
Assign fresh Row IDs during commit

johanl-db:
## Description
This change implements assigning unique fresh Row IDs when committing files on tables that support Row Ids:
- Set the `baseRowId` field of every `add` and `remove` actions in commits.
- Generate `highWaterMark` action to update the Row ID high watermark. 
- Gracefully resolve conflicts between transactions by reassigning overlapping Row IDs before committing.

## How was this patch tested?
- Adding tests to RowIdSuite to cover assigning fresh Row IDs.
- Adding tests to CheckpointSuite to ensure `baseRowId` and `highWaterMark` information survives checkpointing.

## Does this PR introduce _any_ user-facing changes?
No, assigning fresh row ids is hidden from users. Reading row ids will come in a follow-up PR.
Note that row ids are currently hidden behind property `spark.databricks.delta.rowIds.allowForDevOnly` to avoid leaking a partial implementation.

PR #2866, opened 2024-04-08
[Spark] Enable Row Tracking outside of testing

longvu-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Delta Protocol for Row IDs was introduced in this PR: https://github.com/delta-io/delta/pull/1610

Support for writing fresh row IDs / row commit versions was introduced in the following PRs:
- https://github.com/delta-io/delta/pull/1723
- https://github.com/delta-io/delta/pull/1781
- https://github.com/delta-io/delta/pull/1896

**This is sufficient to enable row tracking on a table and write to a table that has row tracking enabled** but not to actually read row IDs / row commit versions back, which is also being added in Delta at the moment ([read BaseRowId](https://github.com/delta-io/delta/commit/283ac02c0510ce67744ff5c410ca416f7fbaa0b9). [read defaultRowCommitVersion](https://github.com/delta-io/delta/pull/2795), [read RowId](https://github.com/delta-io/delta/pull/2856)...)

Using row tracking is currently only allowed in testing, this change allows enabling row tracking outside of testing so that the upcoming Delta 3.2 release includes support for writing to tables with row tracking enabled, making Delta writers future-proof.

## How was this patch tested?
Tests have already been added in previous changes, this only flips the switch to let users enabled Row Tracking outside of tests.

Users are now able to enable Row Tracking when creating a delta table:
```
CREATE TABLE tbl(a int) USING DELTA TBLPROPERTIES ('delta.enableRowTracking' = 'true')
```

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/TableFeature.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/TableFeature.scala
@@ -336,7 +336,8 @@ object TableFeature {
       IcebergCompatV2TableFeature,
       DeletionVectorsTableFeature,
       VacuumProtocolCheckTableFeature,
-      V2CheckpointTableFeature)
+      V2CheckpointTableFeature,
+      RowTrackingFeature)
     if (DeltaUtils.isTesting) {
       features ++= Set(
         TestLegacyWriterFeature,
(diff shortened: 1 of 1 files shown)


tomvanbussel (approved):
Awesome! We have the necessary tests in place now, so it LGTM to move this out of testing.

Issue #2864, opened 2024-04-08
[PROTOCOL RFC] Support Variant data type


gene-db:
### Description of the protocol change

This change adds support for the Variant data type. The Variant data type helps processing and storing semi-structured data in a flexible and efficient way.

[Design Doc](https://docs.google.com/document/d/19QWDzSMFiWnpp3t7phsAWe1qV7tMv-uqtmJDxtMgK4A/edit?usp=sharing)

### Willingness to contribute

The Delta Lake Community encourages protocol innovations. Would you or another member of your organization be willing to contribute this feature to the Delta Lake code base?

- [x] Yes. I can contribute.
- [ ] Yes. I would be willing to contribute with guidance from the Delta Lake community.
- [ ] No. I cannot contribute at this time.

allisonport-db:
This has been merged into `PROTOCOL.md`

PR #2867, opened 2024-04-08
[PROTOCOL RFC] Variant data type

gene-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Adds the protocol changes for the Variant data type (see https://github.com/delta-io/delta/issues/2864) to the RFC folder.

## How was this patch tested?

N/A

N/A

### DIFF
--- a/protocol_rfcs/README.md
+++ b/protocol_rfcs/README.md
@@ -22,6 +22,7 @@ Here is the history of all the RFCs propose/accepted/rejected since Feb 6, 2024,
 | 2023-02-09    | [type-widening.md](https://github.com/delta-io/delta/blob/master/protocol_rfcs/type-widening.md)                                  | https://github.com/delta-io/delta/issues/2623 | Type Widening                 |
 | 2023-02-14    | [managed-commits.md](https://github.com/delta-io/delta/blob/master/protocol_rfcs/managed-commits.md)                              | https://github.com/delta-io/delta/issues/2598 | Managed Commits               |
 | 2023-02-26    | [column-mapping-usage.tracking.md](https://github.com/delta-io/delta/blob/master/protocol_rfcs/column-mapping-usage-tracking.md)) | https://github.com/delta-io/delta/issues/2682 | Column Mapping Usage Tracking |
+| 2023-04-08    | [variant-type.md](https://github.com/delta-io/delta/blob/master/protocol_rfcs/variant-type.md)                                    | https://github.com/delta-io/delta/issues/2864 | Variant Data Type             |
 
 ### Accepted RFCs
 
--- a/protocol_rfcs/README.md
+++ b/protocol_rfcs/README.md
@@ -22,7 +22,7 @@ Here is the history of all the RFCs propose/accepted/rejected since Feb 6, 2024,
 | 2023-02-09    | [type-widening.md](https://github.com/delta-io/delta/blob/master/protocol_rfcs/type-widening.md)                                  | https://github.com/delta-io/delta/issues/2623 | Type Widening                 |
 | 2023-02-14    | [managed-commits.md](https://github.com/delta-io/delta/blob/master/protocol_rfcs/managed-commits.md)                              | https://github.com/delta-io/delta/issues/2598 | Managed Commits               |
 | 2023-02-26    | [column-mapping-usage.tracking.md](https://github.com/delta-io/delta/blob/master/protocol_rfcs/column-mapping-usage-tracking.md)) | https://github.com/delta-io/delta/issues/2682 | Column Mapping Usage Tracking |
-| 2023-04-08    | [variant-type.md](https://github.com/delta-io/delta/blob/master/protocol_rfcs/variant-type.md)                                    | https://github.com/delta-io/delta/issues/2864 | Variant Data Type             |
+| 2023-04-24    | [variant-type.md](https://github.com/delta-io/delta/blob/master/protocol_rfcs/variant-type.md)                                    | https://github.com/delta-io/delta/issues/2864 | Variant Data Type             |
 
 ### Accepted RFCs
 
--- a/protocol_rfcs/variant-type.md
+++ b/protocol_rfcs/variant-type.md
@@ -44,8 +44,8 @@ To support this feature:
 
 ## Variant data in Parquet
 
-The Variant data type is represented as 2 binary encoded values, according to the [Variant binary encoding specification](https://github.com/apache/spark/blob/master/common/variant/README.md).
-The 2 binary values are named `value` and `metadata`.
+The Variant data type is represented as two binary encoded values, according to the [Variant binary encoding specification](https://github.com/apache/spark/blob/master/common/variant/README.md).
+The two binary values are named `value` and `metadata`.
 
 When writing Variant data to parquet files, the Variant data is written as a single Parquet struct, with the following fields:
 
@@ -54,21 +54,22 @@ Struct field name | Parquet primitive type | Description
 value | binary | The binary-encoded Variant value, as described in [Variant binary encoding](https://github.com/apache/spark/blob/master/common/variant/README.md)
 metadata | binary | The binary-encoded Variant metadata, as described in [Variant binary encoding](https://github.com/apache/spark/blob/master/common/variant/README.md)
 
-The parquet struct must include the 2 struct fields `value` and `metadata`.
-Supported writers must write the 2 binary fields, and supported readers must read the 2 binary fields.
+The parquet struct must include the two struct fields `value` and `metadata`.
+Supported writers must write the two binary fields, and supported readers must read the two binary fields.
 Struct fields which start with `_` (underscore) can be safely ignored.
 The only non-ignorable fields must be `value` and `metadata`.
 
 ## Writer Requirements for Variant Data Type
 
 When Variant type is supported (`writerFeatures` field of a table's `protocol` action contains `variantType`), writers:
-- must write the 2 parquet struct fields, `value` and `metadata`, according to the [Variant binary encoding specification](https://github.com/apache/spark/blob/master/common/variant/README.md)
+- must write a column of type `variant` to parquet as a struct containing the fields `value` and `metadata` and storing values that conform to the [Variant binary encoding specification](https://github.com/apache/spark/blob/master/common/variant/README.md)
 - must not write additional, non-ignorable parquet struct fields. Writing additional struct fields with names starting with `_` (underscore) is allowed.
 
 ## Reader Requirements for Variant Data Type
 
 When Variant type is supported (`readerFeatures` field of a table's `protocol` action contains `variantType`), readers:
-- must be able to read the 2 parquet struct fields, `value` and `metadata`
+- must be able to read the two parquet struct fields, `value` and `metadata` and interpret them as a Variant in concordance with the [Variant binary encoding specification](https://github.com/apache/spark/blob/master/common/variant/README.md).
+- It is recommended but not required for a Delta reader to treat the struct as a single indivisible Variant field, if the reader is used in an engine or other context that supports Variant.
 - can ignore any parquet struct field names starting with `_` (underscore)
 
 ## Compatibility with other Delta Features
--- a/protocol_rfcs/variant-type.md
+++ b/protocol_rfcs/variant-type.md
@@ -2,7 +2,7 @@
 **Associated Github issue for discussions: https://github.com/delta-io/delta/issues/2864**
 
 This protocol change adds support for the Variant data type.
-The Variant data type is beneficial for storing and processing semi-structure data.
+The Variant data type is beneficial for storing and processing semi-structured data.
 
 --------
 
@@ -10,12 +10,12 @@ The Variant data type is beneficial for storing and processing semi-structure da
 
 # Variant Data Type
 
-This feature introduces the Variant data type, for processing with semi-structured data.
+This feature enables support for the Variant data type, for storing semi-structured data.
 The schema serialization method is described in [Schema Serialization Format](#schema-serialization-format).
 
 To support this feature:
 - The table must be on Reader Version 3 and Writer Version 7
-- The feature `variantType` must exist in the table `protocol`'s `readerFeatures` and `writerFeatures`, either during its creation or at a later stage.
+- The feature `variantType` must exist in the table `protocol`'s `readerFeatures` and `writerFeatures`.
 
 ## Example JSON-Encoded Delta Table Schema with Variant types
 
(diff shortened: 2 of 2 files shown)


protocol_rfcs/variant-type.md
```diff
@@ -0,0 +1,87 @@
+# Variant Data Type
+**Associated Github issue for discussions: https://github.com/delta-io/delta/issues/2864**
```
tdas, line 2: Would be better to follow the structure of the protocol doc - Requirements for readers and requirements for writers.

protocol_rfcs/variant-type.md
```diff
@@ -0,0 +1,87 @@
...
+
+This protocol change adds support for the Variant data type.
+The Variant data type is beneficial for storing and processing semi-structure data.
+
+--------
+
+> ***New Section after the [Clustered Table](#clustered-table) section***
```
tdas, line 9: Would be better to follow the structure of other features in the protocol doc - Requirements for readers and requirements for writers.

protocol_rfcs/variant-type.md
```diff
@@ -0,0 +1,87 @@
...
+The parquet struct must include the 2 struct fields `value` and `metadata`.
+Supported writers must write the 2 binary fields, and supported readers must read the 2 binary fields.
+Struct fields which start with `_` (underscore) can be safely ignored.
+The only non-ignorable fields must be `value` and `metadata`.
+
+## Compatibility with other Delta Features
+
+Feature | Support for Variant Data Type
+-|-
+Partition Columns | &#x2705; A Variant column is allowed to be a non-partitioned column of a partitioned table. <br/> &#x274C; Variant is not a comparable data type, so it cannot be included in a partition column.
```
tdas, line 67: do we need these unicode characters? they are nice for the design doc, but not useful in a protocol doc.

gene-db: Removed.

protocol_rfcs/variant-type.md
gene-db: Updated with reader and writer requirements.

protocol_rfcs/variant-type.md
```diff
@@ -0,0 +1,98 @@
...
+
+The Variant data type is represented as 2 binary encoded values, according to the [Variant binary encoding specification](https://github.com/apache/spark/blob/master/common/variant/README.md).
+The 2 binary values are named `value` and `metadata`.
+
+When writing Variant data to parquet files, the Variant data is written as a single Parquet struct, with the following fields:
+
+Struct field name | Parquet primitive type | Description
+-|-|-
+value | binary | The binary-encoded Variant value, as described in [Variant binary encoding](https://github.com/apache/spark/blob/master/common/variant/README.md)
+metadata | binary | The binary-encoded Variant  metadata, as described in [Variant binary encoding](https://github.com/apache/spark/blob/master/common/variant/README.md)
```
bart-samwel, line 55:
```suggestion
metadata | binary | The binary-encoded Variant metadata, as described in [Variant binary encoding](https://github.com/apache/spark/blob/master/common/variant/README.md)
```

protocol_rfcs/variant-type.md
```diff
@@ -0,0 +1,98 @@
...
+
+The parquet struct must include the 2 struct fields `value` and `metadata`.
+Supported writers must write the 2 binary fields, and supported readers must read the 2 binary fields.
+Struct fields which start with `_` (underscore) can be safely ignored.
+The only non-ignorable fields must be `value` and `metadata`.
```
bart-samwel, line 60: This seems to say (somewhat implicitly) that any field that is not `value` or `metadata` should start with an underscore? I don't see why that is necessary, if we define that only `value` and `metadata` should be non-ignorable.

protocol_rfcs/variant-type.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Writer Requirements for Variant Data Type
+
+When Variant type is supported (`writerFeatures` field of a table's `protocol` action contains `variantType`), writers:
+- must write the 2 parquet struct fields, `value` and `metadata`, according to the [Variant binary encoding specification](https://github.com/apache/spark/blob/master/common/variant/README.md)
+- must not write additional parquet struct fields
```
bart-samwel, line 66: We're explicitly saying above that other fields can be ignored. So then why do we need to say that writers can't write other fields? One of the two requirements is superfluous.

protocol_rfcs/variant-type.md
```diff
@@ -0,0 +1,98 @@
...
+Delta Column Statistics | **Supported:** A Variant column supports the `nullCount` statistic. <br/> **Unsupported:** Variant is not a comparable data type, so a Variant column does not support the `minValues` and `maxValues` statistics.
+Generated Columns | **Supported:** A Variant column is allowed to be used as a source in a generated column expression, as long as the Variant type is not the result type of the generated column expression. <br/> **Unsupported:** The Variant data type is not allowed to be the result type of a generated column expression.
+Delta CHECK Constraints | **Supported:** A Variant column is allowed to be used for a CHECK constraint expression.
+Default Column Values | **Supported:** A Variant column is allowed to have a default column value.
+Change Data Feed | **Supported:** A table using the Variant data type is allowed to enable the Delta Change Data Feed.
+Convert-to-Delta Command | **Supported:** A Variant Parquet table following the Spark Variant specification, can be converted to Delta with the CONVERT TO DELTA command.
```
bart-samwel, line 85: CONVERT TO DELTA is not part of the protocol spec. It's a connector specific command. This doc is only about the protocol, not about any specific implementation.

protocol_rfcs/variant-type.md
```diff
@@ -0,0 +1,98 @@
...
+Create Table Clone Command | **Supported:** A Variant table following the Spark Variant specification, can be deep or shallow cloned.
```
bart-samwel, line 86: Same here.

protocol_rfcs/variant-type.md
gene-db: I added clarification, but I wanted to say that the writer should not write additional, non-ignorable fields. Writing additional ignorable fields are allowed.

protocol_rfcs/variant-type.md
gene-db: Something we would like to support in the future is to introduce another non-ignorable struct field. That would require a new table feature, but it would depend on this table feature. In that case, how can we specify in the protocol that the new table feature would use an additional, non-ignorable field?

gene-db (commented):
@bart-samwel Thanks! I updated the PR.

I also left a question about how we could specify the non-ignorable fields spec regarding potential future enhancements?

protocol_rfcs/variant-type.md
```diff
@@ -0,0 +1,96 @@
...
+
+The parquet struct must include the 2 struct fields `value` and `metadata`.
+Supported writers must write the 2 binary fields, and supported readers must read the 2 binary fields.
+Struct fields which start with `_` (underscore) can be safely ignored.
+The only non-ignorable fields must be `value` and `metadata`.
+
+## Writer Requirements for Variant Data Type
+
+When Variant type is supported (`writerFeatures` field of a table's `protocol` action contains `variantType`), writers:
+- must write the 2 parquet struct fields, `value` and `metadata`, according to the [Variant binary encoding specification](https://github.com/apache/spark/blob/master/common/variant/README.md)
```
bart-samwel, line 65:
```suggestion
- must write a column of type `variant` to parquet as a struct containing the fields `value` and `metadata` and storing values that conform to the [Variant binary encoding specification](https://github.com/apache/spark/blob/master/common/variant/README.md)
```

protocol_rfcs/variant-type.md
```diff
@@ -0,0 +1,96 @@
...
+- must not write additional, non-ignorable parquet struct fields. Writing additional struct fields with names starting with `_` (underscore) is allowed.
+
+## Reader Requirements for Variant Data Type
+
+When Variant type is supported (`readerFeatures` field of a table's `protocol` action contains `variantType`), readers:
+- must be able to read the 2 parquet struct fields, `value` and `metadata`
```
bart-samwel, line 71: Do we need to say anything about what they are supposed to do with this? Should a Delta reader expose this as a unified VARIANT type? Probably not all readers can do this because their engines don't support this. But how about the suggestion below?

Oh, and style nit: "2" is usually written as a word in prose. Numbers are for higher values.
 
```suggestion
- must be able to read the two parquet struct fields, `value` and `metadata` and interpret them as a Variant in concordance with the [Variant binary encoding specification](https://github.com/apache/spark/blob/master/common/variant/README.md).
- It is recommended but not required for a Delta reader to treat the struct as a single indivisible Variant field, if the reader is used in an engine or other context that supports Variant.
```

protocol_rfcs/variant-type.md
gene-db: Updated.

protocol_rfcs/variant-type.md
gene-db: This suggestion sounds good! Updated.

bart-samwel (approved):

protocol_rfcs/variant-type.md
```diff
@@ -0,0 +1,97 @@
...
+    },
+    "nullable" : false,
+    "metadata" : { }
+  } ]
+}
+```
+
+## Variant data in Parquet
+
+The Variant data type is represented as two binary encoded values, according to the [Variant binary encoding specification](https://github.com/apache/spark/blob/master/common/variant/README.md).
```
scovich, line 47:
```suggestion
The Variant data type is represented as two binary encoded values, according to the [Spark Variant binary encoding specification](https://github.com/apache/spark/blob/master/common/variant/README.md).
```

protocol_rfcs/variant-type.md
scovich: The Delta spec already states that clients should ignore unrecognized fields, does that not suffice?
If the idea is to carve out a "safe" namespace for custom fields, then `_` prefix seems reasonable?

protocol_rfcs/variant-type.md
```diff
@@ -0,0 +1,97 @@
...
+## Writer Requirements for Variant Data Type
+
+When Variant type is supported (`writerFeatures` field of a table's `protocol` action contains `variantType`), writers:
+- must write a column of type `variant` to parquet as a struct containing the fields `value` and `metadata` and storing values that conform to the [Variant binary encoding specification](https://github.com/apache/spark/blob/master/common/variant/README.md)
+- must not write additional, non-ignorable parquet struct fields. Writing additional struct fields with names starting with `_` (underscore) is allowed.
+
+## Reader Requirements for Variant Data Type
+
+When Variant type is supported (`readerFeatures` field of a table's `protocol` action contains `variantType`), readers:
+- must be able to read the two parquet struct fields, `value` and `metadata` and interpret them as a Variant in concordance with the [Variant binary encoding specification](https://github.com/apache/spark/blob/master/common/variant/README.md).
```
scovich, line 71: As worded here, this "must be able to read... and interpret... in accordance with the... specification" would seem to contradict the next line that allows to not fully support the logical datatype?

I guess the main requirements are really:
1. Reader must recognize and tolerate a `variant` data type in a Delta schema
2. Reader must use the correct physical schema (struct-of-binary) when reading a Variant column from file
3. Reader must make the column available to queries:
   * [Recommended] Expose the logical Variant column, if the engine supports Variant.
   * [Alternate] Expose the raw physical column, e.g. if the engine does not support Variant.

gene-db: Thanks! Updated.

gene-db (commented):
@scovich Thanks! Updated the PR.

protocol_rfcs/variant-type.md
```diff
@@ -68,9 +67,11 @@ When Variant type is supported (`writerFeatures` field of a table's `protocol` a
...
-- must be able to read the two parquet struct fields, `value` and `metadata` and interpret them as a Variant in concordance with the [Variant binary encoding specification](https://github.com/apache/spark/blob/master/common/variant/README.md).
-- It is recommended but not required for a Delta reader to treat the struct as a single indivisible Variant field, if the reader is used in an engine or other context that supports Variant.
-- can ignore any parquet struct field names starting with `_` (underscore)
+- must recognize and tolerate a `variant` data type in a Delta schema
+- must use the correct physical schema (struct-of-binary, with fields `value` and `metadata`) when reading a Variant data type from file
+- must make the column available to the engine:
+    - [Recommended] Expose and interpret the struct-of-binary as a single Variant field in accordance with the [Spark Variant binary encoding specification](https://github.com/apache/spark/blob/master/common/variant/README.md).
+    - [Alternate] Expose the raw physical struct-of-binary, e.g. if the engine does not support Variant.
```
scovich, line 74: Is it also an acceptable alternate, to expose a string column by internally converting the struct-of-binary? Gives up all the benefits of the variant encoding, but maximizes compat with engines that don't allow users to load the library code that would interpret the struct-of-binary directly?

scovich (approved):
LGTM. One question.

protocol_rfcs/variant-type.md
gene-db: Yeah, that should work. I added that as another alternate.

protocol_rfcs/README.md
```diff
@@ -22,6 +22,7 @@ Here is the history of all the RFCs propose/accepted/rejected since Feb 6, 2024,
 | 2023-02-09    | [type-widening.md](https://github.com/delta-io/delta/blob/master/protocol_rfcs/type-widening.md)                                  | https://github.com/delta-io/delta/issues/2623 | Type Widening                 |
 | 2023-02-14    | [managed-commits.md](https://github.com/delta-io/delta/blob/master/protocol_rfcs/managed-commits.md)                              | https://github.com/delta-io/delta/issues/2598 | Managed Commits               |
 | 2023-02-26    | [column-mapping-usage.tracking.md](https://github.com/delta-io/delta/blob/master/protocol_rfcs/column-mapping-usage-tracking.md)) | https://github.com/delta-io/delta/issues/2682 | Column Mapping Usage Tracking |
+| 2023-04-08    | [variant-type.md](https://github.com/delta-io/delta/blob/master/protocol_rfcs/variant-type.md)                                    | https://github.com/delta-io/delta/issues/2864 | Variant Data Type             |
```
tdas, line 25: the gap in this date, and when we are ready to merge is large. can you update this to todays; date

gene-db: Updated the date.

tdas (approved):

PR #2870, opened 2024-04-09
[Spark] Don't use fixed tahoe id for testing

larsk-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

 * Do not use the literal "testId" for every single tahoeId in testing. This can hide issues where we use the wrong id (for example as key in a map).
 * Fix one of these issues with the partition stats for auto compaction, where we used a completely random tableId as the key, rather than than the tableId that commitLarge just wrote.
 * Add a (temporary) flag to not include the tahoeId in the equals and hashCode methods of TahoeLogFileIndex. Having an unstable external field there is prone to race conditions (of the form this != this) and losing instances in hash sets/maps. So ideally we should make this the default.

## How was this patch tested?

Testing-only PR.

No.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala
@@ -181,6 +181,18 @@ trait DeltaSQLConfBase {
       .booleanConf
       .createWithDefault(true)
 
+  val DELTA_INCLUDE_TABLE_ID_IN_FILE_INDEX_COMPARISON =
+    buildConf("includeTableIdInFileIndexComparison")
+      .internal()
+      .doc(
+        """
+          |Include the deltaLog.tableId field in equals and hashCode for TahoeLogFileIndex.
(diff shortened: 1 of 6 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala
```diff
@@ -181,6 +181,18 @@ trait DeltaSQLConfBase {
       .booleanConf
       .createWithDefault(true)
 
+  val DELTA_INCLUDE_TABLE_ID_IN_FILE_INDEX_COMPARISON =
```
cstavr, line 184: This is currently unused. I guess it is a debug flag?

larsk-db: Yes, for debugging/testing on a real cluster for now.

cstavr (approved):
LGTM. Thanks!

PR #2871, opened 2024-04-09
[Flink][WIP] Support using with Flink 1.18

auroranockert:

- [ ] Spark
- [ ] Standalone
- [x] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This adds changes to support using Delta with Flink 1.18.

## Work in progress

 - [ ] Decide if the Flink version we build against really should be changed to 1.18.1
 - [ ] Update the README with version numbers that actually support newer Flink
 - [ ] Fix the Mima settings so that it does not think that things in `io.delta.flink.internal` are actually exposed API.

I think these should be done as separate PRs but I am not sure, depending on what you think I'll either do another PR for the Mima things or fix it properly in here.

## How was this patch tested?

It was tested via the test suite (on Java 8, running on arm64 macOS) and some minor manual testing.

```
[info] Passed: Total 568, Failed 0, Errors 0, Passed 565, Skipped 3
[success] Total time: 1142 s (19:02), completed Apr 9, 2024 4:36:35 PM
```

If you have any ideas on how to better test this I would love to know, since I think I can see a patch to support Flink 1.19 in my future and would love to be more confident in that one.

No. It should not beyond supporting a newer version of Flink.

## Thanks

Thanks to @sdahlbac and @oskarskog for helping me with this, I definitely would not have had the patience to run the test suite this many without them.

### DIFF
--- a/connectors/flink/src/main/java/io/delta/flink/internal/table/DeltaCatalogTableHelper.java
+++ b/connectors/flink/src/main/java/io/delta/flink/internal/table/DeltaCatalogTableHelper.java
@@ -6,7 +6,6 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
-import java.util.Optional;
 import java.util.stream.Collectors;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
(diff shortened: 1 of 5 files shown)


connectors/flink/src/main/java/io/delta/flink/internal/table/DeltaCatalogTableHelper.java
```diff
@@ -324,6 +289,41 @@ public static Map<String, String> prepareDeltaTableProperties(
         return deltaLogProperties;
     }
 
+    /**
+     * Prepare catalog table to store in metastore. This table will have only selected
+     * options from DDL and an empty schema.
+     */
+    public static ResolvedCatalogTable prepareMetastoreTable(
```
scottsand-db, line 296: Hey - why move this code from above to below, here?

Could you edit it in the same place in the file? This makes seeing the git diff much easier

scottsand-db (changes requested):
Hey @auroranockert - thanks for this PR! I think we'd need to ask some users how they feel about this Flink 1.18 upgrade. Please respond to my 1 comment on this PR and make the necessary changes and let's come back to this

foss-contributor:
@auroranockert, thank you for your contribution! This pull request has been inactive for over 18 months and will be closed in one week. To keep it open, please leave a comment on this pull request or contact us in the `#contributors` channel on [Delta Lake Slack](https://go.delta.io/slack). We appreciate your involvement in Delta Lake!

PR #2873, opened 2024-04-09
[Spark] Add drop support for InCommitTimestamp table feature

dhruvarya-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Adds drop feature support for InCommitTimestamps. When the user runs DROP FEATURE, the following will happen if ICT is present in the PROTOCOL:
1. If any of the ICT-related properties are present, the first commit will:
   - Set `IN_COMMIT_TIMESTAMPS_ENABLED` = true
   - Remove `IN_COMMIT_TIMESTAMP_ENABLEMENT_VERSION` and `IN_COMMIT_TIMESTAMP_ENABLEMENT_TIMESTAMP`
2. A second commit will remove ICT from the protocol.

## How was this patch tested?

New tests in the DeltaProtocolVersionSuite for the following scenarios:
1. When ICT is enabled from commit 0 onwards.
2. When ICT is enabled in some commit after 0.
3.  Dropping when the feature is not there in protocol
4. Dropping when only one provenance property is present and even the enablement property is not present
5. Dropping when none of the table properties are present

Yes. Users will now be able to run ALTER TABLE <> DROP FEATURE inCommitTimestamps-dev on their tables.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaOperations.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaOperations.scala
@@ -356,14 +356,6 @@ object DeltaOperations {
       "properties" -> JsonUtils.toJson(propKeys),
       "ifExists" -> ifExists)
   }
-  /** Recorded when the table properties are set and/or removed in the same commit. */
-  case class UpdateTableProperties(
-      setProperties: Map[String, String],
-      unsetProperties: Seq[String]) extends Operation("UPDATE TBLPROPERTIES") {
-    override val parameters: Map[String, Any] = Map(
-      "setProperties" -> JsonUtils.toJson(setProperties),
-      "unsetProperties" -> JsonUtils.toJson(unsetProperties))
-  }
   /** Recorded when dropping a table feature. */
   case class DropTableFeature(
       featureName: String,
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaOperations.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaOperations.scala
@@ -356,6 +356,14 @@ object DeltaOperations {
       "properties" -> JsonUtils.toJson(propKeys),
       "ifExists" -> ifExists)
   }
+  /** Recorded when the table properties are set and/or removed in the same commit. */
+  case class UpdateTableProperties(
+      setProperties: Map[String, String],
+      unsetProperties: Seq[String]) extends Operation("UPDATE TBLPROPERTIES") {
+    override val parameters: Map[String, Any] = Map(
+      "setProperties" -> JsonUtils.toJson(setProperties),
+      "unsetProperties" -> JsonUtils.toJson(unsetProperties))
+  }
   /** Recorded when dropping a table feature. */
   case class DropTableFeature(
       featureName: String,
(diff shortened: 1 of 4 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/DeltaOperations.scala
```diff
@@ -356,6 +356,14 @@ object DeltaOperations {
       "properties" -> JsonUtils.toJson(propKeys),
       "ifExists" -> ifExists)
   }
+  /** Recorded when the table properties are set and/or removed in the same commit. */
+  case class UpdateTableProperties(
```
prakharjain09, line 360: Instead of introducing a new operation, could we reuse `UnsetTableProperties` ?

spark/src/test/scala/org/apache/spark/sql/delta/DeltaProtocolVersionSuite.scala
```diff
@@ -3571,6 +3571,150 @@ trait DeltaProtocolVersionSuiteBase extends QueryTest
...
+      // Remove the enablement property.
+      AlterTableUnsetPropertiesDeltaCommand(
+        DeltaTableV2(spark, deltaLog.dataPath),
+        Seq(DeltaConfigs.IN_COMMIT_TIMESTAMPS_ENABLED.key),
+        ifExists = true).run(spark)
+
+      AlterTableDropFeatureDeltaCommand(
+        DeltaTableV2(spark, deltaLog.dataPath),
+        InCommitTimestampTableFeature.name)
+        .run(spark)
```
prakharjain09, line 3711: Can we also add assertion over usage logs?

spark/src/test/scala/org/apache/spark/sql/delta/DeltaProtocolVersionSuite.scala
```diff
@@ -3571,6 +3571,150 @@ trait DeltaProtocolVersionSuiteBase extends QueryTest
...
+        DeltaTableV2(spark, deltaLog.dataPath),
+        Map(DeltaConfigs.IN_COMMIT_TIMESTAMP_ENABLEMENT_VERSION.key -> "1")).run(spark)
+      val snapshot1 = deltaLog.update()
+      assert(snapshot1.protocol.writerFeatureNames.contains(InCommitTimestampTableFeature.name))
+      // Ensure that the enablement property is not set.
+      assert(!DeltaConfigs.IN_COMMIT_TIMESTAMPS_ENABLED.fromMetaData(snapshot1.metadata))
+      assert(snapshot1.metadata.configuration.contains(
+        DeltaConfigs.IN_COMMIT_TIMESTAMP_ENABLEMENT_VERSION.key))
+
+      AlterTableDropFeatureDeltaCommand(
```
prakharjain09, line 3684: Same here - lets put some assertions on usage logs.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaProtocolVersionSuite.scala
dhruvarya-db: Done

spark/src/test/scala/org/apache/spark/sql/delta/DeltaProtocolVersionSuite.scala
dhruvarya-db: Done

prakharjain09 (approved):

PR #2875, opened 2024-04-10
[Delta][Python] Make delta exceptions derive from DeltaConcurrentModficationException

johnnyg:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [x] Other (Delta)

## Description

Made the delta exceptions derive from DeltaConcurrentModficationException to align with what's done in Scala.
This also makes it easy to check/catch all of the delta concurrent related exceptions.
Also removed an exception conversion as the scala equivalent class is abstract meaning it can never be instantiated.

## How was this patch tested?

Shouldn't have any functional difference so will be covered by existing tests.

Yes, users will now be able to catch all the delta exceptions by doing something like:
```python

try:
    delta_operation()
except delta.exceptions.DeltaConcurrentModificationException as e:
    ...
````

### DIFF
--- a/python/delta/exceptions.py
+++ b/python/delta/exceptions.py
@@ -40,7 +40,7 @@ class DeltaConcurrentModificationException(CapturedException):
     """
 
 
-class ConcurrentWriteException(CapturedException):
+class ConcurrentWriteException(DeltaConcurrentModificationException):
     """
     Thrown when a concurrent transaction has written data after the current transaction read the
     table.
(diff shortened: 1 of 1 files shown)


foss-contributor:
@johnnyg, thank you for your contribution! This pull request has been inactive for over 18 months and will be closed in one week. To keep it open, please leave a comment on this pull request or contact us in the `#contributors` channel on [Delta Lake Slack](https://go.delta.io/slack). We appreciate your involvement in Delta Lake!

PR #2879, opened 2024-04-10
[Spark] Add Managed Commit support in getSnapshotAt API

prakharjain09:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Add Managed Commit support in deltaLog.getSnapshotAt() API.

## How was this patch tested?

UTs

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
@@ -463,8 +463,7 @@ trait SnapshotManagement { self: DeltaLog =>
         throw new IllegalStateException(s"Could not find any delta files for version $newVersion")
       }
       if (versionToLoad.exists(_ != newVersion)) {
-        throw new IllegalStateException(
-          s"Trying to load a non-existent version ${versionToLoad.get}")
+        throwNonExistentVersionError(versionToLoad.get)
       }
       val lastCommitTimestamp = deltas.last.getModificationTime
 
@@ -556,6 +555,11 @@ trait SnapshotManagement { self: DeltaLog =>
         s"[${JsonUtils.toJson(eventData)}]")
     }
     deltasAfterCheckpoint
+  }
+
+  def throwNonExistentVersionError(versionToLoad: Long): Unit = {
+    throw new IllegalStateException(
+      s"Trying to load a non-existent version $versionToLoad")
   }
 
   /**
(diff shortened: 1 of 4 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
```diff
@@ -1143,66 +1147,64 @@ trait SnapshotManagement { self: DeltaLog =>
...
-    val current = unsafeVolatileSnapshot
-    if (current.version == version) {
-      return current
+    val currentSnapshot = unsafeVolatileSnapshot
+    val upperBoundSnapshot = if (currentSnapshot.version >= version) {
+      // current snapshot is already newer than what we are looking for. so it could be used as
+      // upper bound.
+      currentSnapshot
+    } else {
+      val latestSnapshot = update()
```
scovich, line 1173: Do I understand correctly, that this `update` is needed in order to find the right commit store, in case the requested version is not yet backfilled?
 
(probably could use a doc comment, it's a subtle point)

spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
```diff
@@ -1143,66 +1147,64 @@ trait SnapshotManagement { self: DeltaLog =>
...
-      return current
+    val (lastCheckpointInfoOpt, lastCheckpointProviderOpt) = lastCheckpointProvider match {
+      // NOTE: We must ignore any hint whose version is higher than the requested version.
+      case Some(checkpointProvider) if checkpointProvider.version <= version =>
+        // Prefer the last checkpoint provider hint, because it doesn't require any I/O to use.
+        None -> Some(checkpointProvider)
+      case _ =>
+        val lastCheckpointInfoForListing = lastCheckpointHint
+            .filter(_.version <= version)
+            .orElse(findLastCompleteCheckpointBefore(version))
```
scovich, line 1191: This does backward listing in advance, so `createLogSegment` won't have to list from 0?
(probably worth a comment, a bit subtle)

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitTestUtils.scala
```diff
@@ -28,6 +29,18 @@ import org.apache.spark.sql.test.SharedSparkSession
 trait ManagedCommitTestUtils
   extends DeltaTestUtilsBase { self: SparkFunSuite with SharedSparkSession =>
 
+  def testWithoutManagedCommits(testName: String)(f: => Unit): Unit = {
```
scovich, line 32: This is needed because `withSqlConf` doesn't support removing properties?
(maybe worth a doc comment)

scovich (approved):
Slick. Just a few nits about code comments.

spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
prakharjain09: Yes.
Adding doc comment in a followup PR.

spark/src/main/scala/org/apache/spark/sql/delta/SnapshotManagement.scala
prakharjain09: Yes - Adding doc comment in a followup PR.

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitTestUtils.scala
prakharjain09: Yes - Adding doc comment in a followup PR.

Issue #2006, opened 2023-08-29
[BUG] Non-Null Columns Implicitly Interpreted As Invariant


emrowirshi:
#### Which Delta project/connector is this regarding?

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

### Describe the problem
We have a table whose schema includes a non-nullable column (example below), and does not explicitly specify any invariants. This table is readable via Spark up until we set the reader version to 3 and the writer version to 7. At that point, reads fail because invariants are not listed in the protocol's writerFeatures, despite the fact that the table doesn't explicitly specify invariants (error code/stack below). Looking at the code, it appears that non-null constraints in the schema are being conflated with invariants, and are then failing a downstream check that if invariants are used, they must also be specified in writerFeatures.

#### Steps to reproduce

1. Publish log with non-nullable column, and without invariants:
```
{"protocol":{"minReaderVersion":1,"minWriterVersion":1}}
{"metaData":{"id":"92F69D3B-ECE2-494A-BD3F-9683AEE759D9","name":"LunchMenu","format":{"provider":"parquet","options":{}},"schemaString":"{\"type\":\"struct\",\"fields\":[{\"name\":\"Dish\",\"type\":\"string\",\"nullable\":false,\"metadata\":{}}]}","partitionColumns":[],"configuration":{}}}

```
2. Update the reader and writer versions.
`{"protocol":{"minReaderVersion":3,"minWriterVersion":7,"readerFeatures":["deletionVectors"],"writerFeatures":["deletionVectors"]}}`

3. Attempt to read the table.

#### Observed results

`DeltaTableFeatureException: Unable to operate on this table because the following table features are enabled in metadata but not listed in protocol: invariants.`

#### Expected results

The table should still be readable, as invariants are not used.

#### Further details

Loose code tracing below, just based on some initial digging in the implementation. 

This is the call which produces [InvariantsTableFeature](https://github.com/delta-io/delta/blob/9726f7f2e3922ecdbe4c3c9a0a1c11174193d14d/spark/src/main/scala/org/apache/spark/sql/delta/TableFeature.scala#L420).

And in the definition of [getFromSchema](https://github.com/delta-io/delta/blob/9726f7f2e3922ecdbe4c3c9a0a1c11174193d14d/spark/src/main/scala/org/apache/spark/sql/delta/constraints/Invariants.scala#L73), not null constraints are returned as invariants.

[...]

PR #2881, opened 2024-04-11
[Spark] Reject type changes on columns referenced by constraints/generated columns

johanl-db:
It is generally not safe to change the type of a column or field that is referenced by a CHECK constraint or a generated column. For example, some functions may produce different results depending on the input data type, e.g. `hash`.

This change adds checks to fail when the type of a column or field that is referenced by a CHECK constraint or a generated column is changed:
- using `ALTER TABLE t CHANGE COLUMN col TYPE type`.
- using schema evolution, in `ImplicitMetadataOperation.mergeSchema()`.

For the latter, a check for generated columns only was already in place in `SchemaMergingUtils.mergeSchemas`. That check is replaced in favor of the more generic check in `ImplicitMetadataOperation` which reuses existing logic already used to block column rename/drop in `ALTER TABLE`.

## How was this patch tested?
- Tests for rejecting type changes with CHECK constraints and generated columns added to `DeltaTypeWideningSuite`.
- Existing tests for rejecting type changes in `GeneratedColumnSuite` are extended.
- Tests covering the updated and newly added error classes are added to `DeltaErrorsSuite`

The type widening table feature isn't available publicly yet, this change isn't user-facing in that regard.
This change update the following error codes:
- `_LEGACY_ERROR_TEMP_DELTA_0004` -> `DELTA_CONSTRAINT_DEPENDENT_COLUMN_CHANGE`
- `_LEGACY_ERROR_TEMP_DELTA_0005` -> `DELTA_GENERATED_COLUMNS_DEPENDENT_COLUMN_CHANGE`
and introduced the following error code:
- `DELTA_CONSTRAINT_DATA_TYPE_MISMATCH`


spark/src/main/scala/org/apache/spark/sql/delta/commands/alterDeltaTableCommands.scala
```diff
@@ -775,6 +775,11 @@ case class AlterTableChangeColumnDeltaCommand(
       }
     }
 
+    if (originalField.dataType != newType) {
+      checkDependentExpressions(
+        spark, columnPath :+ columnName, txn.metadata, txn.protocol, "change")
```
olaky, line 780: Does alter not fit better here because it comes out of the alter table ... alter column command?

spark/src/main/resources/error/delta-error-classes.json
```diff
@@ -529,6 +529,21 @@
     ],
     "sqlState" : "42710"
   },
+  "DELTA_CONSTRAINT_DATA_TYPE_MISMATCH" : {
+    "message" : [
+      "Column <columnName> has data type <columnType> and cannot accept data type <dataType> because this column is referenced by the following check constraint(s):",
```
olaky, line 534: I assume the contraints can already be retrieved by the user and we are not introducing a way to read them now?

spark/src/main/scala/org/apache/spark/sql/delta/schema/ImplicitMetadataOperation.scala
```diff
@@ -206,25 +206,66 @@ object ImplicitMetadataOperation {
...
+          Constraints.findDependentConstraints(sparkSession, columnPath, newMetadata)
+        if (dependentConstraints.nonEmpty) {
+          throw DeltaErrors.constraintDataTypeMismatch(
+            columnPath,
+            currentField.dataType,
+            updateField.dataType,
+            dependentConstraints
+          )
+        }
+        // check if the column to change is referenced by any generated columns
```
olaky, line 257:
```suggestion
        // check if the field to change is referenced by any generated columns
```

spark/src/main/scala/org/apache/spark/sql/delta/schema/ImplicitMetadataOperation.scala
```diff
@@ -206,25 +206,66 @@ object ImplicitMetadataOperation {
...
+      // This condition is actually too strict, structs may be identified as changing because one
+      // of their field is changing even though that field isn't referenced by any constraint or
+      // generated column. This is intentional to keep the check simple and robust, esp. since it
+      // aligns with the historical behavior of this check.
+      if !SchemaMergingUtils.equalsIgnoreCaseAndCompatibleNullability(
+        currentField.dataType,
+        updateField.dataType
+      ) =>
+        val columnPath = fieldPath :+ currentField.name
+        // check if the column to change is referenced by check constraints
```
olaky, line 246:
```suggestion
        // check if the field to change is referenced by check constraints
```

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
```diff
@@ -980,3 +983,241 @@ trait DeltaTypeWideningTableFeatureTests extends BeforeAndAfterEach {
...
+      sql("ALTER TABLE t CHANGE COLUMN a.y TYPE SMALLINT")
+      checkAnswer(sql("SELECT * FROM t"), Row(Row(2, 3)))
+    }
+  }
+
+  test(s"check constraint with type evolution") {
+    withTable("t") {
+      sql(s"CREATE TABLE t (a byte) USING DELTA")
+      sql("ALTER TABLE t ADD CONSTRAINT ck CHECK (hash(a) > 0)")
+      sql("INSERT INTO t VALUES (2)")
```
olaky, line 1039: Let's use a number above 127 to clearly show it will not be a TINYINT

olaky (approved):
Change looks good overall, just some smaller suggestions

spark/src/main/resources/error/delta-error-classes.json
johanl-db: You can display constraints e.g. using DESCRIBE DETAIL when you have SELECT permissions on a table so users already have access to that information (also another error already displays the constraints)

spark/src/main/scala/org/apache/spark/sql/delta/commands/alterDeltaTableCommands.scala
johanl-db: The syntax is `ALTER TABLE tableName (ALTER | CHANGE) COLUMN ..`, alter and change are both equally valid here

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
johanl-db: Now using larger values after the type has been widened

spark/src/main/scala/org/apache/spark/sql/delta/schema/ImplicitMetadataOperation.scala
```diff
@@ -206,25 +206,68 @@ object ImplicitMetadataOperation {
...
+      // generated column. This is intentional to keep the check simple and robust, esp. since it
+      // aligns with the historical behavior of this check.
+      if !SchemaMergingUtils.equalsIgnoreCaseAndCompatibleNullability(
+        currentField.dataType,
+        updateField.dataType
+      ) =>
+        val columnPath = fieldPath :+ currentField.name
+        // check if the field to change is referenced by check constraints
+        val dependentConstraints =
+          Constraints.findDependentConstraints(sparkSession, columnPath, metadata)
```
tomvanbussel, line 248: This seems a bit inefficient. Won't this result in the constraints and generated columns being parsed over and over again? Probably not a big issue in practice though.

tomvanbussel (approved):
Nice! Just one comment.

spark/src/main/scala/org/apache/spark/sql/delta/schema/ImplicitMetadataOperation.scala
johanl-db: True but it allows unifying the checks we do here during schema evolution and the checks in ALTER TABLE so that both use the same helpers.

We can improve in the future, in this change I'd rather focus on correctness and reuse helpers that have been field-tested already.

felipepessoto:
If a column is non-nullable would it be rejected as well? As NOT NULL is considered an invariant? https://github.com/delta-io/delta/issues/2006.

PR #2883, opened 2024-04-11
[Spark] Make DeltaHistory.getActiveCommitAtTime compatible with managed commits

dhruvarya-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Makes DeltaHistory.getActiveCommitAtTime compatible with Managed Commits so that timestamp-based time travel queries work correctly with managed commits. Note that the only major changes are around edge case handling where commits can sometimes be missing.

## How was this patch tested?

New test suite InCommitTimestampWithManagedCommitSuite which also extends the existing ICT suite. Added a test case where we assert that missing unbackfilled files result in an exception.

No.


prakharjain09 (approved):
Lets also extend DeltaHistoryManagerSuite.scala to managed-commits.

PR #2888, opened 2024-04-12
[Spark] DV Reads Performance Improvement in Delta by removing Broadcasting DV Information

longvu-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Back then, we relied on an [expensive Broadcast of DV files](https://github.com/delta-io/delta/pull/1542) to pass the DV files to the associated Parquet Files. With the introduction of [adding custom metadata to files](https://github.com/apache/spark/pull/40677) introduced in Spark 3.5, we can now pass the DV through the custom metadata field, this is expected to improve the performance of DV reads in Delta.
## How was this patch tested?

Adjusted the existing UTs that cover our changes.
No.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaParquetFileFormat.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaParquetFileFormat.scala
@@ -16,8 +16,6 @@
 
 package org.apache.spark.sql.delta
 
-import java.net.URI
-
 import scala.collection.mutable.ArrayBuffer
 import scala.util.control.NonFatal
 
@@ -32,9 +30,9 @@ import org.apache.hadoop.mapreduce.Job
 import org.apache.parquet.hadoop.ParquetOutputFormat
 import org.apache.parquet.hadoop.util.ContextUtil
 
-import org.apache.spark.broadcast.Broadcast
 import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.catalyst.expressions.FileSourceConstantMetadataStructField
 import org.apache.spark.sql.connector.catalog.CatalogV2Implicits.parseColumnPath
 import org.apache.spark.sql.execution.datasources.OutputWriterFactory
 import org.apache.spark.sql.execution.datasources.PartitionedFile
@@ -42,7 +40,7 @@ import org.apache.spark.sql.execution.datasources.parquet.ParquetFileFormat
 import org.apache.spark.sql.execution.vectorized.{OffHeapColumnVector, OnHeapColumnVector, WritableColumnVector}
 import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.sources._
-import org.apache.spark.sql.types.{ByteType, LongType, MetadataBuilder, StructField, StructType}
+import org.apache.spark.sql.types.{ByteType, LongType, MetadataBuilder, StringType, StructField, StructType}
 import org.apache.spark.sql.vectorized.{ColumnarBatch, ColumnarBatchRow, ColumnVector}
 import org.apache.spark.util.SerializableConfiguration
 
@@ -59,13 +57,11 @@ case class DeltaParquetFileFormat(
     nullableRowTrackingFields: Boolean = false,
     isSplittable: Boolean = true,
     disablePushDowns: Boolean = false,
-    tablePath: Option[String] = None,
-    broadcastDvMap: Option[Broadcast[Map[URI, DeletionVectorDescriptorWithFilterType]]] = None,
-    broadcastHadoopConf: Option[Broadcast[SerializableConfiguration]] = None)
+    tablePath: Option[String] = None)
   extends ParquetFileFormat {
   // Validate either we have all arguments for DV enabled read or none of them.
-  if (hasDeletionVectorMap) {
-    require(tablePath.isDefined && !isSplittable && disablePushDowns,
+  if (hasTablePath) {
+    require(!isSplittable && disablePushDowns,
       "Wrong arguments for Delta table scan with deletion vectors")
   }
 
@@ -124,7 +120,7 @@ case class DeltaParquetFileFormat(
   override def isSplitable(
     sparkSession: SparkSession, options: Map[String, String], path: Path): Boolean = isSplittable
 
-  def hasDeletionVectorMap: Boolean = broadcastDvMap.isDefined && broadcastHadoopConf.isDefined
+  def hasTablePath: Boolean = tablePath.isDefined
 
   /**
    * We sometimes need to replace FileFormat within LogicalPlans, so we have to override
@@ -182,10 +178,12 @@ case class DeltaParquetFileFormat(
       require(disablePushDowns, "Cannot generate row index related metadata with filter pushdown")
     }
 
-    if (hasDeletionVectorMap && isRowDeletedColumn.isEmpty) {
+    if (hasTablePath && isRowDeletedColumn.isEmpty) {
         throw new IllegalArgumentException(
           s"Expected a column $IS_ROW_DELETED_COLUMN_NAME in the schema")
     }
+
+    val serializableHadoopConf = new SerializableConfiguration(hadoopConf)
 
     val useOffHeapBuffers = sparkSession.sessionState.conf.offHeapColumnVectorEnabled
     (partitionedFile: PartitionedFile) => {
@@ -196,8 +194,9 @@ case class DeltaParquetFileFormat(
             partitionedFile,
             rowIteratorFromParquet,
             isRowDeletedColumn,
-            useOffHeapBuffers = useOffHeapBuffers,
-            rowIndexColumn = rowIndexColumn)
+            rowIndexColumn,
+            useOffHeapBuffers,
+            serializableHadoopConf)
         iterToReturn.asInstanceOf[Iterator[InternalRow]]
       } catch {
         case NonFatal(e) =>
@@ -276,16 +275,11 @@ case class DeltaParquetFileFormat(
       .updated(DefaultRowCommitVersion.METADATA_STRUCT_FIELD_NAME, extractDefaultRowCommitVersion)
   }
 
-  def copyWithDVInfo(
-      tablePath: String,
-      broadcastDvMap: Broadcast[Map[URI, DeletionVectorDescriptorWithFilterType]],
-      broadcastHadoopConf: Broadcast[SerializableConfiguration]): DeltaParquetFileFormat = {
+  def disableSplittingAndPushdown(tablePath: String): DeltaParquetFileFormat = {
     this.copy(
       isSplittable = false,
       disablePushDowns = true,
-      tablePath = Some(tablePath),
-      broadcastDvMap = Some(broadcastDvMap),
-      broadcastHadoopConf = Some(broadcastHadoopConf))
+      tablePath = Some(tablePath))
   }
 
   /**
@@ -300,28 +294,32 @@ case class DeltaParquetFileFormat(
       iterator: Iterator[Object],
       isRowDeletedColumn: Option[ColumnMetadata],
       rowIndexColumn: Option[ColumnMetadata],
-      useOffHeapBuffers: Boolean): Iterator[Object] = {
-    val pathUri = partitionedFile.pathUri
-
+      useOffHeapBuffers: Boolean,
+      serializableHadoopConf: SerializableConfiguration): Iterator[Object] = {
     val rowIndexFilter = isRowDeletedColumn.map { col =>
       // Fetch the DV descriptor from the broadcast map and create a row index filter
-      broadcastDvMap.get.value
-        .get(pathUri)
-        .map { case DeletionVectorDescriptorWithFilterType(dvDescriptor, filterType) =>
-          filterType match {
-            case i if i == RowIndexFilterType.IF_CONTAINED =>
-              DropMarkedRowsFilter.createInstance(
-                dvDescriptor,
-                broadcastHadoopConf.get.value.value,
-                tablePath.map(new Path(_)))
-            case i if i == RowIndexFilterType.IF_NOT_CONTAINED =>
-              KeepMarkedRowsFilter.createInstance(
-                dvDescriptor,
-                broadcastHadoopConf.get.value.value,
-                tablePath.map(new Path(_)))
-          }
+      val dvDescriptorOpt = partitionedFile.otherConstantMetadataColumnValues
+        .get(FILE_ROW_INDEX_FILTER_ID_ENCODED)
+      val filterTypeOpt = partitionedFile.otherConstantMetadataColumnValues
+        .get(FILE_ROW_INDEX_FILTER_TYPE)
+      if (dvDescriptorOpt.isDefined && filterTypeOpt.isDefined) {
+        val rowIndexFilter = filterTypeOpt.get match {
+          case RowIndexFilterType.IF_CONTAINED => DropMarkedRowsFilter
+          case RowIndexFilterType.IF_NOT_CONTAINED => KeepMarkedRowsFilter
+          case unexpectedFilterType => throw new IllegalStateException(
+            s"Unexpected row index filter type: ${unexpectedFilterType}")
         }
-        .getOrElse(KeepAllRowsFilter)
+        rowIndexFilter.createInstance(
+          DeletionVectorDescriptor.fromJson(dvDescriptorOpt.get.asInstanceOf[String]),
+          serializableHadoopConf.value,
+          tablePath.map(new Path(_)))
+      } else if (dvDescriptorOpt.isDefined || filterTypeOpt.isDefined) {
+        throw new IllegalStateException(
+          s"Both ${FILE_ROW_INDEX_FILTER_ID_ENCODED} and ${FILE_ROW_INDEX_FILTER_TYPE} " +
+            "should either both have values or no values at all.")
+      } else {
+        KeepAllRowsFilter
+      }
     }
 
     val metadataColumns = Seq(isRowDeletedColumn, rowIndexColumn).filter(_.nonEmpty).map(_.get)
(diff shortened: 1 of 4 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/files/TahoeFileIndex.scala
```diff
@@ -126,6 +127,19 @@ abstract class TahoeFileIndex(
...
+    // than filtered out. In such a case, the `filterTypes` map will be populated by [[CDCReader]]
+    // to indicate IF_NOT_CONTAINED filter should be used. In other cases, `filterTypes` will be
+    // empty, so we generate IF_CONTAINED as the default DV behavior
+    if (addFile.deletionVector != null) {
+      metadata.put(DeltaParquetFileFormat.FILE_ROW_INDEX_FILTER_ID_ENCODED,
+        JsonUtils.toJson(addFile.deletionVector))
+    }
+    val filterType = rowIndexFilters.getOrElse(Map.empty)
+      .getOrElse(addFile.path, RowIndexFilterType.IF_CONTAINED)
+    metadata.put(DeltaParquetFileFormat.FILE_ROW_INDEX_FILTER_TYPE, filterType)
```
xupefei, line 142: This will add one metadata entry regardless the existence of DVs on thi file, right?
How about not adding it but later we do `partitionedFile.otherConstantMetadataColumnValues.contains(FILE_ROW_INDEX_FILTER_TYPE)`?

longvu-db: @xupefei True that

spark/src/test/scala/org/apache/spark/sql/delta/DeltaParquetFileFormatSuite.scala
```diff
@@ -33,26 +29,31 @@ import org.apache.parquet.hadoop.ParquetFileReader
...
 class DeltaParquetFileFormatSuite extends QueryTest
-  with SharedSparkSession with DeltaSQLCommandTest {
+  with SharedSparkSession
+  with DeletionVectorsTestUtils
+  with DeltaSQLCommandTest {
   import testImplicits._
 
+  override def beforeAll(): Unit = {
+    super.beforeAll()
+    enableDeletionVectorsInNewTables(spark.conf)
```
xupefei, line 42: Why enable DV for all tests?

spark/src/main/scala/org/apache/spark/sql/delta/files/TahoeFileIndex.scala
longvu-db: Although the map.get and isDefined also do the `contains` trick

spark/src/test/scala/org/apache/spark/sql/delta/DeltaParquetFileFormatSuite.scala
```diff
@@ -112,11 +109,7 @@ class DeltaParquetFileFormatSuite extends QueryTest
           // Select some rows that are deleted and some rows not deleted
           // Deleted row `value`: 0, 200, 300, 756, 10352, 19999
           // Not deleted row `value`: 7, 900
-          val (deletedColumnValue, notDeletedColumnValue) = rowIndexFilterType match {
-            case RowIndexFilterType.IF_CONTAINED => (1, 0)
-            case RowIndexFilterType.IF_NOT_CONTAINED => (0, 1)
-            case _ => (-1, -1) // Invalid, expecting the test to fail.
-          }
+          val (deletedColumnValue, notDeletedColumnValue) = (1, 0)
```
xupefei, line 112: You reduced the coverage because it's partially covered by CDCReader tests, isn't it?
Anyway have more tests are always better. I would prefer to keep it.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaParquetFileFormatSuite.scala
longvu-db: @xupefei What do you mean sorry? Could you further elaborate?

spark/src/test/scala/org/apache/spark/sql/delta/DeltaParquetFileFormatSuite.scala
longvu-db: @xupefei Hmm, but right now I don't have control over the rowIndexFilterType though.

Which case does the CDCReader not cover?

spark/src/main/scala/org/apache/spark/sql/delta/DeltaParquetFileFormat.scala
```diff
@@ -281,27 +295,28 @@ case class DeltaParquetFileFormat(
...
-          }
+      val dvDescriptorOpt = partitionedFile.otherConstantMetadataColumnValues
+        .get(FILE_ROW_INDEX_FILTER_ID_ENCODED)
+      val filterTypeOpt = partitionedFile.otherConstantMetadataColumnValues
+        .get(FILE_ROW_INDEX_FILTER_TYPE)
+      if (dvDescriptorOpt.isDefined && filterTypeOpt.isDefined) {
+        val rowIndexFilter = if (filterTypeOpt.get == RowIndexFilterType.IF_CONTAINED) {
+          DropMarkedRowsFilter
+        } else if (filterTypeOpt.get == RowIndexFilterType.IF_NOT_CONTAINED) {
+          KeepMarkedRowsFilter
```
xupefei, line 309: Case match is better, because we won't call Option.get multiple times.

xupefei (commented):
Very nice, thank you! Looking forward to seeing how much speed-up we can get with this change :)

spark/src/main/scala/org/apache/spark/sql/delta/DeltaParquetFileFormat.scala
```diff
@@ -283,27 +297,31 @@ case class DeltaParquetFileFormat(
...
         }
-        .getOrElse(KeepAllRowsFilter)
+        rowIndexFilter.createInstance(
+          DeletionVectorDescriptor.fromJson(dvDescriptorOpt.get.asInstanceOf[String]),
+          broadcastHadoopConf.get.value.value,
+          tablePath.map(new Path(_)))
+      } else if (dvDescriptorOpt.isDefined || filterTypeOpt.isDefined) {
+        throw new IllegalStateException(
+          s"Both ${FILE_ROW_INDEX_FILTER_ID_ENCODED} and ${FILE_ROW_INDEX_FILTER_TYPE} " +
+            "should both have a value or not have a value together")
```
longvu-db, line 321: Add a period here @longvu-db

spark/src/test/scala/org/apache/spark/sql/delta/DeltaParquetFileFormatSuite.scala
longvu-db: @xupefei I added some tests with DV turned off, would you be so kind to take another look at the PR :)?

spark/src/main/scala/org/apache/spark/sql/delta/DeltaParquetFileFormat.scala
```diff
@@ -205,16 +203,27 @@ case class DeltaParquetFileFormat(
...
+      } else {
+        // It is fine to expose the row_index field as a metadata field when Row Tracking
+        // is enabled because it is needed to generate the Row ID field, and it is not a
+        // big problem if we use 2b+ rows in a single rowgroup, it will throw an exception and
+        // we can then use less rows per rowgroup. Also, 2b+ rows in a single rowgroup is
+        // not a common use case.
+        super.metadataSchemaFields ++ rowTrackingFields
+      }
+
+    val isDVsEnabled = DeltaConfigs.ENABLE_DELETION_VECTORS_CREATION.fromMetaData(metadata)
```
xupefei, line 218: `metadata` shows the current status of the table, i.e., if DV is enabled in the current version, isn't it? How about some old versions that have DV, but now the user has decided to disable it? I think we should look at the table protocol: if DV table feature is turned on, then we assume the table contains DV.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaParquetFileFormatSuite.scala
xupefei: Ahha I see. We no longer broadcast DVs so there's no IF_CONTAINED or IF_NOT_CONTAINED at this level.

xupefei (commented):
After looking at the code, I feel that we need a test case with the following logic:

1. Create a table without DV. Insert some values.
2. Turn on DV (`delta.enableDeletionVectors` = `true`).
3. Delete some values.
4. Turn off DV (`delta.enableDeletionVectors` = `false`).
5. Delete some values.

Check:
1. Version 1 does not contain DV metadata column.
2. Version 2 does not contain DV metadata column (because there's no DV in this table - correct me if I'm wrong).
3. Version 3 does contain DV metadata columns.
3. Version 4 does contain DV metadata columns.
3. Version 5 does contain DV metadata columns.

spark/src/main/scala/org/apache/spark/sql/delta/DeltaParquetFileFormat.scala
longvu-db: @xupefei Quick question, can an existing table add new table feature? Like if a table does not have DV from the beginning, in the property, and then at some point later in time get added DVs.

longvu-db: @xupefei I think you can do this using ALTER TABLE right? But once it is in, you cannot drop it at the moment.

longvu-db:
@xupefei I addressed your comments =)).

Regarding the new test case:
1. Version 2 **should contain** DV metadata column (because having the metadata columns depends on whether or not the table's protocol supports DV, not on whether any DVs is written or not)
2. What is the value of doing action 5? If we disable the DV for the table metadata, and if the DV metadata indeed stay, then whatever actions we do, it would always stay right?

spark/src/test/scala/org/apache/spark/sql/delta/DeltaParquetFileFormatSuite.scala
```diff
@@ -16,140 +16,214 @@
...
+
+      checkDVMetadataFields(tablePath, checkMetadataFields = true)
+    }
+  }
+
+  private val FILE_ROW_INDEX_FILTER_ID_ENCODED = "_metadata.row_index_filter_id_encoded"
+
+  private val FILE_ROW_INDEX_FILTER_TYPE = "_metadata.row_index_filter_type"
+
+  private def checkDVMetadataFields(tablePath: String, checkMetadataFields: Boolean): Unit = {
```
xupefei, line 200:
```suggestion
  private def checkDVMetadataFields(tablePath: String, metadataContainsDVFields: Boolean): Unit = {
```

spark/src/main/scala/org/apache/spark/sql/delta/DeltaParquetFileFormat.scala
```diff
@@ -104,7 +102,7 @@ case class DeltaParquetFileFormat(
   override def isSplitable(
     sparkSession: SparkSession, options: Map[String, String], path: Path): Boolean = isSplittable
 
-  def hasDeletionVectorMap: Boolean = broadcastDvMap.isDefined && broadcastHadoopConf.isDefined
+  def hasBroadcastHadoopConf: Boolean = broadcastHadoopConf.isDefined
```
johanl-db, line 105: I think you can get rid of `broadcastHadoopConf` and `hasBroadcastHadoopConf` altogether here.

The only use we have for it now is to pass it to `rowIndexFilter.createInstance` when creating the DV filter but we also have `hadoopConf: Configuration` around (passed as an argument to `buildReaderWithPartitionValues`) which I believe should be sufficient here: we don't need this configuration to be broadcast first for the purpose of loading DVs from storage

spark/src/test/scala/org/apache/spark/sql/delta/DeltaParquetFileFormatSuite.scala
```diff
@@ -16,140 +16,214 @@
...
+  testWithBothParquetReaders("Existence of Deletion Vectors metadata columns when we " +
+    "add and remove DV from the table's properties") {
+    withTempDir { tempDir =>
+      val tablePath = tempDir.toString
+      val deltaLog = DeltaLog.forTable(spark, tempDir)
+
+      // Version 1: Create a table without DV and insert some values.
+      generateData(tablePath)
+
+      checkDVMetadataFields(tablePath, checkMetadataFields = false)
```
xupefei, line 170: Can we also check the table contents using `checkResult` each time after we check the metadata?

xupefei (commented):
A few improvements to the tests then we're good to go

xupefei:
Cc @larsk-db

spark/src/main/scala/org/apache/spark/sql/delta/PreprocessTableWithDVs.scala
```diff
@@ -93,20 +91,22 @@ object ScanWithDeletionVectors {
 
     // See if the relation is already modified to include DV reads as part of
     // a previous invocation of this rule on this table
-    if (fileFormat.hasDeletionVectorMap) return None
+    if (fileFormat.hasBroadcastHadoopConf) return None
```
johanl-db, line 94: We should update that check since using a broadcast conf isn't a marker of the preprocess step for DVs (we actually remove this need in this change..)

We could instead check if the scan already contains the `IS_ROW_DELETED_STRUCT_FIELD` field, e.g.:


```
    if (scan.output.exists(_.name == IS_ROW_DELETED_COLUMN_NAME)) return None
```

spark/src/main/scala/org/apache/spark/sql/delta/PreprocessTableWithDVs.scala
```diff
@@ -145,7 +144,7 @@ object ScanWithDeletionVectors {
       new SerializableConfiguration(tahoeFileIndex.deltaLog.newDeltaHadoopConf()))
 
     val newFileFormat = fileFormat.copyWithDVInfo(
-      tahoeFileIndex.path.toString, filePathToDVBroadcastMap, hadoopConfBroadcast)
+      tahoeFileIndex.path.toString, hadoopConfBroadcast)
```
johanl-db, line 147: I would remove `copyWithDVInfo` at this point and directly do here:
```
val newFileFormat = fileFormat.copy(isSplittable = false, disablePushDowns = true)
```


Or rename `copyWithDVInfo` to e.g. `disableSplittingAndPushdown`

spark/src/main/scala/org/apache/spark/sql/delta/files/TahoeFileIndex.scala
```diff
@@ -126,6 +127,19 @@ abstract class TahoeFileIndex(
     addFile.defaultRowCommitVersion.foreach(defaultRowCommitVersion =>
       metadata.put(DefaultRowCommitVersion.METADATA_STRUCT_FIELD_NAME, defaultRowCommitVersion))
 
+    // Attach filter types to FileActions, so that later [[DeltaParquetFileFormat]] could pick it up
+    // to decide which kind of rows should be filtered out. This info is necessary for reading CDC
+    // rows that have been deleted (marked in DV), in which case marked rows must be kept rather
+    // than filtered out. In such a case, the `filterTypes` map will be populated by [[CDCReader]]
+    // to indicate IF_NOT_CONTAINED filter should be used. In other cases, `filterTypes` will be
+    // empty, so we generate IF_CONTAINED as the default DV behavior
```
johanl-db, line 135: Move this comment just before `val filterType`

Also it could be a bit shorter, e.g.:
> Set the filter type to `IF_CONTAINED` by default to filter out rows unless a filter type was explicitly provided in `rowIndexFilters`. This can happen e.g. when reading CDC data to keep deleted rows instead of filtering them out.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaParquetFileFormatSuite.scala
```diff
@@ -16,140 +16,214 @@
...
+              rowIndexes(
+                df.filter("value in (0, 7, 200, 300, 756, 900, 10352, 19999)")) ===
+                Seq(0, 7, 200, 300, 756, 900, 10352, 19999).toSet)
+          }
         }
       }
     }
   }
 
+  testWithBothParquetReaders("Existence of Deletion Vectors metadata columns when we " +
```
johanl-db, line 161: This test probably becomes moot if you don't pass the DV information via the metadata struct anymore 9see my previous comment)

larsk-db:
> Cc @larsk-db

I took a quick look, but it seems you and @johanl-db are on top of this. Ping me if you have something specific that needs my input.

spark/src/main/scala/org/apache/spark/sql/delta/DeltaParquetFileFormat.scala
longvu-db: @johanl-db No clue why we had it back in the days hmm

longvu-db: @johanl-db Removing `broadcastHadoopConf` doesn't seem to work, I found this [error message](https://runbot-ci.cloud.databricks.com/build/Delta-OSS-Scala212-Pr/run/17985292)but I haven't investigated yet, changing back to using the `broadcastHadoopConf` makes the tests passed

spark/src/main/scala/org/apache/spark/sql/delta/DeltaParquetFileFormat.scala
```diff
@@ -417,6 +416,12 @@ object DeltaParquetFileFormat {
   val ROW_INDEX_COLUMN_NAME = "__delta_internal_row_index"
   val ROW_INDEX_STRUCT_FIELD = StructField(ROW_INDEX_COLUMN_NAME, LongType)
 
+  /** The name of the metadata column that contains the encoded row index filter identifier. */
```
johanl-db, line 419: Update this comment and the one below, this isn't the name of a metadata column anymore, just the key for the row index filter id / type in the file's `otherConstantMetadataColumnValues` map.

spark/src/main/scala/org/apache/spark/sql/delta/files/TahoeFileIndex.scala
johanl-db: Not addressed ^

longvu-db: @johanl-db I actually addressed, but I was not sure if your version removes too much original info, @xupefei, is it fine if we keep it like

> Set the filter type to IF_CONTAINED by default to filter out rows unless a filter type was explicitly provided in rowIndexFilters. This can happen e.g. when reading CDC data to keep deleted rows instead of filtering them out.

spark/src/main/scala/org/apache/spark/sql/delta/DeltaParquetFileFormat.scala
longvu-db: @johanl-db My bad yeah

spark/src/main/scala/org/apache/spark/sql/delta/files/TahoeFileIndex.scala
xupefei: I would suggest
> Set the filter type to IF_CONTAINED by default to let [[DeltaParquetFileFormat]] filter out rows unless a filter type was explicitly provided in rowIndexFilters. This can happen e.g. when reading CDC data to keep deleted rows instead of filtering them out.

johanl-db (approved):

tdas (approved):
I highly approve this change :D
Thank you for doing this! This will improve the stability on large tables with lots of DVs significantly.

PR #2891, opened 2024-04-15
[SPARK][VARIANT] Add minimal support for variant type in delta-spark

richardc-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Adds the variant table feature to minimally implement the variant type as described in the RFC in #2867.

Also disables using variant columns as partition columns.

## How was this patch tested?

Added some UTs. More UTs will be merged in followup PRs

no


spark/src/main/scala-spark-3.5/shims/VariantShim.scala
scottsand-db, line 1: `VariantShims` (plural)

spark/src/main/scala/org/apache/spark/sql/delta/schema/SchemaUtils.scala
```diff
@@ -1263,6 +1263,13 @@ def normalizeColumnNamesInDataType(
     unsupportedDataTypes.toSeq
   }
 
+  /**
+   * Find VariantType columns in the table schema.
```
scottsand-db, line 1267: `Find` in the description and `Boolean` as the return type seems off to me.

`Returns 'true' if any XXX exists in the table schema`

spark/src/main/scala/org/apache/spark/sql/delta/util/PartitionUtils.scala
```diff
@@ -605,7 +605,7 @@ private[delta] object PartitionUtils {
 
     partitionColumnsSchema(schema, partitionColumns, caseSensitive).foreach {
       field => field.dataType match {
-        case _: AtomicType => // OK
+        case a: AtomicType if !VariantShim.isTypeVariant(a) => // OK
```
scottsand-db, line 608: Please add a comment/context. Why can't Variant's be valid partition columsn?

spark/src/test/scala-spark-master/shims/DeltaVariantSuite.scala
scottsand-db, line 1: This isn't a 'shim'. It's not used to fix compilation differences between Spark 3.5 and Master.

I believe the goal here is to have this file be Spark-Master-Only.

Thus, please put it into `spark/src/test/scala-spark-master/org/apache/spark/sql/delta/DeltaVariantSuite.scala`

scottsand-db (changes requested):
Looking good! Left some comments! Thanks!

PR #2892, opened 2024-04-15
[Spark] Enhance CommitStore Equality with Semantic Comparison

sumeet-db:
- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR revises the CommitStore equality check to use semantic comparison rather than direct object reference comparison.

## How was this patch tested?

UTs

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/managedcommit/CommitStore.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/managedcommit/CommitStore.scala
@@ -106,6 +106,28 @@ trait CommitStore {
       logPath: Path,
       startVersion: Long,
       endVersion: Option[Long]): Unit
+
+  /**
+   * Determines whether this CommitStore is semantically equal to another CommitStore.
+   *
+   * Semantic equality is determined by each CommitStore implementation based on whether the two
+   * instances can be used interchangeably when invoking any of the CommitStore APIs, such as
+   * `commit`, `getCommits`, etc. For e.g., both the instances might be pointing to the same
+   * underlying endpoint.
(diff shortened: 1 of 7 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/managedcommit/CommitStore.scala
```diff
@@ -106,6 +106,28 @@ trait CommitStore {
...
+   * Semantic equality is determined by each CommitStore implementation based on whether the two
+   * instances can be used interchangeably when invoking any of the CommitStore APIs, such as
+   * `commit`, `getCommits`, etc. For e.g., both the instances might be pointing to the same
+   * underlying endpoint.
+   */
+  def semanticEquals(other: CommitStore): Boolean
+}
+
+object CommitStore {
+  def semanticEquals(
```
LukasRupprecht, line 122: Can you clarify why we need this static method? Is it just a helper to simplify dealing with the options? Do we expect this to be used frequently?

LukasRupprecht (approved):
Thanks for the PR! Looks good, just have one clarifying comment.

spark/src/main/scala/org/apache/spark/sql/delta/managedcommit/CommitStore.scala
sumeet-db: It's a helper method used in OptimisticTransaction. Inlining the logic there was making it harder to read the code.

prakharjain09 (commented):
LGTM!

PR #2901, opened 2024-04-17
[Spark] Managed Commit support for DeltaLog's getChanges/getChangeLogFiles API

prakharjain09:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR makes changes to `DeltaLog.getChanges/getChangeLogFiles` API in order to support managed-commits.

## How was this patch tested?

UTs

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala
@@ -31,6 +31,7 @@ import com.databricks.spark.util.TagDefinitions._
 import org.apache.spark.sql.delta.actions._
 import org.apache.spark.sql.delta.commands.WriteIntoDelta
 import org.apache.spark.sql.delta.files.{TahoeBatchFileIndex, TahoeLogFileIndex}
+import org.apache.spark.sql.delta.managedcommit.ManagedCommitUtils
 import org.apache.spark.sql.delta.metering.DeltaLogging
 import org.apache.spark.sql.delta.schema.{SchemaMergingUtils, SchemaUtils}
 import org.apache.spark.sql.delta.sources._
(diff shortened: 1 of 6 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala
```diff
@@ -274,35 +275,64 @@ class DeltaLog private(
   /**
    * Get all actions starting from "startVersion" (inclusive). If `startVersion` doesn't exist,
    * return an empty Iterator.
+   * Callers are encouraged to use the other override which takes the endVersion if available to
+   * avoid I/O and improve performance of this method.
```
scovich, line 279: Do we have (m)any internal Delta call sites of these "deprecated" method signatures? 
Or have they all been updated and we're only worried about external callers?

spark/src/main/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitUtils.scala
```diff
@@ -0,0 +1,88 @@
...
+   */
+  def commitFilesIterator(
+      deltaLog: DeltaLog,
+      startVersion: Long): Iterator[(FileStatus, Long)] = {
+
+    def listDeltas(startVersion: Long, endVersion: Option[Long]): Iterator[(FileStatus, Long)] = {
+      deltaLog
+        .listFrom(startVersion)
+        .collect { case DeltaFile(fileStatus, version) => (fileStatus, version) }
+        .takeWhile { case (_, version) => endVersion.forall(version <= _) }
```
scovich, line 41: Should we adopt `stopAt` here as well? I guess it's pretty unlikely to hit a listing page boundary right at the requested `endVersion`, so probably not worth the trouble?

spark/src/main/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitUtils.scala
```diff
@@ -0,0 +1,88 @@
...
+    var maxVersionSeen = startVersion - 1
+    val listedDeltas = listDeltas(startVersion, endVersion = None).filter { case (_, version) =>
+      maxVersionSeen = math.max(maxVersionSeen, version)
+      true
+    }
+
+    def tailFromSnapshot(): Iterator[(FileStatus, Long)] = {
+      val currentSnapshotInDeltaLog = deltaLog.unsafeVolatileSnapshot
+      if (currentSnapshotInDeltaLog.version == maxVersionSeen &&
+          currentSnapshotInDeltaLog.tableCommitStoreOpt.isEmpty) {
```
scovich, line 53: To make sure I understand: If the volatile snapshot version is _smaller_ than what the listing found, then we don't know whether the table might have managed commits or not, because the feature could have been enabled by one of the later commits we don't yet have a snapshot for?

What if the unsafe snapshot version is _larger_ than what the listing returned tho? In that case it's probably ok to trust the listing result as long as the snapshot doesn't have MC enabled... but I guess it's also a _really_ small window where that could happen, so we probably don't care to optimize for it?

scovich (approved):
Nice. Some clever iterator optimizations there.

spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala
prakharjain09: We do have more call-sites where end is not available. e.g. `DeltaSourceCDCSupport` uses this API.

spark/src/main/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitUtils.scala
prakharjain09: yes correct.

spark/src/main/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitUtils.scala
prakharjain09: > If the volatile snapshot version is smaller than what the listing found, then we don't know whether the table might have managed commits or not, because the feature could have been enabled by one of the later commits we don't yet have a snapshot for?

correct.

> What if the unsafe snapshot version is larger than what the listing returned tho? In that case it's probably ok to trust the listing result as long as the snapshot doesn't have MC enabled... but I guess it's also a really small window where that could happen, so we probably don't care to optimize for it?

If unsafeVolatileSnapshot is larger than what listing returned, then:
1. [common case] Either MC is enabled and commits b/w listing-end and unsafeVolatileSnapshot are not backfilled yet?
2. Or MC just got disabled and backfill has not finished yet. uncommon case - we don't want to optimize this.

scovich: Yes, that's about what I was thinking as well. Not safe to optimize in general, and even the somewhat safe case is uncommon so meh.

prakharjain09: There is this another case:
LIST returns: 1, 2, 3, 4, 5, 6, volatile snapshot is v10 with no commit store
=> the snapshot may have updated after our listing completed (very narrow time window, but maybe the thread got pre-empted by OS or something)
=> by linearizability arg, we can safely claim the latest version as-of our operation is v6, and ignore the snapshot

spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala
scovich: And those are cases where we don't want to filter by end, right? Things like "get everything available" ?

PR #2896, opened 2024-04-16
[Spark][TEST-ONLY] Merge source materialization non-determinism determination in subqueries tests

juliuszsompolski:
#### Which Delta project/connector is this regarding?

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR adds extra tests to consider determinism in subqueries for source merge materialization.

## How was this patch tested?

This patch is test only and adds new tests.

## Does this PR introduce _any_ user-facing changes?

No.

PR #2906, opened 2024-04-17
[Spark] Merge should materialize source with correlated subqueries

juliuszsompolski:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Small fixes for determinism determination for Merge Materialize Source.
* since we consider joins as potentially non-deterministic, we should also consider correlated subqueries as such, because they essentially are also joins.
* non Delta source will not be found in subqueries in `findFirstNonDeltaScan`, but found in `findFirstNonDeterministicNode`. This will make merge source materialization give wrong reason for materialization.

## How was this patch tested?

Added tests with subqueries and nested subqueries in Project and Filters and correlated subqueries.

No.


juliuszsompolski:
@larsk-db @cstavr @tomvanbussel this is a redo of https://github.com/delta-io/delta/pull/2896 after investigating the test failures the new tests caused.

larsk-db (approved):
LGTM

PR #2911, opened 2024-04-18
[Spark] Make update catalog schema truncation threshold configurable

dhruvarya-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Currently, during schema sync to catalog, the whole schema gets truncated if any of the fields is longer than 4000 characters. This PR makes this threshold a configurable through the config `DeltaSQLConf. DELTA_UPDATE_CATALOG_LONG_FIELD_TRUNCATION_THRESHOLD`.

## How was this patch tested?

Created variations of existing test cases that validate that setting the config to a bigger value skips the truncation.

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/hooks/UpdateCatalog.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/hooks/UpdateCatalog.scala
@@ -198,11 +198,14 @@ case class UpdateCatalog(table: CatalogTable) extends UpdateCatalogBase {
 
 
   override protected def schemaHasChanged(snapshot: Snapshot, spark: SparkSession): Boolean = {
-    // We need to check whether the schema in the catalog matches the current schema. If a
-    // field in the schema is very long, we cannot store the schema in the catalog, therefore
-    // here we have to compare what's in the catalog with what we actually can store in the
-    // catalog
-    val schemaChanged = UpdateCatalog.truncateSchemaIfNecessary(snapshot.schema) != table.schema
+    // We need to check whether the schema in the catalog matches the current schema.
+    // Depending on the schema validation policy, the schema might need to be truncated.
+    // Therefore, we should use what we want to store in the catalog for comparison.
+    val truncationThreshold = spark.sessionState.conf.getConf(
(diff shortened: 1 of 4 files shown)


tdas (approved):

spark/src/main/scala/org/apache/spark/sql/delta/hooks/UpdateCatalog.scala
```diff
@@ -261,11 +264,11 @@ object UpdateCatalog {
   // This is the encoding of the database for the Hive MetaStore
   private val latin1 = Charset.forName("ISO-8859-1")
 
-  // Maximum number of characters that a catalog can store.
-  val MAX_CATALOG_TYPE_DDL_LENGTH = 4000
   val ERROR_KEY = "delta.catalogUpdateError"
   val LONG_SCHEMA_ERROR: String = "The schema contains a very long nested field and cannot be " +
```
tdas, line 268: When this error is thrown, make sure that it says which configuration to update so that users can self-service this.

dhruvarya-db: @tdas We don't throw an error when this truncation happens, we just log the error message in an HMS table property. This is because this sync to the catalog is a best-effort async operation. I could try to create a user-visible log here.

dhruvarya-db: @tdas There are a few ways in which we can surface this. I will address this specific issue in a follow up PR.

tdas (approved):

tdas:
delta spark master test failures are unrelated

felipepessoto:
@dhruvarya-db, @tdas @vkorukanti some of these UpdateCatalogSuite tests are flaky, it seems we had similar issue in the past: #2476, I see it happening more as I increase test parallelism.

Have you found any clue of the root cause?

I tried adding this:

```
    if (!UpdateCatalog.awaitCompletion(10000)) {
      logWarning(s"There are active catalog udpate requests after 10000ms")
      // scalastyle:off println
      println(s"There are active catalog udpate requests after 10000ms")
      // scalastyle:on println
    }
```

to `verifySchemaInCatalog`, but it didn't help, and no message is print.

PR #2924, opened 2024-04-19
[Spark] Apply Column Mapping to filters pushed to Parquet scan

tomvanbussel:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR fixes a correctness bug when Column Mapping is enabled. The issue occurs when a column is renamed to the physical name of another column. If a query filters on this column then the Parquet scan incorrectly applies this filter to the other column. This PR fixes this problem by translating the column names in the filters from logical to physical in `DeltaParquetFileFormat` before passing them on to `ParquetFileFormat`. 

## How was this patch tested?

Added a unit test to `DeltaColumnMappingSuite`.

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaParquetFileFormat.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaParquetFileFormat.scala
@@ -102,6 +102,14 @@ case class DeltaParquetFileFormat(
     } else schema
   }
 
+  /**
+   * Prepares filters so that they can be pushed down into the Parquet reader.
+   *
+   * If column mapping is enabled, then logical column names in the filters will be replaced with
+   * their corresponding physical column names. This is necessary as the Parquet files will use
+   * physical column names, and the requested schema pushed down in the Parquet reader will also use
+   * physical column names.
+   */
   private def prepareFiltersForRead(filters: Seq[Filter]): Seq[Filter] = {
     if (disablePushDowns) {
       Seq.empty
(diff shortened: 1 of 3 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/DeltaParquetFileFormat.scala
```diff
@@ -469,6 +479,66 @@ object DeltaParquetFileFormat {
...
+   * Translates the filter to use physical column names instead of logical column names.
+   * This is needed when the column mapping mode is set to `NameMapping` or `IdMapping`
+   * to match the requested schema that's passed to the [[ParquetFileFormat]].
+   */
+  private def translateFilterForColumnMapping(
+       filter: Filter,
+       physicalNameMap: Map[Seq[String], Seq[String]]): Option[Filter] = {
+    def getPhysicalName(attribute: String): String = {
+      import org.apache.spark.sql.connector.catalog.CatalogV2Implicits.MultipartIdentifierHelper
+      physicalNameMap(parseColumnPath(attribute)).quoted
```
johanl-db, line 494: For an extra layer of safety, you can use `Map.get` and drop the filter if there's no entry for that attribute in the map.


Also, this could be implemented as an extractor, e.g.:

```
    object PhysicalAttribute {
      import org.apache.spark.sql.connector.catalog.CatalogV2Implicits.MultipartIdentifierHelper
      def unapply(attribute: String): Option[String] = {
        physicalNameMap.get(parseColumnPath(attribute)).map(_.quoted)
      }
    }

    filter match {
      case EqualTo(PhysicalAttribute(physicalAttr), value) =>
        Some(EqualTo(physicalAttr, value))
      case EqualNullSafe(PhysicalAttribute(physicalAttr), value) =>
        Some(EqualNullSafe(physicalAttr, value))
...
```
not that it makes a big difference, just a matter of personal preference

johanl-db (approved):

tdas (approved):

PR #2925, opened 2024-04-19
[Spark] Add Preserving Row Tracking in Delete

longvu-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Add Preserving Row Tracking in Delete by reading the metadata column and writing it out to the physical column.

## How was this patch tested?
Added UTs.

No.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/RowTracking.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/RowTracking.scala
@@ -16,6 +16,7 @@
 
 package org.apache.spark.sql.delta
 
+import org.apache.spark.sql.delta.DeltaCommitTag.PreservedRowTrackingTag
 import org.apache.spark.sql.delta.actions.{Metadata, Protocol, TableFeatureProtocolUtils}
 
 import org.apache.spark.sql.DataFrame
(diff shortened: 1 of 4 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
```diff
@@ -1108,6 +1108,9 @@ trait OptimisticTransactionImpl extends TransactionalWrite
       val readRowIdHighWatermark =
         RowId.extractHighWatermark(snapshot).getOrElse(RowId.MISSING_HIGH_WATER_MARK)
 
+      val tagsWithPreservingRowTrackingIfEnabled =
+        RowTracking.addPreservedRowTrackingTagIfNotSet(snapshot, tags)
```
johanl-db, line 1112: You should set this tag in DELETE, other operations aren't preserving row tracking information.

Alternatively, set it to true by default here but explicitly set the tag to false in operations that don't preserve row tracking information: UPDATE, MERGE, OPTIMIZE, ..?

longvu-db: @johanl-db Done!

johanl-db (approved):
Looks good!

tdas (approved):

PR #2926, opened 2024-04-19
[Spark] Add FileMetadataMaterializationTracker for Row Tracking Backfill

longvu-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

### Background on Row Tracking Backfill 
In a future PR, we will introduce Row Tracking Backfill, a command in Delta Lake that assigns Row IDs and Row Commit Versions to each rows when we enable Row Tracking on an existing table. This is done by committing `addFile` actions to assign the `baseRowId` and the `defaultRowCommitVersion` for every files in the table. Due to the size of the table, doing it in one commit can be very large, unstable and causes a lot of concurrency conflicts, we propose doing it by batches (that is multiple commits, each commit handles a subset of the `addFile` actions of the table).

### Why we need this for Row Tracking Backfill?
However, we could still hit stability issues when a single batch is large enough to OOM the driver. This would happen when individual tasks batched together would be huge. Think of tables where each `AddFile` is just 1-2 rows.

### What is the solution?
We propose having a global file materialization limit that restricts the number of files that can be materialized at once on the driver, this limit will be added when Row Tracking Backfill is introduced in a future PR. A case we need to consider is what if the task size is more than the materialization limit. In that case we can allow the task to complete as we do not want to break the task boundary (breaks idempotence of Row Tracking Backfill).

The `FileMetadataMaterializationTracker` is the component used in the Row Tracking Backfill process to ensure that a single batch is not large enough to OOM the driver.

### Design

The driver holds a semaphore that can give out permits equal to the materialization limit of the driver.

The driver also holds a overallocation lock that allows only one query to over allocate to complete materializing the task.

Each `RowTrackingBackfillCommand` instance will maintain a `FileMetadataMaterializationTracker` that will keep track of how many files were materialized and this will acquire/release the over provisioning semaphore as well.

A permit to materialize a file is acquired while iterating the files iterator while creating a task. The permits are released upon failure or when the batch completes executing.

## How was this patch tested?
Added UTs.

No.


sabir-akhadov:
Looks good! I like the detailed PR description.
I find it a bit unstructured though. Could you please add paragraphs with subtitles like the original PR to separate the design of this class, how it is going to be used by row ids and which problem it solves.

longvu-db:
@sabir-akhadov Great idea! I added the subtitles to make the paragraphs more clearer.

sabir-akhadov (approved):
lgtm

tomvanbussel (approved):
LGTM!

PR #2939, opened 2024-04-22
[Docs] Add documentation for Row Tracking

tomvanbussel:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [x] Other (Docs)

## Description

This PR adds documentation for the new Row Tracking table feature that is releasing in Delta 3.2.0

## How was this patch tested?

N/A

N/A


docs/source/delta-row-tracking.md
```diff
@@ -0,0 +1,113 @@
...
+
+  -- Using a CLONE statement to copy configuration
+  CREATE TABLE graduate CLONE student;
+```
+
+- **Existing table**: Set the table property `'delta.enableRowTracking' = 'true'` in the `ALTER TABLE` command.
+
+  ```sql
+  ALTER TABLE grade SET TBLPROPERTIES ('delta.enableRowTracking' = 'true');
+  ```
```
johanl-db, line 42: This only works if the table is empty currently.
We should probably call that out in the limitations

docs/source/delta-row-tracking.md
```diff
@@ -0,0 +1,113 @@
...
+
+    ```scala
+    spark.conf.set("spark.databricks.delta.properties.defaults.enableRowTracking", true)
+    ```
+
+.. important:: Because cloning a <Delta> table creates a separate history, the row ids and row commit versions on cloned tables do not match that of the original table.
+
+.. important:: Enabling row tracking on existing table will automatically assign row ids and row commit versions to all existing rows in the table. This process may cause multiple new versions of the table to be created and may take a long time.
+
+.. warning:: Tables created with row tracking enabled have the row tracking <Delta> table feature enabled at creation and use <Delta> writer version 7. Table protocol versions cannot be downgraded, and tables with row tracking enabled are not writeable by <Delta> clients that do not support all enabled <Delta> writer protocol table features. See [_](/versioning.md).
```
johanl-db, line 63: I think this is important enough to be pulled as the first element in "Enable row tracking" section

docs/source/delta-row-tracking.md
```diff
@@ -0,0 +1,113 @@
...
+orphan: 1
+---
+
+# Use row tracking for Delta tables
+
+Row tracking allows <Delta> to track row-level lineage in a <Delta> table. When enabled on a <Delta> table, row tracking adds two new metadata fields to the table:
+
+- **Row IDs** provide rows with an identifier that is unique within the table. A row keeps the same ID whenever it is modified using a `MERGE` or `UPDATE` statement.
+
+- **Row commit versions** record the last version of the table in which the row was modified. A row is assigned a new version whenever it is modified using a `MERGE` or `UPDATE` statement.
```
johanl-db, line 12: It could be worth quickly highlighting here that these can be accessed via `_metadata.row_id` and _metadata.row_commit_version` even if there's more detail about it below

docs/source/delta-row-tracking.md
```diff
@@ -0,0 +1,113 @@
...
+  ```
+
+## What is the schema of the row tracking metadata fields?
+
+Row tracking adds the following metadata fields that can be accessed when reading a table:
+
+| Column name                    | Type | Values                                                                       |
+|--------------------------------|------|------------------------------------------------------------------------------|
+| `_metadata.row_id`             | Long | The unique identifier of the row.                                            |
+| `_metadata.row_commit_version` | Long | The table version at which the row was last inserted or updated. |
```
johanl-db, line 95: I would merge these two sections. Start with describing the metadata fields and then considerations about accessing the hidden `_metadata` column

docs/source/delta-row-tracking.md
```diff
@@ -0,0 +1,113 @@
...
+
+Row tracking can be disabled to reduce the storage overhead of the metadata fields. After disabling row tracking the metadata fields remain available, but all rows always get assigned a new id and commit version whenever they are touched by an operation.
+
+```sql
+ALTER TABLE table_name SET TBLPROPERTIES (delta.enableRowTracking = false);
+```
+
+.. important:: Disabling row tracking does not remove the corresponding table feature and does not downgrade the table protocol version.
+
+## Limitations
```
johanl-db, line 107: Another limitation: we can't drop the row tracking table feature currently

johanl-db (commented):
I don't know if this should target `master`, `branch-3.2` or both. Probably both to have a current and historic version of the doc

docs/source/delta-row-tracking.md
```diff
@@ -0,0 +1,113 @@
...
+  CREATE TABLE course_new
+  TBLPROPERTIES ('delta.enableRowTracking' = 'true')
+  AS SELECT * FROM course_old;
+
+  -- Using a LIKE statement to copy configuration
+  CREATE TABLE graduate LIKE student;
+
+  -- Using a CLONE statement to copy configuration
+  CREATE TABLE graduate CLONE student;
+```
```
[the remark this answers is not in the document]
tomvanbussel: Thank you for catching this! Should be fixed now.

docs/source/delta-row-tracking.md
tomvanbussel: Done

docs/source/delta-row-tracking.md
tomvanbussel: I'm trying to keep this first section a little bit more high-level.

johanl-db (approved):
LGTM!
I deployed the change locally, rendering looks good

PR #2946, opened 2024-04-22
[Spark] Make ICT available outside of test environments

dhruvarya-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Move the InCommitTimestamp feature to the list of supported features so that users can enable it outside of the test environment. Note that InCommitTimestamp is still an RFC and the feature name will still have the -dev suffix in it.

## How was this patch tested?

Existing tests.

Yes, users can now use the `InCommitTimestamp-dev` feature in their environments if they want to try it before the RFC is  merged.


scovich (approved):
Nice that we're ready to start trying this out so soon.

PR #2966, opened 2024-04-24
Fix import order in DeltaFormatSharingSource

linzhou-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [X] Other (Delta Sharing)

## Description
Fix import order in DeltaFormatSharingSource

## How was this patch tested?
Unit Tests

No


scottsand-db (approved):

tdas (changes requested):
I dont think we want to accept trivial changes like this. I do think we need to fix our scala style to better enforce import order.. but in the absence of that, I dont think making trivial import order changes make sense

linzhou-db:
> I dont think we want to accept trivial changes like this. I do think we need to fix our scala style to better enforce import order.. but in the absence of that, I dont think making trivial import order changes make sense

sg, I'll close this.

PR #2977, opened 2024-04-26
[Spark] Enable ICT and VacuumProtocolCheck automatically with Managed Commits

dhruvarya-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Enabling Managed Commits will automatically:

1. Add ICT and VacuumProtocolCheck features to the protocol.
2. Enable ICT

These two features are pre-requisites for managed commit.

## How was this patch tested?

New suite ManagedCommitEnablementSuite. Updated existing tests to work with ICT.

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
@@ -695,6 +695,25 @@ trait OptimisticTransactionImpl extends TransactionalWrite
     MaterializedRowCommitVersion.throwIfMaterializedColumnNameConflictsWithSchema(metadata)
   }
 
+  /**
+   * Some features require their pre-requisite features to not only be present
+   * in the protocol but also be enabled. This method sets the flags required
+   * to enable these pre-requisite features.
+   */
+  private def getMetadataWithDependentFeaturesEnabled(metadata: Metadata): Metadata = {
+    DeltaConfigs.MANAGED_COMMIT_OWNER_NAME.fromMetaData(metadata).map { _ =>
+      // managed-commits requires ICT to be enabled as per the spec.
+      // If ICT is just in Protocol and not in Metadata,
+      // then it is in a 'supported' state but not enabled.
+      // In order to enable ICT, we have to set the table property in Metadata.
+      val ictEnablementConfigOpt =
+        Option.when(!DeltaConfigs.IN_COMMIT_TIMESTAMPS_ENABLED.fromMetaData(metadata))(
+          (DeltaConfigs.IN_COMMIT_TIMESTAMPS_ENABLED.key -> "true"))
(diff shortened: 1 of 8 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
```diff
@@ -1349,8 +1377,6 @@ trait OptimisticTransactionImpl extends TransactionalWrite
             removeFilesHistogram.foreach(_.insert(r.getFileSize))
           case _: SetTransaction =>
             numSetTransaction += 1
-          case m: Metadata =>
-            assertMetadata(m)
```
prakharjain09, line 1353: Just to make sure that I for this right: we don't need this anymore here as `updateMetadataAndProtocolWithRequiredFeatures` already takes care of this check?

spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
```diff
@@ -695,6 +695,25 @@ trait OptimisticTransactionImpl extends TransactionalWrite
...
+   * to enable these pre-requisite features.
+   */
+  private def getMetadataWithDependentFeaturesEnabled(metadata: Metadata): Metadata = {
+    DeltaConfigs.MANAGED_COMMIT_OWNER_NAME.fromMetaData(metadata).map { _ =>
+      // managed-commits requires ICT to be enabled as per the spec.
+      // If ICT is just in Protocol and not in Metadata,
+      // then it is in a 'supported' state but not enabled.
+      // In order to enable ICT, we have to set the table property in Metadata.
+      val ictEnablementConfigOpt =
+        Option.when(!DeltaConfigs.IN_COMMIT_TIMESTAMPS_ENABLED.fromMetaData(metadata))(
```
prakharjain09, line 710: We don't need to explicitly enabled `VacuumProtocolCheckTableFeature` here as TableFeature infra takes care of enabling the requiredFeatures?

prakharjain09 (approved):

spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
dhruvarya-db: Yes, there has been no change in the number of times assertMetadata is called.

spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
dhruvarya-db: Yes, `VacuumProtocolCheckTableFeature` does not have any metadata property that enables it. It is directly enabled/disabled in the protocol.

PR #2981, opened 2024-04-26
[Spark] Better schema validation in MERGE schema evolution tests

johanl-db:
## Description
Address following shortcomings of schema evolution tests in MERGE:
- Tests ignore the nullability of fields when validating the schema of the table after evolution, which prevents checking for nullability of specific fields.

- For struct evolution tests, a result schema can be passed but this schema is only used to parse the result data and not to validate the schema of the table after evolution.

## How was this patch tested?
Updated tests

No, test-only


fred-db (approved):
LGTM, thanks for adding the testing! :)

PR #2987, opened 2024-04-27
[Spark] Support dropping the CHECK constraints table feature

tomvanbussel:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR adds support for dropping the `checkConstraints` table feature using the `ALTER TABLE ... DROP FEATURE` command. It throws an error if the table still contains CHECK constraints (as dropping a feature should never change the logical state of a table).

## How was this patch tested?

Added a test to `CheckConstraintsSuite`

Yes, `ALTER TABLE ... DROP FEATURE checkConstraints` is now supported.


spark/src/main/resources/error/delta-error-classes.json
```diff
@@ -170,6 +170,13 @@
     ],
     "sqlState" : "42703"
   },
+  "DELTA_CANNOT_DROP_CHECK_CONSTRAINT_FEATURE" : {
+    "message" : [
+      "Cannot drop the CHECK constraints table feature.",
+      "The following constraints must be dropped first: <constraints>."
```
andreaschat-db, line 176: I was thinking that this a only a writer feature which means that can be safely ignored by readers that do not understand it. In that sense, dropping the feature should be allowed even if there are constraints in the table. They just need to be ignored. However, we do no need add a guard (if it is not already there) to only check constraints when the feature is enabled/supported. That is for writers that understand the feature.

On the other hand, if we do want to force this behavior it means the feature cannot be removed until the history is truncated. History validation and check is only done at the moment for all reader+writer features but there are plans to allow it for selected writer features.

andreaschat-db: To add to that, it is good to drop constraints before dropping the feature but we should not check the history if there are constraints, i.e. `actionUsesFeature.`

tomvanbussel: Requiring users to drop all constraints first before allowing the feature to be dropped is intentional. Otherwise, some users might not realize that dropping the `checkConstraints` feature will drop all check constraints on their table, or they may not realize that there are check constraints defined on their table.

We do not require all constraints to be removed from the history (see the added test), only that the current state of the table does not contain any check constraints. I have implemented `actionUsesFeature`, but AFAIK it is never called right now, as `historyContainsFeature` is only called for reader features (and this is a writer feature).

spark/src/main/scala/org/apache/spark/sql/delta/PreDowngradeTableFeatureCommand.scala
```diff
@@ -378,3 +379,12 @@ case class ColumnMappingPreDowngradeCommand(table: DeltaTableV2)
     true
   }
 }
+
+case class CheckConstraintsPreDowngradeTableFeatureCommand(table: DeltaTableV2)
+    extends PreDowngradeTableFeatureCommand {
+  override def removeFeatureTracesIfNeeded(): Boolean = {
```
andreaschat-db, line 385: It would be nice to have a short comment here about what you explained above in https://github.com/delta-io/delta/pull/2987/files#r1635146168.

spark/src/main/scala/org/apache/spark/sql/delta/TableFeature.scala
```diff
@@ -457,12 +457,24 @@ object InvariantsTableFeature
...
     Constraints.getCheckConstraints(metadata, spark).nonEmpty
   }
+
+  override def preDowngradeCommand(table: DeltaTableV2): PreDowngradeTableFeatureCommand =
+    CheckConstraintsPreDowngradeTableFeatureCommand(table)
+
+  override def validateRemoval(snapshot: Snapshot): Boolean =
+    Constraints.getCheckConstraintNames(snapshot.metadata).isEmpty
+
+  override def actionUsesFeature(action: Action): Boolean = action match {
```
andreaschat-db, line 474: Shall we then remove `actionUsesFeature` since it is not used? It is confusing to see it implemented.

andreaschat-db (commented):
Left a couple of comments.

spark/src/main/scala/org/apache/spark/sql/delta/constraints/Constraints.scala
```diff
@@ -49,6 +49,13 @@ object Constraints {
   /** A SQL expression to check for when writing out data. */
   case class Check(name: String, expression: Expression) extends Constraint
 
+  def getCheckConstraintNames(metadata: Metadata): Seq[String] = {
```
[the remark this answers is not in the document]
tomvanbussel: I prefer to keep the brackets here, as this is a multi-line method. This helps avoid unhappy surprises in the future when modifying the code. The rest of this class also use brackets for multi-line methods (even when not needed).

spark/src/test/scala/org/apache/spark/sql/delta/schema/CheckConstraintsSuite.scala
```diff
@@ -459,4 +459,35 @@ class CheckConstraintsSuite extends QueryTest
...
+        sql("ALTER TABLE table DROP FEATURE checkConstraints")
+      }
+      checkError(
+        error2,
+        errorClass = "DELTA_CANNOT_DROP_CHECK_CONSTRAINT_FEATURE",
+        parameters = Map("constraints" -> "`c2`")
+      )
+
+      sql("ALTER TABLE table DROP CONSTRAINT c2")
+      sql("ALTER TABLE table DROP FEATURE checkConstraints")
```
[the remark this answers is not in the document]
tomvanbussel: Done!

spark/src/main/scala/org/apache/spark/sql/delta/PreDowngradeTableFeatureCommand.scala
```diff
@@ -378,3 +379,12 @@ case class ColumnMappingPreDowngradeCommand(table: DeltaTableV2)
...
+    val checkConstraintNames = Constraints.getCheckConstraintNames(table.initialSnapshot.metadata)
+    if (checkConstraintNames.isEmpty) return false
+    throw DeltaErrors.cannotDropCheckConstraintFeature(checkConstraintNames)
```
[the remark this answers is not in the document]
tomvanbussel: Agreed that dropping the constraints one-by-one is not great. On the other hand, I do like asking the user to explicitly drop these constraints themselves. A warning can be easily missed/skipped. Perhaps we can add a `ALTER TABLE ... DROP CONSTRAINTS` syntax in the future that allows dropping multiple constraints at once?

spark/src/main/scala/org/apache/spark/sql/delta/PreDowngradeTableFeatureCommand.scala
tomvanbussel: Done!

spark/src/main/scala/org/apache/spark/sql/delta/TableFeature.scala
tomvanbussel: Modified it to throw an exception instead (to make sure it is actually never used).

andreaschat-db (approved):
LGTM. Thanks!

PR #3000, opened 2024-05-01
[Spark] Ensure commit directory is created for old and new tables

sumeet-db:
- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

- Remove `createLogDirectory` and replace all usages with `ensureLogDirectoryExists` since the latter is optimized for creation rather than existence check now. Rename it `createLogDirectoriesIfNotExists`
- Ensure commit directory is created on existing tables from older releases when manged-commits is enabled.

## How was this patch tested?

UTs

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
@@ -1415,7 +1415,7 @@ trait OptimisticTransactionImpl extends TransactionalWrite
         .assignIfMissing(protocol, allActions, getFirstAttemptVersion)
 
       if (readVersion < 0) {
-        deltaLog.createLogDirectory()
+        deltaLog.createLogDirectoriesIfNotExists()
       }
       val fsWriteStartNano = System.nanoTime()
       val jsonActions = allActions.map(_.json)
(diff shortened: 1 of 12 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
```diff
@@ -1694,6 +1694,9 @@ trait OptimisticTransactionImpl extends TransactionalWrite
         }
         logWarning("Detected no metadata in initial commit but commit validation was turned off.")
       }
+    } else {
+      // Ensure Commit Directory exists if managed commits is enabled on an existing table.
+      newMetadata.flatMap(_.managedCommitOwnerName).foreach(_ => deltaLog.ensureLogDirectoryExist())
```
prakharjain09, line 1699: We also need to check that oldMetadata didn't had managed commits? Otherwise we will keep on invoking this function forever for MC tables.

prakharjain09: We can also move the ensureLogDirectory() logic to a separate block outside this existing if condition.

```
  lazy val isFsToMcConversion = oldMetadata....isEmpty && newMetadata....nonEmpty
  val shouldCreateLogDirs = snapshot.version == -1 || isfsToMcConversion
  if (ishouldCreateLogDirs) {
    deltaLog.ensureLogDirectoryExist()
  }

```

tdas:
Yooohoooo. THREE THOUSAND!!

spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
prakharjain09: > We also need to check that oldMetadata didn't had managed commits? Otherwise we will keep on invoking this function forever for MC tables.

Can we also add a test for this case?

sumeet-db: Added.

prakharjain09 (approved):

PR #3002, opened 2024-05-01
[Spark] Makes CommitOwnerClient independent of Delta Spark dependencies

LukasRupprecht:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR is the first in a series of PRs that refactors the current CommitOwnerClient interface to become its own module. This is similar to `storage` module that currently exists and contains the LogStore interface and its implementations. In this PR, we remove any Delta Spark dependencies from the CommitOwnerClient in preparation for it to be moved outside of Delta Spark.

## How was this patch tested?

Added tests to check equivalence of newly introduced AbstractProtocol/AbstractMetadata with Protocol/Metadata.

No

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/CommitOwnerClientSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/CommitOwnerClientSuite.scala
@@ -237,14 +237,14 @@ class CommitOwnerClientSuite extends QueryTest with DeltaSQLTestUtils with Share
 
   /**
    * We expect the Protocol action to have the same fields as AbstractProtocol (part of the
-   * CommitStore interface). With this if any change has happened in the Protocol of the table,
-   * the same change is propagated to the CommitStore as AbstractProtocol. The CommitStore can
-   * access the changes using getters and decide to act on the changes based on the spec of
-   * the commit-owner.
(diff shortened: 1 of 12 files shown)


spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/CommitOwnerClientSuite.scala
```diff
@@ -218,4 +220,54 @@ class CommitOwnerClientSuite extends QueryTest with DeltaSQLTestUtils with Share
...
+    }.toSet
+
+    fields.filterNot { field =>
+      getters.contains(s"get${field.capitalize}")
+    }.toSet
+  }
+
+  /**
+   * We expect the Protocol action to have the same fields as AbstractProtocol (part of the
+   * CommitStore interface). With this if any change has happened in the Protocol of the table,
```
prakharjain09, line 240: CommitStore has been replaced with Commit owner / ComitOwnerClient -- depending on the context in the entire code/comments. Could we follow the same convention here as well.

LukasRupprecht: Updated to reflect the new naming.

prakharjain09 (approved):

PR #3010, opened 2024-05-01
[Spark] Add drop support for managed commits

dhruvarya-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Adds table feature phaseout support for managed commits.
Fixes the ManagedCommit table feature --> correctly marks it as a WriterFeature instead of a ReaderWriterFeature.

## How was this patch tested?

New tests in DeltaProtocolVersionSuite.

Yes, the drop table command now works for managed commits.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/PreDowngradeTableFeatureCommand.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/PreDowngradeTableFeatureCommand.scala
@@ -22,6 +22,7 @@ import scala.util.control.NonFatal
 
 import org.apache.spark.sql.delta.catalog.DeltaTableV2
 import org.apache.spark.sql.delta.commands.{AlterTableSetPropertiesDeltaCommand, AlterTableUnsetPropertiesDeltaCommand, DeltaReorgTableCommand, DeltaReorgTableMode, DeltaReorgTableSpec}
+import org.apache.spark.sql.delta.managedcommit.ManagedCommitUtils
 import org.apache.spark.sql.delta.metering.DeltaLogging
 import org.apache.spark.sql.delta.util.{Utils => DeltaUtils}
 import org.apache.spark.sql.util.ScalaExtensions._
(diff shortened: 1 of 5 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/PreDowngradeTableFeatureCommand.scala
```diff
@@ -213,6 +213,56 @@ case class VacuumProtocolCheckPreDowngradeCommand(table: DeltaTableV2)
...
+  /**
+   * We disable the feature by removing the following table properties:
+   *    1. DeltaConfigs.MANAGED_COMMIT_OWNER_NAME.key
+   *    2. DeltaConfigs.MANAGED_COMMIT_OWNER_CONF.key
+   *    3. DeltaConfigs.MANAGED_COMMIT_TABLE_CONF.key
+   *
+   * @return true if any change to the metadata (the three properties listed above) was made.
+   *         false otherwise.
+   */
+  override def removeFeatureTracesIfNeeded(): Boolean = {
```
prakharjain09, line 229: This function starts returning false as soon as these 3 table properties are not part of the table Metadata. That signals that there are no traces of the feature - which is not true in case there are unbackfilled commits in the Snapshot.

Once this method returns false, the AlterTable command will do a commit to remove managed-commit from the table and that takes care of backfilling any unbackfilled commits. So although it doesn't shouldn't cause any issue but semantically we should not give a green light from here until the commits are also backfilled.

Could we make sure that unbackfilled commits are backfilled by doing:

        table.initialSnapshot.deltaLog.update().ensureCommitFilesBackfilled()

prakharjain09: Actually the workaround that I suggested `snapshot.ensureCommitFilesBackfilled` won't work as there is no commit-owner on the table after `AlterTableUnsetPropertiesDeltaCommand` is done .

dhruvarya-db: Okay, this sounds reasonable. Semantically, `removeFeatureTracesIfNeeded` is not complete if there are any unbackfilled commits after it has finished.

spark/src/main/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitUtils.scala
```diff
@@ -108,4 +110,85 @@ object ManagedCommitUtils {
...
+
+  /**
+   * This method takes care of backfilling any unbackfilled delta files when managed commit is
+   * not enabled on the table (i.e. commit-owner is not present) but there are still unbackfilled
+   * delta files in the table. This can happen if an error occurred during the MC -> FS commit
+   * where the commit-owner was able to register the downgrade commit but it failed to backfill
+   * it. This method must be invoked before doing the next commit as otherwise there will be a
+   * gap in the backfilled commit sequence.
+   */
+  def backfillWhenManagedCommitDisabled(snapshot: Snapshot): Unit = {
```
dhruvarya-db, line 151: Moved as is from OptimisticTransaction.

prakharjain09 (approved):

PR #3013, opened 2024-05-01
[SPARK][VARIANT] Add more variant test cases

richardc-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

adds more tests to ensure that Delta spark can use various delta features with variant

## How was this patch tested?

test only change

no


spark/src/test/scala-spark-master/org/apache/spark/sql/delta/DeltaVariantSuite.scala
```diff
@@ -126,4 +137,417 @@ class DeltaVariantSuite
...
+      DeltaTable.create(spark)
+        .tableName("tbl")
+        .addColumn("v", "VARIANT")
+        .addColumn(
+          DeltaTable.columnBuilder(spark, "vInt")
+            .dataType("INT")
+            .generatedAlwaysAs("v::int")
+            .build()
+        )
+        .execute()
```
johanl-db, line 446: Personal preference: I find using SQL more readable:
```
sql("CREATE TABLE tbl(v VARIANT, vInt INT GENERATED ALWAYS AS v::int)")
```

spark/src/test/scala-spark-master/org/apache/spark/sql/delta/DeltaVariantSuite.scala
```diff
@@ -126,4 +137,417 @@ class DeltaVariantSuite
...
+      val expectedDf = spark.range(0, 100).selectExpr(
+        "parse_json(cast(id as string)) v",
+        "cast(id as int) vInt"
+      )
+      val actualDf = spark.sql("select * from tbl")
+      checkAnswer(actualDf, expectedDf.collect())
+    }
+  }
+
+  test("Variant cannot be created as a generated column") {
```
johanl-db, line 462: It this something that we could eventually support?

spark/src/test/scala-spark-master/org/apache/spark/sql/delta/DeltaVariantSuite.scala
```diff
@@ -126,4 +137,417 @@ class DeltaVariantSuite
...
+  }
+
+  test("variant cannot be used as a clustering column") {
+    withTable("tbl") {
+      val e = intercept[DeltaAnalysisException] {
+        sql("CREATE TABLE tbl(v variant) USING DELTA CLUSTER BY (v)")
+      }
+      checkError(
+        e,
+        "DELTA_CLUSTERING_COLUMN_MISSING_STATS",
```
johanl-db, line 210: This fails because of missing stats but tests in `StatsCollectionSuite` show that we can compute stats for variant columns.
Is this test just not doing some pre-requisite steps to be able to use variant as clustering key or do we actually reject this?

richardc-db: IIUC, clustering requires the type to be comparable. Currently variants are not comparable so the min/max stats are missing. null count stats, however, are present.

spark/src/test/scala-spark-master/org/apache/spark/sql/delta/DeltaVariantSuite.scala
richardc-db: Yeah. Currently its not allowed because writers must be able to enforce that <variant value> <=> <generated expresion>. Currently, variants aren't comparable so its not allowed.

I updated the check in `GeneratedColumn` to be more specific. Before this PR, it was actually possible to `CAST(intVal as variant)` to generate a variant column. The new check completely disallows variant types.

spark/src/test/scala-spark-master/org/apache/spark/sql/delta/DeltaVariantSuite.scala
richardc-db: hmm.. do we expect this exact syntax to work? For some reason I get the error
```
[UNSUPPORTED_FEATURE.TABLE_OPERATION] The feature is not supported: Table `spark_catalog`.`default`.`tbl` does not support generated columns. Please check the current catalog and namespace to make sure the qualified table name is expected, and also check the catalog implementation which is configured by "spark.sql.catalog". SQLSTATE: 0A000
```

even when changing `v` from a variant column to an int column.

spark/src/test/scala-spark-master/org/apache/spark/sql/delta/DeltaVariantSuite.scala
```diff
@@ -126,4 +137,417 @@ class DeltaVariantSuite
...
+      // Adding the constraint should work after deleting the variant that is less than zero.
+      sql("ALTER TABLE tbl ADD CONSTRAINT variantGTEZero CHECK (variant_get(v, '$', 'INT') >= 0)")
+      val newLessThanZeroCount = spark.sql("select * from tbl where v::int < 0").count()
+      assert(newLessThanZeroCount == 0)
+    }
+  }
+
+  // These tests read from Delta tables written by other engines to ensure that delta-spark is
+  // compatible.
+  // The golden tables were written using "variant-table.sql" so running this SQL file
```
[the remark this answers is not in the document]
richardc-db: hmm agree. I've removed them for now from this PR. Maybe they can be added later

spark/src/test/scala-spark-master/org/apache/spark/sql/delta/DeltaVariantSuite.scala
johanl-db: I missed a set of parentheses:
```
sql("CREATE TABLE tbl(v VARIANT, vInt INT GENERATED ALWAYS AS (v::int))")
```

spark/src/main/scala/org/apache/spark/sql/delta/DeltaErrors.scala
```diff
@@ -1765,6 +1765,13 @@ trait DeltaErrorsBase
     )
   }
 
+  def generatedColumnsUnsupportedType(dt: DataType): Throwable = {
+    new DeltaAnalysisException(
+      errorClass = "DELTA_UNSUPPORTED_DATA_TYPE_IN_GENERATED_COLUMN",
+      messageParameters = Array(s"${dt.simpleString}")
```
johanl-db, line 1771: nit:
use `dt.sql` or `toSQLType(dt)`

johanl-db (approved):
One small nit

spark/src/test/scala-spark-master/org/apache/spark/sql/delta/DeltaVariantSuite.scala
richardc-db: ah sorry, I should've been specific. Even with the parentheses I get the same error...

PR #3014, opened 2024-05-01
[Kernel] Refactor all user-facing exceptions to be "KernelExceptions"

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

In order to provide a unified and user-friendly experience we are making all user-facing exceptions instances of one parent class `KernelException`. This PR also attempts to unify/improve some of these error messsages.

## How was this patch tested?

Existing tests should suffice (updated to match the changes here).

Yes, changes exception types + messages.


vkorukanti (approved):
LGTM. I think there is at least one more exception that need to extend the `KernelException`. E.g `CheckpointAlreadyExistsException`

PR #3022, opened 2024-05-02
[Spark] Make txn readPredicates thread safe

cstavr:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

`OptimisticTransaction.readPredicates` may be updated by multiple threads that call `filesForScan`. This commit turns it from an `ArrayBuffer` to a `ConcurrentLinkedQueue` to be thread safe.

## How was this patch tested?

Existing tests.

No


bart-samwel (approved):

felipepessoto:
@cstavr @scottsand-db in this PR we updated `readPredicates`, but not the `readFiles`. Doesn't make sense to update `readFiles` as well as they are usually updated together

felipepessoto:
Created a follow up PR: https://github.com/delta-io/delta/pull/6097

PR #3024, opened 2024-05-02
[Spark] Allow type widening for all supported type changes with Spark 4.0

johanl-db:
This PR adds shims to ungate the remaining type changes that only work with Spark 4.0 / master. Spark 4.0 contains the required changes to Parquet readers to be able to read the data after applying the type changes.

## Description
Extend the list of supported type changes for type widening to include changes that can be supported with Spark 4.0:
- (byte, short, int) -> long
- float -> double
- date -> timestampNTZ
- (byte, short, int) -> double
- decimal -> decimal (with increased precision/scale that doesn't cause precision loss)
- (byte, short, int, long) -> decimal

Shims are added to support these changes when compiling against Spark 4.0/master and to only allow `byte` -> `short`  - > `int` when compiling against Spark 3.5.

## How was this patch tested?
Adding test cases for the new type changes in the existing type widening test suites. The list of supported / unsupported changes covered in tests differs between Spark 3.5 and Spark 4.0, shims are also provided to handle this.

Yes: allow using the listed type changes with type widening, either via `ALTER TABLE CHANGE COLUMN TYPE` or during schema evolution in MERGE and INSERT.


spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
```diff
@@ -89,6 +92,19 @@ trait DeltaTypeWideningTestMixin extends SharedSparkSession with DeltaDMLTestUti
...
+  def checkAnswerWithTolerance(df: DataFrame, expected: DataFrame, dataType: DataType): Unit = {
+    if (Seq(FloatType, DoubleType).contains(dataType)) {
+      checkAggregatesWithTol(
+        df.sort("value"),
+        expected.sort("value").collect(),
+        absTol = 0.001)
+    } else {
+      checkAnswer(df, expected)
+    }
+  }
```
johanl-db, line 107: I was hoping this would work to compare floats but it doesn't.. I haven't found an alternative yet besides rolling out my own comparison

KamilKandzia:
Will be in future an option to change the column type of a table from int to string without overwriting the entire table? Unless such an option is now available (but I don't remember that)

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
johanl-db: Rolled out my own in the end

spark/src/main/scala-spark-3.5/shims/TypeWideningShims.scala
```diff
@@ -0,0 +1,57 @@
...
+      case (ByteType, ShortType) => true
+      case (ByteType | ShortType, IntegerType) => true
+      case _ => false
+    }
+
+  /**
+   * Returns whether the given type change can be applied during schema evolution. Only a
+   * subset of supported type changes are considered for schema evolution.
+   */
+  def isTypeChangeSupportedForSchemaEvolution(fromType: AtomicType, toType: AtomicType): Boolean =
```
sabir-akhadov, line 49: I might be wrong but isn't this method equivalent to `isTypeChangeSupported`?

spark/src/main/scala-spark-master/shims/TypeWideningShims.scala
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+package org.apache.spark.sql.delta
+
+import org.apache.spark.sql.catalyst.expressions.Cast
+import org.apache.spark.sql.types._
+
+/**
+ * Type widening only supports a limited set of type changes with Spark 3.5 due to the parquet
+ * readers lacking the corresponding conversions that were added in Spark 4.0.
```
sabir-akhadov, line 24: Not sure about the mechanics but shouldn't this go in a `scala-spark-4.0` directory instead of master? What happens when the 4.0 is cut/released?

spark/src/main/scala/org/apache/spark/sql/delta/TypeWidening.scala
```diff
@@ -66,28 +65,14 @@ object TypeWidening {
...
-    (fromType, toType) match {
-      case (from, to) if from == to => true
-      // All supported type changes below are supposed to be widening, but to be safe, reject any
-      // non-widening change upfront.
-      case (from, to) if !Cast.canUpCast(from, to) => false
-      case (ByteType, ShortType) => true
-      case (ByteType | ShortType, IntegerType) => true
-      case _ => false
-    }
+    TypeWideningShims.isTypeChangeSupported(fromType, toType)
```
sabir-akhadov, line 68: For my education, how is the TypeWideningShims object visible here?

spark/src/test/scala-spark-3.5/shims/TypeWideningTestCasesShims.scala
```diff
@@ -0,0 +1,141 @@
...
+ * accordingly for delta on Spark 3.5.
+ */
+trait TypeWideningTestCasesShims {
+  self: DeltaTypeWideningTestCases with SQLTestUtils =>
+
+  import testImplicits._
+
+  // Type changes that are supported by all Parquet readers. Byte, Short, Int are all stored as
+  // INT32 in parquet so these changes are guaranteed to be supported.
+  protected val supportedTestCases: Seq[TypeEvolutionTestCase] = Seq(
```
sabir-akhadov, line 34: That's a nice solution using shims to manipulate test cases!

sabir-akhadov (commented):
looks good, left some comments

spark/src/main/scala-spark-3.5/shims/TypeWideningShims.scala
johanl-db: Right, I updated it

sabir-akhadov (approved):
lgtm

PR #3028, opened 2024-05-02
[Spark] Extend Delta feature drop safety check to unbackfilled deltas

sumeet-db:
- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

When dropping a feature, DeltaLog now checks both backfilled and unbackfilled deltas for any traces of the feature before confirming it's safe to drop.

However, feature dropping currently does a checkpoint before detecting feature traces in the history, so there are no unbackfilled deltas at that point.

## How was this patch tested?

UTs

No

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/delta/DeltaTableFeatureSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/DeltaTableFeatureSuite.scala
@@ -22,6 +22,7 @@ import scala.collection.mutable
 
 import org.apache.spark.sql.delta.actions._
 import org.apache.spark.sql.delta.actions.TableFeatureProtocolUtils._
+import org.apache.spark.sql.delta.managedcommit.{CommitOwnerProvider, InMemoryCommitOwnerBuilder}
 import org.apache.spark.sql.delta.sources.DeltaSQLConf
 import org.apache.spark.sql.delta.test.DeltaSQLCommandTest
 import org.apache.spark.sql.delta.test.DeltaTestImplicits._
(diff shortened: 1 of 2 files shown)


spark/src/test/scala/org/apache/spark/sql/delta/DeltaTableFeatureSuite.scala
```diff
@@ -490,6 +491,52 @@ class DeltaTableFeatureSuite
...
+      // Disable feature on the latest snapshot
+      val tblProperties3 = Seq(s"'${TestRemovableReaderWriterFeature.TABLE_PROP_KEY}' = 'false'")
+      sql(buildTablePropertyModifyingCommand(
+        "ALTER", targetTableName = table, sourceTableName = table, tblProperties3))
+
+      val tableFeature =
+        TableFeature.featureNameToFeature(featureName).get.asInstanceOf[RemovableFeature]
+      assert(tableFeature.historyContainsFeature(spark, log.update()))
+
+      // Dropping feature should fail because the feature still has traces in deltas.
```
prakharjain09, line 525: Even without the fix in this PR, the DROP FEATURE will fail as expected as we trigger a checkpoint for feature drop and that backfills the commit?

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTableFeatureSuite.scala
```diff
@@ -490,6 +491,52 @@ class DeltaTableFeatureSuite
...
+      assert(log.update().protocol.readerAndWriterFeatureNames.contains(featureName))
+
+      // Disable feature on the latest snapshot
+      val tblProperties3 = Seq(s"'${TestRemovableReaderWriterFeature.TABLE_PROP_KEY}' = 'false'")
+      sql(buildTablePropertyModifyingCommand(
+        "ALTER", targetTableName = table, sourceTableName = table, tblProperties3))
+
+      val tableFeature =
+        TableFeature.featureNameToFeature(featureName).get.asInstanceOf[RemovableFeature]
+      assert(tableFeature.historyContainsFeature(spark, log.update()))
```
prakharjain09, line 523: This used to return false before your changes?

prakharjain09 (approved):

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTableFeatureSuite.scala
sumeet-db: That's right. DROP FEATURE will still fail because checkpointing ensures there are no unbackfilled files.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTableFeatureSuite.scala
sumeet-db: Correct.

PR #3037, opened 2024-05-03
[Spark] Fix GeneratedColumnSuite tests with type changes

johanl-db:
## Description
Spark started using ANSI mode enabled by default, breaking a couple of tests in `GeneratedColumnSuite` that contained slient overflows.
These tests are updated to remove the overflows.

## How was this patch tested?
Running tests with ANSI mode enabled


xupefei:
Looks good to me!

xupefei (approved):

PR #3039, opened 2024-05-03
[Spark]Block metadata conversion for Iceberg tables with row-level deletes

KaiqiJinWow:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Iceberg V2 spec introduce the row-level deletes, which is not supported yet for metadata conversion. We add the blocker here to throw exception if customer want to convert a iceberg table with row-level deletes.

Also MOR is not necessarily to be enabled for using row-level deletes, so we remove the check for MOR configurations here and only check if the deleteFiles is nonEmpty in `planFiles()`.

## How was this patch tested?

Exist tests.

No


mingdai-db (approved):
LGTM, let's mention in the description that MOR is not necessarily to be enabled for using row-level deletes

PR #3042, opened 2024-05-03
[Doc] Add a short note about the schema sync threshold

dhruvarya-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [X] Other (Doc)

## Description

Adds a short note about truncation of schemas with long fields during sync to HMS.

### DIFF
--- a/docs/source/delta-batch.md
+++ b/docs/source/delta-batch.md
@@ -1227,7 +1227,7 @@ If Schema is large, the schema will be broken down into multiple parts. Appendin
   ```
 
 .. note::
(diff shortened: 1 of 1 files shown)


docs/source/delta-batch.md
```diff
@@ -1226,6 +1226,9 @@ If Schema is large, the schema will be broken down into multiple parts. Appendin
   }
   ```
 
+.. note::
+    By default, if the table schema has a field longer than 4000 characters, an empty schema will be stored in HMS. This truncation threshold can be increased by setting `spark.databricks.delta.catalog.update.longFieldTruncationThreshold` to a larger number.
```
prakharjain09, line 1230: This is an internal conf and should not be documents?

PR #3046, opened 2024-05-03
[Spark] Fix CommitInfo.inCommitTimestamp deserialization for very small timestamps

dhruvarya-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Currently, if we deserialize a CommitInfo with a very small inCommitTimestamp and then try to access this inCommitTimestamp, this exception is thrown

```
java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.Long
```
This PR fixes the CommitInfo so that the inCommitTimestamp field is deserialized correctly.

## How was this patch tested?

Added a new test case that was failing before the fix.

No

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/delta/ActionSerializerSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/ActionSerializerSuite.scala
@@ -194,6 +194,15 @@ class ActionSerializerSuite extends QueryTest with SharedSparkSession with Delta
     assert(Action.fromJson(json1).asInstanceOf[CommitInfo].inCommitTimestamp.get == 123L)
   }
 
+  test("deserialization of CommitInfo with a very large ICT") {
+    val json1 =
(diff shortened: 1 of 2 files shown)


junlee-db (approved):

spark/src/test/scala/org/apache/spark/sql/delta/ActionSerializerSuite.scala
```diff
@@ -185,6 +185,25 @@ class ActionSerializerSuite extends QueryTest with SharedSparkSession with Delta
     assert(Action.fromJson(json1) === expectedCommitInfo)
   }
 
+  test("deserialization of CommitInfo with a very small ICT") {
+    val json1 =
+      """{"commitInfo":{"inCommitTimestamp":123,"timestamp":123,"operation":"CONVERT",""" +
```
prakharjain09, line 190: Can we also have a test case with a Long ICT value.

prakharjain09 (approved):

PR #3049, opened 2024-05-04
[Spark] Extend Fuzz test For Managed Commit

junlee-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR extends Fuzz test to test managed commit features. Specifically, it adds a new event phase inside commit operation, so that we can capture the backfill as a separate operation. By doing so, it is possible that multiple commits can go through before backfill and managed commit is expected to deal with various situation to return the correct output.

## How was this patch tested?
Existing fuzz tests should naturally use the extended backfill phases.

No


prakharjain09 (approved):
LGTM!

PR #3052, opened 2024-05-06
[Spark] Fix logging failure in deltaAssert helper

johanl-db:
## Description
Fix helper method `deltaAssert` introduced in https://github.com/delta-io/delta/pull/2709 to not log failures when the assertion holds.

## How was this patch tested?
The helper intentionally behaves differently in tests and in production (failing vs. logging), there's no easy/meaningful way to test the prod behavior from tests.


fred-db (approved):
LGTM, thanks for fixing!

PR #3053, opened 2024-05-06
[Spark] Test type widening compatibility with other Delta features

johanl-db:
## Description
Additional tests covering type widening and:
- Reading CDF
- Column mapping
- Time travel
- RESTORE
- CLONE

## How was this patch tested?
Test only

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
@@ -20,13 +20,14 @@ import java.util.concurrent.TimeUnit
 
 import com.databricks.spark.util.Log4jUsageLogger
 import org.apache.spark.sql.delta.DeltaOperations.ManualUpdate
-import org.apache.spark.sql.delta.actions.AddFile
+import org.apache.spark.sql.delta.actions.RemoveFile
 import org.apache.spark.sql.delta.catalog.DeltaTableV2
 import org.apache.spark.sql.delta.commands.AlterTableDropFeatureDeltaCommand
+import org.apache.spark.sql.delta.commands.cdc.CDCReader
 import org.apache.spark.sql.delta.rowtracking.RowTrackingTestUtils
 import org.apache.spark.sql.delta.sources.DeltaSQLConf
 import org.apache.spark.sql.delta.test.DeltaSQLCommandTest
-import org.apache.spark.sql.delta.util.JsonUtils
+import org.apache.spark.sql.delta.util.{DeltaFileOperations, JsonUtils}
 import org.apache.hadoop.fs.Path
 
 import org.apache.spark.{SparkConf, SparkException}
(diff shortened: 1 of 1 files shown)


spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
```diff
@@ -505,6 +506,84 @@ trait DeltaTypeWideningNestedFieldsTests {
...
+    assert(readDeltaTable(tempPath).schema("b").dataType === ShortType)
+    checkAnswer(sql(s"SELECT b FROM delta.`$tempPath`"), Seq(Row(1), Row(2)))
+    append(Seq(3).toDF("b").select($"b".cast(ShortType)))
+    checkAnswer(sql(s"SELECT b FROM delta.`$tempPath`"), Seq(Row(1), Row(2), Row(3)))
+
+    // Change column type again, add more data.
+    sql(s"ALTER TABLE delta.`$tempPath` CHANGE COLUMN b TYPE int")
+    assert(readDeltaTable(tempPath).schema("b").dataType === IntegerType)
+    append(Seq(4).toDF("b"))
+    checkAnswer(sql(s"SELECT b FROM delta.`$tempPath`"), Seq(Row(1), Row(2), Row(3), Row(4)))
```
tomvanbussel, line 568: Do we also have a test for dropping the feature with a column that was renamed after its type was changed?

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
```diff
@@ -1252,6 +1399,50 @@ trait DeltaTypeWideningTableFeatureTests {
...
+  test("dropping feature after RESTORE correctly rewrite files with old type") {
+    addSingleFile(Seq(1), ShortType)
+    addSingleFile(Seq(2), ShortType)
+    sql(s"ALTER TABLE delta.`$tempPath` CHANGE COLUMN a TYPE INT")
+    // Delete the first file which will be added back again with RESTORE.
+    sql(s"DELETE FROM delta.`$tempPath` WHERE a = 1")
+    addSingleFile(Seq(Int.MinValue), IntegerType)
+
+    // RESTORE to version 2 -> ALTER TABLE CHANGE COLUMN TYPE.
+    // The type change is then still present and the first file initially added at version 0 is
```
tomvanbussel, line 1433: What will happen when restoring to before the type change? We should still require history truncation in this case.

tomvanbussel (commented):
Added a few suggestions for additional tests. Do we already have tests for constraints and generated columns. AFAIK we disallow type changes on generated columns, but constraints might be interesting as it's possible to create a constraint after changing the type.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
```diff
@@ -505,6 +507,84 @@ trait DeltaTypeWideningNestedFieldsTests {
...
+      parameters = Map(
+        "start" -> "1",
+        "end" -> "1",
+        "readSchema" -> ".*",
+        "readVersion" -> "3",
+        "incompatibleVersion" -> "1",
+        "config" -> ".*defaultSchemaModeForColumnMappingTable"
+      )
+    )
+    checkAnswer(readCDF(start = 3, end = 3), Seq(Row(3, "insert"), Row(4, "insert")))
```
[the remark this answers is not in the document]
johanl-db: You can directly provide a snapshot to use to get the read schema, I added a test for this.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTypeWideningSuite.scala
johanl-db: I added a test to cover this. Dropping the feature will indeed fail as historical versions still contain the feature, asking to wait for the retention period to expire

johanl-db:
> Added a few suggestions for additional tests. Do we already have tests for constraints and generated columns. AFAIK we disallow type changes on generated columns, but constraints might be interesting as it's possible to create a constraint after changing the type.

There are tests already for generated columns and constraints, added as part of https://github.com/delta-io/delta/pull/2881

I added a test here to cover constraints + RESTORE which is an interesting case:
1. Change column `a` type from `byte` to `int`
2. Add a CHECK constraint on column `a`.
3. RESTORE table to before the type change.

The constraint was added on type `int` and the column now has type `byte`. This does work though because CHECK constraints are part of the table metadata and it gets removed during RESTORE.

tomvanbussel (approved):
LGTM!

PR #3056, opened 2024-05-06
[INFRA] Move checkstyle config files to one location

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [X] Other (INFRA)

## Description

Currently the java checkstyle config files are shared by some projects and stored in different locations nested under those projects. Moves the checkstyle config files to a common location. This will make it easier to compare our configs in the future and keep track of the different settings for the different projects.

## How was this patch tested?

Locally ran `checkstyle` for all the affected projects. Also modified one file and checked that the checks failed.


kernel/dev/checkstyle-suppressions.xml
scottsand-db, line 1: why delete this?

scottsand-db: because its empty?

scottsand-db (approved):

PR #3059, opened 2024-05-07
fix hanging behavior for constructDataFilters with expression like EqualTo(Literal, Literal) 

ami7o:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
This fixes the hanging behavior for `constructDataFilters` with expressions like `EqualTo(Literal, Literal)`. This is due to infinite recursion caused by cases like `case EqualTo(v: Literal, a) => constructDataFilters(EqualTo(a, v))`.  This also fixes the same behavior for `EqualTo, Not(EqualTo), EqualNullSafe, LessThan, LessThanOrEqual, GreaterThan, GreaterThanOrEqual`

## How was this patch tested?

new UT `DataSkippingDeltaConstructDataFiltersSuite`

No


imback82 (approved):
LGTM!

PR #3063, opened 2024-05-07
[PROTOCOL] Fix inline DV example in PROTOCOL.md

nicklan:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [X] Other (protocol.md fix)

## Description
The previous example was incorrect. It had the magic encoded in _big-endian_, but it should be little-endian. This fixes it. 

To see that the previous one was incorrect, you can go to https://www.dcode.fr/ascii-85-encoding, and decode `wi5b=000010000siXQKl0rr91000f55c8Xg0@@D72lkbi5=-{L` (be sure to select the zeromq encoding). The first 4 bytes of this decode to `0x64, 0x39, 0xD3, 0xD0`.  This is `1681511376` (the correct magic), but only if you decode as Big endian. 

The new value decodes to `0xD1 0xD3 0x39 0x64`, which is 1681511377 in little Endian. 

## How was this patch tested?

Tested in [this pr](https://github.com/delta-incubator/delta-kernel-rs/pull/201/) in the rust kernel

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -887,8 +887,8 @@ Assuming that this DV is stored relative to an `s3://mytable/` directory, the ab
 ```json
 {
   "storageType" : "i",
-  "pathOrInlineDv" : "wi5b=000010000siXQKl0rr91000f55c8Xg0@@D72lkbi5=-{L",
-  "sizeInBytes" : 40,
+  "pathOrInlineDv" : "^Bg9^0rr910000000000iXQKl0rr91000f55c8Xg0@@D72lkbi5=-{L",
+  "sizeInBytes" : 44,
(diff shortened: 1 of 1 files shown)


felipepessoto:
Is this duplicated? #2050

nicklan:
> Is this duplicated? #2050
> 
> Fix #1905

Ohh yep, did not see that one. We should merge that one :)

felipepessoto:
We ended up not merging none of them :)

Could you review #2050?

felipepessoto:
@nicklan @vkorukanti the other PR is also closed without merge. Can we merge this one?

PR #3077, opened 2024-05-09
[Spark] Fix inCommitTimestamp deserialization for very small timestamps

dhruvarya-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Follow up for https://github.com/delta-io/delta/commit/2ef3a2ae887e4d79c1469c60c54b3ec993b97db0 . There is another site where we don't specify the right way to serialize inCommitTimestamp which will cause failures when deserializing very small timestamps.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/Checksum.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/Checksum.scala
@@ -30,6 +30,7 @@ import org.apache.spark.sql.delta.sources.DeltaSQLConf
 import org.apache.spark.sql.delta.stats.FileSizeHistogram
 import org.apache.spark.sql.delta.storage.LogStore
 import org.apache.spark.sql.delta.util.{FileNames, JsonUtils}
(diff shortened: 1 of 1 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/Checksum.scala
```diff
@@ -54,6 +54,7 @@ case class VersionChecksum(
     numFiles: Long,
     numMetadata: Long,
     numProtocol: Long,
+    @JsonDeserialize(contentAs = classOf[Long])
```
junlee-db, line 57: need `import com.fasterxml.jackson.databind.annotation.JsonDeserialize`?

dhruvarya-db: Thanks for catching that! Fixed.

junlee-db (approved):

Issue #3081, opened 2024-05-10
[Feature Request][Spark] Optimize automated batching


Kimahriman:
#### Which Delta project/connector is this regarding?

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

### Overview

Currently optimize is an all or nothing operation on all files in the table, or limited by a partition filter. The partition filter allows you to do manually batching of subsets of the table, but with clustering now a thing, there is no option to do partition filtering. We should add the ability to enable batch support inside of optimize, so chunks of optimized files can be added to the transaction log incrementally.

### Motivation

Currently you could rewrite an entire petabyte of data, just to fail on the last file and have all that be for naught, wasting a lot of compute time and storage space. With automatic batching, nearly all of the results would be saved along the way, and only the last batch that failed would have to be retried.

### Further details

I think this can be fairly straightforward, just grouping the existing bins into another layer of batches.

### Willingness to contribute

The Delta Lake Community encourages new feature contributions. Would you or another member of your organization be willing to contribute an implementation of this feature?

- [x] Yes. I can contribute this feature independently.
- [ ] Yes. I would be willing to contribute this feature with guidance from the Delta Lake community.
- [ ] No. I cannot contribute this feature at this time.

Kimahriman:
I think this is already a thing in Databricks, so it would be great to know if there are any plans to open-source that before I spend a bunch of time on this! @scottsand-db

felipepessoto:
Adding a DRY RUN option could also help. If you know there are many files to optimize you can plan accordingly.

Kimahriman:
> Adding a DRY RUN option could also help. If you know there are many files to optimize you can plan accordingly.

Probably good for a separate issue/follow on. I don't really have that use case nor want to bloat the current PR

PR #3089, opened 2024-05-14
[Spark] Optimize batching / incremental progress

Kimahriman:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Resolves https://github.com/delta-io/delta/issues/3081

Adds support for splitting an optimize run into batches with a new config `spark.databricks.delta.optimize.batchSize`. Batches will be created by grouping existing bins into groups until `batchSize` is reached. The default behavior remains the same, and batching is only enabled if the `batchSize` is configured.

This will apply to all optimization paths. I don't see any reason it shouldn't apply to to compaction, z-ordering, clustering, auto-compaction, or reorg/DV rewriting if a user configures it.

The way transactions are handled within the optimize executor had to be updated. Instead of creating a transaction upfront, we list all the files in the most recent snapshot, and then create transactions for each batch.

This is very important to add for clustering, as there is no way to manually do a partial set of the table using partition filtering. This could cause a lot of execution time and storage space to be wasted if something fails before optimizing the entire table finishes.

## How was this patch tested?

A simple new UT is added. I can add others as well, just looking for some feedback on the approach and suggestions of what other tests to add.

Yes, adds new capability to optimization that is disabled by default.


spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala
```diff
@@ -401,6 +391,90 @@ class OptimizeExecutor(
...
+
+  private def runOptimizeBatch(
+    bins: Seq[(Map[String, String], Seq[AddFile])],
+    maxFileSize: Long
+  ): (Seq[AddFile], Seq[RemoveFile], Seq[DeletionVectorDescriptor]) = {
+    val txn = snapshot.deltaLog.startTransaction(catalogTable, Some(snapshot))
+
+    val filesToProcess = bins.flatMap(_._2)
+
+    txn.trackFilesRead(filesToProcess)
```
Kimahriman, line 436: This should be the only difference with the existing default behavior. Only the filtered candidates are registered with the transaction, not all matching files.

spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala
```diff
@@ -401,6 +391,90 @@ class OptimizeExecutor(
...
+    txn.trackReadPredicates(partitionPredicate)
```
Kimahriman, line 437: I assume the partition predicate should still be registered even if it's just a subset of the partition that's being processed. This is already what's happening anyway with the candidates being filtered

spark/src/test/scala/org/apache/spark/sql/delta/optimize/OptimizeCompactionSuite.scala
```diff
@@ -536,6 +536,39 @@ trait OptimizeCompactionSuiteBase extends QueryTest
...
+
+          val df = spark.read.format("delta").load(tempDir.getAbsolutePath)
+          val deltaLog = loadDeltaLog(tempDir.getAbsolutePath)
+          val part = "part".phy(deltaLog)
+          val files = groupInputFilesByPartition(df.inputFiles, deltaLog)
+          // All partitions should be one file now
+          assert(files.values.forall(_.length == 1))
+          // The last 5 commits in the history should be optimize batches, one for each partition
+          val commits = deltaLog.history.getHistory(None)
+          assert(commits.filter(_.operation == "OPTIMIZE").length == optimizeCommits)
```
dabao521, line 566: Can you add a check the data before and after OPTIMIZE should be the same?

spark/src/test/scala/org/apache/spark/sql/delta/optimize/OptimizeCompactionSuite.scala
```diff
@@ -536,6 +536,39 @@ trait OptimizeCompactionSuiteBase extends QueryTest
     }
   }
 
+  test("optimize command with batching") {
```
dabao521, line 539: Can you add more tests to cover:
1. `OPTIMIZE WHERE` for patititioned table. This makes sure the batching works correctly with the filter.
2. Since this change also impacts  `zorder by` and `cluster by`, we need to add tests for both of them to validate batching works as expected.
3. OPTIMIZE on an empty table. Make sure it doens't trigger any divide by zero errors.

dabao521 (commented):
Thanks for the change, I left a few comments and we need more tests for this change.

spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala
```diff
@@ -61,17 +62,17 @@ abstract class OptimizeTableCommandBase extends RunnableCommand with DeltaComman
    */
   def validateZorderByColumns(
       spark: SparkSession,
-      txn: OptimisticTransaction,
```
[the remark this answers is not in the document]
Kimahriman: Updated

spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala
```diff
@@ -273,34 +280,17 @@ class OptimizeExecutor(
 
       val maxThreads =
```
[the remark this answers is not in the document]
Kimahriman: Yep good catch

spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala
```diff
@@ -273,34 +280,17 @@ class OptimizeExecutor(
...
-          } else {
-            val deleted = candidateSetOld -- candidateSetNew
-            logWarning(s"The following compacted files were delete " +
-              s"during checkpoint ${deleted.mkString(",")}. Aborting the compaction.")
-            false
-          }
-        }
+      val batchResults = batchSize match {
+        case Some(size) =>
+          groupBinsIntoBatches(jobs, size).map(runOptimizeBatch(_, maxFileSize))
```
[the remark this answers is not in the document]
Kimahriman: Hmm didn't look at the ConflictChecker when working on this, I did have in my mind an improvement to doing multiple batches simultaneously/overlapping to prevent the tail of execution for each batch. There's no conceptual reason the commits should conflict, since they are reading specific files and not changing data. If they would conflict that might just be an improvement that should be made in the ConflictChecker

spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala
```diff
@@ -309,10 +299,10 @@ class OptimizeExecutor(
       optimizeStats.totalConsideredFiles = candidateFiles.size
       optimizeStats.totalFilesSkipped = optimizeStats.totalConsideredFiles - removedFiles.size
       optimizeStats.totalClusterParallelism = sparkSession.sparkContext.defaultParallelism
-      val numTableColumns = txn.snapshot.metadata.schema.size
+      val numTableColumns = snapshot.metadata.schema.size
```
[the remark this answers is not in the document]
Kimahriman: Yeah that makes sense, I also kept a numBatches to actually represent the number of batches? Or is that bad because it changes the meaning of the existing stat

spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala
```diff
@@ -401,6 +391,90 @@ class OptimizeExecutor(
...
+   * Utility methods to group bins into batches for incremental progress.
+   *
+   * @param bins List of bins to group into batches
+   * @return Sequence of batches. Each batch contains one or more bins.
+   */
+  private def groupBinsIntoBatches(
+      bins: Seq[(Map[String, String], Seq[AddFile])],
+      batchSize: Long)
+  : Seq[Seq[(Map[String, String], Seq[AddFile])]] = {
+    val batches = new ArrayBuffer[Seq[(Map[String, String], Seq[AddFile])]]()
```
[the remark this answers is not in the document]
Kimahriman: Yep thought these were getting verbose. Added case classes.

spark/src/test/scala/org/apache/spark/sql/delta/optimize/OptimizeCompactionSuite.scala
Kimahriman: Will do

spark/src/test/scala/org/apache/spark/sql/delta/optimize/OptimizeCompactionSuite.scala
Kimahriman: Thanks I'll work on those

spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala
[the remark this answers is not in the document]
Kimahriman: Though since I'm creating a transaction off the same original snapshot, even doing them serially they should invoke the same conflict checking against themselves right? And the current simple test doesn't seem to have an issue. Will see if I encounter anything in further tests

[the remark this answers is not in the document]
Kimahriman: Actually I guess the Optimize command does it's own conflict checking to resolve that

spark/src/test/scala/org/apache/spark/sql/delta/optimize/OptimizeCompactionSuite.scala
Kimahriman: Added tests for all the above

spark/src/test/scala/org/apache/spark/sql/delta/optimize/OptimizeCompactionSuite.scala
Kimahriman: Add data checks

spark/src/main/scala/org/apache/spark/sql/delta/hooks/AutoCompact.scala
```diff
@@ -195,15 +195,14 @@ trait AutoCompactBase extends PostCommitHook with DeltaLogging {
...
       val optimizeContext = DeltaOptimizeContext(
         reorg = None,
         minFileSizeOpt,
         maxFileSizeOpt,
         maxDeletedRowsRatio = maxDeletedRowsRatio
       )
-      val rows = new OptimizeExecutor(spark, txn, partitionPredicates, Seq(), true, optimizeContext)
-        .optimize()
+      val rows = new OptimizeExecutor(spark, deltaLog.update(), catalogTable, partitionPredicates,
+        Seq(), true, optimizeContext).optimize()
```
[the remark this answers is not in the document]
Kimahriman: Updated

spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala
```diff
@@ -1168,6 +1168,19 @@ trait DeltaSQLConfBase {
         .checkValue(_ > 0, "'optimize.maxThreads' must be positive.")
         .createWithDefault(15)
 
+  val DELTA_OPTIMIZE_BATCH_SIZE =
+    buildConf("optimize.batchSize")
+        .internal()
+        .doc(
+          """
+            |The size of a batch within an OPTIMIZE JOB. After a batch is complete, it's
```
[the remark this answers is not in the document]
Kimahriman: Fixed

Kimahriman:
I think I've addressed everything, let me know how it looks

spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala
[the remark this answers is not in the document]
dabao521: hmm, looks like the comment hasn't been updated at line 60?

spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala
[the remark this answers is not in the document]
dabao521: Yea, that makes sense, adding `numBins` is better

spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala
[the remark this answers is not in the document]
dabao521: Yes, there isn't conflict within a OPITMIZE command since commits from different batches are working on disjoint set of files as you mentioned.  It is more of the conflicts between OPTIMIZE commands.  Say if batch1 from OPTIMIZE A runs concurrently with batch1 from OPTIMIZE B, and the batch mixes multiple partition files, those two batches are likely conflicting since there is high chance files are overlapped in some partition. Had if batches are at partition boundaries,  conflict would happen less likely.  This is due to Delta enhanced conflict resolution for partitioned table.
This won't eliminate conflict between OPTIMIZEs, but more of a optimization. I don't see it is needed now and we can reivist if there is one.

spark/src/test/scala/org/apache/spark/sql/delta/optimize/OptimizeCompactionSuite.scala
```diff
@@ -536,6 +537,94 @@ trait OptimizeCompactionSuiteBase extends QueryTest
...
+
+        val df = spark.read.format("delta").load(tempDir.getAbsolutePath)
+        checkAnswer(df, data)
+
+        val deltaLog = loadDeltaLog(tempDir.getAbsolutePath)
+
+        val commits = deltaLog.history.getHistory(None)
+        assert(commits.filter(_.operation == "OPTIMIZE").length == expectedCommits)
+
+        val part = "part".phy(deltaLog)
```
dabao521, line 572: unused variable?

spark/src/test/scala/org/apache/spark/sql/delta/optimize/OptimizeZOrderSuite.scala
```diff
@@ -193,6 +194,94 @@ trait OptimizeZOrderSuiteBase extends OptimizePartitionTableHelper
     }
   }
 
+  def optimizeWithBatching(
```
dabao521, line 197: Is this function same as the one in spark/src/test/scala/org/apache/spark/sql/delta/optimize/OptimizeCompactionSuite.scala? Can we refactor to share the implemnetation?

dabao521 (commented):
Hi @Kimahriman, Thanks for the change and I left a few more minor comments and overrall it looks good to me

spark/src/test/scala/org/apache/spark/sql/delta/optimize/OptimizeZOrderSuite.scala
Kimahriman: They're the same except for the `executeOptimizePath` call which has a z-order col here. I could try to refactor into a shared function or base class, but it would be a little tricky having to use the different `executeOptimizePath` calls, and the two suites don't currently share any other base optimize related base class. Might just be overkill, but I can try if you really think it's worth it.

spark/src/test/scala/org/apache/spark/sql/delta/optimize/OptimizeCompactionSuite.scala
Kimahriman: removed (and in z-order suite)

spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala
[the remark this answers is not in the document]
Kimahriman: Whoops only updated `OptimizeExecutor` docs, not this one. Fixed

dabao521 (approved):

spark/src/test/scala/org/apache/spark/sql/delta/optimize/OptimizeZOrderSuite.scala
Kimahriman: Made https://github.com/delta-io/delta/issues/3249

spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala
```diff
@@ -400,8 +393,93 @@ class OptimizeExecutor(
...
+  private def groupBinsIntoBatches(
+      bins: Seq[Bin],
+      batchSize: Long)
+  : Seq[Batch] = {
+    val batches = new ArrayBuffer[Batch]()
+
+    val currentBatch = new ArrayBuffer[Bin]()
+    var currentBatchSize = 0L
+
+    bins.foreach { bin =>
```
zedtang, line 415: We have similar bin packing logic in [BinPackingUtils](https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/util/BinPackingUtils.scala), can we reuse it?

spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala
```diff
@@ -1226,6 +1226,19 @@ trait DeltaSQLConfBase {
...
+    buildConf("optimize.batchSize")
+        .internal()
+        .doc(
+          """
+            |The size of a batch within an OPTIMIZE JOB. After a batch is complete, its
+            | progress will be committed to the transaction log, allowing for incremental
+            | progress.
+            |""".stripMargin)
+        .bytesConf(ByteUnit.BYTE)
+        .checkValue(_ >= 0, "batchSize has to be positive")
```
zedtang, line 1239: Nit 
```suggestion
        .checkValue(_ > 0, "batchSize has to be positive")
```

spark/src/test/scala/org/apache/spark/sql/delta/clustering/ClusteredTableClusteringSuite.scala
```diff
@@ -91,7 +92,7 @@ class ClusteredTableClusteringSuite extends SparkFunSuite
         assert(files0.size === 4)
         assertNotClustered(files0)
 
-        // Optimize should cluster the data into two 2 files since MAX_RECORDS_PER_FILE is 2.
+         // Optimize should cluster the data into two 2 files since MAX_RECORDS_PER_FILE is 2.
```
zedtang, line 95: nit: revert

spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala
```diff
@@ -142,12 +143,12 @@ case class OptimizeTableCommand(
 
   override def run(sparkSession: SparkSession): Seq[Row] = {
     val table = getDeltaTable(child, "OPTIMIZE")
-    val txn = table.startTransaction()
```
zedtang, line 145: This is a breaking change - previously we hold a transaction before reading the snapshot, which will detect any concurrent updates ([pointer](https://github.com/delta-io/delta/blob/a24a857d27bbb4b3585814c35fa3002cee929fec/spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala#L181)). With this change, we could be potentially reading a stale snapshot.

I think we should guard this change behind the config you introduced. We can have a wrapper class that contains either the `txn` or the `snapshot` depending on the config and pass it down.

zedtang: nvm, I misread the code 😅 as long as later we create a txn based on this snapshot, we should be good. During commit time we try to commit with snapshot.readVersion + 1 so it should see any updates.

spark/src/test/scala/org/apache/spark/sql/delta/clustering/ClusteredTableClusteringSuite.scala
Kimahriman: sorry, done

spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala
Kimahriman: fixed

spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala
Kimahriman: Hmm yeah I think so. The only difference is it emits a bin before it reaches the maxBinSize, where I currently emit a bit right after I reach it, but I don't think it matters that much either way

Kimahriman: Replaced the function with that and just inlined it above

zedtang (approved):
LGTM!

spark/src/main/scala/org/apache/spark/sql/delta/commands/OptimizeTableCommand.scala
```diff
@@ -207,17 +210,22 @@ case class DeltaOptimizeContext(
   }
 }
 
+case class Bin(partitionValues: Map[String, String], files: Seq[AddFile])
```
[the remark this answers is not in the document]
Kimahriman: Added

PR #3105, opened 2024-05-16
[Spark] Handle case when Checkpoints.findLastCompleteCheckpoint is passed MAX_VALUE

prakharjain09:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Fixes an issue where `Checkpoints.findLastCompleteCheckpoint` goes into an almost infinite loop if it is passed a Checkpoint.MAX_VALUE.

## How was this patch tested?

UT

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/Checkpoints.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/Checkpoints.scala
@@ -429,10 +429,46 @@ trait Checkpoints extends DeltaLogging {
    */
   private[delta] def findLastCompleteCheckpointBefore(
       checkpointInstance: Option[CheckpointInstance] = None): Option[CheckpointInstance] = {
-    val upperBoundCv = checkpointInstance.filterNot(_.version < 0).getOrElse {
-        logInfo(s"Try to find Delta last complete checkpoint")
-        return findLastCompleteCheckpoint()
-      }
+    val eventData = mutable.Map[String, String]()
+    val startTimeMs = System.currentTimeMillis()
+    def sendUsageLog(): Unit = {
+      eventData("totalTimeTakenMs") = (System.currentTimeMillis() - startTimeMs).toString
+      recordDeltaEvent(
+        self, opType = "delta.findLastCompleteCheckpointBefore", data = eventData.toMap)
+    }
+    try {
+      val resultOpt = findLastCompleteCheckpointBeforeInternal(eventData, checkpointInstance)
+      eventData("resultantCheckpointVersion") = resultOpt.map(_.version).getOrElse(-1L).toString
+      sendUsageLog()
+      resultOpt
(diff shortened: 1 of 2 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/Checkpoints.scala
```diff
@@ -429,10 +429,46 @@ trait Checkpoints extends DeltaLogging {
...
+
+  private def findLastCompleteCheckpointBeforeInternal(
+      eventData: mutable.Map[String, String],
+      checkpointInstance: Option[CheckpointInstance]): Option[CheckpointInstance] = {
+    val upperBoundCv =
+      checkpointInstance
+        // If someone passes the upperBound as 0 or sentinel value, we should not do backward
+        // listing. Instead we should list the entire directory from 0 and return the latest
+        // available checkpoint.
+        .filterNot(cv => cv.version < 0 || cv.version == CheckpointInstance.MaxValue.version)
```
prakharjain09, line 461: This is the fix and there is a corresponding test: `findLastCompleteCheckpoint with CheckpointInstance.MAX value` .

spark/src/test/scala/org/apache/spark/sql/delta/FindLastCompleteCheckpointSuite.scala
```diff
@@ -117,11 +139,15 @@ class FindLastCompleteCheckpointSuite
...
-          .contains(CheckpointInstance(version = 1000)))
+      val eventData1 = getLastCompleteCheckpointUsageLog {
+        assert(
+          log.findLastCompleteCheckpointBefore(Some(CheckpointInstance(version = 2000)))
+            .contains(CheckpointInstance(version = 1000)))
+      }
       assert(logStore.listFromCount == 1)
       assert(logStore.elementsConsumedFromListFromIter == 1002 + 2) // commits + checkpoint
+      assert(eventData1("iterations") == "1")
+      assert(eventData1("numFilesScanned") == "1004")
```
scovich, line 150: Does this work for better error messages?
```suggestion
      assert(eventData1("iterations") === "1")
      assert(eventData1("numFilesScanned") === "1004")
```

(bunch of others, including existing code... maybe worth a cleanup PR some day)

scovich (approved):
Nice catch

tdas (approved):

spark/src/test/scala/org/apache/spark/sql/delta/FindLastCompleteCheckpointSuite.scala
prakharjain09: For consistency in the test suite, I am keeping these as `==`. Will open a PR which converts all `==` to `===`.

PR #339, opened 2020-03-03
DynamoDBLogStore

mrk-its:
This PR addresses issue #41 - Support for AWS S3 (multiple clusters/drivers/JVMs)

It implements few ideas from #41 discussion:
  - provides generic base class BaseExternalLogStore for storing listing of commit files
    in external DB. This class may be easly extended for specific DB backend
  - stores contents of commit in temporary file and links to it in DB's row
    to be able to finish uncompleted write operation while reading
  - provides concrete DynamoDBLogStore implementation extending BaseExternalLogStore
  - implementations for other DB backends should be simple to implement
    (ZooKeeper implementation is almost ready, I can create separate PR if anyone is interested)

## DynamoDBLogStore requirements:

To enable DynamoDBLogStore set following spark property:
`
spark.delta.logStore.class=io.delta.storage.DynamoDBLogStore
`

Single dynamodb table is required. Default table name is 'delta_log',
it may be changed by setting spark property.

Required key schema:
  - parentPath: String, HASH
  - filename: String, RANGE

Table may be created with following aws cli command:

```
aws --region us-east-1 dynamodb create-table \
    --table-name delta_log \
    --attribute-definitions AttributeName=tablePath,AttributeType=S \
                            AttributeName=fileName,AttributeType=S \
    --key-schema AttributeName=tablePath,KeyType=HASH \
                AttributeName=fileName,KeyType=RANGE \
    --provisioned-throughput ReadCapacityUnits=5,WriteCapacityUnits=5
```

Following spark properties are recognized:
- spark.delta.DynamoDBLogStore.tableName - table name (defaults to 'delta_log')
- spark.delta.DynamoDBLogStore.region - AWS region (defaults to 'us-east-1')

## Testing

Python integration test is included :examples/python/dynamodb_logstore.py

This solution has been also stress-tested on Amazon's EMR cluster 
(mutiple test jobs writing thousands of parallel transactions to single delta table)
and no data loss has beed observed so far

PR #3107, opened 2024-05-16
[Spark] Managed Commits: add a DynamoDB-based commit owner

dhruvarya-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Taking inspiration from https://github.com/delta-io/delta/pull/339, this PR adds a Commit Owner Client which uses DynamoDB as the backend. Each Delta table managed by a DynamoDB instance will have one corresponding entry in a DynamoDB table. The table schema is as follows:

* tableId: String --- The unique identifier for the entry. This is a UUID.
* path: String --- The fully qualified path of the table in the file system. e.g. s3://bucket/path.
* acceptingCommits: Boolean --- Whether the commit owner is accepting new commits. This will only
*  be set to false when the table is converted from managed commits to file system commits.
* tableVersion: Number --- The version of the latest commit.
* tableTimestamp: Number --- The inCommitTimestamp of the latest commit.
* schemaVersion: Number --- The version of the schema used to store the data.
* commits: --- The list of unbackfilled commits.
  -  version: Number --- The version of the commit.
  -  inCommitTimestamp: Number --- The inCommitTimestamp of the commit.
  -  fsName: String --- The name of the unbackfilled file.
  -  fsLength: Number --- The length of the unbackfilled file.
  -  fsTimestamp: Number --- The modification time of the unbackfilled file.

For a table to be managed by DynamoDB, `registerTable` must be called for that Delta table. This will create a new entry in the db for this Delta table. Every `commit` invocation appends the UUID delta file status to the `commits` list in the table entry. `commit` is performed through a conditional write in DynamoDB.

## How was this patch tested?

Added a new suite called `DynamoDBCommitOwnerClient5BackfillSuite` which uses a mock DynamoDB client. + plus manual testing against a DynamoDB instance.

### DIFF
--- a/spark/src/main/java/io/delta/dynamodbcommitstore/DynamoDBCommitOwnerClient.java
+++ b/spark/src/main/java/io/delta/dynamodbcommitstore/DynamoDBCommitOwnerClient.java
@@ -1,0 +1,681 @@
+/*
+ * Copyright (2021) The Delta Lake Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.delta.dynamodbcommitstore;
+
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDB;
+import com.amazonaws.services.dynamodbv2.model.*;
+import org.apache.spark.sql.delta.managedcommit.*;
+import org.apache.spark.sql.delta.storage.LogStore;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+
+import scala.Option;
+import scala.Tuple2;
+import scala.collection.Iterator;
+import scala.collection.immutable.Map;
+import scala.collection.JavaConverters;
+import scala.collection.immutable.Map$;
+
+import java.io.*;
+import java.nio.file.FileAlreadyExistsException;
+import java.util.ArrayList;
+import java.util.HashMap;
+
+// TODO(dynamodb):
+//  1. figure out whether we need retry logic like DeltaCommitClient
+//  ~~2. data schema versioning~~
+//  3. logging
+//  4. error handling
+//  5. local tests
+//  6. passing hadoop conf to the credential provider
+
+/**
+ * A commit owner client that uses DynamoDB as the commit owner. The table schema is as follows:
+ * tableId: String --- The unique identifier for the table. This is a UUID.
+ * path: String --- The fully qualified path of the table in the file system. e.g. s3://bucket/path.
+ * acceptingCommits: Boolean --- Whether the commit owner is accepting new commits. This will only
+ *  be set to false when the table is converted from managed commits to file system commits.
+ * tableVersion: Number --- The version of the latest commit.
+ * tableTimestamp: Number --- The inCommitTimestamp of the latest commit.
+ * schemaVersion: Number --- The version of the schema used to store the data.
+ * commits: --- The list of unbackfilled commits.
+ *  version: Number --- The version of the commit.
+ *  inCommitTimestamp: Number --- The inCommitTimestamp of the commit.
+ *  fsName: String --- The name of the unbackfilled file.
+ *  fsLength: Number --- The length of the unbackfilled file.
+ *  fsTimestamp: Number --- The modification time of the unbackfilled file.
+ */
+public class DynamoDBCommitOwnerClient implements CommitOwnerClient {
+
+    /**
+     * The name of the DynamoDB table used to store unbackfilled commits.
+     */
+    final String managedCommitsTableName;
+
+    /**
+     * The DynamoDB client used to interact with the DynamoDB table.
+     */
+    final AmazonDynamoDB client;
+
+    /**
+     * The endpoint of the DynamoDB table.
+     */
+    final String endpoint;
+
+    /**
+     * The number of write capacity units to provision for the DynamoDB table if the
+     * client ends up creating a new one.
+     */
+    final long writeCapacityUnits;
+
+    /**
+     * The number of read capacity units to provision for the DynamoDB table if the
+     * client ends up creating a new one.
+     */
+    final long readCapacityUnits;
+
+    /**
+     * The number of commits to batch backfill at once. A backfill is performed
+     * whenever commitVersion % batchSize == 0.
+     */
+    public final long backfillBatchSize;
+
+    /**
+     * Whether we should skip matching the current table path against the one stored in DynamoDB
+     * when interacting with it.
+     */
+    final boolean skipPathCheck;
+
+    /**
+     * The key used to store the tableId in the managed commit table configuration.
+     */
+    final static String TABLE_CONF_TABLE_ID_KEY = "tableId";
+
+    /**
+     * The version of the client. This is used to ensure that the client is compatible with the
+     * schema of the data stored in the DynamoDB table. A client should only be able to
+     * access a table if the schema version of the table matches the client version.
+     */
+    final int CLIENT_VERSION = 1;
+
+
+    public DynamoDBCommitOwnerClient(
+            String managedCommitsTableName,
+            String endpoint,
+            AmazonDynamoDB client,
+            long backfillBatchSize) throws IOException {
+        this(
+            managedCommitsTableName,
+            endpoint,
+            client,
+            backfillBatchSize,
+            5 /* readCapacityUnits */,
+            5 /* writeCapacityUnits */,
+            false /* skipPathCheck */);
+    }
+
+    public DynamoDBCommitOwnerClient(
+            String managedCommitsTableName,
+            String endpoint,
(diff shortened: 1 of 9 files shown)


spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/CommitOwnerClientImplSuiteBase.scala
```diff
@@ -198,15 +198,19 @@ trait CommitOwnerClientImplSuiteBase extends QueryTest
...
 
       val e = intercept[CommitFailedException] {
         commit(version = 0, timestamp = 0, tableCommitOwnerClient)
       }
       assert(e.getMessage === "Commit version 0 must go via filesystem.")
       writeCommitZero(logPath)
-      assert(tableCommitOwnerClient.getCommits() == GetCommitsResponse(Seq.empty, -1))
+      var expectedVersion = -1
+      if (tableCommitOwnerClient.commitOwnerClient.isInstanceOf[DynamoDBCommitOwnerClient]) {
+        // DynamoDBCommitOwnerClient stores attemptVersion as the current table version when
```
prakharjain09, line 209: `DynamoDBCommitOwnerClient.getCommits` starts advertising `attemptVersion` as the current commit version even without knowledge around whether the commit has gone through or not. Practically this won't cause any issue because until the commit is done, no client will invoke `DynamoDBCommitOwnerClient.getCommits` since the table was a FS table. But just for the sake of semantics, we should fix it in a followup PR.

spark/src/main/java/io/delta/dynamodbcommitstore/DynamoDBCommitOwnerClient.java
```diff
@@ -0,0 +1,681 @@
...
+            long backfillBatchSize) throws IOException {
+        this(
+            managedCommitsTableName,
+            endpoint,
+            client,
+            backfillBatchSize,
+            5 /* readCapacityUnits */,
+            5 /* writeCapacityUnits */,
+            false /* skipPathCheck */);
+        tryEnsureTableExists();
```
prakharjain09, line 130: We are calling the other constructor here so it will take care of invoking tryEnsureTableExists ?

spark/src/main/java/io/delta/dynamodbcommitstore/DynamoDBCommitOwnerClient.java
```diff
@@ -0,0 +1,681 @@
...
+                                "attribute_not_exists(%s)", DynamoDBTableEntryConstants.TABLE_ID));
+        client.putItem(request);
+
+        Tuple2<String, String> tableIdEntry = new Tuple2<>(DynamoDBTableEntryConstants.TABLE_ID, tableId);
+        Map<String, String> tableConf = Map$.MODULE$.<String, String>empty().$plus(tableIdEntry);
+
+        return tableConf;
+    }
+
+    // Copied from DynamoDbLogStore. TODO: add the logging back.
```
prakharjain09, line 618: What logging needs to be added back?

spark/src/main/java/io/delta/dynamodbcommitstore/DynamoDBCommitOwnerClientBuilder.java
```diff
@@ -0,0 +1,72 @@
...
+    // TODO: update this interface so that it can take a sparkSession.
+    @Override
+    public CommitOwnerClient build(Map<String, String> conf) {
+        String managedCommitsTableName = conf.get(MANAGED_COMMITS_TABLE_NAME_KEY).getOrElse(() -> {
+            throw new RuntimeException(MANAGED_COMMITS_TABLE_NAME_KEY + " not found");
+        });
+        String dynamoDBEndpoint = conf.get(DYNAMO_DB_ENDPOINT_KEY).getOrElse(() -> {
+            throw new RuntimeException(DYNAMO_DB_ENDPOINT_KEY + " not found");
+        });
+        String awsCredentialsProviderName = conf.get(AWS_CREDENTIALS_PROVIDER_KEY).getOrElse( () ->
```
prakharjain09, line 50:
```suggestion
        String awsCredentialsProviderName = conf.get(AWS_CREDENTIALS_PROVIDER_KEY).getOrElse(() ->
```

prakharjain09, line 50: Should AWS_CREDENTIALS_PROVIDER_KEY come from the commitOwnerConf from DeltaLog? Isn't this more of an authentication config which should be configured at JVM level i.e. via SparkSession.

spark/src/main/java/io/delta/dynamodbcommitstore/DynamoDBCommitOwnerClientBuilder.java
```diff
@@ -0,0 +1,72 @@
...
+    private final long BACKFILL_BATCH_SIZE = 1L;
+
+    @Override
+    public String getName() {
+        return "dynamodb";
+    }
+
+    private static final String MANAGED_COMMITS_TABLE_NAME_KEY = "managedCommitsTableName";
+    private static final String DYNAMO_DB_ENDPOINT_KEY = "dynamoDBEndpoint";
+    private static final String AWS_CREDENTIALS_PROVIDER_KEY = "awsCredentialsProvider";
```
prakharjain09, line 39: Add comments for these properties.

spark/src/main/java/io/delta/dynamodbcommitstore/DynamoDBCommitOwnerClient.java
dhruvarya-db: You are right, will fix

spark/src/main/java/io/delta/dynamodbcommitstore/DynamoDBCommitOwnerClient.java
dhruvarya-db: 1. That the table exists/does not exist
2. Table creation attempt
3. Table creation failure
I will add these and logs in other parts of this class in a follow up PR.

spark/src/main/java/io/delta/dynamodbcommitstore/DynamoDBCommitOwnerClientBuilder.java
dhruvarya-db: Yes, once the changes around passing sparkSession to the CommitOwnerBuilder have baked in, I will update this to read it directly from sparkSession. For now, I will hardcode this to the default credentials provider.

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/CommitOwnerClientImplSuiteBase.scala
dhruvarya-db: Yes, for DynamoDB, we treat registerTable as a valid commit and store the attemptVersion instead of the current Delta table version. This is not risky because the DynamoDB table entry is only treated as valid if the corresponding filesystem commit goes through (which means that the attemptVersion is in fact the Delta table version). In case of a filesystem conflict, it is expected that registerTable will be called again. However, for consistency, we can use another bool to track whether any commits have gone through DynamoDB and return -1.

spark/src/main/java/io/delta/dynamodbcommitstore/DynamoDBCommitOwnerClient.java
```diff
@@ -0,0 +1,681 @@
...
+    }
+
+    /**
+     * Writes the given actions to a file.
+     * logStore.write(overwrite=false) will throw a FileAlreadyExistsException if the file already
+     * exists. However, the scala LogStore interface does not declare this as part of the function
+     * signature. This method wraps the write method and declares the exception to ensure that the
+     * caller is aware of the exception.
+     */
+    private void writeActionsToFile(
```
LukasRupprecht, line 454: Should we call this `writeActionsToBackfilledFile`?

spark/src/main/java/io/delta/dynamodbcommitstore/DynamoDBCommitOwnerClient.java
```diff
@@ -0,0 +1,681 @@
...
+            Option<Object> lastKnownBackfilledVersion) {
+        GetCommitsResponse resp =
+                getCommits(logPath, managedCommitTableConf, lastKnownBackfilledVersion, Option.empty());
+        validateBackfilledFileExists(logPath, hadoopConf, lastKnownBackfilledVersion);
+        if (version > resp.getLatestTableVersion()) {
+            throw new IllegalArgumentException(
+                    "The requested backfill version " + version + " is greater than the latest " +
+                            "version " + resp.getLatestTableVersion() + " for the table.");
+        }
+        // If partial writes a visible in this filesystem, we should not try to overwrite existing
```
LukasRupprecht, line 510:
```suggestion
        // If partial writes are visible in this filesystem, we should not try to overwrite existing
```

spark/src/main/scala/org/apache/spark/sql/delta/managedcommit/CommitOwnerClient.scala
```diff
@@ -43,7 +43,7 @@ case class Commit(
 case class CommitFailedException(
     private val retryable: Boolean,
     private val conflict: Boolean,
-    private val message: String) extends Exception(message) {
+    private val message: String) extends RuntimeException(message) {
```
LukasRupprecht, line 46: Why is this changed needed? Isn't that just making it more restrictive?

dhruvarya-db: This is needed because the CommitOwnerClient trait does not declare that `commit` can throw `CommitFailedException` in the signature (Scala doesn't let us do that). The Java code won't compile if I try to throw an exception of type Exception if it has not been declared in the function signature. `RuntimeException`s can be thrown though.

prakharjain09 (approved):

PR #3111, opened 2024-05-17
[Spark] InCommitTimestamp: Use clock.currentTimeMillis() instead of nanoTime() in commitLarge

dhruvarya-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

We currently use NANOSECONDS.toMillis(System.nanoTime()) for generating the ICT when `commitLarge` is called. However, this usage of System.nanoTime() is not correct as it should only be used for measuring time difference, not to get an approximate wall clock time. This leads to scenarios where the ICT becomes very small (e.g. 1 Jan 1970) sometimes because some systems return a very small number when System.nanoTime() is called. This PR changes this so that clock.getCurrentTimeMillis() is used instead.

## How was this patch tested?

Added a test case to ensure that `clock.getCurrentTimeMillis()` is being used.

No

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/delta/InCommitTimestampSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/InCommitTimestampSuite.scala
@@ -163,7 +163,7 @@ class InCommitTimestampSuite
           Map.empty)
       }
       val ver1Snapshot = deltaLog.snapshot
-      val retrievedTimestamp = getInCommitTimestamp(deltaLog, 1)
+      val retrievedTimestamp = getInCommitTimestamp(deltaLog, version = 1)
       assert(ver1Snapshot.timestamp == retrievedTimestamp)
       assert(ver1Snapshot.timestamp == expectedCommit1Time)
       assert(filterUsageRecords(usageRecords, "delta.commit.large").length == 1)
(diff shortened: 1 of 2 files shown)


spark/src/test/scala/org/apache/spark/sql/delta/InCommitTimestampSuite.scala
```diff
@@ -143,6 +143,33 @@ class InCommitTimestampSuite
...
+        deltaLog.startTransaction().commitLarge(
+          spark,
+          Seq(createTestAddFile("1")).toIterator,
+          None,
+          DeltaOperations.ManualUpdate,
+          Map.empty,
+          Map.empty)
+      }
+      val ver1Snapshot = deltaLog.snapshot
+      val retrievedTimestamp = getInCommitTimestamp(deltaLog, 1)
```
prakharjain09, line 166: [nit] used named parameter.

prakharjain09 (approved):

spark/src/test/scala/org/apache/spark/sql/delta/InCommitTimestampSuite.scala
```diff
@@ -143,6 +143,33 @@ class InCommitTimestampSuite
     }
   }
 
+  test("commitLarge should use clock.currentTimeMillis() for ICT") {
```
prakharjain09, line 146: Could we have a similar test for `txn.commit` API?

PR #3113, opened 2024-05-17
[Kernel] Define more operation for writer.

horizonzy:
The kernel operation is poor, we need more operations for compatibility.


horizonzy:
@vkorukanti /cc

PR #3114, opened 2024-05-17
[Spark] Enable column mapping removal feature flag

sabir-akhadov:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Enable column mapping removal feature flag to allow user to run 
```
ALTER TABLE table_name SET TBLPROPERTIES ('delta.columnMapping.mode' = 'none')
```
or
```
ALTER TABLE table_name UNSET TBLPROPERTIES ('delta.columnMapping.mode')
```
to remove column mapping property from their table and rewrite physical files to match the logical column names.

Also allows column mapping feature to be dropped with

```
ALTER TABLE <table_name> DROP FEATURE columnMapping
```

## How was this patch tested?
existing tests

Yes

 Allows user to run 
```
ALTER TABLE table_name SET TBLPROPERTIES ('delta.columnMapping.mode' = 'none')
```
or
```
ALTER TABLE table_name UNSET TBLPROPERTIES ('delta.columnMapping.mode')
```
to remove column mapping from a Delta table.
Previously, this commands would not run successfully and would return an exception stating such an operation is prohibited.

Also allows column mapping feature to be dropped with

```
ALTER TABLE <table_name> DROP FEATURE columnMapping
```

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala
@@ -1597,7 +1597,7 @@ trait DeltaSQLConfBase {
           |If enabled, allow the column mapping to be removed from a table.
           |""".stripMargin)
       .booleanConf
-      .createWithDefault(false)
+      .createWithDefault(true)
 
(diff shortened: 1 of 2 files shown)


tomvanbussel (approved):
Nice! LGTM

PR #3117, opened 2024-05-20
[SPARK] Support predicates for stats that are not at the top level

olaky:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This refactoring adds support for nested statistics columns. So far, all statistics are keys in the stats struct in AddFiles. This PR adds support for statistics that are part of nested structs. This is a prerequisite for file skipping on collated string columns ([Protocol RFC](https://github.com/delta-io/delta/pull/3068)). Statistics for collated string columns will be wrapped in a struct keyed by the versioned collation that was used to generate them.  For example:

```
"stats": { "statsWithCollation": { "icu.en_US.72": { "minValues": { ...} } } }
```

This PR replaces statType in StatsColumn with pathToStatType, which can be used to represent a path. This way we can re-use all of the existing data skipping code without changes.

## How was this patch tested?

It is not possible to test this change without altering [statsSchema](https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala#L285). I would still like to ship this PR separately because the change is big enough in itself. There is existing test coverage for stats parsing and file skipping, but none of them uses nested statistics yet.

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingReader.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingReader.scala
@@ -69,7 +69,7 @@ case class NumRecords(numPhysicalRecords: java.lang.Long, numLogicalRecords: jav
  * `pathToColumn` is the nested name of the user column whose stats are to be accessed.
  * `columnDataType` is the data type of the column.
  */
-private [stats] case class StatsColumn private(
+private[stats] case class StatsColumn private(
     pathToStatType: Seq[String],
     pathToColumn: Seq[String])
 
--- a/spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingReader.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingReader.scala
@@ -64,12 +64,20 @@ case class NumRecords(numPhysicalRecords: java.lang.Long, numLogicalRecords: jav
  * Represents a stats column (MIN, MAX, etc) for a given (nested) user table column name. Used to
  * keep track of which stats columns a data skipping query depends on.
  *
- * The `statType` is any value accepted by `getStatsColumnOpt()` (see object `DeltaStatistics`);
+ * The `pathToStatType` is path to a stats type accepted by `getStatsColumnOpt()`
+ *  (see object `DeltaStatistics`);
  * `pathToColumn` is the nested name of the user column whose stats are to be accessed.
+ * `columnDataType` is the data type of the column.
  */
-private [stats] case class StatsColumn(
-    statType: String,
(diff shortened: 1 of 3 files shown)


olaky:
cc @vkorukanti @cstavr

spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingReader.scala
```diff
@@ -64,12 +64,20 @@ case class NumRecords(numPhysicalRecords: java.lang.Long, numLogicalRecords: jav
...
- * The `statType` is any value accepted by `getStatsColumnOpt()` (see object `DeltaStatistics`);
+ * The `pathToStatType` is path to a stats type accepted by `getStatsColumnOpt()`
+ *  (see object `DeltaStatistics`);
  * `pathToColumn` is the nested name of the user column whose stats are to be accessed.
+ * `columnDataType` is the data type of the column.
  */
-private [stats] case class StatsColumn(
-    statType: String,
-    pathToColumn: Seq[String] = Nil)
+private [stats] case class StatsColumn private(
```
cstavr, line 72:
```suggestion
private[stats] case class StatsColumn private(
```

cstavr (approved):

spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingReader.scala
```diff
@@ -603,17 +611,35 @@ trait DataSkippingReaderBase
    * Returns an expression to access the given statistics for a specific column, or None if that
    * stats column does not exist.
    *
-   * @param statType One of the fields declared by object `DeltaStatistics`
-   * @param pathToColumn The components of the nested column name to get stats for.
+   * @param pathToStatType Path components of one of the fields declared by the `DeltaStatistics`
+   *                       object. For statistics of collated strings, this path contains the
+   *                       versioned collation identifier. In all other cases the path only has one
+   *                       element. The path is in reverse order.
```
bart-samwel, line 617: Why in reverse order? It's not clear from the names that this is the case -- this information is kind of tucked away as if it's a secret. :)

bart-samwel (approved):

spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingReader.scala
olaky: This is to keep it consistent with pathToColumn. This way we are using foldRight on both of them, which was my main goal here

olaky:
The failing test also fails in a docs only change: https://github.com/delta-io/delta/actions/runs/9167309765/job/25204242662?pr=3123

PR #3125, opened 2024-05-21
Delta python proto codegen

longvu-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)


longvu-db:
Failed tests are in `org.apache.spark.sql.delta.perf.OptimizeGeneratedColumnSuite`, seems unrelated to this PR.

connect/common/src/main/buf.gen.yaml
```diff
@@ -0,0 +1,21 @@
+#
```
tdas, line 1: I suggest the directory name of `spark-connect` to capture that connect is only for spark.

longvu-db: @tdas Hmm, from an outside look, even though the content of the `python` folder is only Spark, there is nothing in the naming that suggests that it is for Spark no? Shall we also do `delta/spark-connect`?

tdas: the entire content of python/delta is for spark. I agree that ideally this entire python/delta should be under spark/, but thats an existing problem that we will solve holistically. 

its not clear to me yet, why all the python files need to be under python/delta/connect/ and not directly under python/delta/

can you tell me more?

tdas: I dont have a strong opinion. for now we can keep it here... and we can move it when we have more clarity.

longvu-db: @tdas Because all the files under `python/delta/connect` are for Delta **Connect**, perhaps we can change the name to `delta-connect` also.

I think naming `name1/name2` might be better than `name1-name2`, so `delta/connect` instead of `delta-connect`, `spark/connect` instead of `spark-connect`.

dev/connect-gen-protos.sh
```diff
@@ -0,0 +1,102 @@
...
+rm -Rf gen
+
+# Now, regenerate the new files
+buf generate --debug -vvv
+
+# We need to edit the generate python files to account for the actual package location and not
+# the one generated by proto.
+for f in `find gen/proto/python/delta/connect -name "*.py*"`; do
+  # First fix the imports.
+  if [[ $f == *_pb2.py || $f == *_pb2_grpc.py ]]; then
```
andreaschat-db, line 69: This is the only part we modify the generated code. How do we know the replacements are correct? Perhaps add that to "How was this patch tested section"?

andreaschat-db (commented):
Left a couple of comments.

dev/tox.ini
```diff
@@ -17,6 +17,6 @@
 [pycodestyle]
```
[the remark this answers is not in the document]
longvu-db: Yes, I think so. https://github.com/delta-io/delta/commit/71606f47d5bc8d06b8acefe1f3a649bc44d38c10

dev/connect-gen-protos.sh
longvu-db: I added a Python script to check if the generated codes are out of sync with current proto message.

In the future, when we set up the Delta Connect server and client, we will have tests for that, and if the generated codes are wrong then it will break those Delta Connect tests

.github/workflows/spark_test.yaml
```diff
@@ -41,6 +41,10 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y make build-essential libssl-dev zlib1g-dev libbz2-dev libreadline-dev libsqlite3-dev wget curl llvm libncurses5-dev libncursesw5-dev xz-utils tk-dev libffi-dev liblzma-dev python-openssl git
           sudo apt install libedit-dev
+          curl -LO https://github.com/bufbuild/buf/releases/download/v1.28.1/buf-Linux-x86_64.tar.gz
```
andreaschat-db, line 44: Do we need to add this in the requirements?

andreaschat-db (approved):
LGTM. Left one comment.

.github/workflows/spark_test.yaml
longvu-db: I think not because the `requirements.txt` is only for Python dependencies, used for the Python `venv`

tomvanbussel (approved):
LGTM

PR #3138, opened 2024-05-22
[Spark] Add CREATE LIKE compatibility with user-provided table properties 

jamesd-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

User provided properties when performing CREATE LIKE commands were being ignored and only the properties from source table were being added. This PR adds/overwrites any applicable properties with the user provided ones.

## How was this patch tested?

Unit tests were created replicating the customer issue for CREATE LIKE commands both originating in Delta tables and other formats.

No


rahulsmahadev:
can you update the title from `CREATE LIKE ` -> `CREATE TABLE LIKE`

rahulsmahadev (approved):
LGTM

Issue #3145, opened 2024-05-23
[Feature Request] Migrate to use Spark Structured Logging


zedtang:
#### Which Delta project/connector is this regarding?

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

### Overview

Spark structured logging ([SPIP](https://docs.google.com/document/d/1rATVGmFLNVLmtxSpWrEceYm7d-ocgu8ofhryVs4g3XU/edit?pli=1#heading=h.rks22ljmjzmt)) aims to provide more analyzable logs by changing the plain text logs to a structured JSON format. Spark has introduced a new set of structured logging APIs and is currently working on [migration](https://issues.apache.org/jira/browse/SPARK-47240).
### Motivation

Currently, analyzing Delta text logs can be challenging. Structured logging will make the logs easier to understand and directly queryable using Spark SQL. It will simplify the debugging process and facilitate more efficient log analysis.
### Willingness to contribute

The Delta Lake Community encourages new feature contributions. Would you or another member of your organization be willing to contribute an implementation of this feature?

- [x] Yes. I can contribute this feature independently.
- [ ] Yes. I would be willing to contribute this feature with guidance from the Delta Lake community.
- [ ] No. I cannot contribute this feature at this time.

PR #3146, opened 2024-05-23
[Spark] Support Spark Structured Logging API

zedtang:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Uber issue: #3145 

Support Spark Structured Logging API by using the shimming layer. Spark Structured Logging is only available in Spark 4.0 snapshot so need to shim the API to make it compile for Spark 3.5.
## How was this patch tested?

Tests that the new API on Spark master can produce structured logs: DeltaStructuredLoggingSuite
Tests that the shimming API on Spark 3.5 still produce the same plain text logs: DeltaPatternLoggingSuite

No.


spark/src/main/scala-spark-3.5/shims/LogKeyShims.scala
```diff
@@ -0,0 +1,48 @@
...
+
+package org.apache.spark.internal
+
+import java.util.Locale
+
+/**
+ * All structured logging `keys` used in `MDC` must extends `LogKeyShims`
+ */
+trait LogKeyShims {
+  val name: String = this.toString.toLowerCase(Locale.ROOT)
```
gengliangwang, line 47: There is a recent changes in https://github.com/apache/spark/pull/46634/files#diff-41ea575e2e1305fa8a92f4e3f549e7b896d97eac0e43e00d3dc65c816846e349R52. Let's wait until the PR is merged (probably this week)

spark/src/main/scala/org/apache/spark/sql/delta/logging/LogKeys.scala
```diff
@@ -0,0 +1,789 @@
...
+
+package org.apache.spark.sql.delta.logging
+
+import org.apache.spark.internal.LogKeyShims
+
+/**
+ * Various keys used for mapped diagnostic contexts(MDC) in logging. All structured logging keys
+ * should be defined here for standardization.
+ */
+object LogKeys {
```
gengliangwang, line 47: QQ: I plan to reduce the number of log keys in Apache Spark. Meanwhile, there will be new log keys from delta in this file. How will we keep sync with the Apache Spark?
Shall we just store the log keys for delta logging in this file?

zedtang: Yeah we can and should just store the keys that Delta uses. This means potentially duplicate some keys with Spark in this file for the keys that Delta uses but that's fine.

I removed the used keys.

gengliangwang (approved):
LGTM

spark/src/main/scala-spark-3.5/shims/LogKeyShims.scala
```diff
@@ -0,0 +1,46 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.internal
+
+import java.util.Locale
+
+// LogKey is part of Spark's Structured Logging API and is not available in Spark 3.5.
+trait LogKeyShims {
```
jaceklaskowski, line 44: Why not singular (`LogKeyShim`)?

spark/src/main/scala-spark-3.5/shims/LogKeyShims.scala
```diff
@@ -0,0 +1,46 @@
...
+ * limitations under the License.
+ */
+
+/*
+ * This file contains code from the Apache Spark project (original license above).
+ * It contains modifications, which are licensed as follows:
+ */
+
+/*
+ * Copyright (2021) The Delta Lake Project Authors.
```
jaceklaskowski, line 24: Since it's a new file, use 2024

spark/src/main/scala-spark-3.5/shims/LoggingShims.scala
```diff
@@ -0,0 +1,154 @@
...
+ */
+
+package org.apache.spark.internal
+
+import org.apache.logging.log4j.CloseableThreadContext
+
+// MDC is part of Spark's Structured Logging API and is not available in Spark 3.5.
+case class MDC(key: LogKeyShims, value: Any) {
+  require(!value.isInstanceOf[MessageWithContext],
+    "the class of value cannot be MessageWithContext")
```
jaceklaskowski, line 46: s/cannot/must

Please add the type to the error OR remove the error message altogether (since it does not add much anyway).

spark/src/main/scala-spark-3.5/shims/LoggingShims.scala
```diff
@@ -0,0 +1,154 @@
...
+}
+
+object LogEntry {
+  import scala.language.implicitConversions
+
+  implicit def from(msgWithCtx: => MessageWithContext): LogEntry =
+    new LogEntry(msgWithCtx)
+}
+
+trait LoggingShims extends Logging {
```
jaceklaskowski, line 76: A singular?

jaceklaskowski (changes requested):

spark/src/main/scala-spark-3.5/shims/LogKeyShims.scala
zedtang: It looks like every file under spark/ always uses 2021.

spark/src/main/scala-spark-3.5/shims/LogKeyShims.scala
zedtang: This follows the naming convention under https://github.com/delta-io/delta/tree/master/spark/src/main/scala-spark-3.5/shims

spark/src/main/scala-spark-3.5/shims/LoggingShims.scala
zedtang: This is a stripped-down version from https://github.com/apache/spark/blob/bdcb79f23da3d09469910508426a54a78adcbda6/common/utils/src/main/scala/org/apache/spark/internal/Logging.scala#L78 to provide the syntax support, without any functionality.
I prefer to keep it as is if it doesn't fix anything.

spark/src/main/scala-spark-3.5/shims/LoggingShims.scala
zedtang: ditto

zedtang (commented):
@jaceklaskowski which of your comments are blocking? I see you have used the request changes option

PR #3153, opened 2024-05-24
Fix concating paths with relative filenames that contain colon ':'

xupefei:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR fixes an issue where the `safeConcatPaths` method throws an exception when the `relativeChildPath` contains a colon `:`. Such a character is not allowed in Hadoop paths due to ambiguity (`aa:bb.csv` can be interpreted as an absolute path like `aa://bb.csv` where `aa` is the scheme), but is allowed in many file systems such as S3. Thus we need to handle this case.

The fix here is to prepend a `/` so that Hadoop will know that everything after `/` belongs to the path, not the scheme.

## How was this patch tested?

New tests.

Nope.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaTable.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaTable.scala
@@ -470,24 +470,31 @@ object DeltaTableUtils extends PredicateHelper
   def withActiveSession[T](spark: SparkSession)(body: => T): T = spark.withActive(body)
 
   /**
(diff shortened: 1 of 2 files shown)


xupefei:
@LukasRupprecht Hey Lukas, I saw you wrote the original code. Could you take a look at this PR?

LukasRupprecht (approved):
LGTM, thanks for the fix!

PR #3155, opened 2024-05-24
[Spark] Drop Type widening feature: read Parquet footers to collect files to rewrite

johanl-db:
The initial approach to identify files that contain a type that differs from the table schema and that must be rewritten before dropping the type widening table feature is convoluted and turns out to be more brittle than intended.

This change switches instead to directly reading the file schema from the Parquet footer and rewriting all files that have a mismatching type.

### Additional Context
Files are identified using their default row commit version (a part of the row tracking feature) and matched against type changes previously applied to the table and recorded in the table metadata: any file written before the latest type change should use a different type and must be rewritten.

This requires multiple pieces of information to be accurately tracked:
- Default row commit versions must be correctly assigned to all files. E.p. files that are copied over without modification must never be assigned a new default row commit version. On the other hand, default row commit versions are preserved across CLONE but these versions don't match anything in the new cloned table.
- Type change history must be reliably recorded and preserved across schema changes, e.g. column mapping.

Any bug will likely lead to files not being correctly rewritten before removing the table feature, potentially leaving the table in an unreadable state.

## How was this patch tested?
Tests added in previous PR to cover CLONE and RESTORE: https://github.com/delta-io/delta/pull/3053
Tests added and updated in this PR to cover rewriting files with different column types when removing the table feature.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/PreDowngradeTableFeatureCommand.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/PreDowngradeTableFeatureCommand.scala
@@ -23,9 +23,9 @@ import scala.util.control.NonFatal
 import org.apache.spark.sql.delta.catalog.DeltaTableV2
 import org.apache.spark.sql.delta.commands.{AlterTableSetPropertiesDeltaCommand, AlterTableUnsetPropertiesDeltaCommand, DeltaReorgTableCommand, DeltaReorgTableMode, DeltaReorgTableSpec}
 import org.apache.spark.sql.delta.commands.columnmapping.RemoveColumnMappingCommand
+import org.apache.spark.sql.delta.commands.optimize.OptimizeMetrics
 import org.apache.spark.sql.delta.managedcommit.ManagedCommitUtils
 import org.apache.spark.sql.delta.metering.DeltaLogging
-import org.apache.spark.sql.delta.schema.SchemaMergingUtils
 import org.apache.spark.sql.delta.util.{Utils => DeltaUtils}
 import org.apache.spark.sql.util.ScalaExtensions._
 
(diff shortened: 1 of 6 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/PreDowngradeTableFeatureCommand.scala
```diff
@@ -309,8 +309,9 @@ case class TypeWideningPreDowngradeCommand(table: DeltaTableV2)
    * @return Return the number of files rewritten.
    */
   private def rewriteFilesIfNeeded(): Long = {
-    val numFilesToRewrite = TypeWidening.numFilesRequiringRewrite(table.initialSnapshot)
-    if (numFilesToRewrite == 0L) return 0L
+    if (!TypeWideningMetadata.containsTypeWideningMetadata(table.initialSnapshot.schema)) {
```
sabir-akhadov, line 312: Is it guaranteed that there are no files to rewrite if the metadata is not contained? 
I guess if there is no metadata but there are files the table is not fully readable anyways?

spark/src/main/scala/org/apache/spark/sql/delta/commands/ReorgTableHelper.scala
```diff
@@ -0,0 +1,102 @@
...
+        /* length */ addFile.size,
+        /* isDir */ false,
+        /* blockReplication */ 0,
+        /* blockSize */ 1,
+        /* modificationTime */ addFile.modificationTime,
+        new Path(absPath)
+      )
+    }
+
+    val footers = DeltaFileOperations.readParquetFootersInParallel(
```
sabir-akhadov, line 88: Looks like this uses multi-threading to read all footers on the driver. 
I wonder how does this compare to launching a spark job? Are there any expectations on the number of footers to read?

johanl-db: There could be a lot of footers to read..
I updated it to read footers on the executors, see `filterParquetFilesOnExecutors`

spark/src/main/scala/org/apache/spark/sql/delta/commands/ReorgTableHelper.scala
```diff
@@ -0,0 +1,111 @@
...
+      configuration,
+      fileStatuses.toList,
+      ignoreCorruptFiles)
```
sabir-akhadov, line 100: So this param will allow it to ignore corrupt files? Should we instead never ignore corrupt files? It could potentially be dangerous because a corrupt file might recover later on and become unreadable. 
Any modification we do on tables are potentially unsafe if we ignore corrupt files.

johanl-db: Good catch, I updated it to always fail on corrupted files + added a test

sabir-akhadov (approved):
lgtm, thanks

PR #3157, opened 2024-05-24
[Spark] Make listDeltaCompactedDeltaCheckpointFilesAndLatestChecksumFile reusable

dhruvarya-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Factors out the main logic from `listDeltaCompactedDeltaCheckpointFilesAndLatestChecksumFile` into a new function `listFromFileSystemInternal` which can potentially be reused by other callers. Also added a new test utility which runs tests with in-memory-tracking commit owner enabled.

## How was this patch tested?

Existing tests should cover this.

No

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitTestUtils.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitTestUtils.scala
@@ -72,6 +72,26 @@ trait ManagedCommitTestUtils
     }
   }
 
+  /**
+   * Run the test against a [[TrackingCommitOwnerClient]] with backfill batch size =
+   * `batchBackfillSize`
+   */
+  def testWithTrackingInMemoryCommitOwner(
+      backfillBatchSize: Int)(testName: String)(f: => Unit): Unit = {
+    test(s"$testName [Backfill batch size: $backfillBatchSize]") {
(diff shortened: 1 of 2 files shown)


spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitTestUtils.scala
```diff
@@ -72,6 +72,26 @@ trait ManagedCommitTestUtils
     }
   }
 
+  /**
+   * Run the test against a [[TrackingCommitOwnerClient]] with backfill batch size =
+   * `batchBackfillSize`
+   */
+  def testWithTrackingInMemoryCommitOwner(
```
sumeet-db, line 79: The other users of this utils class won't fully understand different commit owner implementations etc. Should we call it `def testWithManagedCommits(` instead?

sumeet-db (approved):

spark/src/test/scala/org/apache/spark/sql/delta/managedcommit/ManagedCommitTestUtils.scala
dhruvarya-db: Sounds good, updated.

PR #3158, opened 2024-05-25
[Spark] Add an integration test for DynamoDB Commit Owner

dhruvarya-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Adds an integration test for the DynamoDB Commit Owner. Tests the following scenarios
1. Automated dynamodb table creation
2. Concurrent reads and writes
3. Table upgrade and downgrade

## How was this patch tested?

Test runs successfully with real DynamoDB and S3.


run-integration-tests.py
```diff
@@ -187,6 +187,46 @@ def run_dynamodb_logstore_integration_tests(root_dir, version, test_name, extra_
...
+    test_dir = path.join(root_dir, \
+        path.join("spark", "src", "main", "java", "io", "delta", "dynamodbcommitstore", "integration_tests"))
+    test_files = [path.join(test_dir, f) for f in os.listdir(test_dir)
+                  if path.isfile(path.join(test_dir, f)) and
+                  f.endswith(".py") and not f.startswith("_")]
+
+    python_root_dir = path.join(root_dir, "python")
+    extra_class_path = path.join(python_root_dir, path.join("delta", "testing"))
+    packages = "io.delta:delta-%s_2.12:%s" % (get_artifact_name(version), version)
+    packages += "," + "org.apache.hadoop:hadoop-aws:3.3.4"
```
junlee-db, line 207: please help me understand, why do we need this while the test above doesn't?

in logstore integration test we use 3.3.1, how is the version decided? https://github.com/delta-io/delta/blob/master/storage-s3-dynamodb/integration_tests/dynamodb_logstore.py#L64

spark/src/main/java/io/delta/dynamodbcommitstore/integration_tests/dynamodb_commitowner_integration_test.py
```diff
@@ -0,0 +1,233 @@
...
+        print("Reading {:d} rows ...".format(
+            spark.read.format("delta").load(delta_table_path).distinct().count())
+        )
+        time.sleep(1)
+
+
+def start_read_thread():
+    thread = threading.Thread(target=read_data)
+    thread.start()
+    return thread
```
junlee-db, line 135: Seems to be quite common utils in the other test as well, maybe should we create a dynamo_test_util.py and let both of them look at the files?

spark/src/main/java/io/delta/dynamodbcommitstore/integration_tests/dynamodb_commitowner_integration_test.py
```diff
@@ -0,0 +1,233 @@
...
+    .appName("utilities") \
+    .master("local[*]") \
+    .config("spark.sql.extensions", "io.delta.sql.DeltaSparkSessionExtension") \
+    .config("spark.sql.catalog.spark_catalog", "org.apache.spark.sql.delta.catalog.DeltaCatalog") \
+    .config(f"spark.databricks.delta.properties.defaults.{commit_owner_property_key}{property_key_suffix}", "dynamodb") \
+    .config(f"spark.databricks.delta.properties.defaults.managedCommits.commitOwnerConf{property_key_suffix}", dynamodb_commit_owner_conf) \
+    .config(f"spark.databricks.delta.managedCommits.commitOwner.ddb.awsCredentialsProviderName", "com.amazonaws.auth.profile.ProfileCredentialsProvider") \
+    .getOrCreate()
+
+# spark.sparkContext.setLogLevel("INFO")
```
junlee-db, line 110: redundant line?

spark/src/main/java/io/delta/dynamodbcommitstore/integration_tests/dynamodb_commitowner_integration_test.py
```diff
@@ -0,0 +1,233 @@
...
+
+delta_table_version = num_rows
+validate_table_version_as_per_dynamodb(delta_table_version)
+
+def perform_insert_and_validate(insert_value):
+    spark.sql(f"INSERT INTO delta.`{delta_table_path}` VALUES ({insert_value}, {insert_value})")
+    res = spark.sql(f"SELECT 1 FROM delta.`{delta_table_path}` WHERE id = {insert_value} AND a = {insert_value}").collect()
+    assert(len(res) == 1)
+
+def check_for_delta_file_existence(version, is_backfilled, should_exist):
```
junlee-db, line 189: maybe call them `check_for_delta_file` because existence is decided by `should_exist` param.

junlee-db (commented):
Logic looks good, some mechanicals

spark/src/main/java/io/delta/dynamodbcommitstore/integration_tests/dynamodb_commitowner_integration_test.py
```diff
@@ -0,0 +1,233 @@
...
+check_for_delta_file_existence(delta_table_version, is_backfilled=False, should_exist=False)
+print("[SUCCESS] Downgrade to filesystem based commits worked")
+
+# Upgrade to managed commits should work
+print("===================== Evaluating upgrade to managed commits =====================")
+spark.sql(f"ALTER TABLE delta.`{delta_table_path}` SET TBLPROPERTIES ('delta.{commit_owner_property_key}{property_key_suffix}' = 'dynamodb')")
+delta_table_version += 1
+
+perform_insert_and_validate(9991)
+delta_table_version += 1
```
junlee-db, line 225: run another check `check_for_delta_file_existence`... here?

spark/src/main/java/io/delta/dynamodbcommitstore/integration_tests/dynamodb_commitowner_integration_test.py
```diff
@@ -0,0 +1,233 @@
...
+
+check_for_delta_file_existence(delta_table_version, is_backfilled=True, should_exist=True)
+# No UUID delta file should have been created for this version
+check_for_delta_file_existence(delta_table_version, is_backfilled=False, should_exist=False)
+print("[SUCCESS] Downgrade to filesystem based commits worked")
+
+# Upgrade to managed commits should work
+print("===================== Evaluating upgrade to managed commits =====================")
+spark.sql(f"ALTER TABLE delta.`{delta_table_path}` SET TBLPROPERTIES ('delta.{commit_owner_property_key}{property_key_suffix}' = 'dynamodb')")
+delta_table_version += 1
```
junlee-db, line 222: verify here too?

run-integration-tests.py
dhruvarya-db: That is a good question. I think just like `dynamodb_logstore.py`, we should let the user specify the version for this library.  

Hi @scottsand-db, this integration test is very similar to https://github.com/delta-io/delta/blob/master/storage-s3-dynamodb/integration_tests/dynamodb_logstore.py that you worked on. One difference is that dynamodb-commitowner is built as part of delta-spark. However, I couldn't get the package to work correctly without adding `org.apache.hadoop:hadoop-aws:3.3.4` as an extra dependency. Should we add `org.apache.hadoop:hadoop-aws:3.3.4` as a direct dependency of `io.delta:delta-*` in build.sbt?

spark/src/main/java/io/delta/dynamodbcommitstore/integration_tests/dynamodb_commitowner_integration_test.py
dhruvarya-db: Renamed to `check_for_delta_file_in_filesystem` to make it explicit that we are querying the filesystem not dynamodb

spark/src/main/java/io/delta/dynamodbcommitstore/integration_tests/dynamodb_commitowner_integration_test.py
```diff
@@ -0,0 +1,236 @@
...
+commit_owner_property_key = "managedCommits.commitOwner"
+property_key_suffix = "-preview"
+
+spark = SparkSession \
+    .builder \
+    .appName("utilities") \
+    .master("local[*]") \
+    .config("spark.sql.extensions", "io.delta.sql.DeltaSparkSessionExtension") \
+    .config("spark.sql.catalog.spark_catalog", "org.apache.spark.sql.delta.catalog.DeltaCatalog") \
+    .config(f"spark.databricks.delta.properties.defaults.{commit_owner_property_key}{property_key_suffix}", "dynamodb") \
```
prakharjain09, line 105: Could we also run the same test on top of a new table which is not a managed-commit table at the time of creation?
So CREATE -> INSERT+SELECT -> UPGRADE -> INSERT+SELECT -> DOWNGRADE. -> INSERT+SELECT -> UPGRADE

dhruvarya-db: Done

prakharjain09 (approved):

PR #3162, opened 2024-05-28
Delta 4.0 - Type Widening documentation

johanl-db:
## Description
Update the type widening documentation to list additional type changes supported in Delta 4.0

### DIFF
--- a/docs/source/delta-type-widening.md
+++ b/docs/source/delta-type-widening.md
@@ -23,15 +23,16 @@ The feature introduces a limited set of supported type changes in <Delta> 3.2 an
   "`decimal`", ,"`decimal` with greater precision and scale"
   "`date`", ,"`timestampNTZ`"
 
-To avoid accidental promotion of integer values to decimals, you must manually commit type changes from `byte`, `short`, `int`, or `long` to `decimal` or `double`.
+To avoid accidental promotion of integer values to decimals, type changes from `byte`, `short`, `int`, or `long` to `decimal` or `double` are not eligible to be applied automatically during schema evolution. You must manually alter the type in that case.
 
 .. note::
 
-  When changing any numeric type to `decimal`, the total precision must be equal to or greater than the starting precision. If you also increase the scale, the total precision must increase by a corresponding amount.
+  When changing an integer or decimal type to decimal, the total precision must be equal to or greater than the starting precision. If you also increase the scale, the total precision must increase by a corresponding amount.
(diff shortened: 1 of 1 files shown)


docs/source/delta-type-widening.md
```diff
@@ -4,15 +4,34 @@ description: Learn about type widening in Delta.
...
+
+  "`byte`","`short`, `int`","`short`,`int`,`long`, `decimal`, `double`"
+  "`short`","`int`", "`int`,`long`, `decimal`, `double`"
+  "`int`"," ","`long`, `decimal`, `double`"
+  "`long`", ,"`decimal`"
+  "`float`", ,"`double`"
+  "`decimal`", ,"`decimal` with greater precision and scale"
+  "`date`", ,"`timestampNTZ`"
+
+To avoid accidental promotion of integer values to decimals, you must manually commit type changes from `byte`, `short`, `int`, or `long` to `decimal` or `double`.
```
tomvanbussel, line 26: "Manually commit" sounds like this would require the user to manually create and commit a transaction. Lets change this to something like "you must manually alter the type from ...".

docs/source/delta-type-widening.md
```diff
@@ -4,15 +4,34 @@ description: Learn about type widening in Delta.
...
+
+.. note::
+
+  When changing any numeric type to `decimal`, the total precision must be equal to or greater than the starting precision. If you also increase the scale, the total precision must increase by a corresponding amount.
```
tomvanbussel, line 30: "any numeric type" suggests it's possible to change floats/doubles to decimals.

docs/source/delta-type-widening.md
tomvanbussel, line 26: Is float to double automatic?

docs/source/delta-type-widening.md
```diff
@@ -4,15 +4,34 @@ description: Learn about type widening in Delta.
...
+
+  The minimum target for `byte`, `short`, and `int` types is `decimal(10,0)`. The minimum target for `long` is `decimal(20,0)`.
+
+  If you want to add two decimal places to a field with `decimal(10,1)`, the minimum target is `decimal(12,3)`.
```
tomvanbussel, line 34: Needs a "for example" to make it clear that this applies more generally. Might be worth mentioning decimal(p1, s1) -> decimal(p2, s2) is only allowed if p1 - s1 >= p2 - s2 and s1 >= s2.

tomvanbussel (commented):
Nice! Great to see that type widening now supports more conversions.

tomvanbussel (approved):
Thank you for addressing my feedback. LGTM!

PR #3167, opened 2024-05-28
[Spark] Include violating value in DELTA_EXCEED_CHAR_VARCHAR_LIMIT

tomvanbussel:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR modifies the error message for the `DELTA_EXCEED_CHAR_VARCHAR_LIMIT` error class to include the value that violated the constraint.

## How was this patch tested?

Modified a test in `DeltaErrorsSuite` and added a test to `DeltaConstraintsSuite`.

Yes, the error message for the `DELTA_EXCEED_CHAR_VARCHAR_LIMIT` error class is modified to include the value that violated the constraint.

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/delta/DeltaErrorsSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/DeltaErrorsSuite.scala
@@ -278,16 +278,16 @@ trait DeltaErrorsSuiteBase
         Some("NOT NULL constraint violated for column: col1.\n"))
     }
     {
-      val expr = CatalystSqlParser.parseExpression("concat(\"hello \", \"world\")")
+      val expr = UnresolvedAttribute("col")
       val e = intercept[DeltaInvariantViolationException] {
         throw DeltaInvariantViolationException(
           Constraints.Check(CharVarcharConstraint.INVARIANT_NAME,
             LessThanOrEqual(Length(expr), Literal(5))),
-          Map.empty[String, Any])
+          Map("col" -> "Hello World"))
       }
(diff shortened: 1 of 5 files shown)


spark/src/test/scala/org/apache/spark/sql/delta/DeltaErrorsSuite.scala
```diff
@@ -278,16 +278,16 @@ trait DeltaErrorsSuiteBase
...
-      val expr = CatalystSqlParser.parseExpression("concat(\"hello \", \"world\")")
+      val expr = UnresolvedAttribute("col")
       val e = intercept[DeltaInvariantViolationException] {
         throw DeltaInvariantViolationException(
           Constraints.Check(CharVarcharConstraint.INVARIANT_NAME,
             LessThanOrEqual(Length(expr), Literal(5))),
-          Map.empty[String, Any])
+          Map("col" -> "Hello World"))
       }
       checkErrorMessage(e, Some("DELTA_EXCEED_CHAR_VARCHAR_LIMIT"), Some("22001"),
```
johanl-db, line 288: Can you migrate this to `checkError`?

johanl-db (approved):
Looks good, nice improvement!

spark/src/test/scala/org/apache/spark/sql/delta/DeltaErrorsSuite.scala
tomvanbussel: It seems like the goal of this suite is to actually test the final error messages and error codes. This wouldn't be possible with `checkError` (as its purpose to be able to write tests without having to match the error message).

PR #3168, opened 2024-05-28
[Spark] Delta connect spark connect proto

longvu-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Delta Connect is an extension for Spark Connect that adds support for the DeltaTable API. To be able to communicate back to Spark Connect we need to have access to Spark Connect's Protobufs. 

Since Buf does not support local dependencies, while Delta and Apache Spark are two separate repositories, the only solution for Delta Connect to get access to [Spark Connect's Protobuf messages](https://github.com/apache/spark/tree/master/connector/connect/common/src/main/protobuf/spark/connect) is to have the copies of the messages in the Delta repository. 

This may go out of sync, and we would need to update it manually from the Apache Spark repository, but this should be fine and would not break anything since Protobufs are backward compatible.

## How was this patch tested?
N/A.

No.


andreaschat-db (commented):
High level question: How do we plan on staying in sync with Spark Connect? Can this part later on being pushed to a different repository?

longvu-db:
@andreaschat-db I don't think we have any plans right now, we are fine with having it to be out of sync momentarily with Spark Connect, do we have any plans on having the Spark Connect's protos to stay in sync with their copies in Delta Connect @tomvanbussel?

tomvanbussel:
@longvu-db I am not sure if I have a good answer for now. We should figure this out together with the Apache Spark folks. For now we can just manually update these Protobuf files whenever there is a Spark release.

andreaschat-db (approved):
LGTM.

PR #3178, opened 2024-05-30
[Spark] Remove deprecated FileNames.deltaFile method

sumeet-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

- Clients are expected to use DeltaCommitFileProvider instead.
- FileNames.unsafeDeltaFile can be used in case the clients knows for sure the file has been backfilled.

## How was this patch tested?

Build

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/util/FileNames.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/util/FileNames.scala
@@ -43,18 +43,6 @@ object FileNames {
   def unsafeDeltaFile(path: Path, version: Long): Path = new Path(path, f"$version%020d.json")
 
   /**
-   * Returns the delta (json format) path for a given delta file.
-   * WARNING: This API is unsafe and deprecated. It can resolve to incorrect paths if the table has
-   * Managed Commits. It will be removed in future versions.
-   * Use DeltaCommitFileProvider(snapshot).deltaFile instead to guarantee accurate paths or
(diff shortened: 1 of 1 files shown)


prakharjain09 (approved):

spark/src/main/scala/org/apache/spark/sql/delta/util/FileNames.scala
```diff
@@ -42,18 +42,6 @@ object FileNames {
...
-   * Returns the delta (json format) path for a given delta file.
-   * WARNING: This API is unsafe and deprecated. It can resolve to incorrect paths if the table has
-   * Managed Commits. It will be removed in future versions.
-   * Use DeltaCommitFileProvider(snapshot).deltaFile instead to guarantee accurate paths or
-   * unsafeDeltaFile for potentially incorrect file name resolution.
-   */
-  @deprecated(
-    "This method is deprecated and will be removed in future versions. " +
-      "Use DeltaCommitFileProvider(snapshot).deltaFile or unsafeDeltaFile instead")
-  def deltaFile(path: Path, version: Long): Path = unsafeDeltaFile(path, version)
```
prakharjain09, line 55: Lets get this reviewed by @tdas also.

tdas: This is internal method. no users should be using this. this deprecation i guess is for delta devs, not users. So this should be good to remove.

tdas (approved):

PR #3201, opened 2024-06-03
[Spark] Add Delta Connect Python Client for First 4.0 Preview branch

longvu-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

[Delta Connect Python Client PR](https://github.com/delta-io/delta/pull/3136) but merge into branch-4.0-preview1 instead of master.

## How was this patch tested?
Added UTs.

No.


andreaschat-db (commented):
Hi @longvu-db, the compilation against spark master (which is where spark-connect compiles) is broken :).

python/run-tests.py
```diff
@@ -22,22 +22,28 @@
...
-    extra_class_path = path.join(python_root_dir, path.join("delta", "testing"))
+    extra_class_path = path.join(python_root_dir, path.join(test_path, "testing"))
 
     for test_file in test_files:
         try:
             cmd = ["spark-submit",
                    "--driver-class-path=%s" % extra_class_path,
-                   "--packages", package, test_file]
+                   "--packages", package +
+                   ",io.delta:delta-connect-server_2.13:4.0.0-SNAPSHOT," +
```
tomvanbussel, line 41: How does this work? Does it use the locally built packages for this? Will it continue to use the local packages once we publish these packages to Maven?

python/run-tests.py
```diff
@@ -22,22 +22,28 @@
...
+                   "io.delta:delta-connect-common_2.13:4.0.0-SNAPSHOT," +
+                   "org.apache.spark:spark-connect_2.13:4.0.0-preview1," +
+                   "org.apache.spark:spark-connect-client-jvm_2.13:4.0.0-preview1," +
```
tomvanbussel, line 44: It shouldn't be necessary to include this one.

python/run-tests.py
```diff
@@ -22,22 +22,28 @@
...
+                   "org.apache.spark:spark-connect-common_2.13:4.0.0-preview1," +
+                   "com.google.protobuf:protobuf-java:3.25.1", test_file]
```
tomvanbussel, line 46: In the final product it shouldn't have to be the necessary to manually include this. It's also not necessary for Spark (and somehow also not for our Scala client tests?). Let's look at how Spark assembles its jars (including it's shading of this library, which we will also have to do).

python/run-tests.py
```diff
@@ -22,22 +22,28 @@
...
                   f.endswith(".py") and not f.startswith("_")]
-    extra_class_path = path.join(python_root_dir, path.join("delta", "testing"))
+    extra_class_path = path.join(python_root_dir, path.join(test_path, "testing"))
 
     for test_file in test_files:
         try:
             cmd = ["spark-submit",
                    "--driver-class-path=%s" % extra_class_path,
-                   "--packages", package, test_file]
+                   "--packages", package +
```
tomvanbussel, line 40: We should only include extra packages when running the connect test. The regular tests shouldn't use the extra packages. Is that possible?

python/run-tests.py
```diff
@@ -182,4 +188,7 @@ def run_delta_connect_codegen_python(root_dir):
     run_mypy_tests(root_dir)
     run_pypi_packaging_tests(root_dir)
     run_delta_connect_codegen_python(root_dir)
-    test(root_dir, package)
+    test(root_dir, "delta", package)
+    is_spark_master = os.getenv("IS_SPARK_MASTER")
+    if is_spark_master is not None:
+        test(root_dir, path.join("delta", "connect"), package)
```
tomvanbussel, line 193: What's the significance of this path?

python/run-tests.py
```diff
@@ -182,4 +188,7 @@ def run_delta_connect_codegen_python(root_dir):
     run_mypy_tests(root_dir)
     run_pypi_packaging_tests(root_dir)
     run_delta_connect_codegen_python(root_dir)
-    test(root_dir, package)
+    test(root_dir, "delta", package)
+    is_spark_master = os.getenv("IS_SPARK_MASTER")
+    if is_spark_master is not None:
```
tomvanbussel, line 192: Should we not run the tests if `IS_SPARK_MASTER` is not set?

python/run-tests.py
```diff
@@ -182,4 +188,7 @@ def run_delta_connect_codegen_python(root_dir):
     run_mypy_tests(root_dir)
     run_pypi_packaging_tests(root_dir)
     run_delta_connect_codegen_python(root_dir)
-    test(root_dir, package)
+    test(root_dir, "delta", package)
+    is_spark_master = os.getenv("IS_SPARK_MASTER")
```
tomvanbussel, line 191: Seems like `IS_SPARK_MASTER` is only set when running the connect suites. Can we instead rename it something like `RUN_SPARK_CONNECT_TESTS`?

tomvanbussel (commented):
Great to see that you managed to get everything working! Adding a few comments about the build and testing infra.

python/run-tests.py
longvu-db: Before, we only run tests in `python/delta/tests`, 

`test_files = [os.path.join(test_dir, f) for f in os.listdir(test_dir)
                  if os.path.isfile(os.path.join(test_dir, f)) and
                  f.endswith(".py") and not f.startswith("_")]` does not recursively go through the whole `delta` folder while we are adding tests to `python/delta/connect/tests`, so

python/run-tests.py
longvu-db: @tomvanbussel Yeah I also thought about this 50 minutes ago haha

python/run-tests.py
longvu-db: @tomvanbussel Yes that should be possible, I just hasn't finished polishing the PR

python/run-tests.py
longvu-db: @tomvanbussel My understanding is that this downloads from `Maven`.

> WARNING: Using incubator modules: jdk.incubator.vector
> :: loading settings :: url = jar:file:/home/runner/.local/share/virtualenvs/delta-Jo9PrCI6/lib/python3.9/site-packages/pyspark/jars/ivy-2.5.2.jar!/org/apache/ivy/core/settings/ivysettings.xml
> Ivy Default Cache set to: /home/runner/.ivy2.5.2/cache
> The jars for the packages stored in: /home/runner/.ivy2.5.2/jars
> io.delta#delta-spark_2.13 added as a dependency
> io.delta#delta-connect-server_2.13 added as a dependency
> io.delta#delta-connect-common_2.13 added as a dependency
> org.apache.spark#spark-connect_2.13 added as a dependency
> org.apache.spark#spark-connect-client-jvm_2.13 added as a dependency
> org.apache.spark#spark-connect-common_2.13 added as a dependency
> com.google.protobuf#protobuf-java added as a dependency

> downloading file:/home/runner/.m2/repository/io/delta/delta-connect-server_2.13/4.0.0-SNAPSHOT/delta-connect-server_2.13-4.0.0-SNAPSHOT.jar ...

> 	[SUCCESSFUL ] io.delta#delta-connect-server_2.13;4.0.0-SNAPSHOT!delta-connect-server_2.13.jar (1ms)
> downloading file:/home/runner/.m2/repository/io/delta/delta-connect-common_2.13/4.0.0-SNAPSHOT/delta-connect-common_2.13-4.0.0-SNAPSHOT.jar ...
> 	[SUCCESSFUL ] io.delta#delta-connect-common_2.13;4.0.0-SNAPSHOT!delta-connect-common_2.13.jar (3ms)
> downloading https://repo1.maven.org/maven2/org/apache/spark/spark-connect_2.13/4.0.0-preview1/spark-connect_2.13-4.0.0-preview1.jar ...
> 	[SUCCESSFUL ] org.apache.spark#spark-connect_2.13;4.0.0-preview1!spark-connect_2.13.jar (94ms)
> downloading https://repo1.maven.org/maven2/org/apache/spark/spark-connect-client-jvm_2.13/4.0.0-preview1/spark-connect-client-jvm_2.13-4.0.0-preview1.jar ...
> 	[SUCCESSFUL ] org.apache.spark#spark-connect-client-jvm_2.13;4.0.0-preview1!spark-connect-client-jvm_2.13.jar (144ms)
> downloading https://repo1.maven.org/maven2/org/apache/spark/spark-connect-common_2.13/4.0.0-preview1/spark-connect-common_2.13-4.0.0-preview1.jar ...
> 	[SUCCESSFUL ] org.apache.spark#spark-connect-common_2.13;4.0.0-preview1!spark-connect-common_2.13.jar (37ms)
> downloading https://repo1.maven.org/maven2/com/google/protobuf/protobuf-java/3.25.1/protobuf-java-3.25.1.jar ...
> 	[SUCCESSFUL ] com.google.protobuf#protobuf-java;3.25.1!protobuf-java.jar (19ms)
> downloading https://repo1.maven.org/maven2/io/grpc/grpc-protobuf/1.62.2/grpc-protobuf-1.62.2.jar ...
> 	[SUCCESSFUL ] io.grpc#grpc-protobuf;1.62.2!grpc-protobuf.jar (4ms)

dev/delta-connect-gen-protos.sh
```diff
@@ -71,6 +71,7 @@ for f in `find gen/proto/python/delta/connect -name "*.py*"`; do
   if [[ $f == *_pb2.py || $f == *_pb2_grpc.py ]]; then
     sed \
       -e "s/DESCRIPTOR, 'delta.connect/DESCRIPTOR, 'delta.connect.proto/g" \
+      -e 's/from delta.connect import/from delta.connect.proto import/g' \
```
longvu-db, line 74: This should have been added to https://github.com/delta-io/delta/pull/3125

dev/requirements.txt
```diff
@@ -1,9 +1,14 @@
 # Linter
-mypy==0.982
+mypy==1.8.0
```
longvu-db, line 2: We are now using `1.8.0` so just updating the requirements.

python/delta/connect/proto/commands_pb2.py
```diff
@@ -27,7 +27,7 @@
 _sym_db = _symbol_database.Default()
 
 
-from delta.connect import base_pb2 as delta_dot_connect_dot_base__pb2
+from delta.connect.proto import base_pb2 as delta_dot_connect_dot_base__pb2
```
longvu-db, line 30: This comes from running again the `delta-connect-gen-protos.sh`

python/delta/connect/tests/test_deltatable.py
```diff
@@ -0,0 +1,161 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import unittest
+
+from delta.connect.testing.utils import DeltaTestCase
+import os
+import sys
+
```
longvu-db, line 22: Ran into this issue (https://stackoverflow.com/questions/4383571/importing-files-from-different-folder)

python/run-tests.py
tomvanbussel: Okay, but that's not really what we want. We want these tests to use the locally built package when available. Otherwise you'd have to publish a new Maven artifact every time the server plugin is modified.

python/delta/tests/test_sql.py
```diff
@@ -104,6 +104,7 @@ def read_table() -> DataFrame:
                 return self.spark.sql(f"SELECT * FROM {table}")
 
             self.spark.sql(f"DROP TABLE IF EXISTS {table}")
```
longvu-db, line 106: I think this should be added anyway since we create `table` and `table_part` down below

python/run-tests.py
longvu-db: @tomvanbussel Added as a TODO in the future that should be done after the Preview.

python/run-tests.py
```diff
@@ -61,12 +61,14 @@ def prepare(root_dir):
...
 
+
+def get_local_package(package_name):
     # Get current release which is required to be loaded
     version = '0.0.0'
     with open(os.path.join(root_dir, "version.sbt")) as fd:
         version = fd.readline().split('"')[1]
-    package = "io.delta:delta-spark_2.13:" + version
-    return package
+    local_package = f"io.delta:{package_name}_2.13:" + version
```
andreaschat-db, line 70:
```suggestion
    return f"io.delta:{package_name}_2.13:" + version
```

python/run-tests.py
```diff
@@ -176,10 +178,26 @@ def run_delta_connect_codegen_python(root_dir):
 if __name__ == "__main__":
     print("##### Running python tests #####")
     root_dir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
-    package = prepare(root_dir)
+    prepare(root_dir)
+    delta_spark_package = get_local_package("delta-spark")
```
andreaschat-db, line 182: Shall we move this after line 189?

python/run-tests.py
```diff
@@ -22,22 +22,22 @@
 from os import path
 
 
-def test(root_dir, package):
-    # Run all of the test under test/python directory, each of them
+def test(root_dir, test_path, packages):
```
andreaschat-db, line 25: nit: I would name both for consistency either `root_dir/test_dir` or `root_path/test_path`. Potentially the later.

python/run-tests.py
```diff
@@ -176,10 +178,26 @@ def run_delta_connect_codegen_python(root_dir):
...
     run_python_style_checks(root_dir)
     run_mypy_tests(root_dir)
     run_pypi_packaging_tests(root_dir)
     run_delta_connect_codegen_python(root_dir)
-    test(root_dir, package)
+    run_delta_connect_tests = os.getenv("RUN_DELTA_CONNECT_TESTS")
+    if run_delta_connect_tests is None:
+        test(root_dir, "delta", [delta_spark_package])
+    else:
+        assert(run_delta_connect_tests == "true")
```
andreaschat-db, line 192: There shouldn't be any assertion here. If the user runs the test with RUN_DELTA_CONNECT_TESTS=false should go to the case above. Let's also say that `RUN_DELTA_CONNECT_TESTS ` is by default false.

python/run-tests.py
```diff
@@ -176,10 +178,26 @@ def run_delta_connect_codegen_python(root_dir):
...
+
+        # TODO: In the future, find a way to get these
+        # packages locally instead of downloading from Maven.
+        delta_connect_packages = ["com.google.protobuf:protobuf-java:3.25.1",
+                                  "org.apache.spark:spark-connect_2.13:4.0.0-preview1",
+                                  get_local_package("delta-connect-server")]
+
+        test(
```
andreaschat-db, line 200: nit: Does not fit in a single line? Shall we use the same style for both `path` calls?

andreaschat-db (commented):
Left a couple of comments.

python/run-tests.py
longvu-db: @andreaschat-db I wanted to avoid the case where the user can set RUN_DELTA_CONNECT_TESTS=*** some random string ***, it should either not be specified (which is false), false or true.

andreaschat-db: Now with `or run_delta_connect_tests == "false"` should work. 

nit: It is still a bit weird position for the assertion since it is asymmetric. What about:
```suggestion
        if run_delta_connect_tests is None or run_delta_connect_tests == "false":
          ...
       elif run_delta_connect_tests == "true":
          ...
       else:
          error msg  
```

python/run-tests.py
```diff
@@ -22,16 +22,16 @@
 from os import path
 
 
-def test(root_dir, test_path, packages):
-    # Run all of the tests under test_path directory, each of them
-    # has main entry point to execute, which is python's unittest testing
+def test(root_dir, codes_dir, packages):
```
andreaschat-db, line 25: nit: 
1. I believe "code" is an uncountable noun, i.e. there is no plural.
2. Shall we just call it `toTest_dir`?

andreaschat-db (approved):
LGMT. Left 2 nits.

python/run-tests.py
longvu-db: @andreaschat-db Not sure if toTest is following the python naming convention.

python/run-tests.py
```diff
@@ -176,10 +177,24 @@ def run_delta_connect_codegen_python(root_dir):
...
+    run_delta_connect_tests = os.getenv("RUN_DELTA_CONNECT_TESTS")
+    if run_delta_connect_tests is None or run_delta_connect_tests == "false":
+        delta_spark_package = get_local_package("delta-spark")
+        test(root_dir, "delta", [delta_spark_package])
+    elif run_delta_connect_tests == "true":
+        # TODO: In the future, find a way to get these
+        # packages locally instead of downloading from Maven.
+        delta_connect_packages = ["com.google.protobuf:protobuf-java:3.25.1",
+                                  "org.apache.spark:spark-connect_2.13:4.0.0-preview1",
+                                  get_local_package("delta-connect-server")]
```
allisonport-db, line 195: Why do we need the server package here?

python/delta/connect/tests/test_deltatable.py
```diff
@@ -0,0 +1,162 @@
...
+from delta.connect.testing.utils import DeltaTestCase
+
+path_to_delta_connect_tests_folder = os.path.dirname(os.path.abspath(__file__))
+path_to_delta_folder = os.path.dirname(os.path.dirname(path_to_delta_connect_tests_folder))
+sys.path.append(path_to_delta_folder)
+
+from tests.test_deltatable import DeltaTableTestsMixin
+
+
+class DeltaTableTests(DeltaTableTestsMixin, DeltaTestCase):
```
allisonport-db, line 30: A little confused, isn't nothing tested here?

python/run-tests.py
longvu-db: @allisonport-db We need the server code to test the client locally just like the Scala Client one.

python/delta/connect/tests/test_deltatable.py
longvu-db: @allisonport-db https://github.com/delta-io/delta/blob/master/python/delta/tests/test_deltatable.py `test_forPath`, `test_forPathWithOptions, `test_forName`, and `test_alias_and_toDF` but under Delta Connect mode

PR #3236, opened 2024-06-07
[Spark] use map lookup in createPhysicalSchema

andrewxue-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Instead of calling `SchemaUtils.findNestedFieldIgnoreCase` for each column, we prepare a map with `SchemaUtils.explode` before, and perform map lookups during iteration. 

This speeds up this function on wide tables. It may still be slow for tables with deeply nested schemas because the path needs to be built every time, but there should be no regression.

## How was this patch tested?

Manual profiling for an alter table add columns query:

Before: (~13s)
Screenshot 2024-06-06 at 5 50 01 PM

After: (~3s)
Screenshot 2024-06-06 at 5 50 17 PM

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaColumnMapping.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaColumnMapping.scala
@@ -32,7 +32,7 @@ import org.json4s.jackson.JsonMethods._
 import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.catalyst.analysis.UnresolvedAttribute
 import org.apache.spark.sql.catalyst.expressions.Attribute
-import org.apache.spark.sql.catalyst.util.CaseInsensitiveMap
+import org.apache.spark.sql.catalyst.util.{CaseInsensitiveMap, QuotingUtils}
 import org.apache.spark.sql.execution.datasources.parquet.ParquetFileFormat
 import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.types.{ArrayType, DataType, MapType, Metadata => SparkMetadata, MetadataBuilder, StructField, StructType}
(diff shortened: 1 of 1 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/DeltaColumnMapping.scala
```diff
@@ -511,10 +511,15 @@ trait DeltaColumnMappingBase extends DeltaLogging {
       throw DeltaErrors.unsupportedColumnMappingMode(columnMappingMode.name)
     }
 
+    val referenceSchemaColumnMap: Map[String, StructField] =
+      SchemaMergingUtils.explode(referenceSchema).map { case (path, field) =>
+        QuotingUtils.quoteNameParts(path).toLowerCase() -> field
```
bart-samwel, line 516: We should avoid depending on the default locale.

```suggestion
        QuotingUtils.quoteNameParts(path).toLowerCase(Locale.ROOT) -> field
```

spark/src/main/scala/org/apache/spark/sql/delta/DeltaColumnMapping.scala
```diff
@@ -511,10 +511,15 @@ trait DeltaColumnMappingBase extends DeltaLogging {
...
+      }.toMap
+
     SchemaMergingUtils.transformColumns(schema) { (path, field, _) =>
       val fullName = path :+ field.name
-      val inSchema = SchemaUtils
-        .findNestedFieldIgnoreCase(referenceSchema, fullName, includeCollections = true)
+      val inSchema =
+        referenceSchemaColumnMap.get(QuotingUtils.quoteNameParts(fullName).toLowerCase())
```
bart-samwel, line 522:
```suggestion
        referenceSchemaColumnMap.get(QuotingUtils.quoteNameParts(fullName).toLowerCase(Locale.ROOT))
```

bart-samwel (approved):

PR #3237, opened 2024-06-07
[Spark] Rename managed commit to coordinated commits

dhruvarya-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [x] Other (RFC)

## Description

Renames Managed Commit to Coordinated Commits and Commit Owner to Commit Coordinator to better express the meaning of the feature. Configs, table feature name, classes, functions, comments, file names, and directory names have all been updated to reflect the new terminology.

## How was this patch tested?

Existing tests should cover these changes.

Yes, the feature name and config names have changed.


prakharjain09 (approved):

connectors/standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
```diff
@@ -527,7 +527,9 @@ private[internal] class OptimisticTransactionImpl(
   private def shouldCheckpoint(committedVersion: Long): Boolean = {
     val checkpointingEnabled =
       deltaLog.hadoopConf.getBoolean(StandaloneHadoopConf.CHECKPOINTING_ENABLED, true)
-    checkpointingEnabled && committedVersion != 0 && committedVersion % deltaLog.checkpointInterval == 0
+    checkpointingEnabled &&
```
scottsand-db, line 530: Why this change?

dhruvarya-db: My bad, I have reverted this change.

PR #3283, opened 2024-06-18
[Kernel] Update ConflictChecker to perform conflict resolution of ICT

EstherBear:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

Update ConflictChecker to perform conflict resolution of inCommitTimestamp and complete the inCommitTimestamp support in Kernel.

## How was this patch tested?

Add unit tests to verify the conflict resolution of timestamps and enablement version.

Yes, user can enable monotonic inCommitTimestamp by enabling its property.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
@@ -152,12 +152,13 @@ public class TransactionImpl
                             rebaseState.getLatestCommitTimestamp(),
                             attemptCommitInfo.getInCommitTimestamp());
                     if (updatedInCommitTimestamp.isPresent()) {
-                        Optional<Metadata> metadataWithICTInfo = getUpdatedMetadataAfterConflict(
-                                engine,
-                                updatedInCommitTimestamp.get(),
-                                metadata,
-                                rebaseState.getLatestVersion()
-                        );
+                        Optional<Metadata> metadataWithICTInfo =
+                                getMetadataWithUpdatedICTEnablementInfo(
+                                        engine,
+                                        updatedInCommitTimestamp.get(),
+                                        metadata,
+                                        rebaseState.getLatestVersion()
+                                );
                         metadataWithICTInfo.ifPresent(this::updateMetadata);
                     }
                     attemptCommitInfo.setInCommitTimestamp(updatedInCommitTimestamp);
@@ -208,7 +209,7 @@ public class TransactionImpl
         return attemptInCommitTimestamp;
     }
 
-    private Optional<Metadata> getUpdatedMetadataAfterConflict(
+    private Optional<Metadata> getMetadataWithUpdatedICTEnablementInfo(
             Engine engine,
             Long updatedInCommitTimestamp,
             Metadata attemptMetadata,
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
@@ -41,8 +41,7 @@ import io.delta.kernel.internal.util.Clock;
 import io.delta.kernel.internal.util.FileNames;
 import io.delta.kernel.internal.util.InCommitTimestampUtils;
 import io.delta.kernel.internal.util.VectorUtils;
-import static io.delta.kernel.internal.TableConfig.CHECKPOINT_INTERVAL;
-import static io.delta.kernel.internal.TableConfig.IN_COMMIT_TIMESTAMPS_ENABLED;
+import static io.delta.kernel.internal.TableConfig.*;
 import static io.delta.kernel.internal.actions.SingleAction.*;
 import static io.delta.kernel.internal.util.Preconditions.checkArgument;
 import static io.delta.kernel.internal.util.Preconditions.checkState;
@@ -152,13 +151,13 @@ public class TransactionImpl
                                     rebaseState.getLatestCommitTimestamp(),
                                     attemptCommitInfo.getInCommitTimestamp());
                     if (updatedInCommitTimestamp.isPresent()) {
-                        Optional<Metadata> metadataWithICTInfo =
-                                getMetadataWithUpdatedICTEnablementInfo(
+                        Optional<Metadata> metadataWithICTInfo = InCommitTimestampUtils
+                                .getUpdatedMetadataWithICTEnablementInfo(
                                         engine,
                                         updatedInCommitTimestamp.get(),
+                                        readSnapshot,
                                         metadata,
-                                        rebaseState.getLatestVersion()
-                                );
+                                        rebaseState.getLatestVersion() + 1L);
                         metadataWithICTInfo.ifPresent(this::updateMetadata);
                     }
                     attemptCommitInfo.setInCommitTimestamp(updatedInCommitTimestamp);
@@ -207,20 +206,6 @@ public class TransactionImpl
             return Optional.of(updatedInCommitTimestamp);
         }
         return attemptInCommitTimestamp;
-    }
-
-    private Optional<Metadata> getMetadataWithUpdatedICTEnablementInfo(
-            Engine engine,
-            Long updatedInCommitTimestamp,
-            Metadata attemptMetadata,
-            Long lastWinningVersion) {
-        long nextAvailableVersion = lastWinningVersion + 1L;
-        return InCommitTimestampUtils.getUpdatedMetadataWithICTEnablementInfo(
-                engine,
-                updatedInCommitTimestamp,
-                readSnapshot,
-                attemptMetadata,
-                nextAvailableVersion);
     }
 
     private TransactionCommitResult doCommit(
@@ -286,7 +271,7 @@ public class TransactionImpl
      */
     private Optional<Long> generateInCommitTimestampForFirstCommitAttempt(
             Engine engine, long currentTimestamp) {
-        if (IN_COMMIT_TIMESTAMPS_ENABLED.fromMetadata(metadata)) {
+        if (isICTEnabled(metadata)) {
             long lastCommitTimestamp = readSnapshot.getTimestamp(engine);
             return Optional.of(Math.max(currentTimestamp, lastCommitTimestamp + 1));
         } else {
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
@@ -149,8 +149,8 @@ public class TransactionImpl
                     commitAsVersion = newCommitAsVersion;
                     Optional<Long> updatedInCommitTimestamp =
                             getUpdatedInCommitTimestampAfterConflict(
-                            rebaseState.getLatestCommitTimestamp(),
-                            attemptCommitInfo.getInCommitTimestamp());
+                                    rebaseState.getLatestCommitTimestamp(),
+                                    attemptCommitInfo.getInCommitTimestamp());
                     if (updatedInCommitTimestamp.isPresent()) {
                         Optional<Metadata> metadataWithICTInfo =
                                 getMetadataWithUpdatedICTEnablementInfo(
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
@@ -136,8 +136,6 @@ public class TransactionImpl
             do {
                 logger.info("Committing transaction as version = {}.", commitAsVersion);
                 try {
-                    // TODO Update the attemptCommitInfo and metadata based on the conflict
-                    // resolution.
                     return doCommit(engine, commitAsVersion, attemptCommitInfo, dataActions);
                 } catch (FileAlreadyExistsException fnfe) {
                     logger.info("Concurrent write detected when committing as version = {}. " +
@@ -149,6 +147,20 @@ public class TransactionImpl
                             "New commit version %d should be greater than the previous commit " +
                                     "attempt version %d.", newCommitAsVersion, commitAsVersion);
                     commitAsVersion = newCommitAsVersion;
+                    Optional<Long> updatedInCommitTimestamp =
+                            getUpdatedInCommitTimestampAfterConflict(
+                            rebaseState.getLatestCommitTimestamp(),
+                            attemptCommitInfo.getInCommitTimestamp());
+                    if (updatedInCommitTimestamp.isPresent()) {
+                        Optional<Metadata> metadataWithICTInfo = getUpdatedMetadataAfterConflict(
+                                engine,
+                                updatedInCommitTimestamp.get(),
+                                metadata,
+                                rebaseState.getLatestVersion()
+                        );
+                        metadataWithICTInfo.ifPresent(this::updateMetadata);
+                    }
+                    attemptCommitInfo.setInCommitTimestamp(updatedInCommitTimestamp);
                 }
                 numRetries++;
             } while (numRetries < NUM_TXN_RETRIES);
@@ -184,6 +196,30 @@ public class TransactionImpl
                     metadataWithICTInfo.ifPresent(this::updateMetadata);
                 }
         );
+    }
+
+    private Optional<Long> getUpdatedInCommitTimestampAfterConflict(
+            long winningCommitTimestamp, Optional<Long> attemptInCommitTimestamp) {
+        if (attemptInCommitTimestamp.isPresent()) {
+            long updatedInCommitTimestamp = Math.max(
+                    attemptInCommitTimestamp.get(), winningCommitTimestamp + 1);
+            return Optional.of(updatedInCommitTimestamp);
+        }
+        return attemptInCommitTimestamp;
+    }
+
+    private Optional<Metadata> getUpdatedMetadataAfterConflict(
+            Engine engine,
+            Long updatedInCommitTimestamp,
+            Metadata attemptMetadata,
+            Long lastWinningVersion) {
+        long nextAvailableVersion = lastWinningVersion + 1L;
+        return InCommitTimestampUtils.getUpdatedMetadataWithICTEnablementInfo(
+                engine,
+                updatedInCommitTimestamp,
+                readSnapshot,
+                attemptMetadata,
+                nextAvailableVersion);
     }
 
     private TransactionCommitResult doCommit(
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
@@ -130,7 +130,10 @@ public class TransactionImpl
             long commitAsVersion = readSnapshot.getVersion(engine) + 1;
             // Generate the commit action with the inCommitTimestamp if ICT is enabled.
             CommitInfo attemptCommitInfo = generateCommitAction(engine);
-            updateMetadataWithICTIfRequired(engine, attemptCommitInfo);
+            updateMetadataWithICTIfRequired(
+                    engine,
+                    attemptCommitInfo.getInCommitTimestamp(),
+                    readSnapshot.getVersion(engine));
             int numRetries = 0;
             do {
                 logger.info("Committing transaction as version = {}.", commitAsVersion);
@@ -150,16 +153,10 @@ public class TransactionImpl
                             getUpdatedInCommitTimestampAfterConflict(
                                     rebaseState.getLatestCommitTimestamp(),
                                     attemptCommitInfo.getInCommitTimestamp());
-                    if (updatedInCommitTimestamp.isPresent()) {
-                        Optional<Metadata> metadataWithICTInfo = InCommitTimestampUtils
-                                .getUpdatedMetadataWithICTEnablementInfo(
-                                        engine,
-                                        updatedInCommitTimestamp.get(),
-                                        readSnapshot,
-                                        metadata,
-                                        rebaseState.getLatestVersion() + 1L);
-                        metadataWithICTInfo.ifPresent(this::updateMetadata);
-                    }
+                    updateMetadataWithICTIfRequired(
+                            engine,
+                            updatedInCommitTimestamp,
+                            rebaseState.getLatestVersion());
                     attemptCommitInfo.setInCommitTimestamp(updatedInCommitTimestamp);
                 }
                 numRetries++;
@@ -181,10 +178,11 @@ public class TransactionImpl
         this.shouldUpdateMetadata = true;
     }
 
-    private void updateMetadataWithICTIfRequired(Engine engine, CommitInfo attemptCommitInfo) {
+    private void updateMetadataWithICTIfRequired(
+            Engine engine, Optional<Long> inCommitTimestampOpt, long lastCommitVersion) {
         // If ICT is enabled for the current transaction, update the metadata with the ICT
         // enablement info.
-        attemptCommitInfo.getInCommitTimestamp().ifPresent(
+        inCommitTimestampOpt.ifPresent(
                 inCommitTimestamp -> {
                     Optional<Metadata> metadataWithICTInfo =
                             InCommitTimestampUtils.getUpdatedMetadataWithICTEnablementInfo(
@@ -192,7 +190,7 @@ public class TransactionImpl
                                     inCommitTimestamp,
                                     readSnapshot,
                                     metadata,
-                                    readSnapshot.getVersion(engine) + 1L);
+                                    lastCommitVersion + 1L);
                     metadataWithICTInfo.ifPresent(this::updateMetadata);
                 }
         );
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
@@ -31,7 +31,7 @@ import io.delta.kernel.internal.*;
 import io.delta.kernel.internal.actions.CommitInfo;
 import io.delta.kernel.internal.actions.SetTransaction;
 import io.delta.kernel.internal.util.FileNames;
-import static io.delta.kernel.internal.TableConfig.IN_COMMIT_TIMESTAMPS_ENABLED;
+import static io.delta.kernel.internal.TableConfig.isICTEnabled;
 import static io.delta.kernel.internal.actions.SingleAction.CONFLICT_RESOLUTION_SCHEMA;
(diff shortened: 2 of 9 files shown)


EstherBear:
This PR is based on https://github.com/delta-io/delta/pull/3282 merge it first.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/InCommitTimestampSuite.scala
```diff
@@ -266,4 +266,78 @@ class InCommitTimestampSuite extends DeltaTableWriteSuiteBase {
...
+      appendData(
+        engine,
+        tablePath,
+        data = Seq(Map.empty[String, Literal] -> dataBatches2),
+        clock = clock
+      )
+      clock.setTime(startTime - 1000)
+      commitAppendData(engine, txn1, Seq(Map.empty[String, Literal] -> dataBatches1))
+      assert(
+        getInCommitTimestamp(engine, table, 2).get > getInCommitTimestamp(engine, table, 1).get)
```
dhruvarya-db, line 295: Can you assert that 
`getInCommitTimestamp(engine, table, 2).get == getInCommitTimestamp(engine, table, 1).get) + 1`

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/InCommitTimestampSuite.scala
```diff
@@ -266,4 +266,78 @@ class InCommitTimestampSuite extends DeltaTableWriteSuiteBase {
...
+
+      commitAppendData(engine, txn1, Seq(Map.empty[String, Literal] -> dataBatches1))
+
+      val ver1Snapshot = table.getSnapshotAsOfVersion(engine, 1).asInstanceOf[SnapshotImpl]
+      val ver2Snapshot = table.getLatestSnapshot(engine).asInstanceOf[SnapshotImpl]
+      val observedEnablementTimestamp =
+        TableConfig.IN_COMMIT_TIMESTAMP_ENABLEMENT_TIMESTAMP.fromMetadata(ver2Snapshot.getMetadata)
+      val observedEnablementVersion =
+        TableConfig.IN_COMMIT_TIMESTAMP_ENABLEMENT_VERSION.fromMetadata(ver2Snapshot.getMetadata)
+      assert(observedEnablementTimestamp.get > ver1Snapshot.getTimestamp(engine))
```
dhruvarya-db, line 337: Again, 
 `assert(observedEnablementTimestamp.get == ver1Snapshot.getTimestamp(engine) + 1)`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -172,6 +185,30 @@ private void updateMetadata(Metadata metadata) {
         this.shouldUpdateMetadata = true;
     }
 
+    private Optional<Long> getUpdatedCommitTimestampAfterConflict(
```
dhruvarya-db, line 188: s/getUpdatedCommitTimestampAfterConflict/getUpdatedInCommitTimestampAfterConflict

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
```diff
@@ -113,7 +124,9 @@ public TransactionRebaseState resolveConflicts(Engine engine) throws ConcurrentW
 
         // if we get here, we have successfully rebased (i.e no logical conflicts)
         // against the winning transactions
-        return new TransactionRebaseState(getLastWinningTxnVersion(winningCommits));
+        return new TransactionRebaseState(
+                getLastWinningTxnVersion(winningCommits),
+                getLastCommitTimestamp(engine, winningCommits, winningCommitInfoOpt.get()));
```
dhruvarya-db, line 129: I see that `getLastWinningTxnVersion` is being invoked inside `getLastCommitTimestamp` as well. Can you store it in variable and pass it to both places from here instead?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
```diff
@@ -88,6 +92,8 @@ public static TransactionRebaseState resolveConflicts(
 
     public TransactionRebaseState resolveConflicts(Engine engine) throws ConcurrentWriteException {
         List<FileStatus> winningCommits = getWinningCommitFiles(engine);
+        AtomicReference<Optional<CommitInfo>> winningCommitInfoOpt = new AtomicReference<>(
```
dhruvarya-db, line 95: Any reason why this needs to an `AtomicReference`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
```diff
@@ -102,6 +108,11 @@ public TransactionRebaseState resolveConflicts(Engine engine) throws ConcurrentW
             actionsIterator.forEachRemaining(actionBatch -> {
                 checkArgument(!actionBatch.isFromCheckpoint());  // no checkpoints should be read
                 ColumnarBatch batch = actionBatch.getColumnarBatch();
+                CommitInfo commitInfo = CommitInfo.fromColumnVector(
+                        batch.getColumnVector(COMMITINFO_ORDINAL),0);
+                if (commitInfo != null) {
+                    winningCommitInfoOpt.set(Optional.of(commitInfo));
```
dhruvarya-db, line 114: I am a bit concerned about the scenario where the last WinningCommit does not have `CommitInfo` but the one before it does. In that case, we will unintentionally end up using the wrong inCommitTimestamp instead of throwing an exception.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
EstherBear: Because of this error:
image

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
EstherBear: Thank you for pointing this out! I've updated the code and added a test case to test the scenario where the last WinningCommit does not have `CommitInfo` but the one before it does.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
```diff
@@ -99,21 +103,33 @@ public TransactionRebaseState resolveConflicts(Engine engine) throws ConcurrentW
                 CONFLICT_RESOLUTION_SCHEMA,
                 Optional.empty())) {
 
-            actionsIterator.forEachRemaining(actionBatch -> {
+            List<ActionWrapper> actionBatchList = new ArrayList<>();
```
vkorukanti, line 106: We can't materialize the actions into a list. Kernel-API module should avoid holding large chunks of memory.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
```diff
@@ -99,21 +103,33 @@ public TransactionRebaseState resolveConflicts(Engine engine) throws ConcurrentW
...
+            actionsIterator.forEachRemaining(actionBatchList::add);
+            for (int i = 0; i < actionBatchList.size(); i++) {
+                ActionWrapper actionBatch = actionBatchList.get(i);
                 checkArgument(!actionBatch.isFromCheckpoint());  // no checkpoints should be read
                 ColumnarBatch batch = actionBatch.getColumnarBatch();
+                if (i == actionBatchList.size() - 1) {
+                    CommitInfo commitInfo =
+                            getCommitInfo(batch.getColumnVector(COMMITINFO_ORDINAL));
+                    winningCommitInfoOpt = Optional.ofNullable(commitInfo);
+                }
```
vkorukanti, line 116: If the goal is to get the last commit info, why not always look for the commit info, store in a variable, and the last value of the variable is the one we need?

Also, we can't assume the commit info to be present in the last action batch. A commit file could be read could generate multiple batches and the first batch in the list contains the commit info.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
```diff
@@ -99,21 +103,33 @@ public TransactionRebaseState resolveConflicts(Engine engine) throws ConcurrentW
...
             throw new UncheckedIOException("Error reading actions from winning commits.", ioe);
         }
 
         // if we get here, we have successfully rebased (i.e no logical conflicts)
         // against the winning transactions
-        return new TransactionRebaseState(getLastWinningTxnVersion(winningCommits));
+        long lastWinningVersion = getLastWinningTxnVersion(winningCommits);
+        return new TransactionRebaseState(
+                lastWinningVersion,
+                getLastCommitTimestamp(
```
vkorukanti, line 131: why do we need to fetch the ICT again? aren't we reading already above?

vkorukanti (changes requested):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
EstherBear: We couldn't always store commit info in a variable and use the last value of the variable because this [scenario](https://github.com/delta-io/delta/pull/3283#discussion_r1645227258) mentioned by @dhruvarya-db. But you are right, it's not necessarily in the last action batch.

EstherBear: Does it make sense to check if a batch is generated from the last commit file by checking its version from ActionWrapper?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
EstherBear: We just read the CommitInfo above but not the ICT? We still need to extract the ICT in CommitInfo?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
EstherBear: I also add a [test](https://github.com/delta-io/delta/blob/2748b6357ddb958be49b74ceec933b0b267ac89a/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/InCommitTimestampSuite.scala#L490) for this [scenario](https://github.com/delta-io/delta/pull/3283#discussion_r1645227258).

EstherBear: Pushed the solution by checking version from ActionWrapper.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -186,6 +198,30 @@ private void updateMetadataWithICTIfRequired(Engine engine, CommitInfo attemptCo
...
+            long winningCommitTimestamp, Optional<Long> attemptInCommitTimestamp) {
+        if (attemptInCommitTimestamp.isPresent()) {
+            long updatedInCommitTimestamp = Math.max(
+                    attemptInCommitTimestamp.get(), winningCommitTimestamp + 1);
+            return Optional.of(updatedInCommitTimestamp);
+        }
+        return attemptInCommitTimestamp;
+    }
+
+    private Optional<Metadata> getUpdatedMetadataAfterConflict(
```
dhruvarya-db, line 211: Let's make the name more specific? Something like `getMetadataWithUpdatedICTEnablementInfo`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
dhruvarya-db: @vkorukanti From what I understand, this will not perform an additional IO. This simply extracts the timestamp from the `CommitInfo` action (or the file modification timestamp if ICT is not enabled.)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
```diff
@@ -218,6 +259,25 @@ private List<FileStatus> getWinningCommitFiles(Engine engine) {
         }
     }
 
+    private long getLastCommitTimestamp(
```
dhruvarya-db, line 262: Please add a docstring

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
```diff
@@ -141,6 +158,15 @@ public TransactionRebaseState(long latestVersion) {
         public long getLatestVersion() {
             return latestVersion;
         }
+
+        /**
+         * Return the latest CommitTimestamp of the table.
```
dhruvarya-db, line 163: s/CommitTimestamp/commit timestamp. Also, please clarify what this means for both ICT and non-ICT tables.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -149,6 +147,21 @@ public TransactionCommitResult commit(Engine engine, CloseableIterable<Row> data
                             "New commit version %d should be greater than the previous commit " +
                                     "attempt version %d.", newCommitAsVersion, commitAsVersion);
                     commitAsVersion = newCommitAsVersion;
+                    Optional<Long> updatedInCommitTimestamp =
+                            getUpdatedInCommitTimestampAfterConflict(
+                            rebaseState.getLatestCommitTimestamp(),
```
dhruvarya-db, line 152: [nit] fix indentation

dhruvarya-db (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -186,6 +199,30 @@ private void updateMetadataWithICTIfRequired(Engine engine, CommitInfo attemptCo
...
+            Metadata attemptMetadata,
+            Long lastWinningVersion) {
+        long nextAvailableVersion = lastWinningVersion + 1L;
+        return InCommitTimestampUtils.getUpdatedMetadataWithICTEnablementInfo(
+                engine,
+                updatedInCommitTimestamp,
+                readSnapshot,
+                attemptMetadata,
+                nextAvailableVersion);
+    }
```
vkorukanti, line 224: This method isn't adding any value. Use call the `InCommitTimestampUtils.getUpdatedMetadataWithICTEnablementInfo` directly.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
```diff
@@ -113,7 +125,10 @@ public TransactionRebaseState resolveConflicts(Engine engine) throws ConcurrentW
 
         // if we get here, we have successfully rebased (i.e no logical conflicts)
         // against the winning transactions
-        return new TransactionRebaseState(getLastWinningTxnVersion(winningCommits));
+        return new TransactionRebaseState(
+                lastWinningVersion,
+                getLastCommitTimestamp(
+                        engine, lastWinningVersion, winningCommits, winningCommitInfoOpt.get()));
```
vkorukanti, line 131: Do you want to check that `winningCommitInfoOpt` is not empty?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
```diff
@@ -88,10 +92,13 @@ public static TransactionRebaseState resolveConflicts(
 
     public TransactionRebaseState resolveConflicts(Engine engine) throws ConcurrentWriteException {
         List<FileStatus> winningCommits = getWinningCommitFiles(engine);
+        AtomicReference<Optional<CommitInfo>> winningCommitInfoOpt =
```
vkorukanti, line 95: why is this a atmoic reference? Why can't we just use the Optional?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
```diff
@@ -173,6 +201,21 @@ private void handleMetadata(ColumnVector metadataVector) {
         }
     }
 
+    /**
+     * Get the commit info from the winning transactions.
+     *
+     * @param commitInfoVector commit info rows from the winning transactions
+     * @return the commit info
+     */
+    private CommitInfo getCommitInfo(ColumnVector commitInfoVector) {
```
vkorukanti, line 210: Return `Optional<CommitInfo>` directly, rather than using `Optional.ofNullable` at the caller.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
```diff
@@ -218,6 +261,36 @@ private List<FileStatus> getWinningCommitFiles(Engine engine) {
...
+     */
+    private long getLastCommitTimestamp(
+            Engine engine,
+            long lastWinningVersion,
+            List<FileStatus> winningCommits,
+            Optional<CommitInfo> winningCommitInfoOpt) {
+        FileStatus lastWinningTxn = winningCommits.get(winningCommits.size() - 1);
+        long winningCommitTimestamp = -1L;
+        if (snapshot.getVersion(engine) == -1 ||
+                !IN_COMMIT_TIMESTAMPS_ENABLED.fromMetadata(snapshot.getMetadata())) {
```
vkorukanti, line 283: I think adding utility method on `TableConfig` will help: `isICTEnabled(metadata)`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
```diff
@@ -218,6 +261,36 @@ private List<FileStatus> getWinningCommitFiles(Engine engine) {
...
+            winningCommitTimestamp = lastWinningTxn.getModificationTime();
+        } else {
+            winningCommitTimestamp = CommitInfo.getRequiredInCommitTimestamp(
+                    winningCommitInfoOpt,
+                    String.valueOf(lastWinningVersion),
+                    snapshot.getDataPath());
```
vkorukanti, line 289: why do you need the datapath here?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
```diff
@@ -218,6 +261,36 @@ private List<FileStatus> getWinningCommitFiles(Engine engine) {
...
+            Optional<CommitInfo> winningCommitInfoOpt) {
+        FileStatus lastWinningTxn = winningCommits.get(winningCommits.size() - 1);
+        long winningCommitTimestamp = -1L;
+        if (snapshot.getVersion(engine) == -1 ||
+                !IN_COMMIT_TIMESTAMPS_ENABLED.fromMetadata(snapshot.getMetadata())) {
+            winningCommitTimestamp = lastWinningTxn.getModificationTime();
+        } else {
+            winningCommitTimestamp = CommitInfo.getRequiredInCommitTimestamp(
+                    winningCommitInfoOpt,
+                    String.valueOf(lastWinningVersion),
```
vkorukanti, line 288: why convert to string?

vkorukanti (changes requested):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
EstherBear: Just for the error message print

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
EstherBear: Just for the error message print

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
EstherBear: This get is for the atomic reference but not for the optional.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
EstherBear: Because of this error:
https://github.com/delta-io/delta/pull/3283#discussion_r1645254811

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/InCommitTimestampSuite.scala
```diff
@@ -406,4 +411,124 @@ class InCommitTimestampSuite extends DeltaTableWriteSuiteBase {
...
+        engine,
+        tablePath,
+        data = Seq(Map.empty[String, Literal] -> dataBatches2),
+        clock = clock
+      )
+      clock.setTime(startTime - 1000)
+      commitAppendData(engine, txn1, Seq(Map.empty[String, Literal] -> dataBatches1))
+      assert(
+        getInCommitTimestamp(
+          engine, table, 2).get == getInCommitTimestamp(engine, table, 1).get + 1)
```
vkorukanti, line 441:
```suggestion
      assert(
        getInCommitTimestamp(engine, table, 2).get ===
        getInCommitTimestamp(engine, table, 1).get + 1)
```

vkorukanti: nit: also use named literals `version = 2` (or `2 /* version*/` for Java). Easy to read the code.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/InCommitTimestampSuite.scala
```diff
@@ -406,4 +411,124 @@ class InCommitTimestampSuite extends DeltaTableWriteSuiteBase {
...
+      val txn1 = createTxn(
+        engine,
+        tablePath,
+        schema = testSchema,
+        partCols = Seq.empty,
+        tableProperties = Map(IN_COMMIT_TIMESTAMPS_ENABLED.getKey -> "true"),
+        clock = clock)
+
+      clock.setTime(startTime)
+      appendData(
```
vkorukanti, line 468: Can you add few more winning txn. May be make this test paramterized with different number of winning commits.

vkorukanti (changes requested):
LGTM. Thanks for working this.

Last few comments. Once addressed, this PR is good to go.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
EstherBear: It starts with `optional.empty`. And in the [getLastCommitTimestamp](https://github.com/delta-io/delta/blob/de0155896cdacd64d8ec87aa260f2e38a002b62e/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java#L275) function it will check if it's ict enabled. And if it's ict enabled and the winningCommitInfoOpt is empty, the `CommitInfo.getRequiredInCommitTimestamp` will raise an error.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
EstherBear: The `getRequiredInCommitTimestamp` has some logic to check if `winningCommitInfoOpt` is empty and if it contains ict and raises different errors accordingly. So I think it's better to leave this function handle with the error messages?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/InCommitTimestampSuite.scala
```diff
@@ -406,4 +411,124 @@ class InCommitTimestampSuite extends DeltaTableWriteSuiteBase {
...
+
+      val ver1Snapshot = table.getSnapshotAsOfVersion(engine, 1).asInstanceOf[SnapshotImpl]
+      val ver2Snapshot = table.getLatestSnapshot(engine).asInstanceOf[SnapshotImpl]
+      val observedEnablementTimestamp =
+        IN_COMMIT_TIMESTAMP_ENABLEMENT_TIMESTAMP.fromMetadata(ver2Snapshot.getMetadata)
+      val observedEnablementVersion =
+        IN_COMMIT_TIMESTAMP_ENABLEMENT_VERSION.fromMetadata(ver2Snapshot.getMetadata)
+      assert(observedEnablementTimestamp.get == ver1Snapshot.getTimestamp(engine) + 1)
+      assert(
+        observedEnablementTimestamp.get == getInCommitTimestamp(engine, table, version = 2).get)
```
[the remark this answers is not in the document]
EstherBear: They are different actually. Left is Optional[Long] and right is Option[Long].

vkorukanti (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -169,23 +178,34 @@ private void updateMetadata(Metadata metadata) {
...
                     metadataWithICTInfo.ifPresent(this::updateMetadata);
                 }
         );
     }
 
+    private Optional<Long> getUpdatedInCommitTimestampAfterConflict(
+            long winningCommitTimestamp, Optional<Long> attemptInCommitTimestamp) {
+        if (attemptInCommitTimestamp.isPresent()) {
+            long updatedInCommitTimestamp = Math.max(
+                    attemptInCommitTimestamp.get(), winningCommitTimestamp + 1);
```
scottsand-db, line 203: @EstherBear

- Suppose the first commit attempt was at time T
- Suppose we conflicted with 10 commits, and the 10th winning commit is at time T+100
- Now, we are trying **again** to commit. But reading all of those files took a while. Now it is time T+150
- Currently in the code, we are just setting the ICT to T+101.
- But shouldn't we really be checking against the **current** time too? And still doing math.max against that value?

scottsand-db: cc @dhruvarya-db @vkorukanti

PR #3285, opened 2024-06-20
[PROTOCOL] Allow CDC actions to register data files

cstavr:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [X] Other (PROTOCOL)

## Description

Update the Delta PROTOCOL to allow `AddCDCFile` actions that do not add Change Data Files, but instead register Data Files that are also added by `AddFile` actions. In this case the `_change_type` column in the Data Files might not be `null`. Non-change data readers should disregard this column and only process columns defined within the table's schema.

## How was this patch tested?

N/A

Protocol clarification.

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -477,7 +477,6 @@ The following is an example `remove` action.
 ```
 
 ### Add CDC File
-
 The `cdc` action is used to add a [file](#change-data-files) containing only the data that was changed as part of the transaction. The `cdc` action is allowed to add a [Data File](#data-files) that is also added by an `add` action, when it does not contain any copied rows and the `_change_type` column is filled for all rows.
 
 When change data readers encounter a `cdc` action in a particular Delta table version, they must read the changes made in that version exclusively using the `cdc` files. If a version has no `cdc` action, then the data in `add` and `remove` actions are read as inserted and deleted rows, respectively.
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -477,7 +477,7 @@ The following is an example `remove` action.
 ```
 
 ### Add CDC File
-The `cdc` action is used to add a [file](#change-data-files) containing only the data that was changed as part of the transaction. When change data readers encounter a `cdc` action in a particular Delta table version, they must read the changes made in that version exclusively using the `cdc` files. If a version has no `cdc` action, then the data in `add` and `remove` actions are read as inserted and deleted rows, respectively.
+The `cdc` action is used to register a file containing only the data that was changed as part of the transaction. The `cdc` action can either add a [Change Data File](#change-data-files) or register a [Data File](#data-files) that is also added by an `add` action. When change data readers encounter a `cdc` action in a particular Delta table version, they must read the changes made in that version exclusively using the `cdc` files. If a version has no `cdc` action, then the data in `add` and `remove` actions are read as inserted and deleted rows, respectively.
 
 The schema of the `cdc` action is as follows:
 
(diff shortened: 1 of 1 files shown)


PROTOCOL.md
```diff
@@ -477,7 +477,7 @@ The following is an example `remove` action.
 ```
 
 ### Add CDC File
-The `cdc` action is used to add a [file](#change-data-files) containing only the data that was changed as part of the transaction. When change data readers encounter a `cdc` action in a particular Delta table version, they must read the changes made in that version exclusively using the `cdc` files. If a version has no `cdc` action, then the data in `add` and `remove` actions are read as inserted and deleted rows, respectively.
+The `cdc` action is used to register a file containing only the data that was changed as part of the transaction. The `cdc` action can either add a [Change Data File](#change-data-files) or register a [Data File](#data-files) that is also added by an `add` action. When change data readers encounter a `cdc` action in a particular Delta table version, they must read the changes made in that version exclusively using the `cdc` files. If a version has no `cdc` action, then the data in `add` and `remove` actions are read as inserted and deleted rows, respectively.
```
tdas, line 480: I dont get what this means "The `cdc` action can either add a [Change Data File](#change-data-files) or register a [Data File](#data-files) that is also added by an `add` action."

"cdc action" is 1 action - AddCdcFile. How can it "register" another action?

cstavr: @tdas I mean that an `AddCDCFile` action can contain the path for a Parquet file that is also added by an `AddFile` action.

cstavr: @tdas I update the PR to make this more clear.

PROTOCOL.md
```diff
@@ -477,7 +477,7 @@ The following is an example `remove` action.
 ```
 
 ### Add CDC File
-The `cdc` action is used to add a [file](#change-data-files) containing only the data that was changed as part of the transaction. When change data readers encounter a `cdc` action in a particular Delta table version, they must read the changes made in that version exclusively using the `cdc` files. If a version has no `cdc` action, then the data in `add` and `remove` actions are read as inserted and deleted rows, respectively.
+The `cdc` action is used to add a file containing only the data that was changed as part of the transaction. The file might be a [Change Data File](#change-data-files) or a [Data File](#data-file) that does not contain any copied rows. When change data readers encounter a `cdc` action in a particular Delta table version, they must read the changes made in that version exclusively using the `cdc` files. If a version has no `cdc` action, then the data in `add` and `remove` actions are read as inserted and deleted rows, respectively.
```
tdas, line 480: i still dont get it. is it your goal to say that the same parquet file can be put as both AddFile and AddCDCFile? if so, then this is not clear. 


Also if that is the goal, then its important also write down when a parquet data file can be considered as a CDC file? what constraints does that parquet file need to satisfy.

cstavr: Thanks @tdas. I reworded this section to make it more clear and clarify the requirements for the file added by the `AddCDCFile` action. 

In general the protocol does not refer to "Parquet files". It defines that:
- An `AddFile` action adds a `Data File` which is a file that is stored in the root directory.
- An `AddCDCFile` action adds a `Change Data File` which is a file stored in the `_change_data` directory, contains only data that was changes as part of a transaction (no copied rows) and has an extra `_change_type` column.

This is why I now define that:
"The `cdc` action is allowed to add a [Data File](#data-files) that is also added by an `add` action, when it does not contain any copied rows and the `_change_type` column is filled for all rows."

tdas (approved):

tdas:
I agree with @bart-samwel on this change being safe and backward compatible. LGTM. Merging it.

PR #3297, opened 2024-06-24
[Protocol RFC] Finalize type widening protocol

johanl-db:
## Description
Finalize the protocol RFC for type widening:
- Add missing supported type changes: `byte`,`short`,`int` to `double` and integers to decimals.
- Remove `tableVersion` from the type change metadata fields.
- Remove requirements around populating default row commit versions.

The two last requirements were initially intended to allow matching each file against type changes that happened before or after it was written. This didn't prove useful in practice - it was temporarily used to collect files to rewrite when dropping the feature but this now relies on fetching parquet footer as a more robust and simpler approach.

### DIFF
--- a/protocol_rfcs/type-widening.md
+++ b/protocol_rfcs/type-widening.md
@@ -8,10 +8,9 @@ This protocol change introduces the Type Widening feature, which enables changin
 # Type Widening
 > ***New Section after the [Clustered Table](#clustered-table) section***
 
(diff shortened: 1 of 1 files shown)


tomvanbussel (approved):
LGTM

bart-samwel (approved):

tdas (approved):

PR #3298, opened 2024-06-24
[Delta Uniform] Support expireSnapshot in uniform iceberg table automatically when OPTIMIZE

ChengJi-db:
## Description
**_Issue_**: the current uniform iceberg table doesn't have a mechanism to cleanup old manifest/manifest list files, which adds great storage maintenance overhead
**_Proposed_**: when `OPTIMIZE` is running on uniform delta table, it will trigger the `expireSnapshot` operation on corresponding iceberg table to do cleanup on manifests. The `OPTIMIZE` is chosen as the trigger since it's recommended to run frequently on delta table and iceberg's `expireSnapshot` is also recommended to run frequently (once every day) 

## How was this patch tested?
Manually tested


lzlfred:
lgtm

lzlfred (approved):

PR #3305, opened 2024-06-25
[Spark] Extend coordinated commit changes to other suites

junlee-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Extends various existing Delta suite with coordinated commits. This helps to add more coverage with coordinated commit changes.

## How was this patch tested?
This is test PR.

No

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/delta/DescribeDeltaHistorySuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/DescribeDeltaHistorySuite.scala
@@ -21,10 +21,11 @@ import java.io.File
 
 import org.apache.spark.sql.delta.DescribeDeltaHistorySuiteShims._
 import org.apache.spark.sql.delta.actions.{Action, AddCDCFile, AddFile, Metadata, Protocol, RemoveFile}
+import org.apache.spark.sql.delta.coordinatedcommits.CoordinatedCommitsBaseSuite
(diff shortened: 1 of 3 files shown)


spark/src/test/scala/org/apache/spark/sql/delta/DescribeDeltaHistorySuite.scala
```diff
@@ -139,6 +141,21 @@ trait DescribeDeltaHistorySuiteBase
       .asInstanceOf[Map[String, String]]
   }
 
+  // Returns necessary delta property json expected for the test. If coordinated commit is enabled,
+  // a few properties will be automatically populated, and this method will take care of it.
+  protected def getPropertiesJson(extraProperty: Option[Map[String, String]] = None): String = {
+    val coordinatedCommitsProperty = if (coordinatedCommitsEnabledInTests) {
+      getCoordinatedCommitsDefaultProperties()
```
prakharjain09, line 148: Why is this call-site needs `withICT` as false while the other one in `DescribeDeltaDetailSuite.scala` passes it as true?

junlee-db: `DESCRIBE DETAIL` returns ICT information as well, but HISTORY doesn't. The helper just evolves to cover both cases.

prakharjain09 (approved):

PR #3306, opened 2024-06-25
[Spark] Handle concurrent CREATE TABLE IF NOT EXISTS ... LIKE ... table commands

sumeet-db:
- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

When 2 or more CREATE TABLE IF NOT EXISTS table commands are run concurrently, they both think the table doesn't exist yet and the second command fails with TABLE_ALREADY_EXISTS error.

With this PR, we aim to make sure the second command end up in a no-op instead of a failure.

## How was this patch tested?

UTs

No


LukasRupprecht (approved):
LGTM!

PR #3309, opened 2024-06-25
[Hudi] setCommitFileUpdates bug fix

anniewang-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [x] Other (Hudi read support)

## Description

This PR fixes a bug in the Delta->Hudi metadata conversion logic. When the number of actions to convert is greater than the action batch size (can be changed by user), the previous code incorrectly only converted the last batch instead of converting all batches.
## How was this patch tested?
Unit test

No


harperjiang (approved):
lgtm

PR #3314, opened 2024-06-26
[Spark] Introduce stable type widening table feature

johanl-db:
## Description

This is the first step in transitioning the type widening table feature from preview to stable.
Requirement: the protocol spec is finalized and the final version doesn't contain any backward-incompatible changes. See https://github.com/delta-io/delta/pull/3297

This change introduces a new table feature `typeWidening` for the final, stable phase of type widening. Both the preview and stable table features are supported simultaneously - the current implementation respects both the protocol spec used during the preview and the finalized protocol spec.

The stable feature isn't enabled automatically though as type widening is still in preview. The preview feature `typeWidening-preview` is the default feature enabled when the type widening table property is set. 

The second step of the transition will be at GA time, to make the stable feature the default, automatically enabled one.
During the transition period, all clients will support any table that enabled the feature during both the preview, and later on during the GA phase.

## How was this patch tested?
Added tests covering enabling in combination preview / stable features
Added a test in particular to ensure that the preview feature doesn't get enabled automatically if the stable feature is already enabled.

DBR now supports all tables that enable the type widening table feature, both now during the preview and in the future after the feature transitions to GA.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala
@@ -430,7 +430,7 @@ class DeltaLog private(
     val protocolEnabledFeatures = targetProtocol.writerFeatureNames
       .flatMap(TableFeature.featureNameToFeature)
     val activeFeatures =
-      Protocol.extractAutomaticallyEnabledFeatures(spark, targetMetadata, Some(targetProtocol))
+      Protocol.extractAutomaticallyEnabledFeatures(spark, targetMetadata, targetProtocol)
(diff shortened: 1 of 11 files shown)


tomvanbussel (approved):
Overall the code changes LGTM. The one thing that I'm wondering about is whether we should duplicate all the type widening tests for both features. I guess that there are so few differences between them that in practice the current situation is fine.

johanl-db:
> Overall the code changes LGTM. The one thing that I'm wondering about is whether we should duplicate all the type widening tests for both features. I guess that there are so few differences between them that in practice the current situation is fine.

That wouldn't bring much value here, I don't think it's worth the price of duplicating and running the tests twice as part of each CI run.

Issue #2738, opened 2024-03-08
[Feature Request] Improve UniForm Hudi support for Lists/Maps and Schema Evolution


the-other-tim-brown:
#### Which Delta project/connector is this regarding?

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (Unfiorm)

### Overview

The UniForm support for Hudi does not properly translate Map & List fields and the testing does not validate that the schema can be evolved in the source Delta Lake table and properly reflected in the Hudi table.

### Motivation

UniForm users will benefit from a more complete translation between Delta Lake and Hudi.

### Further details

The [XTable](https://github.com/apache/incubator-xtable/blob/main/core/src/main/java/io/onetable/avro/AvroSchemaConverter.java) project has already handled such translations and the approach can be ported over to this repo.

### Willingness to contribute

The Delta Lake Community encourages new feature contributions. Would you or another member of your organization be willing to contribute an implementation of this feature?

- [ ] Yes. I can contribute this feature independently.
- [ ] Yes. I would be willing to contribute this feature with guidance from the Delta Lake community.
- [ ] No. I cannot contribute this feature at this time.

anniewang-db:
List/map support has been added in #3320

MrPowers:
Great, looks like this one has been completed!

PR #3320, opened 2024-06-28
[Hudi] Support list/map data type conversion

anniewang-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [x] Other (Hudi)

## Description

This PR adds functionality to convert Delta tables with list or map type columns to be Hudi-readable. 

## How was this patch tested?

Added unit tests in ConvertToHudiSuite and tested manually with external Hudi Spark reader.

Yes. Previously users could not enable the Delta table property for Hudi conversion on tables containing list/map columns and would receive an unsupportedType error but now they can.


harperjiang (approved):
lgtm. Please also comment on the issue https://github.com/delta-io/delta/issues/2738 to indicate that this has been implemented.

PR #3331, opened 2024-07-02
[Spark] Use binary encoding for DV descriptor in file metadata

cstavr:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

The Deletion Vector descriptor of each file is currently serialized in file's custom metadata as a JSON string. We can reduce the size of the descriptor by using a custom binary encoding. Note that the serialized DV descriptor is never persisted to disk so this change is safe.

## How was this patch tested?

Updated existing new tests.

No


larsk-db (approved):
LGTM, thanks for the improvement

PR #3334, opened 2024-07-04
[Spark] Deprecate tableVersion field in type widening metadata

johanl-db:
## Description
The protocol specification for type widening dropped the requirement to populate a `tableVersion` field as part of type change history to track in which version of the table a given change was applied.
See protocol update: https://github.com/delta-io/delta/pull/3297
This field was used at some point during the preview but isn't needed anymore.

It is now deprecated:
- The field is preserved in table metadata that already contains it.
- The field isn't set anymore when the stable table feature is active on the table.
- The field is still set when only the preview table feature is active on the table.

The last point is necessary to avoid breaking preview clients (Delta 3.2 & Delta 4.0 preview) that require the field to be set.

## How was this patch tested?
- Updated existing metadata tests to cover `tableVersion` not being set by default.
- Added metadata tests to explicitly cover `tableVersion` being set.
- Added tests covering `tableVersion` when using the preview and stable table features.

Yes. As of this change, a table that supports the stable table feature `typeWidening` won't have a `tableVersion` field in the type change history stored in the table metadata.

Tables that only support the preview table feature `typeWidening-preview` don't see any change.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMetadata.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMetadata.scala
@@ -26,7 +26,8 @@ import org.apache.spark.sql.types._
 
 /**
  * Information corresponding to a single type change.
- * @param version   The version of the table where the type change was made.
+ * @param version   (Deprecated) The version of the table where the type change was made. This is
+ *                  only populated by clients using the preview of type widening.
  * @param fromType  The original type before the type change.
  * @param toType    The new type after the type change.
  * @param fieldPath The path inside nested maps and arrays to the field where the type change was
@@ -34,7 +35,7 @@ import org.apache.spark.sql.types._
  *                  arrays. The path is empty if the type change was applied inside a map or array.
  */
 private[delta] case class TypeChange(
-    version: Long,
+    version: Option[Long],
     fromType: DataType,
     toType: DataType,
     fieldPath: Seq[String]) {
@@ -43,7 +44,8 @@ private[delta] case class TypeChange(
   /** Serialize this type change to a [[Metadata]] object. */
   def toMetadata: Metadata = {
     val builder = new MetadataBuilder()
-      .putLong(TABLE_VERSION_METADATA_KEY, version)
+    version.foreach(builder.putLong(TABLE_VERSION_METADATA_KEY, _))
+    builder
       .putString(FROM_TYPE_METADATA_KEY, fromType.typeName)
       .putString(TO_TYPE_METADATA_KEY, toType.typeName)
     if (fieldPath.nonEmpty) {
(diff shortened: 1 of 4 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMetadata.scala
```diff
@@ -54,6 +56,8 @@ private[delta] case class TypeChange(
 }
 
 private[delta] object TypeChange {
+  // tableVersion was a field present during the preview and removed afterwards. We preserve it if
+  // it's already present in the type change metadata of the table to avoid breaking older clients.
```
tomvanbussel, line 60: What do they still use it for? Would it be possible to always set it to some constant number (e.g. zero)?

spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMetadata.scala
```diff
@@ -130,11 +139,16 @@ private[delta] object TypeWideningMetadata extends DeltaLogging {
...
-        // Record the version the transaction will attempt to use in the type change metadata. If
-        // there's a conflict with another transaction, the version in the metadata will be updated
-        // during conflict resolution. See [[ConflictChecker.updateTypeWideningMetadata()]].
-        val typeChanges =
-          collectTypeChanges(oldField.dataType, newField.dataType, txn.getFirstAttemptVersion)
+        var typeChanges = collectTypeChanges(oldField.dataType, newField.dataType)
+        // The version field isn't used anymore but we need to populate it in case the table doesn't
+        // use the stable feature, as preview clients may then still access the table and rely on
+        // the field being present.
+        if (!txn.protocol.isFeatureSupported(TypeWideningTableFeature)) {
```
tomvanbussel, line 146: Why do we check for `TypeWideningTableFeature` not being there instead of `TypeWideningPreviewTableFeature` being there?

spark/src/test/scala/org/apache/spark/sql/delta/typewidening/TypeWideningMetadataSuite.scala
```diff
@@ -49,8 +50,9 @@ trait TypeWideningMetadataTests extends QueryTest with DeltaSQLCommandTest {
 
   override protected def beforeAll(): Unit = {
     super.beforeAll()
-    sql(s"CREATE TABLE $testTableName (a int) USING delta " +
-      s"TBLPROPERTIES ('${DeltaConfigs.ENABLE_TYPE_WIDENING.key}' = 'true')")
+    sql(s"CREATE TABLE $testTableName (a int) USING delta TBLPROPERTIES (" +
+      s"'${DeltaConfigs.ENABLE_TYPE_WIDENING.key}' = 'true', " +
+      s"'${propertyKey(TypeWideningTableFeature)}' = 'supported')")
```
tomvanbussel, line 55: Why is this necessary? Isn't the feature automatically enabled when setting the property?

tomvanbussel (approved):
Just a few comments, LGTM otherwise

spark/src/test/scala/org/apache/spark/sql/delta/typewidening/TypeWideningMetadataSuite.scala
johanl-db: The *preview* feature is enabled by default, this forces the stable to be added instead.

I added a comment to clarify that and that we're testing with stable feature by default and only explicit set the preview feature in specific tests that cover the old behavior re: `tableVersion`.
+ added a missing test for the preview feature

bart-samwel (approved):
Looks good to me!

PR #3342, opened 2024-07-08
[Kernel][Writes] Add support for writing data file stats

raveeram-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description
Serializes stats to the data file on writes. 

## How was this patch tested?
Unit tests. 

No


raveeram-db:
Will fix the tests.

kernel/kernel-api/src/main/java/io/delta/kernel/statistics/DataFileStatistics.java
```diff
@@ -0,0 +1,261 @@
...
+      StructType schema,
+      Map<Column, T> values,
+      Column parentColPath,
+      JsonUtil.JsonValueWriter<T> writer)
+      throws IOException {
+    if (schema == null) {
+      return;
+    }
+    for (StructField field : schema.fields()) {
+      Column colPath = parentColPath.append(field.getName());
```
vkorukanti, line 165: you can actually just mantain the path as a list and we can avoid the extra API on the `Column`

kernel/kernel-api/src/main/java/io/delta/kernel/statistics/DataFileStatistics.java
```diff
@@ -0,0 +1,261 @@
...
+      if (field.getDataType() instanceof StructType) {
+        generator.writeObjectFieldStart(field.getName());
+        writeJsonValues(generator, (StructType) field.getDataType(), values, colPath, writer);
+        generator.writeEndObject();
+      } else {
+        T value = values.get(colPath);
+        if (value != null) {
```
vkorukanti, line 172: we may want to type check the Literal matches the data type to make sure to not write incorrect stats

kernel/kernel-api/src/main/java/io/delta/kernel/utils/JsonUtil.java
```diff
@@ -0,0 +1,66 @@
...
+  public static JsonFactory factory() {
+    return FACTORY;
+  }
+
+  public static ObjectMapper mapper() {
+    return OBJECT_MAPPER;
+  }
+
+  @FunctionalInterface
+  public interface ToJson {
```
vkorukanti, line 41: what is this interface for?

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/util/DataFileStatisticsSuite.scala
```diff
@@ -0,0 +1,265 @@
...
+      .add("BinaryType", BinaryType.BINARY)
+      .add("NestedStruct", nestedStructType)
+
+    val minValues = Map(
+      new Column("ByteType") -> Literal.ofByte(1.toByte),
+      new Column("ShortType") -> Literal.ofShort(1.toShort),
+      new Column("IntegerType") -> Literal.ofInt(1),
+      new Column("LongType") -> Literal.ofLong(1L),
+      new Column("FloatType") -> Literal.ofFloat(0.1f),
+      new Column("DoubleType") -> Literal.ofDouble(0.1),
```
vkorukanti, line 71: please add all variants of the double: NaN, infinity. Same for float.

vkorukanti (changes requested):
Looks great!

Also it would be good if we can add an integration tests and make sure it is readable/usable in Kernel/Delta-Spark read tests.

kernel/kernel-api/src/main/java/io/delta/kernel/expressions/Column.java
```diff
@@ -61,6 +61,20 @@ public String toString() {
...
   }
 
+  /**
+   * Returns a new column that appends the input column name to the current column. Corresponds to
+   * an additional level of nested reference.
+   *
+   * @param name the column name to append
+   * @return the new column
+   */
+  public Column append(String name) {
```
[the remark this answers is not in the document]
raveeram-db: thanks, makes sense, updated to `appendNestedField`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -484,6 +497,23 @@ private Optional<CRCInfo> buildPostCommitCrcInfoIfCurrentCrcAvailable(
...
+                    DataSkippingUtils.DATA_SKIPPING_NUM_INDEXED_COLS,
+                    String.valueOf(DataSkippingUtils.DEFAULT_DATA_SKIPPING_NUM_INDEXED_COLS)));
+
+    // Get the list of partition columns to exclude
+    Set<String> partitionColumns =
+        TransactionStateRow.getPartitionColumnsList(transactionState).stream()
+            .collect(Collectors.toSet());
+
+    // For now, only support the first numIndexedCols columns
+    return TransactionStateRow.getLogicalSchema(engine, transactionState).fields().stream()
```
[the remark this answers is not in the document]
raveeram-db: Just the top level columns, looking at the code, `fields` does seem to only traverse top level columns but i'll double check

kernel/kernel-api/src/main/java/io/delta/kernel/statistics/DataFileStatistics.java
raveeram-db: Yeah felt like that'd just be extra state to maintain and tradeoff readability a bit; the API could come in handy in other places as well perhaps. Lmk if you feel strongly and I can update

allisonport-db (commented):
Haven't finished looking but submitting pending comments for now. JSON generation looks very clean!

So far mainly have questions around what kind of validation should we do to prevent bad statistics input from the connector?

Also +1 on integration tests, let's make sure Kernel & Spark can read these statistics and can use them to skip files

kernel/kernel-api/src/main/java/io/delta/kernel/statistics/DataFileStatistics.java
```diff
@@ -0,0 +1,261 @@
...
+   * Get the number of nulls of columns in the data file. The map may contain statistics for only a
+   * subset of columns in the data file.
+   *
+   * @return Map of column to number of nulls in the data file.
+   */
+  public Map<Column, Long> getNullCounts() {
+    return nullCounts;
+  }
+
+  public String serializeAsJson() {
```
allisonport-db, line 118: should we take in the expected schema here (from the txn that is generating the stats) and make sure it's the same? or any other ideas on if/where we should validate the `dataSchema`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -484,6 +497,23 @@ private Optional<CRCInfo> buildPostCommitCrcInfoIfCurrentCrcAvailable(
    */
   public static List<Column> getStatisticsColumns(Engine engine, Row transactionState) {
     // TODO: implement this once we start supporting collecting stats
-    return Collections.emptyList();
+    int numIndexedCols =
+        Integer.parseInt(
+            TransactionStateRow.getConfiguration(transactionState)
+                .getOrDefault(
+                    DataSkippingUtils.DATA_SKIPPING_NUM_INDEXED_COLS,
+                    String.valueOf(DataSkippingUtils.DEFAULT_DATA_SKIPPING_NUM_INDEXED_COLS)));
```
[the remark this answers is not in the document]
scottsand-db: Hey @allisonport-db -- this is a static method. I believe we do have to use the TxnStateRow? Metadata isn't an input to the method.

Let me know if I'm missing something!

kernel/kernel-api/src/main/java/io/delta/kernel/statistics/DataFileStatistics.java
```diff
@@ -0,0 +1,261 @@
...
+
+          gen.writeEndObject();
+        });
+  }
+
+  private <T> void writeJsonValues(
+      JsonGenerator generator,
+      StructType schema,
+      Map<Column, T> values,
+      Column parentColPath,
```
scottsand-db, line 158: nit: `parentCol`

kernel/kernel-api/src/main/java/io/delta/kernel/statistics/DataFileStatistics.java
scottsand-db, line 118: It would be nice to include a method docs comment explaining what the serialized JSON will look like? This will help give an intuition as to what the method is doing.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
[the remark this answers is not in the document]
allisonport-db: oops! ignore me haha :)

kernel/kernel-api/src/main/java/io/delta/kernel/statistics/DataFileStatistics.java
```diff
@@ -0,0 +1,261 @@
...
+   * @param dataSchema Schema of the data file.
+   * @param numRecords Number of records in the data file.
+   * @param minValues Map of column to minimum value of it in the data file. If the data file has
+   *     all nulls for the column, the value will be null or not present in the map.
+   * @param maxValues Map of column to maximum value of it in the data file. If the data file has
+   *     all nulls for the column, the value will be null or not present in the map.
+   * @param nullCounts Map of column to number of nulls in the data file.
+   */
+  public DataFileStatistics(
+      StructType dataSchema,
```
scottsand-db, line 67: can we include any nullability checks?

kernel/kernel-api/src/main/java/io/delta/kernel/statistics/DataFileStatistics.java
```diff
@@ -0,0 +1,261 @@
...
+   *
+   * @param dataSchema Schema of the data file.
+   * @param numRecords Number of records in the data file.
+   * @param minValues Map of column to minimum value of it in the data file. If the data file has
+   *     all nulls for the column, the value will be null or not present in the map.
+   * @param maxValues Map of column to maximum value of it in the data file. If the data file has
+   *     all nulls for the column, the value will be null or not present in the map.
+   * @param nullCounts Map of column to number of nulls in the data file.
+   */
+  public DataFileStatistics(
```
allisonport-db, line 66: I think at a high-level we should define what we expect connectors to input here, and what rules we expect them to follow in order to write correct and useful statistics. And then we should enforce them :) We should treat this constructor as a public API as connectors will need to create these. I think I left general comments all over to this nature but this is kind of the gist of it.

kernel/kernel-api/src/main/java/io/delta/kernel/statistics/DataFileStatistics.java
```diff
@@ -0,0 +1,261 @@
...
+
+  @Override
+  public String toString() {
+    return serializeAsJson();
+  }
+
+  /**
+   * Utility method to deserialize statistics from a JSON string. NOTE: Currently, this method only
+   * deserializes the numRecords field and ignores other statistics (minValues, maxValues,
+   * nullCounts). Full deserialization support for all statistics will be added in a future update.
```
scottsand-db, line 233: Question: This means we do not really have data skipping, right? cc @allisonport-db to confirm my understanding?

kernel/kernel-api/src/main/java/io/delta/kernel/statistics/DataFileStatistics.java
```diff
@@ -0,0 +1,261 @@
...
+   */
+  public DataFileStatistics(
+      StructType dataSchema,
+      long numRecords,
+      Map<Column, Literal> minValues,
+      Map<Column, Literal> maxValues,
+      Map<Column, Long> nullCounts) {
+    this.dataSchema = dataSchema;
+    this.numRecords = numRecords;
+    this.minValues = Collections.unmodifiableMap(minValues);
```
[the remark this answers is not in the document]
allisonport-db: Also `requireNonNull` for all of these? Maybe depends on what we decide are the expectations for connectors. Maybe we should have another constructor that only takes in `numRecords`?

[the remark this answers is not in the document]
allisonport-db: > I think we only store stats for certain column data types right? should we validate that?

Ah I see we throw an error upon writing. I guess we should decide whether that's a good place to do it or if we should do some overall validation upon construction

scottsand-db (commented):
Looks great! Left only a few minor comments / questions.

Seems like we need some tests for `DEFAULT_DATA_SKIPPING_NUM_INDEXED_COLS` and, in particular, need to test + document the expected behavior for how nested columns impact the total column count.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/util/DataFileStatisticsSuite.scala
allisonport-db: I think when I added `DefaultJsonHandlerSuite` I was pretty thorough about edge cases for various types if it helps the tests are here https://github.com/delta-io/delta/blob/master/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/engine/DefaultJsonHandlerSuite.scala

kernel/kernel-api/src/main/java/io/delta/kernel/statistics/DataFileStatistics.java
allisonport-db: Haha I commented a similar thing here. From looking at the code, it seems like this is only used by row tracking during writes to get the numRecords. Our data skipping uses the JsonHandler to parse JSON in a completely different code path. I think this is a fine utility for that specific use case (as it's for 1 add-file, and not over batches) but we should document this better to be clear what it is and what it isn't used for.

raveeram-db: Thank you both, that's right, the data skipping path goes through the JsonHandler and this is only used by row tracking and checked with @qiyuandong-db - they only care about the `numRecords` field in `AddFile` for now. I'll create a task to also add support for deserializing stats but given we won't have the dataSchema here, we'll have to be mindful of safety and all.

allisonport-db: I think maybe we don't need to do that (yet at least)? Is there any requirement for the other stats during writes?

Maybe we can just make this a smaller scoped utility that only gets the numRecords from a stats json?

raveeram-db: None yet, yeah, i'll move this out to a Utils class to avoid confusion

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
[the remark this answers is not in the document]
scottsand-db: I spoke with Rahul Mahadev. Nested fields _should_ count towards the total columnd included. (this is what delta-spark does)

e.g. Suppose I have a table `c1: INT, c2: STRUCT with nested columns sub1, ..., sub33, c2: INT`

Then if I have 32 columns indexed for stats collection, only `c1`, `c2.sub1`, ..., `c2.sub31` will be included.

`c2.sub32, c2.sub33, c3` will be excluded from stats collection.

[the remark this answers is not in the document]
raveeram-db: Ack, yup thanks Scott. I'll update

kernel/kernel-api/src/main/java/io/delta/kernel/statistics/DataFileStatistics.java
```diff
@@ -0,0 +1,261 @@
...
+    } else if (type instanceof ShortType) {
+      generator.writeNumber(((Number) value).shortValue());
+    } else if (type instanceof IntegerType) {
+      generator.writeNumber(((Number) value).intValue());
+    } else if (type instanceof LongType) {
+      generator.writeNumber(((Number) value).longValue());
+    } else if (type instanceof FloatType) {
+      generator.writeNumber(((Number) value).floatValue());
+    } else if (type instanceof DoubleType) {
+      generator.writeNumber(((Number) value).doubleValue());
```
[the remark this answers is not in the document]
raveeram-db: Updated.

kernel/kernel-api/src/main/java/io/delta/kernel/utils/JsonUtil.java
```diff
@@ -0,0 +1,66 @@
+/*
```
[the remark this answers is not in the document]
raveeram-db: Yup, merged the two classes.

kernel/kernel-api/src/main/java/io/delta/kernel/utils/JsonUtil.java
raveeram-db: Ah just to allow for the JsonWriter and Jackson to be decoupled in all call sites (should we ever choose to remove it, although I hope not :))

raveeram-db:
> Looks great! Left only a few minor comments / questions.
> 
> Seems like we need some tests for `DEFAULT_DATA_SKIPPING_NUM_INDEXED_COLS` and, in particular, need to test + document the expected behavior for how nested columns impact the total column count.

Thanks @scottsand-db , added a test and a clarifying comment for the same

kernel/kernel-api/src/main/java/io/delta/kernel/statistics/DataFileStatistics.java
raveeram-db: Thanks added.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/util/DataFileStatisticsSuite.scala
```diff
@@ -0,0 +1,265 @@
...
+import io.delta.kernel.exceptions.KernelException
+import io.delta.kernel.expressions.{Column, Literal}
+import io.delta.kernel.statistics.DataFileStatistics
+import io.delta.kernel.types._
+
+import com.fasterxml.jackson.databind.{JsonNode, ObjectMapper}
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatest.matchers.must.Matchers
+
+class DataFileStatisticsSuite extends AnyFunSuite with Matchers {
```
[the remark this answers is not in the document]
raveeram-db: Makes sense, added.

kernel/kernel-api/src/main/java/io/delta/kernel/statistics/DataFileStatistics.java
raveeram-db: Updated the flow here to ensure that the right schema flows through the kernel.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableConfig.java
```diff
@@ -241,6 +241,16 @@ public class TableConfig<T> {
...
           true);
 
+  /** This table property is the number of indexed columns for data skipping. */
+  public static final TableConfig<Integer> DATA_SKIPPING_NUM_INDEXED_COLS =
+      new TableConfig<>(
+          "delta.dataSkippingNumIndexedCols",
+          "32",
+          Integer::valueOf,
+          value -> value > 0,
+          "needs to be a positive integer.",
```
scottsand-db, line 251: I think we should follow exactly what delta-spark does?

https://github.com/delta-io/delta/blob/50aa783c64c7b727e13ad744f82cebd705b99e01/spark/src/main/scala/org/apache/spark/sql/delta/DeltaConfig.scala#L573

i.e. value needs to be >= -1. -1 means collect for all columns.

In general, I think we should copy more context from the delta-spark config comments, and also mention the relationship with nested fields?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -482,12 +490,38 @@ private Optional<CRCInfo> buildPostCommitCrcInfoIfCurrentCrcAvailable(
...
-    // TODO: implement this once we start supporting collecting stats
-    return Collections.emptyList();
+  public static List<Column> getStatisticsColumns(Row transactionState) {
+    int numIndexedCols =
+        TableConfig.DATA_SKIPPING_NUM_INDEXED_COLS.fromMetadata(
+            TransactionStateRow.getConfiguration(transactionState));
+
+    // Get the list of partition columns to exclude
+    Set<String> partitionColumns =
+        TransactionStateRow.getPartitionColumnsList(transactionState).stream()
```
scottsand-db, line 503: nit: can we just do `new HashSet<>(TransactionStateRow.getPartitionColumnsList(transactionState))` ? it avoids the overhead of all the stream creation

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -303,6 +315,37 @@ private static String escapeDots(String name) {
...
+      Column parentColumn,
+      Set<String> excludedColumns,
+      List<Column> result,
+      int maxColumns) {
+    for (StructField field : schema.fields()) {
+      if (result.size() >= maxColumns) {
+        return;
+      }
+
+      Column currentColumn = null;
```
scottsand-db, line 329: nit: you can do `final Column currentColumn;` // do can avoid assigning to null and then re-assigning later

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/StatsUtils.java
```diff
@@ -0,0 +1,58 @@
...
+   * nullCounts). Full deserialization support for all statistics will be added in a future update.
+   *
+   * @param json Data statistics JSON string to deserialize.
+   * @return An {@link Optional} containing the deserialized {@link DataFileStatistics} if present.
+   * @throws KernelException if JSON parsing fails
+   */
+  public static Optional<DataFileStatistics> deserializeFromJson(String json) {
+    JsonNode root;
+    try {
+      root = new ObjectMapper().readTree(json);
```
scottsand-db, line 43: It's actually better to save ObjectMapper to a static final field once, than to keep recreating new instances of it each invocation

kernel/kernel-api/src/main/java/io/delta/kernel/statistics/DataFileStatistics.java
```diff
@@ -0,0 +1,313 @@
...
+import java.util.Objects;
+
+/** Statistics about data file in a Delta Lake table. */
+public class DataFileStatistics {
+  private final long numRecords;
+  private final Map<Column, Literal> minValues;
+  private final Map<Column, Literal> maxValues;
+  private final Map<Column, Long> nullCounts;
+
+  public static final int MICROSECONDS_PER_SECOND = 1_000_000;
```
scottsand-db, line 48: nit (sorry) -- static variables before member variables

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/AddFile.java
```diff
@@ -99,6 +102,7 @@ public static AddFile convertDataFileStatus(
 
   /** Utility to generate an 'AddFile' row from the given fields. */
   public static Row createAddFileRow(
+      StructType tableSchema,
```
scottsand-db, line 105: logial schema? physial schema?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/AddFile.java
```diff
@@ -125,7 +129,8 @@ public static Row createAddFileRow(
     baseRowId.ifPresent(id -> fieldMap.put(FULL_SCHEMA.indexOf("baseRowId"), id));
     defaultRowCommitVersion.ifPresent(
         version -> fieldMap.put(FULL_SCHEMA.indexOf("defaultRowCommitVersion"), version));
-    stats.ifPresent(stat -> fieldMap.put(FULL_SCHEMA.indexOf("stats"), stat.serializeAsJson()));
+    stats.ifPresent(
+        stat -> fieldMap.put(FULL_SCHEMA.indexOf("stats"), stat.serializeAsJson(tableSchema)));
```
scottsand-db, line 133: serializeAsJson takes the `dataSchema` but here we are passing in the `tableSchema` -- just calling this out. I think this is fine?

kernel/kernel-api/src/main/java/io/delta/kernel/statistics/DataFileStatistics.java
```diff
@@ -0,0 +1,313 @@
...
+import java.nio.charset.StandardCharsets;
+import java.time.Instant;
+import java.time.LocalDate;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Collections;
+import java.util.Map;
+import java.util.Objects;
+
+/** Statistics about data file in a Delta Lake table. */
```
scottsand-db, line 41: I still think it is worth calling out a bit more how this class is used?

`Notably, this class it NOT used during data skipping -- that operations on columnar batches (or something, etc etc etc). Instead, this class is used for XXX YYY`

scottsand-db (changes requested):
Looks great! Left some minor comments -- after those are resolved, LGTM

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableConfig.java
vkorukanti: If that is the defined behavior, lets change that. Should be relatively easy to do that in the current PR or follow up PR is fine too.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
```diff
@@ -34,6 +34,9 @@
 
 public class DataSkippingUtils {
 
+  public static final String DATA_SKIPPING_NUM_INDEXED_COLS = "delta.dataSkippingNumIndexedCols";
+  public static final int DEFAULT_DATA_SKIPPING_NUM_INDEXED_COLS = 32;
```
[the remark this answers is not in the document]
vkorukanti: remove this?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/StatsUtils.java
vkorukanti: +1

kernel/kernel-api/src/main/java/io/delta/kernel/statistics/DataFileStatistics.java
```diff
@@ -0,0 +1,313 @@
...
+   *
+   * @param dataSchema the optional data schema (optional). If provided, all min/max values and null
+   *     counts will be included and validated. If null, only numRecords will be serialized without
+   *     validation.
+   * @return a JSON representation of the statistics.
+   * @throws KernelException if dataSchema is provided and there's a type mismatch between the
+   *     Literal values and the expected types in the schema, or if an unsupported data type is
+   *     found.
+   */
+  public String serializeAsJson(StructType dataSchema) {
```
vkorukanti, line 152: We should take the physical schema as the input. The data files get the physical data and collects stats of the written physical column.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -482,12 +490,38 @@ private Optional<CRCInfo> buildPostCommitCrcInfoIfCurrentCrcAvailable(
...
+    //     ├─ col2 (string)
+    //     └─ col3 (struct)
+    //           ├─ a (int)
+    //           └─ b (double)
+    //
+    // And if 'numIndexedCols' is set to 2 with no partition columns to exclude, then the returned
+    // stats columns
+    // would be: [col1, col2]. If 'col1' were a partition column, the returned list would be:
+    // [col2, col3.a] (assuming col3.a is encountered before col3.b).
+    return SchemaUtils.collectLeafColumns(
```
vkorukanti, line 524: We should get the physical column names and not the logical columns names. The data that is being written to Parquet files has the physical schema and the stats are collected on the physical columns.

vkorukanti (commented):
One major comment handling the physical column names when column mapping is enabled.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/AddFile.java
raveeram-db: yeah I think we need to clean-up some of this terminology and settle on one. I see different variants of table, data, physical and logicalSchema being used. I'll have a look and if it's a bigger change, i'll fix it in a follow-up PR

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/StatsUtils.java
raveeram-db: Yup makes sense, previous util I refactored, i'll fix it.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableConfig.java
raveeram-db: Good point and makes sense. Will update.

raveeram-db: Fix and added a test, thanks folks.

kernel/kernel-api/src/main/java/io/delta/kernel/statistics/DataFileStatistics.java
raveeram-db: makes sense, updated to take the physical schema as input.

kernel/kernel-api/src/main/java/io/delta/kernel/Transaction.java
```diff
@@ -213,6 +215,7 @@ static CloseableIterator<Row> generateAppendActions(
           }
           AddFile addFileRow =
               AddFile.convertDataFileStatus(
+                  TransactionStateRow.getPhysicalSchema(transactionState),
```
vkorukanti, line 218: can this be created at line #210 and we can avoid the cost to construct physical schema for every add file row.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/DataSkippingUtils.java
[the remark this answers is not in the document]
vkorukanti: we still need these?

kernel/kernel-api/src/main/java/io/delta/kernel/statistics/DataFileStatistics.java
vkorukanti: one thing to emphasize in the docs is the stats are for the data written in the file and column names here are the same as the ones used in the Parquet file.

Later on when we move to id based stats, we can update it.

kernel/kernel-api/src/main/java/io/delta/kernel/statistics/DataFileStatistics.java
```diff
@@ -0,0 +1,318 @@
...
+      long numRecords,
+      Map<Column, Literal> minValues,
+      Map<Column, Literal> maxValues,
+      Map<Column, Long> nullCounts) {
+    Objects.requireNonNull(minValues, "minValues must not be null to serialize stats.");
+    Objects.requireNonNull(maxValues, "maxValues must not be null to serialize stats.");
+    Objects.requireNonNull(nullCounts, "nullCounts must not be null to serialize stats.");
+
+    this.numRecords = numRecords;
+    this.minValues = Collections.unmodifiableMap(minValues);
```
vkorukanti, line 76:
```suggestion
    this.minValues = Collections.unmodifiableMap(requireNonNull(minValues, "minValues is null"));
```

vkorukanti: same for others

kernel/kernel-api/src/main/java/io/delta/kernel/statistics/DataFileStatistics.java
```diff
@@ -0,0 +1,318 @@
...
+      }
+    } else if (type instanceof DoubleType) {
+      double d = ((Number) value).doubleValue();
+      if (Double.isNaN(d) || Double.isInfinite(d)) {
+        generator.writeString(String.valueOf(d));
+      } else {
+        generator.writeNumber(d);
+      }
+    } else if (type instanceof StringType) {
+      generator.writeString((String) value);
```
vkorukanti, line 295: do we need to truncate the string length? What does the delta-spark do?

vkorukanti (changes requested):
Looks great overall! Just few last minor comments.

kernel/kernel-api/src/main/java/io/delta/kernel/Transaction.java
raveeram-db: Yup that makes sense. Will update

kernel/kernel-api/src/main/java/io/delta/kernel/statistics/DataFileStatistics.java
raveeram-db: Looks like the truncation happens during stats collection in `delta-spark` and perhaps does make sense for the Kernel to leave that to the connectors/engines to define.

vkorukanti: Does Delta protocol specify any requirements? Also do we need to escape any special characters in the value?

kernel/kernel-api/src/main/java/io/delta/kernel/statistics/DataFileStatistics.java
```diff
@@ -0,0 +1,318 @@
...
+    } else if (type instanceof BinaryType) {
+      generator.writeString(new String((byte[]) value, StandardCharsets.UTF_8));
```
vkorukanti, line 297: does Delta-Spark do the same? Convert as UTF-8 string?

vkorukanti (approved):
LGTM.

Couple of comments on the string/binary serialization to check it conforms to the spec (or Delta-Spark behavior)

kernel/kernel-api/src/main/java/io/delta/kernel/statistics/DataFileStatistics.java
raveeram-db: That's handled by Jackson serialization and delta-spark also appears to follow the same

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -482,12 +497,37 @@ private Optional<CRCInfo> buildPostCommitCrcInfoIfCurrentCrcAvailable(
...
+
+    // Get the list of partition columns to exclude
+    Set<String> partitionColumns =
+        new HashSet<>(TransactionStateRow.getPartitionColumnsList(transactionState));
+
+    // Collect the leaf-level columns for statistics calculation.
+    // This call selects only the first 'numIndexedCols' leaf columns from the logical schema,
+    // excluding any column whose top-level name appears in 'partitionColumns'.
+    // NOTE: Nested columns (i.e. each leaf within a StructType) count individually toward the
+    // numIndexedCols limit (not Map/ArrayTypes - they're not stats compatible types).
```
allisonport-db, line 516: It seems like we do include the complex type columns in this list right? I don't see anywhere that we exclude them

kernel/kernel-api/src/test/scala/io/delta/kernel/TransactionSuite.scala
```diff
@@ -122,7 +123,66 @@ class TransactionSuite extends AnyFunSuite with VectorTestUtils with MockEngineU
...
         }
 
-      assert(actStats === Seq("{\"numRecords\":10}", "{\"numRecords\":20}"))
+      assert(actStats === Seq(
+        "{\"numRecords\":10,\"minValues\":{},\"maxValues\":{},\"nullCounts\":{}}",
+        "{\"numRecords\":20,\"minValues\":{},\"maxValues\":{},\"nullCounts\":{}}"))
+    }
+  }
+
+  Seq(0, -1).foreach { numIndexedCols =>
```
allisonport-db, line 132: I feel like we should also test something like 4 here? To make sure we do in fact limit (and nested columns do count toward that limit)

allisonport-db: Also should we test that partition columns are excluded?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWritesSuite.scala
```diff
@@ -911,6 +914,108 @@ class DeltaTableWritesSuite extends DeltaTableWriteSuiteBase with ParquetSuiteBa
     }
   }
 
+  test("insert into table - write stats and validate they can be read by Spark ") {
```
allisonport-db, line 917: How does this validate that spark can read the stats? I might be missing something but I don't think we use data skipping with spark over the table?

allisonport-db: It would also be nice if we could (easily) validate this for all data types. I'm wondering if we can just copy over a spark test suite that tests data skipping - but generate the base table with Kernel instead?

Not sure how easy/difficult this might be to do; if it's simple it'd be great test coverage

allisonport-db (commented):
Minor comments/questions

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWritesSuite.scala
raveeram-db: Ah `verifyWrittenContent` does a `SELECT * from..` on the table which should also validate that the stats get parsed at least. That's a great idea though and let me see if we can add one similar to that for all types

kernel/kernel-api/src/test/scala/io/delta/kernel/TransactionSuite.scala
raveeram-db: Yup that makes sense, I did add one with 3 but didn't push the changes, will update this in a follow up PR. Thanks Alli.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/StatsUtils.java
```diff
@@ -0,0 +1,57 @@
...
+  /**
+   * Utility method to deserialize statistics from a JSON string. NOTE: Currently, this method only
+   * deserializes the numRecords field and ignores other statistics (minValues, maxValues,
+   * nullCounts). Full deserialization support for all statistics will be added in a future update.
+   *
+   * @param json Data statistics JSON string to deserialize.
+   * @return An {@link Optional} containing the deserialized {@link DataFileStatistics} if present.
+   * @throws KernelException if JSON parsing fails
+   */
+  public static Optional<DataFileStatistics> deserializeFromJson(String json) {
```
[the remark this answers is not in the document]
raveeram-db: Makes sense, will update

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
raveeram-db: Yeah I think this comment needs to be clarified, meant to say these are stats ineligible types but do count towards the limit (Delta-Spark atleast appears to do that)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableConfig.java
```diff
@@ -241,6 +241,23 @@ public class TableConfig<T> {
           "needs to be a boolean.",
           true);
 
+  /**
+   * The number of columns to collect stats on for data skipping. A value of -1 means collecting
+   * stats for all columns.
+   *
+   * <p>For Struct types, all leaf fields count individually toward this limit in depth-first order.
+   * For example, if a table has columns a, b.c, b.d, and e, then the first three indexed columns
+   * would be a, b.c, and b.d. Map and array types are not supported for statistics collection.
```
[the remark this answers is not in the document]
raveeram-db: Ah no, we do not (verified with Delta-Spark as well)

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWritesSuite.scala
allisonport-db: I don't actually know, but I have an inkling that spark might only lazily parse those stats if we do data skipping. So I'm not confident the stats are parsed from the select * query without a filter on stats columns...

raveeram-db: Yeah good point, thanks, will add a few more targeted queries/tests that validate data skipping specifically.

PR #3346, opened 2024-07-09
[Spark] Fix VARCHAR/CHAR to string conversion

dhruvarya-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Context: VARCHAR/CHAR to String column conversion was broken. https://github.com/delta-io/delta/pull/2731 was an attempted fix where we strip CHAR/VARCHAR metadata from columns in `alterDeltaTableCommands.scala` during column conversion if the target type is String.

However, this introduced a regression --- updating any metadata for a CHAR/VARCHAR column will convert it into a StringType. This is because the original fix did not account for the fact that the target data type of a VARCHAR column appears to be a String even during other types of column DDL updates that are not data type changes.

This PR:
1. Removes the changes from https://github.com/delta-io/delta/pull/2731
2. Makes it so that StringType is temporarily converted into CHAR/VARCHAR Type and the CHAR/VARCHAR metadata is stripped during column DDL updates. This fixes the original issue.
3. Guards this new change (2) behind a new flag `alterTable.bypassCharVarcharToStringFix`. In case this fix introduces a new regression, this flag can be flipped.

## How was this patch tested?

Introduces tests for:
1. Adding comment on a varchar column
2. Adding a default for a varchar column
3. Test for the new flag

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/commands/alterDeltaTableCommands.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/commands/alterDeltaTableCommands.scala
@@ -586,16 +586,31 @@ case class AlterTableChangeColumnDeltaCommand(
     recordDeltaOperation(deltaLog, "delta.ddl.alter.changeColumns") {
       val txn = startTransaction()
       val metadata = txn.metadata
-      val oldSchema = metadata.schema
+      val bypassCharVarcharToStringFix =
+        sparkSession.conf.get(DeltaSQLConf.DELTA_BYPASS_CHARVARCHAR_TO_STRING_FIX)
+      val oldSchema = if (bypassCharVarcharToStringFix) {
+          metadata.schema
+        } else {
+          SchemaUtils.getRawSchemaWithoutCharVarcharMetadata(metadata.schema)
+        }
       val resolver = sparkSession.sessionState.conf.resolver
 
       // Verify that the columnName provided actually exists in the schema
       SchemaUtils.findColumnPosition(columnPath :+ columnName, oldSchema, resolver)
 
-      val newSchema = transformSchema(oldSchema, Some(columnName)) {
+      val transformedSchema = transformSchema(oldSchema, Some(columnName)) {
         case (`columnPath`, struct @ StructType(fields), _) =>
           val oldColumn = struct(columnName)
-          verifyColumnChange(sparkSession, struct(columnName), resolver, txn)
+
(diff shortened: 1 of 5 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/commands/alterDeltaTableCommands.scala
```diff
@@ -586,16 +586,31 @@ case class AlterTableChangeColumnDeltaCommand(
...
+        } else {
+          SchemaUtils.getRawSchemaWithoutCharVarcharMetadata(metadata.schema)
+        }
       val resolver = sparkSession.sessionState.conf.resolver
 
       // Verify that the columnName provided actually exists in the schema
       SchemaUtils.findColumnPosition(columnPath :+ columnName, oldSchema, resolver)
 
-      val newSchema = transformSchema(oldSchema, Some(columnName)) {
+      var newSchema = transformSchema(oldSchema, Some(columnName)) {
```
juliuszsompolski, line 600: nit: maybe make it a `val transformedSchema = ...` here
and
```
    val newSchema =  if (!sparkSession.conf.get(DeltaSQLConf.DELTA_BYPASS_CHARVARCHAR_TO_STRING_FIX)) {
        // Fields with type CHAR/VARCHAR had been converted from
        // (StringType, metadata = 'VARCHAR(n)') into (VARCHAR(n), metadata = '')
        // so that CHAR/VARCHAR to String conversion can be handled correctly. We should
        // invert this conversion so that downstream operations are not affected.
        newSchema = CharVarcharUtils.replaceCharVarcharWithStringInSchema(transformedSchema)
      } else {
        transformedSchema
      }
```
below

juliuszsompolski (approved):
Thanks! Looks good to me with optional nit and pending CI failure investigation.

spark/src/main/scala/org/apache/spark/sql/delta/commands/alterDeltaTableCommands.scala
dhruvarya-db: Done. Also, updated the code so that we invoke sparkSession.conf.get(DeltaSQLConf.DELTA_BYPASS_CHARVARCHAR_TO_STRING_FIX) once and store it in a val which is reused:
https://github.com/delta-io/delta/pull/3346/files#diff-3f4163d260854433e36f91a05268df8404b5479a2de432044e732f2563a9e66eR589

PR #3347, opened 2024-07-09
[Spark] Fix CDC Commit Timestamp value under different Timezones

longvu-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

The CDC's `_commit_timestamp` is incorrect when we try to read/display it under a different Spark Session's timezone `spark.sql.session.timeZone` (e.g. `America/Chicago`, `Asia/Ho_Chi_Minh`, ...). 

In this PR, we address this issue by taking into account timezone to capture the precise point in time when we convert `CDCDataSpec`'s [Java Timestamp](https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/commands/cdc/CDCReader.scala#L173) field to [Spark's Timestamp](https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/commands/cdc/CDCReader.scala#L80) for the `_commit_timestamp` column for all CDC's file indexes (`CDCAddFileIndex`, `TahoeRemoveFileIndex`, `TahoeChangeFileIndex`).

This is needed in order for CDF to work properly under a different timezone than `UTC`.

## How was this patch tested?
Added UT, some minor UTs fix to take into account timezone.

Yes.

- CDC's `_commit_timestamp` should now be correct when we try to read/display it under a different Spark Session's timezone `spark.sql.session.timeZone` (e.g.` America/Chicago`, `Asia/Ho_Chi_Minh`, ...).
- This is a user-facing change compared to the released Delta Lake versions and within the unreleased branches such as master.

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/delta/DeltaCDCSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/DeltaCDCSuite.scala
@@ -34,7 +34,8 @@ import org.apache.spark.sql.delta.util.FileNames
 import org.apache.spark.SparkConf
 import org.apache.spark.sql.{AnalysisException, DataFrame, QueryTest, Row}
 import org.apache.spark.sql.catalyst.TableIdentifier
-import org.apache.spark.sql.functions.{col, current_timestamp, floor, lit}
+import org.apache.spark.sql.functions.{col, current_timestamp, floor, lit, unix_timestamp}
+import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.streaming.StreamingQueryException
 import org.apache.spark.sql.test.SharedSparkSession
 import org.apache.spark.sql.types.{LongType, StringType, StructType}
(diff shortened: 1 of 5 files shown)


spark/src/test/scala/org/apache/spark/sql/delta/DeltaCDCSuite.scala
```diff
@@ -250,13 +251,11 @@ abstract class DeltaCDCSuiteBase
       // modify timestamps
       // version 0
       modifyDeltaTimestamp(deltaLog, 0, 0)
-      val tsAfterV0 = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss")
-        .format(new Date(1))
+      val tsAfterV0 = dateFormat.format(new Date(1))
```
longvu-db, line 254: In this PR, we change so that we take into account timezone, these test failed because they didn't take into account timezones, so I'm modifying them to take into account timezone.

larsk-db (approved):
LGTM, thanks

PR #3349, opened 2024-07-09
[Spark] Add list utils to DeltaCommitFileProvider

sumeet-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Add a list util to DeltaCommitFileProvider.

## How was this patch tested?

Build

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/util/DeltaCommitFileProvider.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/util/DeltaCommitFileProvider.scala
@@ -61,6 +61,9 @@ case class DeltaCommitFileProvider(
     }
   }
 
+  /**
+   * Lists unbackfilled delta files in a sorted order without incurring additional IO operations.
+   */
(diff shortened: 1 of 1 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/util/DeltaCommitFileProvider.scala
```diff
@@ -60,6 +60,17 @@ case class DeltaCommitFileProvider(
       case _ => FileNames.unsafeDeltaFile(resolvedPath, version)
     }
   }
+
+  def listSortedUnbackfilledDeltaFiles(startVersionOpt: Option[Long] = None): Seq[(Long, Path)] = {
```
dhruvarya-db, line 64: Can you add a short docstring to at least indicate that this not do any additional IO and only returns what it already knows?

dhruvarya-db (approved):

PR #3350, opened 2024-07-09
Update delta-streaming.md to reflect `maxBytesPerTrigger` is ignored for `Trigger.AvailableNow`

frosforever:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
`maxBytesPerTrigger` is currently ignored for streaming delta table reads when using `Trigger.Once` (see https://github.com/delta-io/delta/issues/843) `Trigger.Once` is now deprecated in favor of `Trigger.AvailableNow` (https://github.com/apache/spark/commit/66b1f79b72855af35351ff995492f2c13872dac5). Based on my testing this behavior is carried over to `AvailableNow` as well. If this is indeed correct and expected behavior then this PR updates the docs. If this is not correct then this is an issue to be resolved and the PR should be rejected.

## How was this patch tested?
Observed behavior of stream while using the different triggers. 

No. Docs only

### DIFF
--- a/docs/source/delta-streaming.md
+++ b/docs/source/delta-streaming.md
@@ -41,7 +41,7 @@ spark.readStream.delta("/tmp/delta/events")
 The following options are available to control micro-batches:
 
 - `maxFilesPerTrigger`: How many new files to be considered in every micro-batch. The default is 1000.
-- `maxBytesPerTrigger`: How much data gets processed in each micro-batch. This option sets a "soft max", meaning that a batch processes approximately this amount of data and may process more than the limit in order to make the streaming query move forward in cases when the smallest input unit is larger than this limit. If you use `Trigger.Once` for your streaming, this option is ignored. This is not set by default.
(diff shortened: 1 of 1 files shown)


frosforever:
On second thought, it appears that neither `maxBytesPerTrigger` nor `maxFilesPerTrigger` is respected regardless of the trigger type for delta data sources in spark similar to what's observed here https://github.com/delta-io/delta/issues/843#issuecomment-1308484029.

frosforever:
Closing as further testing indicates that the options are indeed respected with `AvailableNow`. My original test case was flawed as it reused an existing `checkpointLocation`. While https://spark.apache.org/docs/3.3.1/structured-streaming-programming-guide.html#recovery-semantics-after-changes-in-a-streaming-query indicates that "Addition/deletion/modification of rate limits is allowed", not all sources support that and it appears that delta is one such source. Using a fresh `checkpointLocation` on my test query respected the `maxBytesPerTrigger`. 

Apologies for any confusion with this PR!

PR #3365, opened 2024-07-11
[Spark] Remove timeout source from withTempDir in Delta tests

johanl-db:
## Description
`withTempDir` is widely used across Delta tests to create temporary directories. The spark version in `SQLTestUtil` waits for all running spark tasks to finish with a 10s timeout, but seems to be prone to under or over counting the number of running tasks, sometime causing it to timeout and fail the test.

Change:
- Use Delta's version of `withTempDir`/`withTempPath`/withTempPaths`, which immediately deletes the temp directory when the code returns. This version is already in many Delta tests.

## How was this patch tested?
Test-only change.


xupefei (approved):
Looks good to me!

PR #3366, opened 2024-07-11
[Spark] Batch resolve DeltaMergeAction in ResolveMergeInto

c27kwan:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
In this PR, we batch resolve `DeltaMergeActions` in `ResolveMergeInto` to avoid redundant invocations of the analyzer.

For example, if we have the clause 
```
WHEN NOT MATCHED THEN INSERT (c1, c2) VALUES (src.c1, 22) 
```
Previously, we would call `resolveSingleExprOrFail` which invokes the analyzer 4 times: 1 call to resolve `c1`, 1 call to resolve `src.c1`, 1 call to resolve `c2`, and 1 call to resolve `22`. 
With this PR, we batch the resolution of the target column name parts (`[c1, c2]`) against the target relation and the resolution of assignment expression (`[src.c1, 22]`) together. We can resolve a Merge clause with 2 calls using batching. This helps us with the analyzer performance on wide tables.

## How was this patch tested?
Existing tests pass.

No.

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/delta/MergeIntoSchemaEvolutionSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/MergeIntoSchemaEvolutionSuite.scala
@@ -502,7 +502,7 @@ trait MergeIntoSchemaEvolutionBaseTests {
   testEvolution(s"case-sensitive insert")(
     targetData = Seq((0, 0), (1, 10), (3, 30)).toDF("key", "value"),
     sourceData = Seq((1, 1), (2, 2)).toDF("key", "VALUE"),
-    clauses = insert("(key, value, VALUE) VALUES (s.key, s.value, s.VALUE)") :: Nil,
+    clauses = insert("(key, value) VALUES (s.key, s.value)") :: Nil,
     expectErrorContains = "Cannot resolve s.value in INSERT clause",
     expectErrorWithoutEvolutionContains = "Cannot resolve s.value in INSERT clause",
     confs = Seq(SQLConf.CASE_SENSITIVE.key -> "true")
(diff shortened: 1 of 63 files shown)


spark/src/test/scala/org/apache/spark/sql/delta/MergeIntoSchemaEvolutionSuite.scala
```diff
@@ -502,7 +502,7 @@ trait MergeIntoSchemaEvolutionBaseTests {
   testEvolution(s"case-sensitive insert")(
     targetData = Seq((0, 0), (1, 10), (3, 30)).toDF("key", "value"),
     sourceData = Seq((1, 1), (2, 2)).toDF("key", "VALUE"),
-    clauses = insert("(key, value, VALUE) VALUES (s.key, s.value, s.VALUE)") :: Nil,
+    clauses = insert("(key, value) VALUES (s.key, s.value)") :: Nil,
```
c27kwan, line 505: There were two different possible errors here. It can throw a different message dependent on whether we batch. I removed the ambiguity so that the error is always the same.

c27kwan:
I got rid of the memoization because batching will make memoization at that level obsolete. The opportunity for memoization within merge at that level is probably not very useful. We don't resolve the same expression multiple times against the same plans, unless we have many different conditional when matched/when not matched clauses.

spark/src/test/scala/org/apache/spark/sql/delta/MergeIntoSchemaEvolutionSuite.scala
c27kwan: I've added a TODO for the other test, since it's an unrelated new test and we didn't have coverage for it before.

larsk-db (approved):
LGTM

PR #3368, opened 2024-07-12
[Spark] Expose uncertainty in isDeltaTable resolution via exceptions

larsk-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

- Add a new codepath (disabled by default except in testing) to `findDeltaTableRoot` (which is the implementation of `isDeltaTable`), that throws the first exception if the passed in path is not accessible, and no accessible parent is found to contain a `_delta_log` folder.
- This improves error reporting on permission errors, which current often get misreported as `t is not a DeltaTable`.
- This also reduces the risk for transient cloud storage errors causing us to silently resolve a DeltaTable as a non-Delta table, which can lead to inconsistencies in how we process it when we check multiple times.
  - Note that this is still not ideal, we could still end up with a situation where we the first path is accessible but a child of the Delta Table, such as a partition, but then when we check the path that actually does contain the `_delta_log` we get a transient error and wrongly resolve it as not a Delta table. Ultimately, this kind of parent resolution really has been and continues to be best-effort. But at least the case where we are pointed at the correct folder behaves more sane now.

The new path is only enabled in unit tests for now. Flag flip PR will follow separately.

## How was this patch tested?

Existing tests (showing that there is no behaviour change in the "happy path").

No, they will only apply on the flag-flip PR.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaTable.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaTable.scala
@@ -18,13 +18,16 @@ package org.apache.spark.sql.delta
 
 // scalastyle:off import.ordering.noEmptyLine
 import scala.util.{Failure, Success, Try}
+import scala.util.control.NonFatal
 
 import org.apache.spark.sql.delta.files.{TahoeFileIndex, TahoeLogFileIndex}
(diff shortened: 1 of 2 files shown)


xupefei (approved):
Looks good to me, thank you for making this improvement!

PR #3373, opened 2024-07-14
[Spark] Write INT64 by default for Timestamps

Fokko:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

- Iceberg states in the spec that Timestamps should be written with INT64 physical types. There were already flags to enable this, but this PR makes this behavior the default.
- `INT96` is discouraged: https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift#L981

## How was this patch tested?

- Added new unit-tests to validate the correct behavior

Yes, it will default to INT64 if nothing has been set.


lzlfred (approved):
LGTM ! lets stick with parquet spec !

felipepessoto:
@Fokko @lzlfred should we update the protocol to recommend INT64?

Context: https://github.com/delta-io/delta/pull/2048/files#r1476851129

PR #3374, opened 2024-07-14
[Spark] Only attempt parsing partition path into different types if type inference is enabled

amogh-jahagirdar:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
I was looking at this code path for another issue and noticed that when parsing partition values from paths (only used for the write path) without type inference we still attempt to due parsing for date/decima/timestamp types even though it's not required. We can avoid that work with a small refactoring.

## How was this patch tested?
This is a refactoring/minor optimization where existing unit tests will exercise this path

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/util/PartitionUtils.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/util/PartitionUtils.scala
@@ -528,7 +528,7 @@ private[delta] object PartitionUtils {
       timeZone: TimeZone,
       dateFormatter: DateFormatter,
       timestampFormatter: TimestampFormatter): Literal = {
-    val decimalTry = Try {
+    def decimalTry = Try {
       // `BigDecimal` conversion can fail when the `field` is not a form of number.
       val bigDecimal = new JBigDecimal(raw)
       // It reduces the cases for decimals by disallowing values having scale (eg. `1.1`).
@@ -539,7 +539,7 @@ private[delta] object PartitionUtils {
       Literal(bigDecimal)
     }
 
-    val dateTry = Try {
+    def dateTry = Try {
       // try and parse the date, if no exception occurs this is a candidate to be resolved as
       // DateType
       dateFormatter.parse(raw)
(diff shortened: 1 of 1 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/util/PartitionUtils.scala
```diff
@@ -528,46 +528,46 @@ private[delta] object PartitionUtils {
...
-      // `BigDecimal` conversion can fail when the `field` is not a form of number.
-      val bigDecimal = new JBigDecimal(raw)
-      // It reduces the cases for decimals by disallowing values having scale (eg. `1.1`).
-      require(bigDecimal.scale <= 0)
-      // `DecimalType` conversion can fail when
-      //   1. The precision is bigger than 38.
-      //   2. scale is bigger than precision.
-      Literal(bigDecimal)
-    }
+    if (typeInference) {
```
scottsand-db, line 531: Good refactor.

You could also make the `val decimalTry` a `def` instead so that it is only evaluated if necessary.

spark/src/main/scala/org/apache/spark/sql/delta/util/PartitionUtils.scala
```diff
@@ -528,46 +528,46 @@ private[delta] object PartitionUtils {
...
+        val timestampValue = Cast(Literal(unescapedRaw), TimestampType, Some(timeZone.getID)).eval()
+        // Disallow TimestampType if the cast returned null
+        require(timestampValue != null)
+        Literal.create(timestampValue, TimestampType)
+      }
 
-    if (typeInference) {
       // First tries integral types
       Try(Literal.create(Integer.parseInt(raw), IntegerType))
         .orElse(Try(Literal.create(JLong.parseLong(raw), LongType)))
```
scottsand-db, line 573: you could also make the Long try and the Double try also be `def`s

spark/src/main/scala/org/apache/spark/sql/delta/util/PartitionUtils.scala
amogh-jahagirdar: Good suggestion, I like the def better it makes the diff easier on the eyes

scottsand-db (approved):

PR #3378, opened 2024-07-15
[Protocol, Spark] UTC normalize timestamp partition values

amogh-jahagirdar:

- [ X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Currently, in the Delta Protocol, timestamps are not stored with their time zone. This leads to unexpected behavior when querying across systems with different timezones configured (e.g. different spark sessions for instance). For instance in Spark, the timestamp value will be adjusted to spark session time zone and written to the delta log partition values without TZ. If someone were to query the same "timestamp" from a different session timezone,  the same time zone value it can fail to surface results due to partition pruning.

What this change proposes to the delta lake protocol is to allow timestamp partition values to be adjusted to UTC and explicitly stored in partition values with a UTC suffix. The original approach is still supported for compatibility but it is recommended for newer writers to write with UTC suffix.

This is also important for Iceberg Uniform conversion because Iceberg timestamps must be UTC adjusted. Now we have a well defined format for UTC in delta, we can convert string partition values to Iceberg longs to make Uniform conversion succeed.

This change updates the Spark-Delta integration to write out the UTC adjusted values for timestamp types.

This also addresses an issue of microsecond partitions where previously microsecond partitioning (not recommended but technically allowed) would not work and be truncated to seconds.

## How was this patch tested?

Added unit tests for the following cases:

1.) UTC timestamp partition values round trip across different session TZ
2.) A delta log with a mix of Non-UTC and UTC partition values round trip across the same session TZ
3.) Timestamp No Timezone round trips across timezones (kind of a tautology but important to make sure that the timestamp_ntz does not get written with UTC timestamp unintentionally)
4.) Timestamp round trips across same session time zone: UTC normalized
5.) Timestamp round trips across same session time zone: session time normalized (this case worked before this change, so it's important that it keeps working after this change)

Mix of microsecond/second level precision and dates before epoch (to test if everything works with negative)

Yes in the sense that new timestamp partition values will be the normalized UTC values.

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -1791,13 +1791,16 @@ Type | Serialization Format
 string | No translation required
 numeric types | The string representation of the number
 date | Encoded as `{year}-{month}-{day}`. For example, `1970-01-01`
-timestamp | Encoded as `{year}-{month}-{day} {hour}:{minute}:{second}` or `{year}-{month}-{day} {hour}:{minute}:{second}.{microsecond}` For example: `1970-01-01 00:00:00`, or `1970-01-01 00:00:00.123456`
+timestamp | Encoded as `{year}-{month}-{day} {hour}:{minute}:{second}` or `{year}-{month}-{day} {hour}:{minute}:{second}.{microsecond}`. For example: `1970-01-01 00:00:00`, or `1970-01-01 00:00:00.123456`. Timestamps may also be encoded as an ISO8601 formatted timestamp adjusted to UTC timestamp such as `1970-01-01T00:00:00.123456Z`
 timestamp without timezone | Encoded as `{year}-{month}-{day} {hour}:{minute}:{second}` or `{year}-{month}-{day} {hour}:{minute}:{second}.{microsecond}` For example: `1970-01-01 00:00:00`, or `1970-01-01 00:00:00.123456` To use this type, a table must support a feature `timestampNtz`. See section [Timestamp without timezone (TimestampNtz)](#timestamp-without-timezone-timestampNtz) for more information.
 boolean | Encoded as the string "true" or "false"
 binary | Encoded as a string of escaped binary values. For example, `"\u0001\u0002\u0003"`
 
-Note: A `timestamp` value in a partition value doesn't store the time zone due to historical reasons.
-It means its behavior looks similar to `timestamp without time zone` when it is used in a partition column.
+Note: A timestamp value in a partition value may be stored in one of the following ways:
+1. Without a timezone, where the timestamp should be interpreted using the time zone of the system which wrote to the table.
+2. Adjusted to UTC and stored in ISO8601 format.
+
+It is highly recommended that modern writers adjust the timestamp to UTC and store the timestamp in ISO8601 format as outlined in 2.
 
 ## Schema Serialization Format
 
--- a/spark/src/main/scala/org/apache/spark/sql/delta/files/DelayedCommitProtocol.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/files/DelayedCommitProtocol.scala
@@ -33,7 +33,9 @@ import org.apache.spark.internal.Logging
 import org.apache.spark.internal.io.FileCommitProtocol
 import org.apache.spark.internal.io.FileCommitProtocol.TaskCommitMessage
 import org.apache.spark.sql.catalyst.expressions.Cast
-import org.apache.spark.sql.types.StringType
+import org.apache.spark.sql.delta.files.DeltaFileFormatWriter.PartitionedTaskAttemptContextImpl
+import org.apache.spark.sql.delta.sources.DeltaSQLConf
+import org.apache.spark.sql.types.{DataType, StringType, TimestampType}
 
 /**
  * Writes out the files to `path` and returns a list of them in `addedStatuses`. Includes
@@ -67,7 +69,7 @@ class DelayedCommitProtocol(
   // since there's no guarantee the stats will exist.
   @transient val addedStatuses = new ArrayBuffer[AddFile]
 
-  val timestampPartitionPattern = "yyyy-MM-dd HH:mm:ss[.S]"
+  val timestampPartitionPattern = "yyyy-MM-dd HH:mm:ss[.SSSSSS][.S]"
 
   // Constants for CDC partition manipulation. Used only in newTaskTempFile(), but we define them
   // here to avoid building a new redundant regex for every file.
@@ -123,23 +125,48 @@ class DelayedCommitProtocol(
     }
   }
 
-  protected def parsePartitions(dir: String): Map[String, String] = {
-    // TODO: timezones?
+  protected def parsePartitions(
+      dir: String,
+      taskContext: TaskAttemptContext): Map[String, String] = {
     // TODO: enable validatePartitionColumns?
+    val useUtcNormalizedTimestamps = taskContext match {
+      case _: PartitionedTaskAttemptContextImpl => taskContext.getConfiguration.getBoolean(
+        DeltaSQLConf.UTC_TIMESTAMP_PARTITION_VALUES.key, true)
+      case _ => false
+    }
+
+    val partitionColumnToDataType: Map[String, DataType] =
+      taskContext.asInstanceOf[PartitionedTaskAttemptContextImpl]
+        .partitionColToDataType
+        .filter(partitionCol => partitionCol._2 == TimestampType)
+
     val dateFormatter = DateFormatter()
     val timestampFormatter =
       TimestampFormatter(timestampPartitionPattern, java.util.TimeZone.getDefault)
+
+    /**
+     * ToDo: Remove the use of this PartitionUtils API with type inference logic
+     * since the types are already known from the Delta metadata!
+     *
+     * Currently types are passed to the PartitionUtils.parsePartition API to facilitate
+     * timestamp conversion to UTC. In all other cases, the type is just inferred as a String.
+     * Note: the passed in timestampFormatter and timezone detail
+     * is used for parsing from the string timestamp.
+     * If utc normalization is enabled the parsed partition value will be adjusted to UTC
+     * and output in iso8601 format.
+     */
     val parsedPartition =
       PartitionUtils
         .parsePartition(
           new Path(dir),
           typeInference = false,
           Set.empty,
-          Map.empty,
+          userSpecifiedDataTypes = partitionColumnToDataType,
           validatePartitionColumns = false,
           java.util.TimeZone.getDefault,
           dateFormatter,
-          timestampFormatter)
+          timestampFormatter,
+          useUtcNormalizedTimestamps)
         ._1
         .get
     parsedPartition
@@ -164,7 +191,8 @@ class DelayedCommitProtocol(
    */
   override def newTaskTempFile(
       taskContext: TaskAttemptContext, dir: Option[String], ext: String): String = {
-    val partitionValues = dir.map(parsePartitions).getOrElse(Map.empty[String, String])
+    val partitionValues = dir.map(dir => parsePartitions(dir, taskContext))
+      .getOrElse(Map.empty[String, String])
     val filename = getFileName(taskContext, ext, partitionValues)
     val relativePath = randomPrefixLength.map { prefixLength =>
       DeltaUtils.getRandomPrefix(prefixLength) // Generate a random prefix as a first choice
--- a/spark/src/main/scala/org/apache/spark/sql/delta/files/DeltaFileFormatWriter.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/files/DeltaFileFormatWriter.scala
@@ -44,6 +44,7 @@ import org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanExec
 import org.apache.spark.sql.execution.datasources._
 import org.apache.spark.sql.execution.datasources.FileFormatWriter._
 import org.apache.spark.sql.internal.SQLConf
+import org.apache.spark.sql.types.DataType
 import org.apache.spark.util.{SerializableConfiguration, Utils}
 
 /**
@@ -261,6 +262,8 @@ object DeltaFileFormatWriter extends LoggingShims {
 
       val jobTrackerID = SparkHadoopWriterUtils.createJobTrackerID(new Date())
       val ret = new Array[WriteTaskResult](rddWithNonEmptyPartitions.partitions.length)
+      val partitionColumnToDataType = description.partitionColumns
+        .map(attr => (attr.name, attr.dataType)).toMap
       sparkSession.sparkContext.runJob(
         rddWithNonEmptyPartitions,
         (taskContext: TaskContext, iter: Iterator[InternalRow]) => {
@@ -272,7 +275,8 @@ object DeltaFileFormatWriter extends LoggingShims {
             sparkAttemptNumber = taskContext.taskAttemptId().toInt & Integer.MAX_VALUE,
             committer,
             iterator = iter,
-            concurrentOutputWriterSpec = concurrentOutputWriterSpec
+            concurrentOutputWriterSpec = concurrentOutputWriterSpec,
+            partitionColumnToDataType
           )
         },
         rddWithNonEmptyPartitions.partitions.indices,
@@ -377,6 +381,14 @@ object DeltaFileFormatWriter extends LoggingShims {
     }
   }
 
+   class PartitionedTaskAttemptContextImpl(
+       conf: Configuration,
+       taskId: TaskAttemptID,
+       partitionColumnToDataType: Map[String, DataType])
+     extends TaskAttemptContextImpl(conf, taskId) {
+     val partitionColToDataType: Map[String, DataType] = partitionColumnToDataType
+  }
+
   /** Writes data out in a single Spark task. */
   private def executeTask(
       description: WriteJobDescription,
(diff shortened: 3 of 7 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/files/DeltaFileFormatWriter.scala
```diff
@@ -377,6 +378,13 @@ object DeltaFileFormatWriter extends LoggingShims {
     }
   }
 
+   class PartitionedTaskAttemptContextImpl(conf: Configuration,
```
amogh-jahagirdar, line 381: Needed a way to actually send the types of the partition columns as part of the task context, that's important to distinguish the timestamp vs timestamp_ntz cases for this change.

spark/src/main/scala/org/apache/spark/sql/delta/util/TimestampFormatter.scala
```diff
@@ -67,15 +67,15 @@ sealed trait TimestampFormatter extends Serializable {
 
 class Iso8601TimestampFormatter(
     pattern: String,
-    timeZone: TimeZone,
+    timeZone: ZoneId,
```
amogh-jahagirdar, line 70: I don't think changing this poses a compatibility concern as long as the various apply methods still are compatible (My scala is a bit rusty so let me know, but I'm unable to initialize this outside anyways)?

As for why the change:

Spark SQL supports setting any zoneID like "UTC-08:00" etc. However, if one tries to convert this into java timezone type via TimeZone.of(ZoneId.of("UTC-08:00")) then the timezone actualy ends up being just "UTC" because timezone API can't handle the offset -08:00. So passing a zoneID avoids a correctness issue there.

spark/src/main/scala/org/apache/spark/sql/delta/util/PartitionUtils.scala
```diff
@@ -528,46 +545,47 @@ private[delta] object PartitionUtils {
...
-      // so if there are extra-characters after the date, it returns correctly.
-      // We need to check that we can cast the raw string since we later can use Cast to get
-      // the partition values with the right DataType (see
-      // org.apache.spark.sql.execution.datasources.PartitioningAwareFileIndex.inferPartitioning)
-      val dateValue = Cast(Literal(raw), DateType).eval()
-      // Disallow DateType if the cast returned null
-      require(dateValue != null)
-      Literal.create(dateValue, DateType)
-    }
+    if (typeInference) {
```
amogh-jahagirdar, line 549: This is a refactoring to only attempt the attempts to parse into a specific type only if type inference is enabled. I raised it separately: https://github.com/delta-io/delta/pull/3374

after that gets in first, I can rebase so this change is easier to review

spark/src/main/scala/org/apache/spark/sql/delta/util/PartitionUtils.scala
```diff
@@ -279,7 +280,9 @@ private[delta] object PartitionUtils {
       validatePartitionColumns: Boolean,
       timeZone: TimeZone,
       dateFormatter: DateFormatter,
-      timestampFormatter: TimestampFormatter): (Option[PartitionValues], Option[Path]) = {
+      timestampFormatter: TimestampFormatter,
+      utcNormalizeTimestamps: Boolean = false,
+      skipCastingToUserType: Boolean = false): (Option[PartitionValues], Option[Path]) = {
```
amogh-jahagirdar, line 285: I will try and simplify this API as best I can, it's still backwards compatible in its current state but dislike having to specify 2 more boolean parameters.

spark/src/main/scala/org/apache/spark/sql/delta/util/PartitionUtils.scala
```diff
@@ -117,7 +117,8 @@ object PartitionSpec {
 
 private[delta] object PartitionUtils {
 
-  val timestampPartitionPattern = "yyyy-MM-dd HH:mm:ss[.S]"
+  lazy val timestampPartitionPattern = "yyyy-MM-dd HH:mm:ss[.S]"
+  lazy val utcFormatter = TimestampFormatter("yyyy-MM-dd HH:mm:ss[.S]z", ZoneId.of("UTC"))
```
amogh-jahagirdar, line 121: Another aspect to consider:

Delta supports microsecond timestamps, and those can also be used for partitioning. Kernel stores in microseconds https://github.com/delta-io/delta/blob/573a57f62918d0cb8937ca8c9b4047f8d696cc9c/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/PartitionUtils.java#L45 and the same in Iceberg itself since the partition tuple is always just a timestamp type microseconds since epoch.

Perhaps we should do the same here?

spark/src/main/scala/org/apache/spark/sql/delta/files/DeltaFileFormatWriter.scala
amogh-jahagirdar: If there's another way to get the partition column type details without this addition, that'd be awesome but didn't see any where else to get these detials

spark/src/main/scala/org/apache/spark/sql/delta/util/PartitionUtils.scala
```diff
@@ -117,7 +117,8 @@ object PartitionSpec {
 
 private[delta] object PartitionUtils {
 
-  val timestampPartitionPattern = "yyyy-MM-dd HH:mm:ss[.S]"
+  lazy val timestampPartitionPattern = "yyyy-MM-dd HH:mm:ss[.S]"
+  lazy val utcFormatter = TimestampFormatter("yyyy-MM-dd HH:mm:ss.SSSSSSz", ZoneId.of("UTC"))
```
amoghjaha-db, line 121: Normalized partition values will be in microseconds, similar to what Kernel is doing https://github.com/delta-io/delta/blob/573a57f62918d0cb8937ca8c9b4047f8d696cc9c/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/PartitionUtils.java#L45

amogh-jahagirdar, line 121: Normalizes to microseconds like in Delta kernel: https://github.com/delta-io/delta/blob/573a57f62918d0cb8937ca8c9b4047f8d696cc9c/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/PartitionUtils.java#L45

PROTOCOL.md
```diff
@@ -1754,13 +1754,16 @@ Type | Serialization Format
...
 boolean | Encoded as the string "true" or "false"
 binary | Encoded as a string of escaped binary values. For example, `"\u0001\u0002\u0003"`
 
-Note: A `timestamp` value in a partition value doesn't store the time zone due to historical reasons.
-It means its behavior looks similar to `timestamp without time zone` when it is used in a partition column.
+Note: A `timestamp` value in a partition value may be stored in one of the following ways:
+1. Without a timezone, where the timestamp should be interpreted using the time zone of the system which wrote to the table.
+2. Adjusted to UTC where the partition value must have the suffix "UTC".
+
+It is highly recommended that modern writers adjust the timestamp value to UTC as outlined in 2.
```
amogh-jahagirdar, line 1766: If we'd like to do the protocol changes separately, happy to do so, just let me know @scottsand-db @allisonport-db @vkorukanti

spark/src/main/scala/org/apache/spark/sql/delta/files/DelayedCommitProtocol.scala
```diff
@@ -67,7 +70,7 @@ class DelayedCommitProtocol(
   // since there's no guarantee the stats will exist.
   @transient val addedStatuses = new ArrayBuffer[AddFile]
 
-  val timestampPartitionPattern = "yyyy-MM-dd HH:mm:ss[.S]"
+  val timestampPartitionPattern = "yyyy-MM-dd HH:mm:ss[.SSSSSS][.S]"
```
amogh-jahagirdar, line 73: This also addresses an issue of microsecond partitions; not that I'd reccomend identity partitions on microseconds due to extremely small files but it is technically supported by the delta protocol.

spark/src/main/scala/org/apache/spark/sql/delta/files/DelayedCommitProtocol.scala
```diff
@@ -164,7 +183,8 @@ class DelayedCommitProtocol(
    */
   override def newTaskTempFile(
       taskContext: TaskAttemptContext, dir: Option[String], ext: String): String = {
-    val partitionValues = dir.map(parsePartitions).getOrElse(Map.empty[String, String])
+    val partitionValues = dir.map(dir => parsePartitions(dir, taskContext))
```
Fokko, line 186: Scala esthetics nit. You could use currying: 
```suggestion
    val partitionValues = dir.map(parsePartitions(taskContext))
```

And, then the signature of the parsePartitions would look like:
```scala
protected def parsePartitions(taskContext: TaskAttemptContext)(dir: String): Map[String, String] = {
```

spark/src/test/scala/org/apache/spark/sql/delta/DeltaInsertIntoTableSuite.scala
```diff
@@ -18,17 +18,14 @@ package org.apache.spark.sql.delta
 
 // scalastyle:off import.ordering.noEmptyLine
 import java.io.File
-
 import scala.collection.JavaConverters._
```
Fokko, line 21: I think we want to minimize the number of unrelated changes

spark/src/main/scala/org/apache/spark/sql/delta/files/DelayedCommitProtocol.scala
```diff
@@ -123,33 +126,49 @@ class DelayedCommitProtocol(
...
         .get
     parsedPartition
-        .columnNames
-        .zip(
-          parsedPartition
-            .literals
-            .map(l => Cast(l, StringType).eval())
-            .map(Option(_).map(_.toString).orNull))
-        .toMap
+      .columnNames
```
Fokko, line 165: I think we want to minimize the number of unrelated changes

spark/src/main/scala/org/apache/spark/sql/delta/files/DelayedCommitProtocol.scala
```diff
@@ -123,33 +126,49 @@ class DelayedCommitProtocol(
...
     val parsedPartition =
       PartitionUtils
         .parsePartition(
           new Path(dir),
           typeInference = false,
           Set.empty,
-          Map.empty,
+          partitionCols,
           validatePartitionColumns = false,
           java.util.TimeZone.getDefault,
```
bart-samwel, line 157: Interestingly, this is the timezone that is to be used to _parse the input_. It is worth a comment that this is different from the output formatting, and that we expect Spark to produce the partitions using the JVM default timezone. (This probably means that Spark has the same issue with partition columns on any other table.)

spark/src/main/scala/org/apache/spark/sql/delta/files/DelayedCommitProtocol.scala
```diff
@@ -123,33 +126,49 @@ class DelayedCommitProtocol(
...
     // TODO: enable validatePartitionColumns?
+    val utcTimestampPartitionValues = taskContext.getConfiguration.getBoolean(
+      DeltaSQLConf.UTC_TIMESTAMP_PARTITION_VALUES.key, true) &&
+      taskContext.isInstanceOf[PartitionedTaskAttemptContextImpl] &&
+      taskContext.asInstanceOf[PartitionedTaskAttemptContextImpl]
+        .partitionAttr.exists(attribute => attribute.dataType.isInstanceOf[TimestampType])
+    val sessionTimeZone = getZoneId(SQLConf.get.sessionLocalTimeZone)
+    val partitionCols: Map[String, DataType] = if (utcTimestampPartitionValues) {
+      taskContext.asInstanceOf[PartitionedTaskAttemptContextImpl]
+          .partitionAttr.groupBy(attr => attr.name)
```
bart-samwel, line 140: This is for user-defined data types for partition columns. To preserve existing behavior as much as possible (and therefore add as little risk as possible), we should only include timestamp-valued columns in this list. The code will use the existing inference logic for any other columns.

spark/src/main/scala/org/apache/spark/sql/delta/files/DelayedCommitProtocol.scala
```diff
@@ -123,33 +126,49 @@ class DelayedCommitProtocol(
...
+          .map(tup => (tup._1, tup._2.head.dataType))
+    } else {
+      Map.empty
+    }
+
     val dateFormatter = DateFormatter()
     val timestampFormatter =
-      TimestampFormatter(timestampPartitionPattern, java.util.TimeZone.getDefault)
+      TimestampFormatter(timestampPartitionPattern, sessionTimeZone)
```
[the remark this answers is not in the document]
amogh-jahagirdar: That's a great point. I'll leave it as is

spark/src/main/scala/org/apache/spark/sql/delta/files/DelayedCommitProtocol.scala
```diff
@@ -123,33 +126,49 @@ class DelayedCommitProtocol(
...
     val parsedPartition =
       PartitionUtils
         .parsePartition(
           new Path(dir),
           typeInference = false,
           Set.empty,
-          Map.empty,
+          partitionCols,
```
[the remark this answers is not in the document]
amogh-jahagirdar: +100, I'll add some comments explaining all of this since it's def not an intuitive API to work with and in the end we probably want to get away from using it since we know the exact partition column data types for Delta.

PROTOCOL.md
```diff
@@ -1754,13 +1754,16 @@ Type | Serialization Format
 string | No translation required
 numeric types | The string representation of the number
 date | Encoded as `{year}-{month}-{day}`. For example, `1970-01-01`
-timestamp | Encoded as `{year}-{month}-{day} {hour}:{minute}:{second}` or `{year}-{month}-{day} {hour}:{minute}:{second}.{microsecond}` For example: `1970-01-01 00:00:00`, or `1970-01-01 00:00:00.123456`
+timestamp | Encoded as `{year}-{month}-{day} {hour}:{minute}:{second}` or `{year}-{month}-{day} {hour}:{minute}:{second}.{microsecond}` For example: `1970-01-01 00:00:00`, or `1970-01-01 00:00:00.123456` or `2024-06-14 04:00:00UTC` or `2024-06-14 04:00:00.123456UTC`
```
lzlfred, line 1757: plz add the Z postfix to the protocol.

spark/src/main/scala/org/apache/spark/sql/delta/files/DelayedCommitProtocol.scala
```diff
@@ -123,33 +126,49 @@ class DelayedCommitProtocol(
...
 
-  protected def parsePartitions(dir: String): Map[String, String] = {
-    // TODO: timezones?
+  protected def parsePartitions(dir: String,
+                                taskContext: TaskAttemptContext): Map[String, String] = {
     // TODO: enable validatePartitionColumns?
+    val utcTimestampPartitionValues = taskContext.getConfiguration.getBoolean(
+      DeltaSQLConf.UTC_TIMESTAMP_PARTITION_VALUES.key, true) &&
+      taskContext.isInstanceOf[PartitionedTaskAttemptContextImpl] &&
+      taskContext.asInstanceOf[PartitionedTaskAttemptContextImpl]
```
lzlfred, line 135: plz avoid as/isInstanceOf; in almost all cases they can be replaced by pattern matching. 

val utcTimestampPartitionValues = taskContext match {
   case context:PartitionedTaskAttemptContextImpl if DeltaSQLConf....  =>
   case _ => Map.empty
}

spark/src/main/scala/org/apache/spark/sql/delta/files/DeltaFileFormatWriter.scala
```diff
@@ -377,6 +378,13 @@ object DeltaFileFormatWriter extends LoggingShims {
...
+                                           taskId: TaskAttemptID,
```
lzlfred, line 382: indentation.

spark/src/main/scala/org/apache/spark/sql/delta/files/TransactionalWrite.scala
```diff
@@ -17,7 +17,6 @@
 package org.apache.spark.sql.delta.files
 
 import scala.collection.mutable.ListBuffer
```
lzlfred, line 19: revert changes in this file.

spark/src/main/scala/org/apache/spark/sql/delta/util/PartitionUtils.scala
```diff
@@ -360,12 +364,21 @@ private[delta] object PartitionUtils {
           dateFormatter,
           timestampFormatter)
         val columnValue = columnValueLiteral.eval()
-        val castedValue = Cast(columnValueLiteral, dataType, Option(timeZone.getID)).eval()
-        if (validatePartitionColumns && columnValue != null && castedValue == null) {
-          throw DeltaErrors.partitionColumnCastFailed(
-            columnValue.toString, dataType.toString, columnName)
+        if (dataType == DataTypes.TimestampType && utcNormalizeTimestamp) {
```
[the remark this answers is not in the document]
amogh-jahagirdar: I think this ended up being answered in https://github.com/delta-io/delta/pull/3378#discussion_r1686554387 but let me know if this part is still unclear.

spark/src/main/scala/org/apache/spark/sql/delta/files/DelayedCommitProtocol.scala
amogh-jahagirdar: I've commented this section in more detail!

spark/src/main/scala/org/apache/spark/sql/delta/files/DelayedCommitProtocol.scala
amogh-jahagirdar: Agreed, that minimizes the risk by preserving the existing behavior as much as possible

amogh-jahagirdar:
This change depends on the protocol changes from https://github.com/delta-io/delta/pull/3398/commits

lzlfred (approved):
LGTM !!!

PR #3379, opened 2024-07-15
[Spark] Enable UniForm Without Rewrite

xzhseh:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Now users could directly enable Delta UniForm using `ALTER TABLE SET TBLPROPERTIES` command, this only converts the corresponding metadata from `delta` to `iceberg` without rewriting all the underlying parquet files.

## How was this patch tested?
Through manual tests and e2e tests.

Yes, this PR let users enable Delta UniForm directly via `ALTER TABLE SET TBLPROPERTIES` command.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/UniversalFormat.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/UniversalFormat.scala
@@ -55,6 +55,35 @@ object UniversalFormat extends DeltaLogging {
   val HUDI_FORMAT = "hudi"
   val SUPPORTED_FORMATS = Set(HUDI_FORMAT, ICEBERG_FORMAT)
 
+  /**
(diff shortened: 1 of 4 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/UniversalFormat.scala
```diff
@@ -257,6 +286,7 @@ object UniversalFormat extends DeltaLogging {
     }
   }
 }
+
```
lzlfred, line 289: not intended ?

lzlfred (approved):
lgtm. plz fix the empty line.

spark/src/main/scala/org/apache/spark/sql/delta/UniversalFormat.scala
xzhseh: this is intended, to separate the two.

PR #3381, opened 2024-07-16
[Kernel] Update the Kernel read with coordinated commit support

EstherBear:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

Update the Kernel read with coordinated commit support and prepare for kernel coordinated commit write support.
## How was this patch tested?

Unit tests

Yes. Users can enable coordinated commit read support by configuring its properties.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java
@@ -15,10 +15,12 @@
  */
 package io.delta.kernel.internal;
 
+import static io.delta.kernel.internal.TableConfig.*;
 import static io.delta.kernel.internal.TableConfig.TOMBSTONE_RETENTION;
 
 import io.delta.kernel.ScanBuilder;
 import io.delta.kernel.Snapshot;
+import io.delta.kernel.engine.CommitCoordinatorClientHandler;
 import io.delta.kernel.engine.Engine;
 import io.delta.kernel.internal.actions.CommitInfo;
 import io.delta.kernel.internal.actions.Metadata;
@@ -27,6 +29,7 @@ import io.delta.kernel.internal.fs.Path;
 import io.delta.kernel.internal.replay.CreateCheckpointIterator;
 import io.delta.kernel.internal.replay.LogReplay;
 import io.delta.kernel.internal.snapshot.LogSegment;
+import io.delta.kernel.internal.snapshot.TableCommitCoordinatorClientHandler;
 import io.delta.kernel.types.StructType;
 import java.util.Optional;
 
@@ -139,5 +142,29 @@ public class SnapshotImpl implements Snapshot {
     } else {
       return logSegment.lastCommitTimestamp;
     }
+  }
+
+  /**
+   * Returns the commit coordinator client handler based on the table metadata in this snapshot.
+   *
+   * @param engine the engine to use for IO operations
+   * @return the commit coordinator client handler for this snapshot or empty if the metadata is not
+   *     configured to use the commit coordinator.
+   */
+  public Optional<TableCommitCoordinatorClientHandler> getTableCommitCoordinatorClientHandlerOpt(
+      Engine engine) {
+    return COORDINATED_COMMITS_COORDINATOR_NAME
+        .fromMetadata(engine, metadata)
+        .map(
+            commitCoordinatorStr -> {
+              CommitCoordinatorClientHandler handler =
+                  engine.getCommitCoordinatorClientHandler(
+                      commitCoordinatorStr,
+                      COORDINATED_COMMITS_COORDINATOR_CONF.fromMetadata(engine, metadata));
+              return new TableCommitCoordinatorClientHandler(
+                  handler,
+                  logPath.toString(),
+                  COORDINATED_COMMITS_TABLE_CONF.fromMetadata(engine, metadata));
+            });
   }
 }
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
@@ -26,6 +26,7 @@ import static java.lang.String.format;
 
 import io.delta.kernel.*;
 import io.delta.kernel.engine.Engine;
+import io.delta.kernel.engine.coordinatedcommits.Commit;
 import io.delta.kernel.exceptions.CheckpointAlreadyExistsException;
 import io.delta.kernel.exceptions.InvalidTableException;
 import io.delta.kernel.exceptions.TableNotFoundException;
@@ -42,6 +43,7 @@ import io.delta.kernel.utils.FileStatus;
 import java.io.*;
 import java.nio.file.FileAlreadyExistsException;
 import java.util.*;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
@@ -128,14 +130,30 @@ public class SnapshotManager {
   public Snapshot getSnapshotAt(Engine engine, long version) throws TableNotFoundException {
 
     Optional<LogSegment> logSegmentOpt =
-        getLogSegmentForVersion(
+        getLogSegmentAtOrBeforeVersion(
             engine,
             Optional.empty(), /* startCheckpointOpt */
-            Optional.of(version) /* versionToLoadOpt */);
+            Optional.of(version) /* versionToLoadOpt */,
+            Optional.empty() /* tableCommitHandlerOpt */);
 
-    return logSegmentOpt
-        .map(logSegment -> createSnapshot(logSegment, engine))
-        .orElseThrow(() -> new TableNotFoundException(tablePath.toString()));
+    // For non-coordinated commit table, the {@code getCoodinatedCommitsAwareSnapshot} will
+    // create the snapshot with the {@code logSegmentOpt} built here and will not trigger other
+    // operations. For coordinated commit table, the {@code getCoodinatedCommitsAwareSnapshot}
+    // will create the snapshot with the {@code logSegmentOpt} built here and will build the
+    // logSegment again by also fetching the unbackfilled commits from the commit coordinator.
+    // With the unbackfilled commits plus the backfilled commits in Delta log, a new snapshot
+    // will be created.
+    SnapshotImpl snapshot =
+        logSegmentOpt
+            .map(
+                logSegment ->
+                    getCoordinatedCommitsAwareSnapshot(engine, logSegment, Optional.of(version)))
+            .orElseThrow(() -> new TableNotFoundException(tablePath.toString()));
+    long snapshotVer = snapshot.getVersion(engine);
+    if (snapshotVer != version) {
+      throw DeltaErrors.versionAfterLatestCommit(tablePath.toString(), version, snapshotVer);
+    }
+    return snapshot;
   }
 
   /**
@@ -280,38 +298,70 @@ public class SnapshotManager {
    * @param startVersion the version to start. Inclusive.
    * @param versionToLoad the optional parameter to set the max version we should return. Inclusive.
    *     Must be >= startVersion if provided.
+   *     <p>This method lists all delta files and checkpoint files with the following steps:
+   *     <ul>
+   *       <li>When the table is set up with a commit coordinator, retrieve any files that haven't
+   *           been backfilled by initiating a request to the commit coordinator service. If the
+   *           table is not configured to use a commit coordinator, this list will be empty.
+   *       <li>Collect commit files (aka backfilled commits) and checkpoint files by listing the
+   *           contents of the Delta log on storage.
+   *       <li>Filter un-backfilled files to exclude overlapping delta files collected from both
+   *           commit-coordinator and file-system to avoid duplicates.
+   *       <li>Merge and return the backfilled files and filtered un-backfilled files.
+   *     </ul>
+   *     <p>*Note*: If table is a coordinated-commits table, the commit-coordinator client MUST be
+   *     passed to correctly list the commits.
+   * @param startVersion the version to start. Inclusive.
+   * @param versionToLoad the optional parameter to set the max version we should return. Inclusive.
+   *     Must be >= startVersion if provided.
+   * @param tableCommitHandlerOpt the optional commit-coordinator client handler to use for fetching
+   *     un-backfilled commits.
    * @return Some array of files found (possibly empty, if no usable commit files are present), or
    *     None if the listing returned no files at all.
    */
   protected final Optional<List<FileStatus>> listDeltaAndCheckpointFiles(
-      Engine engine, long startVersion, Optional<Long> versionToLoad) {
+      Engine engine,
+      long startVersion,
+      Optional<Long> versionToLoad,
+      Optional<TableCommitCoordinatorClientHandler> tableCommitHandlerOpt) {
     versionToLoad.ifPresent(
         v ->
             checkArgument(
                 v >= startVersion,
                 format("versionToLoad=%s provided is less than startVersion=%s", v, startVersion)));
-    logger.debug("startVersion: {}, versionToLoad: {}", startVersion, versionToLoad);
+    logger.debug(
+        "startVersion: {}, versionToLoad: {}, coordinated commits enabled: {}",
+        startVersion,
+        versionToLoad,
+        tableCommitHandlerOpt.isPresent());
 
-    return listFromOrNone(engine, startVersion)
-        .map(
+    // Fetching the unbackfilled commits before doing the log directory listing to avoid a gap
+    // in delta versions if some delta files are backfilled after the log directory listing but
+    // before the unbackfilled commits listing
+    List<Commit> unbackfilledCommits =
+        getUnbackfilledCommits(tableCommitHandlerOpt, startVersion, versionToLoad);
+
+    final AtomicLong maxDeltaVersionSeen = new AtomicLong(startVersion - 1);
+    Optional<CloseableIterator<FileStatus>> listing = listFromOrNone(engine, startVersion);
+    Optional<List<FileStatus>> resultFromFsListingOpt =
+        listing.map(
             fileStatusesIter -> {
               final List<FileStatus> output = new ArrayList<>();
 
               while (fileStatusesIter.hasNext()) {
                 final FileStatus fileStatus = fileStatusesIter.next();
+                final String fileName = getName(fileStatus.getPath());
 
                 // Pick up all checkpoint and delta files
-                if (!isDeltaCommitOrCheckpointFile(getName(fileStatus.getPath()))) {
+                if (!isDeltaCommitOrCheckpointFile(fileName)) {
                   continue;
                 }
 
                 // Checkpoint files of 0 size are invalid but may be ignored silently when read,
                 // hence we drop them so that we never pick up such checkpoints.
-                if (FileNames.isCheckpointFile(getName(fileStatus.getPath()))
-                    && fileStatus.getSize() == 0) {
+                if (FileNames.isCheckpointFile(fileName) && fileStatus.getSize() == 0) {
                   continue;
                 }
-
                 // Take files until the version we want to load
                 final boolean versionWithinRange =
                     versionToLoad
@@ -331,11 +381,63 @@ public class SnapshotManager {
                   break;
                 }
 
+                // Ideally listFromOrNone should return lexiographically sorted
+                // files and so maxDeltaVersionSeen should be equal to fileVersion.
+                // But we are being defensive here and taking max of all the
+                // fileVersions seen.
+                if (FileNames.isCommitFile(fileName)) {
+                  maxDeltaVersionSeen.set(
+                      Math.max(
+                          maxDeltaVersionSeen.get(), FileNames.deltaVersion(fileStatus.getPath())));
+                }
                 output.add(fileStatus);
               }
 
               return output;
             });
+
+    if (!tableCommitHandlerOpt.isPresent()) {
+      return resultFromFsListingOpt;
+    }
+    List<FileStatus> relevantUnbackfilledCommits =
+        unbackfilledCommits.stream()
+            .filter((commit) -> commit.getVersion() > maxDeltaVersionSeen.get())
+            .filter(
+                (commit) ->
+                    !versionToLoad.isPresent() || commit.getVersion() <= versionToLoad.get())
+            .map(Commit::getFileStatus)
+            .collect(Collectors.toList());
+
+    return resultFromFsListingOpt.map(
+        fsListing -> {
+          fsListing.addAll(relevantUnbackfilledCommits);
+          return fsListing;
+        });
+  }
+
+  private List<Commit> getUnbackfilledCommits(
+      Optional<TableCommitCoordinatorClientHandler> tableCommitHandlerOpt,
+      long startVersion,
+      Optional<Long> versionToLoad) {
+    try {
+      return tableCommitHandlerOpt
+          .map(
+              commitCoordinatorClientHandler -> {
+                logger.info(
+                    "Getting un-backfilled commits from commit coordinator for " + "table: {}",
+                    tablePath);
+                return commitCoordinatorClientHandler
+                    .getCommits(startVersion, versionToLoad.orElse(null))
+                    .getCommits();
+              })
+          .orElse(Collections.emptyList());
+    } catch (Exception e) {
+      logger.error(
+          "Failed to get unbackfilled commits of table {} with commit coordinator: {}",
+          tablePath,
+          e);
+      throw e;
+    }
   }
 
   /**
@@ -354,8 +456,39 @@ public class SnapshotManager {
     Optional<LogSegment> logSegmentOpt = getLogSegmentFrom(engine, lastCheckpointOpt);
 
     return logSegmentOpt
-        .map(logSegment -> createSnapshot(logSegment, engine))
+        .map(logSegment -> getCoordinatedCommitsAwareSnapshot(engine, logSegment, Optional.empty()))
         .orElseThrow(() -> new TableNotFoundException(tablePath.toString()));
+  }
+
+  /**
+   * This can be optimized by making snapshot hint optimization to work with coordinated commits.
+   *
+   * @see <a href="https://github.com/delta-io/delta/issues/3437">issue #3437</a>.
+   */
+  private SnapshotImpl getCoordinatedCommitsAwareSnapshot(
+      Engine engine, LogSegment initialSegmentForNewSnapshot, Optional<Long> versionToLoadOpt) {
+    SnapshotImpl newSnapshot = createSnapshot(initialSegmentForNewSnapshot, engine);
+
+    if (versionToLoadOpt.isPresent() && newSnapshot.getVersion(engine) == versionToLoadOpt.get()) {
+      return newSnapshot;
+    }
+
+    Optional<TableCommitCoordinatorClientHandler> newTableCommitCoordinatorClientHandlerOpt =
+        newSnapshot.getTableCommitCoordinatorClientHandlerOpt(engine);
+
+    if (newTableCommitCoordinatorClientHandlerOpt.isPresent()) {
+      Optional<LogSegment> segmentOpt =
+          getLogSegmentAtOrBeforeVersion(
+              engine,
+              newSnapshot.getLogSegment().checkpointVersionOpt, /* startCheckpointOpt */
+              versionToLoadOpt /* versionToLoadOpt */,
+              newTableCommitCoordinatorClientHandlerOpt /* tableCommitHandlerOpt */);
+      newSnapshot =
+          segmentOpt
+              .map(segment -> createSnapshot(segment, engine))
+              .orElseThrow(() -> new TableNotFoundException(tablePath.toString()));
+    }
+    return newSnapshot;
   }
 
   private SnapshotImpl createSnapshot(LogSegment initSegment, Engine engine) {
@@ -407,26 +540,31 @@ public class SnapshotManager {
    */
   private Optional<LogSegment> getLogSegmentFrom(
       Engine engine, Optional<CheckpointMetaData> startingCheckpoint) {
-    return getLogSegmentForVersion(
-        engine, startingCheckpoint.map(x -> x.version), Optional.empty());
+    return getLogSegmentAtOrBeforeVersion(
+        engine, startingCheckpoint.map(x -> x.version), Optional.empty(), Optional.empty());
   }
 
   /**
-   * Get a list of files that can be used to compute a Snapshot at version `versionToLoad`, If
-   * `versionToLoad` is not provided, will generate the list of files that are needed to load the
-   * latest version of the Delta table. This method also performs checks to ensure that the delta
-   * files are contiguous.
+   * Get a list of files that can be used to compute a Snapshot at or before version
+   * `versionToLoad`, If `versionToLoad` is not provided, will generate the list of files that are
+   * needed to load the latest version of the Delta table. This method also performs checks to
+   * ensure that the delta files are contiguous.
    *
    * @param startCheckpoint A potential start version to perform the listing of the DeltaLog,
    *     typically that of a known checkpoint. If this version's not provided, we will start listing
    *     from version 0.
-   * @param versionToLoad A specific version to load. Typically used with time travel and the Delta
-   *     streaming source. If not provided, we will try to load the latest version of the table.
+   * @param versionToLoad A specific version we try to load, but we may only load a version before
+   *     this version if this version of commit is un-backfilled. Typically used with time travel
+   *     and the Delta streaming source. If not provided, we will try to load the latest version of
+   *     the table.
    * @return Some LogSegment to build a Snapshot if files do exist after the given startCheckpoint.
    *     None, if the delta log directory was missing or empty.
    */
-  public Optional<LogSegment> getLogSegmentForVersion(
-      Engine engine, Optional<Long> startCheckpoint, Optional<Long> versionToLoad) {
+  protected Optional<LogSegment> getLogSegmentAtOrBeforeVersion(
+      Engine engine,
+      Optional<Long> startCheckpoint,
+      Optional<Long> versionToLoad,
+      Optional<TableCommitCoordinatorClientHandler> tableCommitHandlerOpt) {
     // Only use startCheckpoint if it is <= versionToLoad
     Optional<Long> startCheckpointToUse =
         startCheckpoint.filter(v -> !versionToLoad.isPresent() || v <= versionToLoad.get());
@@ -456,7 +594,7 @@ public class SnapshotManager {
 
     long startTimeMillis = System.currentTimeMillis();
     final Optional<List<FileStatus>> newFiles =
-        listDeltaAndCheckpointFiles(engine, startVersion, versionToLoad);
+        listDeltaAndCheckpointFiles(engine, startVersion, versionToLoad, tableCommitHandlerOpt);
     logger.info(
         "{}: Took {}ms to list the files after starting checkpoint",
         tablePath,
@@ -464,7 +602,8 @@ public class SnapshotManager {
 
     startTimeMillis = System.currentTimeMillis();
     try {
-      return getLogSegmentForVersion(engine, startCheckpointToUse, versionToLoad, newFiles);
+      return getLogSegmentAtOrBeforeVersion(
+          engine, startCheckpointToUse, versionToLoad, newFiles, tableCommitHandlerOpt);
     } finally {
       logger.info(
           "{}: Took {}ms to construct a log segment",
@@ -477,11 +616,12 @@ public class SnapshotManager {
    * Helper function for the getLogSegmentForVersion above. Called with a provided files list, and
    * will then try to construct a new LogSegment using that.
    */
-  protected Optional<LogSegment> getLogSegmentForVersion(
+  protected Optional<LogSegment> getLogSegmentAtOrBeforeVersion(
       Engine engine,
       Optional<Long> startCheckpointOpt,
       Optional<Long> versionToLoadOpt,
-      Optional<List<FileStatus>> filesOpt) {
+      Optional<List<FileStatus>> filesOpt,
+      Optional<TableCommitCoordinatorClientHandler> tableCommitHandlerOpt) {
     final List<FileStatus> newFiles;
     if (filesOpt.isPresent()) {
       newFiles = filesOpt.get();
@@ -515,7 +655,8 @@ public class SnapshotManager {
     } else if (newFiles.isEmpty()) {
       // The directory may be deleted and recreated and we may have stale state in our
       // DeltaLog singleton, so try listing from the first version
-      return getLogSegmentForVersion(engine, Optional.empty(), versionToLoadOpt);
+      return getLogSegmentAtOrBeforeVersion(
+          engine, Optional.empty(), versionToLoadOpt, tableCommitHandlerOpt);
     }
 
     Tuple2<List<FileStatus>, List<FileStatus>> checkpointsAndDeltas =
@@ -630,7 +771,7 @@ public class SnapshotManager {
     }
 
     versionToLoadOpt
-        .filter(v -> v != newVersion)
+        .filter(v -> v < newVersion)
         .ifPresent(
             v -> {
               throw DeltaErrors.versionAfterLatestCommit(tablePath.toString(), v, newVersion);
@@ -648,10 +789,14 @@ public class SnapshotManager {
                 "Unable to reconstruct table state: missing log file for version %s",
                 newCheckpointVersion + 1));
       }
+      logger.info(
+          "Verified delta files are contiguous from version {} to {}",
+          newCheckpointVersion + 1,
+          newVersion);
       verifyDeltaVersions(
           deltaVersionsAfterCheckpoint,
           Optional.of(newCheckpointVersion + 1),
-          versionToLoadOpt,
+          Optional.of(newVersion),
           tablePath);
     }
 
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/TableCommitCoordinatorClientHandler.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/TableCommitCoordinatorClientHandler.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright (2024) The Delta Lake Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.delta.kernel.internal.snapshot;
+
+import io.delta.kernel.data.Row;
+import io.delta.kernel.engine.CommitCoordinatorClientHandler;
+import io.delta.kernel.engine.coordinatedcommits.CommitFailedException;
+import io.delta.kernel.engine.coordinatedcommits.CommitResponse;
+import io.delta.kernel.engine.coordinatedcommits.GetCommitsResponse;
+import io.delta.kernel.engine.coordinatedcommits.UpdatedActions;
+import io.delta.kernel.utils.CloseableIterator;
+import java.io.IOException;
+import java.util.Map;
+
+/**
+ * A wrapper around {@link CommitCoordinatorClientHandler} that provides a more user-friendly API
+ * for committing/ accessing commits to a specific table. This class takes care of passing the table
+ * specific configuration to the underlying {@link CommitCoordinatorClientHandler} e.g. logPath /
+ * coordinatedCommitsTableConf.
+ */
+public class TableCommitCoordinatorClientHandler {
+  private final CommitCoordinatorClientHandler commitCoordinatorClientHandler;
+  private final String logPath;
+  private final Map<String, String> tableConf;
+
+  public TableCommitCoordinatorClientHandler(
+      CommitCoordinatorClientHandler commitCoordinatorClientHandler,
+      String logPath,
+      Map<String, String> tableConf) {
+    this.commitCoordinatorClientHandler = commitCoordinatorClientHandler;
+    this.logPath = logPath;
+    this.tableConf = tableConf;
+  }
+
+  public CommitResponse commit(
+      long commitVersion, CloseableIterator<Row> actions, UpdatedActions updatedActions)
+      throws CommitFailedException {
+    return commitCoordinatorClientHandler.commit(
+        logPath, tableConf, commitVersion, actions, updatedActions);
+  }
+
+  public GetCommitsResponse getCommits(Long startVersion, Long endVersion) {
+    return commitCoordinatorClientHandler.getCommits(logPath, tableConf, startVersion, endVersion);
+  }
+
+  public void backfillToVersion(long version, Long lastKnownBackfilledVersion) throws IOException {
+    commitCoordinatorClientHandler.backfillToVersion(
+        logPath, tableConf, version, lastKnownBackfilledVersion);
+  }
+
+  public boolean semanticEquals(
+      CommitCoordinatorClientHandler otherCommitCoordinatorClientHandler) {
+    return commitCoordinatorClientHandler.semanticEquals(otherCommitCoordinatorClientHandler);
+  }
+
+  public boolean semanticEquals(TableCommitCoordinatorClientHandler otherCommitCoordinatorClient) {
+    return semanticEquals(otherCommitCoordinatorClient.commitCoordinatorClientHandler);
+  }
+}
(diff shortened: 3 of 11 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -130,10 +134,16 @@ public Snapshot buildLatestSnapshot(Engine engine)
...
             long version) throws TableNotFoundException {
+        SnapshotImpl upperBoundSnapshot = getSnapshotAtInit(engine);
 
         Optional<LogSegment> logSegmentOpt = getLogSegmentForVersion(engine,
-            Optional.empty(), /* startCheckpointOpt */
-            Optional.of(version) /* versionToLoadOpt */);
+                Optional.empty(), /* startCheckpointOpt */
+                Optional.of(version) /* versionToLoadOpt */,
+                upperBoundSnapshot.getCommitCoordinatorClientHandlerOpt(engine),
+                /* commitCoordinatorClientHandlerOpt */
```
dhruvarya-db, line 143: [nit] /* commitCoordinatorClientHandlerOpt */ should either be on the right of the parameter or above it.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/FileNames.java
```diff
@@ -29,6 +29,9 @@ private FileNames() {}
     private static final Pattern DELTA_FILE_PATTERN =
         Pattern.compile("\\d+\\.json");
 
+    public static final Pattern UUID_DELTA_FILE_REGEX =
```
dhruvarya-db, line 32: This should be private

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -363,13 +420,50 @@ private SnapshotImpl getSnapshotAtInit(Engine engine)
             logger.warn("{}: Last checkpoint file is missing or corrupted. " +
                     "Will search for the checkpoint files directly.", tablePath);
         }
-        Optional<LogSegment> logSegmentOpt =
-            getLogSegmentFrom(engine, lastCheckpointOpt);
+        Optional<LogSegment> logSegmentOpt = getLogSegmentFrom(engine, lastCheckpointOpt);
```
dhruvarya-db, line 423: noise?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java
```diff
@@ -154,4 +155,20 @@ public long getTimestamp(Engine engine) {
...
+
+    /**
+     * Returns the commit coordinator client handler for this snapshot.
+     * If the snapshot is not configured to use a commit coordinator, this returns an empty
+     * Optional.
+     *
+     * @param engine the engine to use for IO operations
+     * @return the commit coordinator client handler for this snapshot
+     */
+    public Optional<CommitCoordinatorClientHandler> getCommitCoordinatorClientHandlerOpt(
```
dhruvarya-db, line 167: I wonder if it would better to replicate the `TableCommitCoordinator` from Delta-Spark in Kernel? `TableCommitCoordinator` wraps around CoordinatedCommitsHandler and stores logpath + configs so that you don't have to pass these to parameters to it on every commit/getCommits call. What do you think @vkorukanti ?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -306,48 +318,93 @@ protected final Optional<List<FileStatus>> listDeltaAndCheckpointFiles(
...
+
+
+        AtomicLong maxDeltaVersionSeen = new AtomicLong(startVersion - 1);
+        Optional<List<FileStatus>> resultFromFsListingOpt = listFromOrNone(engine,
+                startVersion).map(fileStatusesIter -> {
+                    final List<FileStatus> output = new ArrayList<>();
+
+                    while (fileStatusesIter.hasNext()) {
+                        final FileStatus fileStatus = fileStatusesIter.next();
+                        final String fileName = getName(fileStatus.getPath());
```
dhruvarya-db, line 339: Can you try undoing this change? I feel that the diff will become less complex with that.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -306,48 +318,93 @@ protected final Optional<List<FileStatus>> listDeltaAndCheckpointFiles(
...
+                        .getCommits(
+                                logPath.toString(),
+                                coordinatedCommitsTableConf,
+                                startVersion,
+                                versionToLoad.orElse(null))
+                        .getCommits())
+                .orElse(Collections.emptyList());
+
+
+        AtomicLong maxDeltaVersionSeen = new AtomicLong(startVersion - 1);
```
dhruvarya-db, line 332: This does not need to be Atomic?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -363,13 +420,50 @@ private SnapshotImpl getSnapshotAtInit(Engine engine)
...
-            .orElseThrow(() -> new TableNotFoundException(tablePath.toString()));
+                .map(logSegment -> getUpdatedSnapshot(
+                        engine,
+                        Optional.empty(), /* oldSnapshotOpt */
+                        logSegment,
+                        Optional.empty() /* initialCommitCoordinatorClientHandler */))
+                .orElseThrow(() -> new TableNotFoundException(tablePath.toString()));
+    }
+
+    private SnapshotImpl getUpdatedSnapshot(
```
dhruvarya-db, line 434: Kernel does not have the concept of a post-commit snapshot, so I don't think that this is useful? So, kernel will not have an `oldSnapshotOpt` nor an `initialCommitCoordinatorClientHandler`, the logic can be simplified a lot.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -306,48 +318,93 @@ protected final Optional<List<FileStatus>> listDeltaAndCheckpointFiles(
...
                     }
 
-                    output.add(fileStatus);
-                }
+                    return output;
+                });
 
-                return output;
-            });
+        if (!commitCoordinatorClientHandlerOpt.isPresent()) {
```
dhruvarya-db, line 386: can this be `commitCoordinatorClientHandlerOpt.isEmpty() ?

EstherBear: It seems that `Optional#isEmpty()` is a Java 11 method, which is a shortcut for Java 8's `!Optional#isPresent()`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -364,12 +419,32 @@ private SnapshotImpl getSnapshotAtInit(Engine engine)
...
 
         return logSegmentOpt
-            .map(logSegment -> createSnapshot(
-                logSegment, engine))
-            .orElseThrow(() -> new TableNotFoundException(tablePath.toString()));
+                .map(logSegment -> getSnapshot(engine, logSegment))
+                .orElseThrow(() -> new TableNotFoundException(tablePath.toString()));
+    }
+
+    private SnapshotImpl getSnapshot(Engine engine, LogSegment initialSegmentForNewSnapshot) {
```
dhruvarya-db, line 429: Let's be more explicit about what this does? How about `getCoordinatedCommitsAwareSnapshot`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -364,12 +419,32 @@ private SnapshotImpl getSnapshotAtInit(Engine engine)
...
+        SnapshotImpl newSnapshot = createSnapshot(initialSegmentForNewSnapshot, engine);
+
+        Optional<CommitCoordinatorClientHandler> newCommitCoordinatorClientHandlerOpt =
+                newSnapshot.getCommitCoordinatorClientHandlerOpt(engine);
+        boolean shouldUpdate = newCommitCoordinatorClientHandlerOpt.isPresent();
```
dhruvarya-db, line 434: s/shouldUpdate/shouldRegenerateWithCoordinatedCommits ? If not, you can just move the .isPresent() to the if () and remove the additional boolean variable.

kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/engine/DefaultCommitCoordinatorClientHandler.java
```diff
@@ -173,4 +173,8 @@ public boolean semanticEquals(CommitCoordinatorClientHandler other) {
                 commitCoordinatorClient.semanticEquals(
                 ((DefaultCommitCoordinatorClientHandler) other).commitCoordinatorClient);
     }
+
+    public CommitCoordinatorClient getCommitCoordinatorClient() {
```
dhruvarya-db, line 177: noise?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/internal/coordinatedcommits/CoordinatedCommitsSuite.scala
```diff
@@ -0,0 +1,314 @@
...
+
+  test("cold snapshot initialization") {
+    InMemoryCommitCoordinatorBuilder.clearInMemoryInstances()
+    withTempDirAndEngine ({ (tablePath, engine) =>
+      val handler =
+        engine.getCommitCoordinatorClientHandler("tracking-in-memory", Collections.emptyMap())
+      val logPath = new Path("file:" + tablePath, "_delta_log")
+      val tableSpark = Table.forPath(engine, tablePath)
+
+      spark.range(0, 10).write.format("delta").mode("overwrite").save(tablePath) // version 0
```
dhruvarya-db, line 56: Can you move the setup code to a common function? And please add some comments to describe what is happening here.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -306,48 +318,92 @@ protected final Optional<List<FileStatus>> listDeltaAndCheckpointFiles(
...
+            return resultFromFsListingOpt;
+        }
+
+        List<FileStatus> unbackfilledCommitsFiltered = new ArrayList<>();
+        boolean dropConditionMet = false;
+        for (Commit commit : unbackfilledCommits) {
+            if (!dropConditionMet && commit.getVersion() <= maxDeltaVersionSeen.get()) {
+                continue;
+            } else {
+                dropConditionMet = true;
```
dhruvarya-db, line 395: Can you add a test for whether the `dropConditionMet` logic works correctly? You can use the existing test setup logic and instead of deleting some of the backfilled commits, you can retain them to ensure that there is an overlap.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/internal/coordinatedcommits/CoordinatedCommitsSuite.scala
```diff
@@ -0,0 +1,260 @@
...
+    spark.range(0, 10).write.format("delta").mode("overwrite").save(tablePath) // version 0
+    spark.range(10, 20).write.format("delta").mode("append").save(tablePath) // version 1
+    spark.range(20, 30).write.format("delta").mode("append").save(tablePath) // version 2
+    checkAnswer(
+      spark.sql(s"SELECT * FROM delta.`$tablePath`").collect().map(TestRow(_)),
+      (0L to 29L).map(TestRow(_)))
+
+    var tableConf: util.Map[String, String] = null
+
+    (0 to 2).foreach{ version =>
```
dhruvarya-db, line 69: Add a comment briefly explaining what this block does.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/internal/coordinatedcommits/CoordinatedCommitsSuite.scala
```diff
@@ -0,0 +1,260 @@
...
+
+      val table = Table.forPath(engine, tablePath)
+      intercept[RuntimeException] {
+        table.getLatestSnapshot(engine)
+      }
+    }, Map(CommitCoordinatorProvider.getCommitCoordinatorNameConfKey("tracking-in-memory") ->
+      classOf[TrackingInMemoryCommitCoordinatorBuilder].getName,
+      TrackingInMemoryCommitCoordinatorBuilder.BATCH_SIZE_CONF_KEY -> "10"))
+  }
+
```
dhruvarya-db, line 165: Please add a direct test for `getLogSegmentForVersion`

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/internal/coordinatedcommits/CoordinatedCommitsSuite.scala
```diff
@@ -0,0 +1,260 @@
...
+  val expCoordinatorConf: util.Map[String, String] = Map(
+    "coordinatorKey1" -> "string_value",
+    "coordinatorKey2Int" -> "2",
+    "coordinatorKey3ComplexStr" -> "\"hello\"").asJava
+
+  val coordinator = new TestCommitCoordinatorClient()
+}
+
+class TestCommitCoordinatorClient extends InMemoryCommitCoordinator(10) {
```
dhruvarya-db, line 222: Add comment with the parameter name

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/internal/coordinatedcommits/CoordinatedCommitsSuite.scala
```diff
@@ -0,0 +1,260 @@
...
+    spark.range(10, 20).write.format("delta").mode("append").save(tablePath) // version 1
+    spark.range(20, 30).write.format("delta").mode("append").save(tablePath) // version 2
+    checkAnswer(
+      spark.sql(s"SELECT * FROM delta.`$tablePath`").collect().map(TestRow(_)),
+      (0L to 29L).map(TestRow(_)))
+
+    var tableConf: util.Map[String, String] = null
+
+    (0 to 2).foreach{ version =>
+      val delta = getHadoopDeltaFile(logPath, version)
```
dhruvarya-db, line 70: s/delta/commitFilePath

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/internal/coordinatedcommits/CoordinatedCommitsSuite.scala
```diff
@@ -0,0 +1,260 @@
...
+}
+
+object TestCommitCoordinator {
+  val expTableConf: util.Map[String, String] = Map(
+    "tableKey1" -> "string_value",
+    "tableKey2Int" -> "2",
+    "tableKey3ComplexStr" -> "\"hello\""
+  ).asJava
+
+  val expCoordinatorConf: util.Map[String, String] = Map(
```
dhruvarya-db, line 214: Use an ALL UPPERCASE name for hardcoded values like these.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -306,48 +318,92 @@ protected final Optional<List<FileStatus>> listDeltaAndCheckpointFiles(
...
 
-                return output;
-            });
+        if (!commitCoordinatorClientHandlerOpt.isPresent()) {
+            return resultFromFsListingOpt;
+        }
+
+        List<FileStatus> unbackfilledCommitsFiltered = new ArrayList<>();
+        boolean dropConditionMet = false;
+        for (Commit commit : unbackfilledCommits) {
```
dhruvarya-db, line 391: Would this work?
```
List<FileStatus> relevantUnbackfilledCommits = unbackfilledCommits
 .stream()
 .filter((commit) -> commit.getVersion() > maxDeltaVersionSeen.get())
 .takeWhile((commit) -> !versionToLoad.isPresent() || commit.getVersion() <= versionToLoad.get())
 .toArray()
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -367,9 +423,29 @@ private SnapshotImpl getSnapshotAtInit(Engine engine)
...
+            Engine engine, LogSegment initialSegmentForNewSnapshot) {
+        SnapshotImpl newSnapshot = createSnapshot(initialSegmentForNewSnapshot, engine);
+
+        Optional<CommitCoordinatorClientHandler> newCommitCoordinatorClientHandlerOpt =
+                newSnapshot.getCommitCoordinatorClientHandlerOpt(engine);
+
+        if (newCommitCoordinatorClientHandlerOpt.isPresent()) {
+            Optional<LogSegment> segmentOpt = getLogSegmentForVersion(engine,
+                    Optional.empty(), /* startCheckpointOpt */
+                    newSnapshot.getLogSegment().checkpointVersionOpt /* versionToLoadOpt */,
```
dhruvarya-db, line 440: This seems wrong?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -428,8 +504,10 @@ private Optional<LogSegment> getLogSegmentFrom(
         Engine engine,
```
dhruvarya-db, line 504: Looks like there is only one callsite of this function. My preference would be to inline it or change the name to indicate that the function will not return CC commits. My preference would be just to inline it to reduce the complexity. What do you think @vkorukanti?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/internal/coordinatedcommits/CoordinatedCommitsSuite.scala
```diff
@@ -0,0 +1,260 @@
...
+
+import java.{lang, util}
+import java.util.{Collections, Optional}
+import scala.collection.convert.ImplicitConversions.`iterator asScala`
+import scala.collection.JavaConverters._
+
+class CoordinatedCommitsSuite extends DeltaTableWriteSuiteBase
+  with CoordinatedCommitsTestUtils {
+
+  def setupCoordinatedCommitFilesForTest(
```
dhruvarya-db, line 48: Let's add a test which has a checkpoint as well. First create some commits, then create a checkpoint, then a few commits again. Manually convert the table to CC at the version after the checkpoint.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
EstherBear: Thanks for pointing it out!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
EstherBear: The `takeWhile` API is since 9+. So I use `filter` instead.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/FileNames.java
```diff
@@ -29,6 +29,9 @@ private FileNames() {}
     private static final Pattern DELTA_FILE_PATTERN =
         Pattern.compile("\\d+\\.json");
 
+    private static final Pattern UUID_DELTA_FILE_REGEX =
```
vkorukanti, line 32: nit: add an example  file name in method header.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
```diff
@@ -220,7 +220,9 @@ class SnapshotManagerSuite extends AnyFunSuite with MockFileSystemClientUtils {
           new MockSidecarParquetHandler(expectedSidecars),
           new MockSidecarJsonHandler(expectedSidecars)),
         Optional.empty(),
-        versionToLoad
+        versionToLoad,
+        Optional.empty(),
```
vkorukanti, line 224: This class tests the specific cases by mocking the Json and Parquet handler.

We need to add tests with a mocked Commitcoordinator with different scenarios:
1) a version is missing from the getCommits
2) commit coordinator throws error
3) commit coordinator returns a backfilled commit

etc..

vkorukanti (changes requested):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -306,48 +318,86 @@ protected final Optional<List<FileStatus>> listDeltaAndCheckpointFiles(
...
+                    return output;
+                });
 
-                return output;
-            });
+        if (!commitCoordinatorClientHandlerOpt.isPresent()) {
+            return resultFromFsListingOpt;
+        }
+
+        List<FileStatus> relevantUnbackfilledCommits = unbackfilledCommits
```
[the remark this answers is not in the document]
EstherBear: I think there will be a gap in delta versions if some delta files are backfilled after the file-system listing but before the commit-coordinator listing if we change the order?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -306,48 +318,86 @@ protected final Optional<List<FileStatus>> listDeltaAndCheckpointFiles(
...
-                    final FileStatus fileStatus = fileStatusesIter.next();
+        List<Commit> unbackfilledCommits = commitCoordinatorClientHandlerOpt
+                .map(commitCoordinatorClientHandler -> commitCoordinatorClientHandler
+                        .getCommits(
+                                logPath.toString(),
+                                coordinatedCommitsTableConf,
+                                startVersion,
+                                versionToLoad.orElse(null))
+                        .getCommits())
+                .orElse(Collections.emptyList());
```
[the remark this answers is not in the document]
EstherBear: I don't think we have checked exception for the `getCommit` API. But I've added some log statements for it.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -367,9 +446,29 @@ private SnapshotImpl getSnapshotAtInit(Engine engine)
...
-            .map(logSegment -> createSnapshot(
-                logSegment, engine))
-            .orElseThrow(() -> new TableNotFoundException(tablePath.toString()));
+                .map(logSegment -> getCoordinatedCommitsAwareSnapshot(engine, logSegment))
+                .orElseThrow(() -> new TableNotFoundException(tablePath.toString()));
+    }
+
+    private SnapshotImpl getCoordinatedCommitsAwareSnapshot(
+            Engine engine, LogSegment initialSegmentForNewSnapshot) {
+        SnapshotImpl newSnapshot = createSnapshot(initialSegmentForNewSnapshot, engine);
```
vkorukanti, line 455: Can you check if this breaks the snapshot hint optimization we have to avoid reloading the Protocol and Metadata for each get latest snapshot call?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/TableCommitCoordinatorClientHandler.java
```diff
@@ -0,0 +1,73 @@
...
+
+import io.delta.kernel.data.Row;
+import io.delta.kernel.engine.CommitCoordinatorClientHandler;
+import io.delta.kernel.engine.coordinatedcommits.CommitFailedException;
+import io.delta.kernel.engine.coordinatedcommits.CommitResponse;
+import io.delta.kernel.engine.coordinatedcommits.GetCommitsResponse;
+import io.delta.kernel.engine.coordinatedcommits.UpdatedActions;
+import io.delta.kernel.utils.CloseableIterator;
+
+public class TableCommitCoordinatorClientHandler {
```
vkorukanti, line 30: Add a one line comment explaining the use of this class.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/TableCommitCoordinatorClientHandler.java
```diff
@@ -0,0 +1,73 @@
...
+    private final CommitCoordinatorClientHandler commitCoordinatorClientHandler;
+    private final String logPath;
+    private final Map<String, String> tableConf;
+
+    public TableCommitCoordinatorClientHandler(
+            CommitCoordinatorClientHandler commitCoordinatorClientHandler,
+            String logPath,
+            Map<String, String> tableConf) {
```
vkorukanti, line 38: @dhruvarya-db @prakharjain09 Why do we need to pass the `tableConf` to each API on the commit coordinator? Isn't it used when creating the commit coordinator?

vkorukanti (changes requested):
Please check if we need to make any changes for the snapshot protocol and metadata optimization to work in this new model of snapshot loading.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/TableCommitCoordinatorClientHandler.java
dhruvarya-db: @vkorukanti Commit Coordinator is supposed to be reuseable across multiple tables (like JSONHandler). `tableConf` can be thought of as a `tablePath`/`tableId` which is used to uniquely identify the table.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
EstherBear: Thanks for pointing this out! It will break the snapshot hint optimization. Created issue #3437 for this.

dhruvarya-db (approved):
LGTM, thanks for working on this!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -130,10 +132,13 @@ public Snapshot buildLatestSnapshot(Engine engine)
...
             long version) throws TableNotFoundException {
+        SnapshotImpl upperBoundSnapshot = getSnapshotAtInit(engine);
 
         Optional<LogSegment> logSegmentOpt = getLogSegmentForVersion(engine,
-            Optional.empty(), /* startCheckpointOpt */
-            Optional.of(version) /* versionToLoadOpt */);
+                Optional.empty(), /* startCheckpointOpt */
+                Optional.of(version) /* versionToLoadOpt */,
+                /* tableCommitHandlerOpt */
+                upperBoundSnapshot.getTableCommitCoordinatorClientHandlerOpt(engine));
```
vkorukanti, line 141:
```suggestion
                Optional.empty(), /* startCheckpointOpt */
                Optional.of(version) /* versionToLoadOpt */,
                upperBoundSnapshot.getTableCommitCoordinatorClientHandlerOpt(engine));
```

vkorukanti: nit: you don't need to add the comment if the value is self describing.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
vkorukanti: Add the issue as a comment here and specify this can be optimized.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -306,48 +328,113 @@ protected final Optional<List<FileStatus>> listDeltaAndCheckpointFiles(
...
-            });
+        // Fetching the unbackfilled commits before doing the log directory listing to avoid a gap
+        // in delta versions if some delta files are backfilled after the log directory listing but
+        // before the unbackfilled commits listing
+        List<Commit> unbackfilledCommits = getUnbackfilledCommits(
+                tableCommitHandlerOpt,
+                startVersion,
+                versionToLoad);
+
+
```
vkorukanti, line 339:
```suggestion

```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -130,10 +132,13 @@ public Snapshot buildLatestSnapshot(Engine engine)
     public Snapshot getSnapshotAt(
             Engine engine,
             long version) throws TableNotFoundException {
+        SnapshotImpl upperBoundSnapshot = getSnapshotAtInit(engine);
```
vkorukanti, line 135: This regresses the current path where the table is not a commit coordinated table with the extra snapshot loading. Discussed offline with @dhruvarya-db. Lets figure out a way to avoid the regression.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -306,48 +328,113 @@ protected final Optional<List<FileStatus>> listDeltaAndCheckpointFiles(
...
+        List<Commit> unbackfilledCommits = getUnbackfilledCommits(
+                tableCommitHandlerOpt,
+                startVersion,
+                versionToLoad);
+
+
+        final AtomicLong maxDeltaVersionSeen = new AtomicLong(startVersion - 1);
+        Optional<List<FileStatus>> resultFromFsListingOpt =
+                listFromOrNone(engine, startVersion)
+                        .map(fileStatusesIter -> {
```
vkorukanti, line 343: minor nit: 
```
Optional<List<FileStatus>> listing = listFromOrNone(engine, startVersion);
Optional<List<FileStatus>> resultFromFsListingOpt = listing(engine, startVersion)
           .map(fileStatusesIter -> {

```

vkorukanti: just finding a way to show a large diff.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/internal/coordinatedcommits/CoordinatedCommitsSuite.scala
```diff
@@ -0,0 +1,390 @@
...
+      }
+
+      TrackingCommitCoordinatorClient.numGetCommitsCalled.set(0)
+      val snapshot3 = table.getLatestSnapshot(engine)
+      val result3 = readSnapshot(snapshot3, snapshot3.getSchema(engine), null, null, engine)
+      checkAnswer(result3, (0L to 39L).map(TestRow(_)))
+      assert(TrackingCommitCoordinatorClient.numGetCommitsCalled.get === 1)
+    }, Map(CommitCoordinatorProvider.getCommitCoordinatorNameConfKey("tracking-in-memory") ->
+      classOf[TrackingInMemoryCommitCoordinatorBuilder].getName,
+      TrackingInMemoryCommitCoordinatorBuilder.BATCH_SIZE_CONF_KEY -> "10"))
```
vkorukanti, line 205: Can we make these tests a bit cleaner to read them?

```
testWithCoordinatorCommits(config) {
   (tablePath, engine) => {
         // test code.
    }
}
```

vkorukanti (changes requested):
Thank you for working through the comments. One last pending issue the regression of timetravel for non-coordinated commits enabled tables.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
EstherBear: Added two tests for the new way of snapshot loading for the `getSnapshotAt` API.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -306,48 +344,112 @@ protected final Optional<List<FileStatus>> listDeltaAndCheckpointFiles(
...
                         if (output.isEmpty()) {
-                            long earliestVersion = DeltaHistoryManager.getEarliestRecreatableCommit(
-                                engine, logPath);
+                            long earliestVersion = DeltaHistoryManager
+                                    .getEarliestRecreatableCommit(engine, logPath);
                             throw DeltaErrors.versionBeforeFirstAvailableCommit(
-                                tablePath.toString(), versionToLoad.get(), earliestVersion);
+                                    tablePath.toString(),
+                                    versionToLoad.get(),
+                                    earliestVersion);
```
vkorukanti, line 394: these changes can be reverted right?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -132,12 +134,31 @@ public Snapshot getSnapshotAt(
...
-            Optional.empty(), /* startCheckpointOpt */
-            Optional.of(version) /* versionToLoadOpt */);
-
-        return logSegmentOpt
-            .map(logSegment -> createSnapshot(logSegment, engine))
-            .orElseThrow(() -> new TableNotFoundException(tablePath.toString()));
+                Optional.empty(), /* startCheckpointOpt */
+                Optional.of(version) /* versionToLoadOpt */,
+                /* tableCommitHandlerOpt */
+                Optional.empty());
```
vkorukanti, line 140:
```suggestion
                Optional.empty() /* tableCommitHandlerOpt */);
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -132,12 +134,31 @@ public Snapshot getSnapshotAt(
             long version) throws TableNotFoundException {
 
         Optional<LogSegment> logSegmentOpt = getLogSegmentForVersion(engine,
-            Optional.empty(), /* startCheckpointOpt */
-            Optional.of(version) /* versionToLoadOpt */);
-
-        return logSegmentOpt
-            .map(logSegment -> createSnapshot(logSegment, engine))
-            .orElseThrow(() -> new TableNotFoundException(tablePath.toString()));
+                Optional.empty(), /* startCheckpointOpt */
```
vkorukanti, line 137: Add the details of steps to load a snapshot at the given version.
1) for non-coordinated commit table
2) for coordinated commit table.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
vkorukanti: same comments for above. lets try to remove the un-needed changes

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
vkorukanti: Basically convey the point that for non-coordinated tables this change is a no-op

vkorukanti (commented):
Looks great!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -306,48 +350,107 @@ protected final Optional<List<FileStatus>> listDeltaAndCheckpointFiles(
...
-                        fileStatus.getSize() == 0) {
+                    if (FileNames.isCheckpointFile(fileName) && fileStatus.getSize() == 0) {
                         continue;
                     }
 
                     // Take files until the version we want to load
                     final boolean versionWithinRange = versionToLoad
-                        .map(v -> FileNames.getFileVersion(new Path(fileStatus.getPath())) <= v)
-                        .orElse(true);
-
```
dhruvarya-db, line 332: noise?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -306,48 +350,107 @@ protected final Optional<List<FileStatus>> listDeltaAndCheckpointFiles(
...
+                            .map(v -> FileNames.getFileVersion(new Path(fileStatus.getPath())) <= v)
+                            .orElse(true);
                     if (!versionWithinRange) {
                         // If we haven't taken any files yet and the first file we see is greater
                         // than the versionToLoad then the versionToLoad is not reconstructable
                         // from the existing logs
                         if (output.isEmpty()) {
-                            long earliestVersion = DeltaHistoryManager.getEarliestRecreatableCommit(
-                                engine, logPath);
+                            long earliestVersion =
```
dhruvarya-db, line 391: noise?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -306,48 +350,107 @@ protected final Optional<List<FileStatus>> listDeltaAndCheckpointFiles(
...
+                                    DeltaHistoryManager.getEarliestRecreatableCommit(
+                                            engine, logPath);
                             throw DeltaErrors.versionBeforeFirstAvailableCommit(
-                                tablePath.toString(), versionToLoad.get(), earliestVersion);
+                                    tablePath.toString(), versionToLoad.get(), earliestVersion);
```
dhruvarya-db, line 395: looks like more noise? please remove

dhruvarya-db (approved):

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
```diff
@@ -177,7 +181,10 @@ class SnapshotManagerSuite extends AnyFunSuite with MockFileSystemClientUtils {
       numParts: Int = -1,
       startCheckpoint: Optional[java.lang.Long] = Optional.empty(),
       versionToLoad: Optional[java.lang.Long] = Optional.empty(),
-      v2CheckpointSpec: Seq[(Long, Boolean, Int)] = Seq.empty): Unit = {
+      v2CheckpointSpec: Seq[(Long, Boolean, Int)] = Seq.empty,
```
vkorukanti, line 184: method docs still refers to deleted method name `getLogSegmentVersion`

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
```diff
@@ -177,7 +181,10 @@ class SnapshotManagerSuite extends AnyFunSuite with MockFileSystemClientUtils {
...
+      unbackfilledDeltaVersions: Seq[Long] = Seq.empty,
+      tableCommitCoordinatorClientHandlerOpt:
+        Optional[TableCommitCoordinatorClientHandler] = Optional.empty()): Unit = {
```
vkorukanti, line 187: why do you need to pass the `tableCommitCoordianterClientHandlerOpt`? Can we just pass the `unbackfilledDeltaVersions` as `() -> unbackfilledDeltaVersions` and create the mock handler in this method. Just like how we do for the checkpoint v2.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
```diff
@@ -849,13 +851,62 @@ class SnapshotManagerSuite extends AnyFunSuite with MockFileSystemClientUtils {
...
+    val unbackfilledCommits2 = 51L to 60L
+    testWithCheckpoints(
+      (0L to 50L),
+      Seq(10, 30, 50),
+      Seq(20, 40),
+      numParts = 5,
+      unbackfilledDeltaVersions = unbackfilledCommits2,
+      tableCommitCoordinatorClientHandlerOpt =
+        Optional.of(new MockTableCommitCoordinatorClientHandler(logPath, unbackfilledCommits2))
+    )
```
vkorukanti, line 896: what about the case where the backfilled and unbackfilled commits are overlapped?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/internal/coordinatedcommits/CoordinatedCommitsSuite.scala
```diff
@@ -0,0 +1,388 @@
...
+
+  def testWithCoordinatorCommits(
+    hadoopConf: Map[String, String] = Map.empty, f: (String, Engine) => Unit): Unit = {
+    InMemoryCommitCoordinatorBuilder.clearInMemoryInstances()
+    withTempDirAndEngine(f, hadoopConf)
+  }
+
+  test("cold snapshot initialization") {
+    val config = trackingInMemoryBatchSize10Config
+    testWithCoordinatorCommits(config, {
```
vkorukanti, line 130: you can actually move the `test("..")` inside `testWithCoordinatorCommits`.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/internal/coordinatedcommits/CoordinatedCommitsSuite.scala
```diff
@@ -0,0 +1,388 @@
...
+
+/**
+ * A [[CommitCoordinatorClient]] that tests can use to check the coordinator configuration and
+ * table configuration.
+ *
+ * @param EXP_TABLE_CONF The expected table configuration that the builder should receive.
+ * @param EXP_COORDINATOR_CONF The expected coordinator configuration that the builder should
+ *                             receive.
+ * @param coordinatedCommitsTableConf The table configuration that the coordinator receives.
+ * @param conf The coordinator configuration that the builder receives.
```
vkorukanti, line 348: don't see any of these params actually used.

vkorukanti (changes requested):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -131,13 +133,38 @@ public Snapshot getSnapshotAt(
...
+        // will be created.
+        SnapshotImpl snapshot = logSegmentOpt
+                .map(logSegment -> getCoordinatedCommitsAwareSnapshot(
+                        engine, logSegment, Optional.of(version)))
+                .orElseThrow(() -> new TableNotFoundException(tablePath.toString()));
+        long snapshotVer = snapshot.getVersion(engine);
+        if (snapshotVer != version) {
+            throw DeltaErrors.versionAfterLatestCommit(tablePath.toString(), version, snapshotVer);
+        }
+        List<FileStatus> snapshotDeltas = snapshot.getLogSegment().deltas;
```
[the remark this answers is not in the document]
EstherBear: This check was in `getLogSegmentAtOrBeforeVersion` before and now it's moved up to `getSnapshotAt`.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
```diff
@@ -895,3 +946,18 @@ class MockSidecarJsonHandler(sidecars: Seq[FileStatus])
       predicate: Optional[Predicate]): CloseableIterator[ColumnarBatch] =
     singletonSidecarIterator(sidecars)
 }
+
+class MockTableCommitCoordinatorClientHandler(
+  logPath: Path, versions: Seq[Long] = Seq.empty, e: Throwable = null)
+  extends TableCommitCoordinatorClientHandler(null, null, null) {
+  override def getCommits(
+    startVersion: javaLang.Long, endVersion: javaLang.Long): GetCommitsResponse = {
```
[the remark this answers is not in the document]
EstherBear: We need to override the getCommits of TableCommitCoordinatorClientHandler which uses `javaLang.Long` here?

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
```diff
@@ -462,21 +477,6 @@ class SnapshotManagerSuite extends AnyFunSuite with MockFileSystemClientUtils {
...
-        "Cannot load table version 15 as it does not exist. The latest available version is 0"
-    )
-    testExpectedError[RuntimeException](
-      files = deltaFileStatuses((10L until 13L)) ++ singularCheckpointFileStatuses(Seq(10L)),
-      versionToLoad = Optional.of(15),
-      expectedErrorMessageContains =
-        "Cannot load table version 15 as it does not exist. The latest available version is 12"
-    )
-  }
-
```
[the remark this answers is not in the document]
EstherBear: Yes, I added in `CoordinatedCommitsSuite`:
image

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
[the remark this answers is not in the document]
EstherBear: I think I can remove this check because of [this comment](https://github.com/delta-io/delta/pull/3381#discussion_r1704699389).

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -367,9 +462,35 @@ private SnapshotImpl getSnapshotAtInit(Engine engine)
...
+                .map(logSegment -> getCoordinatedCommitsAwareSnapshot(
+                        engine, logSegment, Optional.empty()))
+                .orElseThrow(() -> new TableNotFoundException(tablePath.toString()));
+    }
+
+    /**
+     * This can be optimized by making snapshot hint optimization to work with coordinated commits.
+     * @see <a href="https://github.com/delta-io/delta/issues/3437">issue #3437</a>.
+     */
+    private SnapshotImpl getCoordinatedCommitsAwareSnapshot(
```
allisonport-db, line 474: For coordinated commit table then do we list the log dir twice? Can we not reuse the log segment we've already built? If this is because there may have been changes I'm wondering if it's possible (maybe to much additional complexity) to list from the last file we've seen and reuse existing segment.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -367,9 +462,35 @@ private SnapshotImpl getSnapshotAtInit(Engine engine)
...
+            Engine engine,
+            LogSegment initialSegmentForNewSnapshot,
+            Optional<Long> versionToLoadOpt) {
+        SnapshotImpl newSnapshot = createSnapshot(initialSegmentForNewSnapshot, engine);
```
allisonport-db, line 478: Can we exit early if the `newSnapshot.version = versionToLoad`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
EstherBear: We must list the log dir twice otherwise there might be gap because of this [issue](https://github.com/delta-io/delta/pull/3381#discussion_r1693689307).

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
EstherBear: Sure, we can make explicit check. But I think that's the case without CC. So it should exit early with current logic.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
EstherBear: So the logic should be unbackfilled commit list -> deltalog list -> filter out duplicated unbackfilled commits using the results from deltalog list otherwise there might be gap or duplicated commits.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
allisonport-db: But if we have CC and are getting a snapshot version that is available on the filesystem we dont need to get anything from CC right?

EstherBear: Yes, you are right. Thanks for pointing this out! I'll add early exit.

vkorukanti (approved):
Thank you for adding this feature to Delta Kernel!

PR #3388, opened 2024-07-17
Move OutputTimestampType to `DeltaSQLConf`

Fokko:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaParquetFileFormat.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaParquetFileFormat.scala
@@ -265,11 +265,8 @@ case class DeltaParquetFileFormat(
        dataSchema: StructType): OutputWriterFactory = {
     val factory = super.prepareWrite(sparkSession, job, options, dataSchema)
     val conf = ContextUtil.getConfiguration(job)
-    // If the key is not set explicitly, default to INT64 for Iceberg compatability
-    if (!sparkSession.conf.contains(SQLConf.PARQUET_OUTPUT_TIMESTAMP_TYPE.key)) {
-      conf.set(SQLConf.PARQUET_OUTPUT_TIMESTAMP_TYPE.key,
-        SQLConf.ParquetOutputTimestampType.TIMESTAMP_MICROS.toString)
-    }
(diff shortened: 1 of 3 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/DeltaParquetFileFormat.scala
```diff
@@ -258,6 +258,8 @@ case class DeltaParquetFileFormat(
     }
   }
 
+  private val UNDEFINED = "<undefined>"
```
lzlfred, line 261: i would suggest not to introduce this constant...

spark/src/main/scala/org/apache/spark/sql/delta/DeltaParquetFileFormat.scala
```diff
@@ -266,7 +268,11 @@ case class DeltaParquetFileFormat(
     val factory = super.prepareWrite(sparkSession, job, options, dataSchema)
     val conf = ContextUtil.getConfiguration(job)
     // If the key is not set explicitly, default to INT64 for Iceberg compatability
-    if (!sparkSession.conf.contains(SQLConf.PARQUET_OUTPUT_TIMESTAMP_TYPE.key)) {
+    val parquetOutputTimestampType = sparkSession.conf.get(
```
lzlfred, line 271: uhm i am not sure how this would work... 
https://github.com/apache/spark/blob/5d16c3134c442a5546251fd7c42b1da9fdf3969e/sql/catalyst/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala#L1056

iiuc. the spark conf should already gives a default, so idk if here it will return int96 or undefined... 

I think maybe a cleaner way is to duplicate PARQUET_OUTPUT_TIMESTAMP_TYPE to DeltaSQLConf, with default of 64, and only use that in this DeltaParquetFileFormat ?

Fokko: Thanks @lzlfred  for the suggestion, and I like that very much. I've updated the PR accordingly 👍

lzlfred (approved):
lgtm !

PR #3398, opened 2024-07-19
[Protocol]: Allow storing UTC adjusted timestamp partition values

amogh-jahagirdar:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [X ] Other Protocol

## Description

Currently, in the Delta Protocol, timestamps are not stored with their time zone. This leads to unexpected behavior when querying across systems with different timezones configured (e.g. different spark sessions for instance). For instance in Spark, the timestamp value will be adjusted to spark session time zone and written to the delta log partition values without TZ. If someone were to query the same "timestamp" from a different session timezone, the same time zone value it can fail to surface results due to partition pruning.

What this change proposes to the delta lake protocol is to allow timestamp partition values to be adjusted to UTC and explicitly stored in partition values with a UTC suffix. This way values can reasonably round trip across different TZs without any strange correctness bugs. The original approach is still supported for compatibility but it is recommended for newer writers to write with UTC suffix.

This is also important for Iceberg Uniform conversion because Iceberg timestamps must be UTC adjusted. Now we have a well defined format for UTC in delta, we can convert string partition values to Iceberg longs to make Uniform conversion succeed.

## How was this patch tested?

This is a protocol change. The integration with Spark is here: https://github.com/delta-io/delta/pull/3378/files

This is a protocol change, integrations which leverage these new capabilities will have user facing changes in the sense of timestamp partition values being written in a new form.

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -1754,13 +1754,16 @@ Type | Serialization Format
 string | No translation required
 numeric types | The string representation of the number
 date | Encoded as `{year}-{month}-{day}`. For example, `1970-01-01`
-timestamp | Encoded as `{year}-{month}-{day} {hour}:{minute}:{second}` or `{year}-{month}-{day} {hour}:{minute}:{second}.{microsecond}` For example: `1970-01-01 00:00:00`, or `1970-01-01 00:00:00.123456`
+timestamp | Encoded as `{year}-{month}-{day} {hour}:{minute}:{second}` or `{year}-{month}-{day} {hour}:{minute}:{second}.{microsecond}`. For example: `1970-01-01 00:00:00`, or `1970-01-01 00:00:00.123456`. Timestamps may also be encoded as an ISO8601 formatted timestamp adjusted to UTC timestamp such as `1970-01-01T00:00:00.123456Z`
 timestamp without timezone | Encoded as `{year}-{month}-{day} {hour}:{minute}:{second}` or `{year}-{month}-{day} {hour}:{minute}:{second}.{microsecond}` For example: `1970-01-01 00:00:00`, or `1970-01-01 00:00:00.123456` To use this type, a table must support a feature `timestampNtz`. See section [Timestamp without timezone (TimestampNtz)](#timestamp-without-timezone-timestampNtz) for more information.
 boolean | Encoded as the string "true" or "false"
 binary | Encoded as a string of escaped binary values. For example, `"\u0001\u0002\u0003"`
 
-Note: A `timestamp` value in a partition value doesn't store the time zone due to historical reasons.
-It means its behavior looks similar to `timestamp without time zone` when it is used in a partition column.
+Note: A timestamp value in a partition value may be stored in one of the following ways:
+1. Without a timezone, where the timestamp should be interpreted using the time zone of the system which wrote to the table.
+2. Adjusted to UTC and stored in ISO8601 format.
+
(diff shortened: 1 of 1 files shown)


PROTOCOL.md
```diff
@@ -1754,13 +1754,16 @@ Type | Serialization Format
 string | No translation required
 numeric types | The string representation of the number
 date | Encoded as `{year}-{month}-{day}`. For example, `1970-01-01`
-timestamp | Encoded as `{year}-{month}-{day} {hour}:{minute}:{second}` or `{year}-{month}-{day} {hour}:{minute}:{second}.{microsecond}` For example: `1970-01-01 00:00:00`, or `1970-01-01 00:00:00.123456`
+timestamp | Encoded as `{year}-{month}-{day} {hour}:{minute}:{second}` or `{year}-{month}-{day} {hour}:{minute}:{second}.{microsecond}`. Each of these options may also be suffixed with UTC for UTC adjusted timezones. For example: `1970-01-01 00:00:00`, or `1970-01-01 00:00:00.123456` or `2024-06-14 04:00:00UTC` or `2024-06-14 04:00:00.123456UTC`
```
lzlfred, line 1757: do we also want to put the Z here ?  eg 2024-06-14 04:00:00Z ?
I do not have opinion..

bart-samwel: I think we can allow both. Using Z is actually slightly better because it's shorter. :)

amogh-jahagirdar: I was thinking about this a bit more, we probably want the protocol to just be pretty strict here since imo there's not much value in enabling multiple ways to store the partition value. Since we are just now defining UTC adjusted timestamps as allowed, we don't have to worry about compatibility and being strict ensures that everyone is speaking the same language here.

I agree that "Z" is better, not only because it's shorter but also we probably want to stick to well established industry standards such as ISO8601 where there's no concept of adding the "UTC" suffix. It must be "Z". 

So I've updated the protocol proposal change so that the UTC adjusted timestamp must be stored as ISO8601 format.

This has some other implications:

ISO8601 does *not* allow space between the time and the date. There must be a "T" character. 

Again, if you compare this format to the current format which is not UTC adjusted it's a bit different but I think it's better since ISO8601 is already a standard and people's existing timestamp formatters would *just work*.

amogh-jahagirdar: I was thinking about this a bit more, we probably want the protocol to just be pretty strict here since imo there's not much value in enabling multiple ways to store the partition value. Since we are just now defining UTC adjusted timestamps as allowed, we don't have to worry about compatibility and being strict ensures that everyone is speaking the same language here.

I agree that "Z" is better, not only because it's shorter but also we probably want to stick to well established industry standards such as ISO8601 where there's no concept of adding the "UTC" suffix. It must be "Z". 

So I've updated the protocol proposal change so that the UTC adjusted timestamp must be stored as ISO8601 format.

This has some other implications:

ISO8601 does *not* allow space between the time and the date. There must be a "T" character. 

Again, if you compare this format to the current format which is not UTC adjusted it's a bit different but I think it's better since ISO8601 is already a standard and people's existing timestamp formatters would *just work*.

lzlfred (approved):
LGTM !

amogh-jahagirdar:
I will actually close this in favor of https://github.com/delta-io/delta/pull/3378 which includes these changes.

PR #3330, opened 2024-07-02
[Spark] Use location as the table path

Fokko:
#### Which Delta project/connector is this regarding?

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [x] Other (fill in here)

## Description

This will include the table path, to avoid defaulting to the path that Hive suggests.

PR #3399, opened 2024-07-19
[UNIFORM] Remove unnecessary iceberg patch and disable cleanup of files in expire snapshots via cleanExpiredFiles API

amogh-jahagirdar:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ X] Other Uniform
## Description
We don't need a custom patch to disable cleaning up data files for Iceberg's expire snapshots API. icebergShaded/iceberg_src_patches/0002-iceberg-core-must-not-delete-any-delta-data-files.patch

An option to disable file cleanup already exists through the `cleanExpiredFiles` API.

## How was this patch tested?
Iceberg's cleanFiles option is already tested by Iceberg. I can add separate tests here to make sure that the custom transaction logic for Uniform uses that option.

No, it preserves the existing behavior. Snapshots can be removed, but files will never be deleted.


amogh-jahagirdar:
Since this is related to file cleanup, let me add a test here to make sure the Uniform integration with the Iceberg logic works as expected to be really safe.

iceberg/src/test/scala/org/apache/spark/sql/delta/ConvertToIcebergSuite.scala
```diff
@@ -109,6 +111,7 @@ class ConvertToIcebergSuite extends QueryTest with Eventually {
       runDeltaSql(
         s"""CREATE TABLE `${testTableName}` (col1 INT) USING DELTA
            |TBLPROPERTIES (
+           |  'delta.enableIcebergCompatV2' = 'true',
```
amogh-jahagirdar, line 114: @Fokko already fixed this in https://github.com/delta-io/delta/pull/3330/files, I think we'd want to get that in first. These existing tests do not pass at the moment without this.

iceberg/src/test/scala/org/apache/spark/sql/delta/ConvertToIcebergSuite.scala
```diff
@@ -123,21 +126,82 @@ class ConvertToIcebergSuite extends QueryTest with Eventually {
...
           .option("path", testTablePath)
+          .option("delta.enableIcebergCompatV2", "true")
           .saveAsTable(testTableName)
       }
       verifyReadWithIceberg(testTableName, 0 to 19 map (Row(_)))
     }
   }
 
-  def runDeltaSql(sqlStr: String): Unit = {
+  test("Expire Snapshots") {
```
amogh-jahagirdar, line 144: Added a new test since we don't really have coverage on this path currently

iceberg/src/test/scala/org/apache/spark/sql/delta/ConvertToIcebergSuite.scala
Fokko: I don't think those tests run in CI because it will skip over them when the Docker container is not there. Nevertheless, I think it would be good to get #3330 in to simplify testing.

lzlfred (approved):
lgtm ! thanks for the clean up

amogh-jahagirdar:
Thanks for the review @lzlfred !

PR #3412, opened 2024-07-23
[Spark] Fix the semantic of `shouldRewriteToBeIcebergCompatible` in REORG UPGRADE UNIFORM

xzhseh:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
currently we utilize the helper function `shouldRewriteToBeIcebergCompatible` to filter the portion of parquet files that need to be rewritten when running `REORG UPGRADE UNIFORM` based on the tags in the `AddFile`.

however, the `DeltaUpgradeUniformOperation.icebergCompatVersion` is accidentally shadowed, which will make `shouldRewriteToBeIcebergCompatible` always return `false` if the `AddFile.tags` is not `null` - this is not the expected semantic of this function.

this PR introduces the fix for this problem and add unit tests to ensure the correctness.

## How was this patch tested?
through unit tests in `UniFormE2ESuite.scala`.

no.

### DIFF
--- a/iceberg/src/test/scala/org/apache/spark/sql/delta/uniform/UniFormE2ESuite.scala
+++ b/iceberg/src/test/scala/org/apache/spark/sql/delta/uniform/UniFormE2ESuite.scala
@@ -18,9 +18,12 @@ package org.apache.spark.sql.delta.uniform
 
 import org.apache.spark.sql.delta.sources.DeltaSQLConf
 import org.apache.spark.sql.delta.uniform.{UniFormE2EIcebergSuiteBase, UniFormE2ETest}
-
 import org.apache.spark.SparkConf
+
 import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.catalyst.TableIdentifier
+import org.apache.spark.sql.delta.DeltaLog
+import org.apache.spark.sql.delta.actions.AddFile
 import org.apache.spark.sql.delta.test.DeltaSQLCommandTest
 import org.apache.spark.sql.delta.uniform.hms.HMSTest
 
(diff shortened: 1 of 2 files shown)


iceberg/src/test/scala/org/apache/spark/sql/delta/uniform/UniFormE2ESuite.scala
```diff
@@ -38,6 +41,56 @@ trait WriteDeltaHMSReadIceberg extends UniFormE2ETest with DeltaSQLCommandTest w
...
+  private def runReorgUpgradeUniform(icebergCompatVersion: Int): Unit = {
+    write(
+      s"""
+         | REORG TABLE $testTableName APPLY
+         | (UPGRADE UNIFORM (ICEBERG_COMPAT_VERSION = $icebergCompatVersion))
+         |""".stripMargin
+    )
+  }
+
+  private def assertTagsExistForLatestSnapshot(isNull: Boolean): Unit = {
```
KaiqiJinWow, line 54:
```suggestion
  private def assertTagsExistForLatestSnapshot(tagShouldExist: Boolean): Unit = {
```

iceberg/src/test/scala/org/apache/spark/sql/delta/uniform/UniFormE2ESuite.scala
```diff
@@ -38,6 +41,56 @@ trait WriteDeltaHMSReadIceberg extends UniFormE2ETest with DeltaSQLCommandTest w
...
+    val snapshot = DeltaLog.forTable(spark, new TableIdentifier(testTableName)).update()
+    snapshot.allFiles.collect().forall { addFile =>
+      if (isNull) {
+        addFile.tags == null
+      } else {
+        addFile.tags.getOrElse(AddFile.Tags.ICEBERG_COMPAT_VERSION.name, "0").contains("2")
```
KaiqiJinWow, line 60: Can you use `==` instead of `contains`?

johanl-db (approved):

PR #3416, opened 2024-07-24
[Spark][Kernel][Protocol] Merge InCommitTimestamp RFC, and remove the -preview suffix from feature name and configs.

dhruvarya-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [X] Other (Protocol)

## Description

1. Merges the InCommitTimestamp RFC
2. Removes the -preview suffix from the feature name and properties.

## How was this patch tested?

Existing tests should cover this change.

No


PROTOCOL.md
```diff
@@ -521,7 +521,7 @@ Specifically, to read the row-level changes made in a version, the following str
     Field Name | Data Type | Description
     -|-|-
     _commit_version|`Long`| The table version containing the change. This can be derived from the name of the Delta log file that contains actions.
-    _commit_timestamp|`Timestamp`| The timestamp associated when the commit was created. This can be derived from the file modification time of the Delta log file that contains actions.
+    _commit_timestamp|`Timestamp`| The timestamp associated when the commit was created. Depending on whether [In-Commit Timestamps](#in-commit-timestamps) are enabled, this is derived from either the `inCommitTimestamp` field of the `commitInfo` action of the version's Delta log, or from the Delta log's file modification time.
```
prakharjain09, line 524:
```suggestion
    _commit_timestamp|`Timestamp`| The timestamp associated when the commit was created. Depending on whether [In-Commit Timestamps](#in-commit-timestamps) are enabled, this is derived from either the `inCommitTimestamp` field of the `commitInfo` action of the version's commit file, or from the commit file's modification time.
```

PROTOCOL.md
```diff
@@ -620,6 +620,8 @@ A delta file can optionally contain additional provenance information about what
 
 Implementations are free to store any valid JSON-formatted data via the `commitInfo` action.
 
+When In-Commit Timestamps are enabled, writers are required to include a commitInfo action with every commit, which must include the `inCommitTimestamp` field.
```
prakharjain09, line 623: Cross link `In-Commit Timestamps` to right section.

prakharjain09: Do we also enforce commitInfo to be the first line in commit file as part of this feature?

prakharjain09 (approved):

PROTOCOL.md
dhruvarya-db: Should it be Delta log file instead? To keep it consistent with the existing line above:
"_commit_version|`Long`| The table version containing the change. This can be derived from the name of the Delta log file that contains actions."

prakharjain09 (approved):

PR #3418, opened 2024-07-24
[Spark] Add Delta log throttling class

larsk-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

- Add a token bucket based throttler implementation that can be used to throttle log messages that are suspected of potential log flooding issues.

## How was this patch tested?

New test suite: `LogThrottlingSuite`

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaTable.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaTable.scala
@@ -22,7 +22,7 @@ import scala.util.control.NonFatal
 
 import org.apache.spark.sql.delta.files.{TahoeFileIndex, TahoeLogFileIndex}
 import org.apache.spark.sql.delta.logging.DeltaLogKeys
-import org.apache.spark.sql.delta.metering.DeltaLogging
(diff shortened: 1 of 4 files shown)


zhipengmao-db (approved):

xupefei (approved):
LGTM!

felipepessoto:
@larsk-db @vkorukanti I wonder what the motivation for adding the throttling logic, and specifically to a single log message.

Is there any known issue?

Thanks

larsk-db:
> @larsk-db @vkorukanti I wonder what the motivation for adding the throttling logic, and specifically to a single log message.
> 
> Is there any known issue?
> 
> Thanks

We had some concerns that this particular message could flood the logs really badly if there's some transient outage on a deep path where it just keeps throwing errors as we traverse the hierarchy. And that many times in parallel for parallel queries potentially.
We really don't need to know a million times a second that this particular bucket isn't accessible right now.

There's some expectation that we will re-use this throttling code in more places in the future where similar issues are a concern.

PR #3422, opened 2024-07-25
[Spark] Flip isDeltaTable.throwOnError to true

larsk-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

- Change the default value of the isDeltaTable.throwOnError flag to `true`.

## How was this patch tested?

Existing tests (was already `true` in testing).

Resolving a Delta that is accessed by path (e.g. DeltaTable.forPath() or SELECT ... FROM delta.<path>) will now forward exceptions thrown while accessing , instead of always throwing a DELTA_MISSING_DELTA_TABLE exception.
This helps locating issues such as missing access permissions without having to through support.


bart-samwel (approved):
Woohoo!

PR #3443, opened 2024-07-30
[Spark] Handle type mismatches in Delta streaming sink

johanl-db:
## Description
This change enables writing data to a Delta streaming sink using data types that differ from the actual Delta table schema. This is achieved by adding an implicit cast to columns when needed. Casting behavior respects the configuration `spark.sql.storeAssignmentPolicy`, similar to batch INSERT.

## How was this patch tested?
- Added test suite `DeltaSinkImplicitCastSuite` covering writing to a Delta sink using mismatching types. Covers e.p. interactions with: schema evolution, schema overwrite, column mapping, partition columns, case sensitivity.

Previously, writing to a Delta sink using a type that doesn't match the column type in the Delta table failed with `DELTA_FAILED_TO_MERGE_FIELDS`:
```
spark.readStream
    .table("delta_source")
    # Column 'a' has type INT in 'delta_sink'.
    .select(col("a").cast("long").alias("a"))
    .writeStream
    .format("delta")
    .option("checkpointLocation", "<location>")
    .toTable("delta_sink")

DeltaAnalysisException: [DELTA_FAILED_TO_MERGE_FIELDS] Failed to merge fields 'a' and 'a'
```
With this change, writing to the sink now succeeds and data is cast from `LONG` to `INT`. If any value overflows, the stream fails with (assuming default `storeAssignmentPolicy=ANSI`):
```
SparkArithmeticException: [CAST_OVERFLOW_IN_TABLE_INSERT] Fail to assign a value of 'LONG' type to the 'INT' type column or variable 'a' due to an overflow. Use `try_cast` on the input value to tolerate overflow and return NULL instead."
```


spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala
```diff
@@ -1466,6 +1466,15 @@ trait DeltaSQLConfBase {
       .booleanConf
       .createWithDefault(true)
 
+  val DELTA_STREAMING_SINK_ALLOW_TYPE_CHANGES =
```
tomvanbussel, line 1469: This name confused me a little bit. "Type changes" to me suggest that the types in the table will be changed by streaming sink, but IIUC then this merely allows the types of the written dataframe to differ from the types in the Delta table. Maybe call this something like "add implicit casts"?

spark/src/test/scala/org/apache/spark/sql/delta/DeltaSinkTypeChangeSuite.scala
```diff
@@ -0,0 +1,495 @@
...
+import org.apache.spark.sql.functions.{col, lit}
+import org.apache.spark.sql.internal.SQLConf
+import org.apache.spark.sql.internal.SQLConf.StoreAssignmentPolicy
+import org.apache.spark.sql.streaming.{OutputMode, StreamingQueryException}
+import org.apache.spark.sql.types._
+
+/**
+ * Covers handling type differences when writing data to a Delta sink.
+ */
+abstract class DeltaSinkTypeChangeTest extends DeltaSinkTest {
```
tomvanbussel, line 36: Let's also rename this to be less confusing.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaSinkTypeChangeSuite.scala
```diff
@@ -0,0 +1,495 @@
...
+  def checkOperationHistory[T](stream: TestDeltaStream[T], expectedOperations: Seq[String])
+  : Unit = {
+    val history = sql(s"DESCRIBE HISTORY delta.`${stream.deltaLog.dataPath}`")
+      .sort("version")
+      .select("operation")
+    checkAnswer(history, expectedOperations.map(Row(_)))
+  }
+}
+
+class DeltaSinkTypeChangeSuite extends DeltaSinkTypeChangeTest {
```
tomvanbussel, line 108: And here. (and all the test cases)

tomvanbussel (commented):
Nice! The tests are very extensive! Left a few comments.

spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSink.scala
```diff
@@ -158,6 +163,65 @@ case class DeltaSink(
...
+      tableSchema = metadata.schema,
+      dataSchema = dataSchema,
+      allowImplicitConversions = true
+    )
+  }
+
+  /**
+   * Override [[UpdateExpressionsSupport.cast]] to always apply ANSI cast when casting data,
+   * ignoring 'spark.sql.storeAssignmentPolicy' - we don't need to support LEGACY which is a footgun
+   * and STRICT brings little value.
```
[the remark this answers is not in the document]
johanl-db: I started having second thoughts about also, I'm not really happy with batch insert respecting storeAssignmentPolicy and streaming insert not respecting it. I realistically won't introduce a try_cast option for storeAssignmentPolicy here, the best is probably to fully align batch insert and streaming insert casting behavior and have both respect `storeAssignmentPolicy`

spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSink.scala
```diff
@@ -158,6 +163,65 @@ case class DeltaSink(
...
+   */
+  override protected def cast(child: Expression, dataType: DataType, columnName: String)
+    : Expression = {
```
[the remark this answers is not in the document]
johanl-db: This method was removed in latest changes

tomvanbussel (approved):
LGTM!

tdas (approved):

PR #3452, opened 2024-07-31
[Spark] Add annotation for merge materialize source stage of merge

juliuszsompolski:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Originally MERGE source materialization was lazy, and triggered when the source was used for the first time. Because of that, it couldn't be cleanly separated as a stage. Since it was changed to be eager, we can now annotate it, which should make it easier to find in Spark UI.

## How was this patch tested?

Merge materialize source stage is now annotated:
image

Unit tests validate that the new stage is present in MERGE commit metrics.

No


juliuszsompolski:
cc @larsk-db

larsk-db (approved):
LGTM

PR #3454, opened 2024-08-01
[Spark] Allow stale reads when commit coordinator has not been registered

dhruvarya-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Snapshot construction will now be allowed even if the table has a configured coordinator but the coordinator implementation has not been registered. Writes will still be blocked in such cases.

The change has been gated behind a new flag.

## How was this patch tested?

Added a new test in CoordinatedCommitsSuite.

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
@@ -448,9 +448,7 @@ trait OptimisticTransactionImpl extends TransactionalWrite
   // The commit-coordinator of a table shouldn't change. If it is changed by a concurrent commit,
   // then it will be detected as a conflict and the transaction will anyway fail.
   private[delta] val readSnapshotTableCommitCoordinatorClientOpt:
-    Option[TableCommitCoordinatorClient] = {
-      snapshot.tableCommitCoordinatorClientOpt
-  }
+    Option[TableCommitCoordinatorClient] = snapshot.getTableCommitCoordinatorForWrites
 
   /**
    * Generates a timestamp which is greater than the commit timestamp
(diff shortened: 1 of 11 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/coordinatedcommits/CommitCoordinatorClient.scala
```diff
@@ -237,11 +237,26 @@ object CommitCoordinatorProvider {
       name: String,
       conf: Map[String, String],
       spark: SparkSession): CommitCoordinatorClient = synchronized {
-    nameToBuilderMapping.get(name).map(_.build(spark, conf)).getOrElse {
+    getCommitCoordinatorClientOpt(name, conf, spark).getOrElse {
```
prakharjain09, line 240: Also update the documentation to say - throws the exception in case the give commit coordinator is not found.

spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
```diff
@@ -440,8 +440,25 @@ trait OptimisticTransactionImpl extends TransactionalWrite
...
+      if (snapshot.version != -1 && coordinatorName.isDefined && coordinatorOpt.isEmpty) {
+        recordDeltaEvent(
+          deltaLog,
+          CoordinatedCommitsUsageLogs.COMMIT_COORDINATOR_MISSING_IMPLEMENTATION_WRITE,
+          data = Map(
+            "commitCoordinatorName" -> coordinatorName.get,
+            "readVersion" -> readVersion.toString
+          )
+        )
+        throw new IllegalStateException(
```
prakharjain09, line 455: User could face this error. This should follow the error class framework. We should fix this in same PR or a followup PR.

dhruvarya-db: Done

spark/src/main/scala/org/apache/spark/sql/delta/coordinatedcommits/CommitCoordinatorClient.scala
dhruvarya-db: Updated

prakharjain09 (approved):

PR #3466, opened 2024-08-02
[Kernel] Configure code formatter for Java Code

nastra:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This configures a code formatter using *google-java-format* for the Java code in the Kernel (similarly to how it was done in https://github.com/unitycatalog/unitycatalog/commit/54b76d88255dd2baa3f11e515159cfd34cb295e2).

Code can be checked by running `build/sbt javafmtCheckAll`. Code can automatically be formatted `build/sbt javafmtAll`.
Once this PR is in, we should open follow-up PRs where the code in the Kernel is properly formatted. 
After that we can enforce the new code style during compilation

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -229,6 +229,7 @@ def runTaskOnlyOnSparkMaster[T](
 }
 
 lazy val connectCommon = (project in file("spark-connect/common"))
+  .disablePlugins(JavaFormatterPlugin)
   .settings(
     name := "delta-connect-common",
     commonSettings,
@@ -267,6 +268,7 @@ lazy val connectCommon = (project in file("spark-connect/common"))
   )
 
 lazy val connectClient = (project in file("spark-connect/client"))
+  .disablePlugins(JavaFormatterPlugin)
   .dependsOn(connectCommon % "compile->compile;test->test;provided->provided")
   .settings(
     name := "delta-connect-client",
@@ -355,6 +357,7 @@ lazy val connectClient = (project in file("spark-connect/client"))
 lazy val connectServer = (project in file("spark-connect/server"))
   .dependsOn(connectCommon % "compile->compile;test->test;provided->provided")
   .dependsOn(spark % "compile->compile;test->test;provided->provided")
+  .disablePlugins(JavaFormatterPlugin)
   .settings(
     name := "delta-connect-server",
     commonSettings,
@@ -398,6 +401,7 @@ lazy val connectServer = (project in file("spark-connect/server"))
 lazy val spark = (project in file("spark"))
   .dependsOn(storage)
   .enablePlugins(Antlr4Plugin)
+  .disablePlugins(JavaFormatterPlugin)
   .settings (
     name := "delta-spark",
     commonSettings,
@@ -485,6 +489,7 @@ lazy val spark = (project in file("spark"))
 
 lazy val contribs = (project in file("contribs"))
   .dependsOn(spark % "compile->compile;test->test;provided->provided")
+  .disablePlugins(JavaFormatterPlugin)
   .settings (
     name := "delta-contribs",
     commonSettings,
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -229,6 +229,7 @@ def runTaskOnlyOnSparkMaster[T](
 }
 
 lazy val connectCommon = (project in file("spark-connect/common"))
+  .disablePlugins(JavaFormatterPlugin)
```
scottsand-db, line 232: so, this is opt-out? We cannot have delta-kernel-api and delta-kernel-defaults opt-in?

kernel/README.md
```diff
@@ -68,3 +68,13 @@ We use [GitHub Issues](https://github.com/delta-io/delta/issues) to track commun
 
 # Contributing
 We welcome contributions to Delta Lake and we accept contributions via Pull Requests. See our [CONTRIBUTING.md](https://github.com/delta-io/delta/blob/master/CONTRIBUTING.md) for more details. We also adhere to the [Delta Lake Code of Conduct](https://github.com/delta-io/delta/blob/master/CODE_OF_CONDUCT.md).
+
+# Setting up IDE
+
+Java code adheres to the [Google style](https://google.github.io/styleguide/javaguide.html), which is verified via `build/sbt javafmtCheckAll` during builds.
```
scottsand-db, line 74: let's say and use `build/sbt kernelGroup/javafmtCheckAll` and `build/sbt kernelGroup/javafmtAll` instead (i.e. specify kernelGroup, don't just run this against the root)

scottsand-db (changes requested):
Looking great! Left two important comments.

Can we make it so we apply this check/formatting by default? (Note: not necessarily right now in this PR, but perhaps in a followup PR) Can we add a param --skip-fmt (or something else) that skips formatting for quick local iteration? i.e. to let devs code quickly and then just worry about foramtting at the end. and our CI will check foramtting by default.

scottsand-db:
Btw @nastra  - from your experience, what benefit is there to javafmtchecks as well as checkstyle checks?

nastra:
> Can we make it so we apply this check/formatting by default? (Note: not necessarily right now in this PR, but perhaps in a followup PR)

Yes that is the plan once this PR is in and we re-formatted the code for the kernel project(s)


> Can we add a param --skip-fmt (or something else) that skips formatting for quick local iteration? i.e. to let devs code quickly and then just worry about foramtting at the end. and our CI will check foramtting by default.

I don't think this makes sense to have. If a Dev types `--skip-fmt` then he can also type `javafmtAll`to apply formatting. Besides that, we can also run formatting as part of compilation so that you don't have to type anything to get the code formatted (well you need to trigger compilation at least)

build.sbt
nastra: I don't know sbt well enough, so from first experience it seems that projects need to opt-out from using a certain plugin

kernel/README.md
nastra: IMO this is just more text to type for a Dev. Given that all other projects are being ignored there's no harm running this against the root project. Pretty much all projects that I've been involved in run this against the root and exclude the formatting from being applied on subprojects.
I can still update the wording if you think it's worth it, but people will typically just type the shortest command to get the code formatted

nastra:
> Btw @nastra - from your experience, what benefit is there to javafmtchecks as well as checkstyle checks?

Both can co-exist and I see lots of projects that use both. Checkstyle has still lots of value in enforcing things like naming, using imports from shaded/non-shaded dependencies and a bunch of other things while the code formatter takes care of the style of the code.

build.sbt
scottsand-db: Yeah; i couldn't figure this out either

scottsand-db (approved):

PR #3467, opened 2024-08-02
[Spark] Add Row Tracking Backfill Conflict Checker RemoveFile rule

longvu-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
`RowTrackingBackfill` (or backfill for short) is a special operation that materializes and recommits all existing files in table using one or several commits to ensure that every AddFile has a base row ID and default Row Commit Version.

In this PR, we add a new rule to `ConflictChecker` that resolve concurrent conflicts involving Backfill. We check that RowTrackingBackfill is not resurrecting files that were removed concurrently and that an AddFile and its corresponding RemoveFile have the same base row ID and default RCV. We also add logic to skip certain concurrency checks if it involves Backfill.

This opens up a lot of interesting Conflict Resolutions cases to test, so we will add more UTs revolving conflict checking/resolutions between Backfill and other operations with different scenario in the next PRs.

## How was this patch tested?
Added ConflictResolution UTs.

No.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
@@ -312,10 +312,10 @@ private[delta] class ConflictChecker(
     if (winningCommitSummary.isRowTrackingBackfillTxn) {
       recordTime(timerPhaseName) {
         val backfillActionMap = winningCommitSummary.addedFilePathToActionMap
-        // Copy over the base row ID and row commit version assigned so that the AddFiles and
-        // RemoveFiles have matching base row ID and row commit version. If an AddFile is
-        // re-committed, it should have the same base row ID and row commit version as the one
-        // assigned by Backfill.
+        // Copy over the base row ID and default row commit version assigned so that the AddFiles
+        // and RemoveFiles have matching base row ID and default row commit version.
+        // If an AddFile is re-committed, it should have the same base row ID and
+        // default row commit version as the one assigned by Backfill.
         val newActions = currentTransactionInfo.actions.map {
           case a: AddFile if backfillActionMap.contains(a.path) =>
             val backfillAction = backfillActionMap(a.path)
--- a/spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
@@ -260,11 +260,12 @@ private[delta] class ConflictChecker(
    *    3. Table property and metadata are updated.
    * RowTrackingBackfill does not do any data change. It doesn't matter whether a file is
    * recommitted after the table feature support from Backfill or some other concurrent transaction;
-   * every AddFile just needs to have a base row ID somehow. However, correctness issues can arise
-   * if we don't do the checks in this method.
+   * every AddFile just needs to have a base row ID and a default row commit version somehow.
+   * However, correctness issues can arise if we don't do the checks in this method.
    *
    * Check that RowTrackingBackfill is not resurrecting files that were removed concurrently and
-   * that an AddFile and its corresponding RemoveFile have the same base row ID. To do this, we:
+   * that an AddFile and its corresponding RemoveFile have the same base row ID and
+   * default row commit version. To do this, we:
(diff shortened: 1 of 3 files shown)


spark/src/test/scala/org/apache/spark/sql/delta/rowtracking/RowTrackingConflictResolutionSuite.scala
```diff
@@ -147,4 +157,172 @@ class RowTrackingConflictResolutionSuite extends QueryTest
...
+      executeBackfill(log, backfillTxn)
+
+      val deleteActions = Seq(file1.remove, file2.remove)
+      // Truncate is a data-changing operation.
+      deleteTxn.commit(deleteActions, Truncate())
+
+      val finalFiles = log.update().allFiles.collect()
+      finalFiles.foreach(a => assert(a.baseRowId.nonEmpty))
+      assertRowIdsAreValid(log)
+      assert(finalFiles.map(_.path).toSet == Seq(file3, file4).map(_.path).toSet)
```
longvu-db, line 265:
```suggestion
      assert(finalFiles.map(_.path).toSet === Seq(file3, file4).map(_.path).toSet)
```

spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
```diff
@@ -238,6 +250,90 @@ private[delta] class ConflictChecker(
...
+          case a: AddFile => a.path
+          case r: RemoveFile => r.path
+        }.toSet
+
+        // Remove files from this Backfill commit if they were removed or re-committed by
+        // a concurrent winning txn.
+        if (filePathsToRemoveFromBackfill.nonEmpty) {
+          val newActions = currentTransactionInfo.actions.filterNot {
+            case a: AddFile => filePathsToRemoveFromBackfill.contains(a.path)
+            case d: DomainMetadata if RowTrackingMetadataDomain.isSameDomain(d) => false
```
johanl-db, line 300: We keep the row tracking metadata action but it's not accurate anymore though - the winning transaction probably increased the high-water mark. I assume we run the conflict resolution rule to re-assign the high-water mark afterwards?

Might be worth calling out that ordering dependency in a comment where these rules are called

spark/src/test/scala/org/apache/spark/sql/delta/rowtracking/RowTrackingConflictResolutionSuite.scala
```diff
@@ -147,4 +157,172 @@ class RowTrackingConflictResolutionSuite extends QueryTest
...
+
+      val finalFiles = log.update().allFiles.collect()
+      finalFiles.foreach(a => assert(a.baseRowId.nonEmpty))
+      assertRowIdsAreValid(log)
+      val allFiles = Seq(file1, file2, file3, file4)
+      assert(finalFiles.map(_.path).toSet === allFiles.map(_.path).toSet)
+    }
+  }
+
+  test("Backfill conflicts with a DV delete, Backfill wins") {
```
johanl-db, line 304: The PR description mentions you'll add more tests later so I'll be patient, but it would be interesting to test other conflict cases where the other transaction isn't a plain delete

johanl-db (approved):
Looks good!

spark/src/test/scala/org/apache/spark/sql/delta/rowtracking/RowTrackingConflictResolutionSuite.scala
longvu-db: @johanl-db Yes, this requires extensive testing so this PR will be huge if we try to add them all here, while Row Tracking Backfill is not enabled we can add them one by one.

spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
longvu-db: @johanl-db Yes, this will be handled by the reassignOverlappingRowIds() that we on purpose place after the resolveRowTrackingBackfillConflicts() in checkConflicts(), adding a comment here for extra clarity.

PR #3472, opened 2024-08-04
[Spark] Add tests that check row tracking preservation works on DMLs for backfilled enabled tables

longvu-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
- Add tests that check row tracking preservation works on DMLs for backfilled enabled tables, as opposed to having row tracking enabled from the start.
- `FileMetadataMaterializationTrackerSuite` has a test that sets a different number of permits than the default in the `FileMetadataMaterializationTracker` but forgot to set it back, which causes unwanted failure in other test. We address that by using the default number of permits in that test instead of a different one.

## How was this patch tested?
Added UTs.

No.

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/delta/FileMetadataMaterializationTrackerSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/FileMetadataMaterializationTrackerSuite.scala
@@ -18,13 +18,17 @@ package org.apache.spark.sql.delta
 
 import java.util.concurrent.Semaphore
 
+import org.apache.spark.sql.delta.sources.DeltaSQLConf
 import org.scalatest.concurrent.TimeLimits
 import org.scalatest.time.{Seconds, Span}
 
 import org.apache.spark.SparkFunSuite
+import org.apache.spark.sql.test.SharedSparkSession
 
-class FileMetadataMaterializationTrackerSuite extends SparkFunSuite with TimeLimits {
-
+class FileMetadataMaterializationTrackerSuite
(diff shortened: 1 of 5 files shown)


spark/src/test/scala/org/apache/spark/sql/delta/FileMetadataMaterializationTrackerSuite.scala
```diff
@@ -35,27 +39,29 @@ class FileMetadataMaterializationTrackerSuite extends SparkFunSuite with TimeLim
     }
 
     // Initialize the semaphore for tests
-    val semaphore = new Semaphore(10)
+    val totalAvailablePermits = spark.sessionState.conf.getConf(
+      DeltaSQLConf.DELTA_COMMAND_FILE_MATERIALIZATION_LIMIT)
+    val semaphore = new Semaphore(totalAvailablePermits)
```
johanl-db, line 44: Wait, that means test suites aren't being run in isolation?
I'm actually surprised we don't have a lot more issues with our tests if that's the case

johanl-db (approved):

spark/src/test/scala/org/apache/spark/sql/delta/FileMetadataMaterializationTrackerSuite.scala
longvu-db: Yes, that's the current state of our Github Workflow Actions is, we just brute-force all the tests sequentially for a thread of execution.

PR #3473, opened 2024-08-04
[Spark] Allowing Row Tracking enablement only txn to not fail concurrent no metadata update txns

longvu-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Introducing a new `DeltaCommitTag` which allows ALTER TABLE commands that do row tracking enablement only not to fail concurrent transactions if they do not do metadata update. This is a special case of metadata update conflict that we can safely resolve.

## How was this patch tested?
Added UTs.

No.


tomvanbussel (approved):
LGTM!

PR #3474, opened 2024-08-05
[Spark][Backport 3.2] Fix the semantic of `shouldRewriteToBeIcebergCompatible` in REORG UPGRADE UNIFORM

xzhseh:
## Which Delta project/connector is this regarding?

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
currently we utilize the helper function
`shouldRewriteToBeIcebergCompatible` to filter the portion of parquet files that need to be rewritten when running `REORG UPGRADE UNIFORM` based on the tags in the `AddFile`.

however, the `DeltaUpgradeUniformOperation.icebergCompatVersion` is accidentally shadowed, which will make
`shouldRewriteToBeIcebergCompatible` always return `false` if the `AddFile.tags` is not `null` - this is not the expected semantic of this function.

this PR introduces the fix for this problem and add unit tests to ensure the correctness.

## How was this patch tested?
through unit tests in `UniFormE2ESuite.scala`.

no.


johanl-db (approved):
Stamping the backport

tdas:
these are known python test infra issue failures. since java tests passed, merging this.

PR #3486, opened 2024-08-05
[UNIFORM] Remove timestamp partition patch as it is not effective currently

amogh-jahagirdar:
Remove timestamp partition patch as it is not effective currently

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [X ] Other Uniform

## Description
Delta-Iceberg uniform currently does not support partitioning on timestamps. There was originally patch intended to address that from the beginning of the project but it's not effective as Iceberg internally relies on long timestamp values since epoch, and the patch converts to java.sql.Timestamp; as a result the conversion currently fails with
```
IllegalArgumentException: Wrong class, expected java.lang.Long, but was java.sql.Timestamp, for object
```

Since this patch is essentially ineffective, I think it makes the most sense to remove it.
This is also very important since it is required to enable us to upgrade Iceberg versions since this patch does not cleanly apply on anything after Iceberg 1.2! 

 Note: I am also working towards adding this support so this gap will be closed soon.

## How was this patch tested?
Existing CI

Technically now a user with an timestamp partitioned table will encounter a different error:

Before this change the error would manifest as:

```
IllegalArgumentException: Wrong class, expected java.lang.Long, but was java.sql.Timestamp, for object: 
```

After this change the error would be 
```
"Unsupported type for fromPartitionString: Timestamp"
```

Considering it's unsupported, the new error change is a bit more clear. Note: I am also working towards adding this support so this gap will be closed soon.


amogh-jahagirdar:
cc @Fokko @lzlfred

lzlfred (approved):
lgtm !

Fokko (approved):
Great catch @amogh-jahagirdar 👍

PR #3491, opened 2024-08-06
[Kernel] Always read the checkpoint manifest for CPV2 if sidecar files are present

chirag-s-db:
- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description
The sidecar actions in the checkpoint manifest of a CPV2 table should always be read regardless of the checkpoint predicate, as actions satisfying the predicate may be found in the sidecar file.

## How was this patch tested?
See added test.

No.


chirag-s-db:
See https://github.com/delta-io/delta/pull/3509

PR #3115, opened 2024-05-18
[INFRA] Improve the java style checks log the errors to sbt console

tlm365:
#### Which Delta project/connector is this regarding?

- [ ] Spark
- [x] Standalone
- [x] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description
Resolves #3067.

## How was this patch tested?
On local machine, intentionally create checkstyle errors in module `kernelDefaults` (for experimental), then run the `build/sbt compile` and `build/sbt kernelDefaults/test`.

## Does this PR introduce _any_ user-facing changes?
No.

PR #3494, opened 2024-08-07
[Infra] Make `checkstyle` fail when there are style errors

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [X] Other (INFRA)

## Description

https://github.com/delta-io/delta/pull/3115 added logging for the java style checks, but incidentally made it such that `kernelApi / checkstyle` doesn't fail when there are errors (only `kernelApi / compile` fails).

This PR makes it so that `kernelApi / checkstyle` will also fail, but it will NOT log the errors. 
- I can't find a way to override the `checkstyle` task defined by the plugin. I _can_ do it for `kernelApi / Compile / checkstyle` but that doesn't seem very helpful.

## How was this patch tested?

Checked the commands locally.

### DIFF
--- a/project/Checkstyle.scala
+++ b/project/Checkstyle.scala
@@ -57,21 +57,19 @@ object Checkstyle {
     // and during tests (e.g. build/sbt test)
     Seq(
       checkstyleConfigLocation := CheckstyleConfigLocation.File(checkstyleFile),
-      // if we keep the Error severity, `build/sbt` will throw an error and immediately stop at
-      // the `checkstyle` phase (if error) -> never execute the `check-report` phase of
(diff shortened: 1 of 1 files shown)


felipepessoto:
I wonder if we should change the 100 characters limit per line. 100 chars are not a lot, even more when the code in indented because it is inside several blocks

foss-contributor:
@allisonport-db, thank you for your contribution! This pull request has been inactive for over 18 months and will be closed in one week. To keep it open, please leave a comment on this pull request or contact us in the `#contributors` channel on [Delta Lake Slack](https://go.delta.io/slack). We appreciate your involvement in Delta Lake!

PR #3516, opened 2024-08-09
[KERNEL][VARIANT] Basic read variant

richardc-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

Adds new `VariantValue` interface and `getVariant` methods to `ColumnVector` interface so column vectors consisting of `VariantValue`s can be created.

In kernel-defaults, we implementthe default implementation such that it understands variants completely for now and can directly scan variants with its own `VariantColumnReader`. This is subject to change in the future to remove the `VariantColumnReader` and instead return structs from the scan.

This PR also does the following

* Cross compiles kernel defaults with Spark 3.5 and Spark 4.0. This is done so the tests can write variants using spark 4.0 rather than needing to manually creating and comparing against variant values
* Allows compilation on Java 8 and Java 17 for the above.

## How was this patch tested?

added UTs against Spark

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -59,6 +59,9 @@ spark / sparkVersion := getSparkVersion()
 connectCommon / sparkVersion := getSparkVersion()
 connectClient / sparkVersion := getSparkVersion()
 connectServer / sparkVersion := getSparkVersion()
+kernelDefaults / sparkVersion := getSparkVersion()
+storage / sparkVersion := getSparkVersion()
+goldenTables / sparkVersion := getSparkVersion()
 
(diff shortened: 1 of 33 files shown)


foss-contributor:
@richardc-db, thank you for your contribution! This pull request has been inactive for over 18 months and will be closed in one week. To keep it open, please leave a comment on this pull request or contact us in the `#contributors` channel on [Delta Lake Slack](https://go.delta.io/slack). We appreciate your involvement in Delta Lake!

PR #3517, opened 2024-08-09
Infra: Shard spark and spark master tests across 3 actions

amogh-jahagirdar:
This change adds a matrix of 3 shards to github CI for the delta spark/spark master tests and plumbs those through to our parallelization strategy which will then select tests to run if they are assigned to that shard. This brings down CI times from ~4 hours to 2 hours.

3 shards seemsd to be the point of diminishing returns, and we should look at the actual tests and possibly higher instance types if we really want to get those down further.

No sharding: 4 hours
2 shards: 3 hours
3 shards:  ~2 hours
4 shards: ~2 hours

### How to revert

To undo this change without a full revert commit there are a few ways (in order of recommendation)

1. Remove/comment out *both* 
the shard row in the GH matrix https://github.com/delta-io/delta/pull/3517/files#diff-b77047dc65c62d814f7f67eec57c23c4cf9d9796e6a8a018558e0e935739d8d8R10-R11

AND

the NUM_SHARDS env variable
https://github.com/delta-io/delta/pull/3517/files#diff-b77047dc65c62d814f7f67eec57c23c4cf9d9796e6a8a018558e0e935739d8d8R15

2. Comment out this block https://github.com/delta-io/delta/pull/3517/files#diff-00b77d52981e1ba8cc5ac9b80b22d8f7bb9814222121b714269e14e139daaae0R141-R152 in TestParallelization

### DIFF
--- a/.github/workflows/spark_master_test.yaml
+++ b/.github/workflows/spark_master_test.yaml
@@ -7,8 +7,12 @@ jobs:
       matrix:
         # These Scala versions must match those in the build.sbt
         scala: [2.13.13]
+        # Important: This list of shards must be [0..NUM_SHARDS - 1]
+        shard: [0, 1, 2]
     env:
       SCALA_VERSION: ${{ matrix.scala }}
+      # Important: This must be the same as the length of shards in matrix
+      NUM_SHARDS: 3
     steps:
       - uses: actions/checkout@v3
       - uses: technote-space/get-diff-action@v4
@@ -46,7 +50,7 @@ jobs:
       - name: Run Spark Master tests
         # when changing TEST_PARALLELISM_COUNT make sure to also change it in spark_test.yaml
         run: |
-          TEST_PARALLELISM_COUNT=2 build/sbt -DsparkVersion=master "++ ${{ matrix.scala }}" clean spark/test
+          TEST_PARALLELISM_COUNT=2 SHARD_ID=${{matrix.shard}} build/sbt -DsparkVersion=master "++ ${{ matrix.scala }}" clean spark/test
           TEST_PARALLELISM_COUNT=2 build/sbt -DsparkVersion=master "++ ${{ matrix.scala }}" clean connectServer/test
           TEST_PARALLELISM_COUNT=2 build/sbt -DsparkVersion=master "++ ${{ matrix.scala }}" clean connectServer/assembly connectClient/test
         if: steps.git-diff.outputs.diff
--- a/.github/workflows/spark_test.yaml
+++ b/.github/workflows/spark_test.yaml
@@ -7,8 +7,12 @@ jobs:
       matrix:
         # These Scala versions must match those in the build.sbt
         scala: [2.12.18, 2.13.13]
+        # Important: This list of shards must be [0..NUM_SHARDS - 1]
+        shard: [0, 1, 2]
     env:
       SCALA_VERSION: ${{ matrix.scala }}
+      # Important: This must be the same as the length of shards in matrix
+      NUM_SHARDS: 3
     steps:
       - uses: actions/checkout@v3
       - uses: technote-space/get-diff-action@v4
@@ -78,5 +82,5 @@ jobs:
       - name: Run Scala/Java and Python tests
         # when changing TEST_PARALLELISM_COUNT make sure to also change it in spark_master_test.yaml
         run: |
-          TEST_PARALLELISM_COUNT=2 pipenv run python run-tests.py --group spark
+          TEST_PARALLELISM_COUNT=2 pipenv run python run-tests.py --group spark --shard ${{ matrix.shard }}
         if: steps.git-diff.outputs.diff
--- a/build.sbt
+++ b/build.sbt
@@ -1506,7 +1506,8 @@ val createTargetClassesDir = taskKey[Unit]("create target classes dir")
 
 // Don't use these groups for any other projects
 lazy val sparkGroup = project
-  .aggregate(contribs, storage, storageS3DynamoDB, iceberg, testDeltaIcebergJar, sharing, hudi)  .settings(
+  .aggregate(spark, contribs, storage, storageS3DynamoDB, iceberg, testDeltaIcebergJar, sharing, hudi)
+  .settings(
     // crossScalaVersions must be set to Nil on the aggregating project
     crossScalaVersions := Nil,
     publishArtifact := false,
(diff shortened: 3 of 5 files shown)


amogh-jahagirdar:
cc @nastra @scottsand-db @tdas

build.sbt
```diff
@@ -1506,8 +1506,7 @@ val createTargetClassesDir = taskKey[Unit]("create target classes dir")
 
 // Don't use these groups for any other projects
 lazy val sparkGroup = project
-  .aggregate(spark, contribs, storage, storageS3DynamoDB, iceberg, testDeltaIcebergJar, sharing, hudi)
-  .settings(
+  .aggregate(spark, contribs, storage, storageS3DynamoDB, iceberg, testDeltaIcebergJar, sharing, hudi)  .settings(
```
nastra, line 1509: nit: unnecessary change

.github/workflows/spark_master_test.yaml
```diff
@@ -46,7 +48,7 @@ jobs:
       - name: Run Spark Master tests
         # when changing TEST_PARALLELISM_COUNT make sure to also change it in spark_test.yaml
         run: |
-          TEST_PARALLELISM_COUNT=2 build/sbt -DsparkVersion=master "++ ${{ matrix.scala }}" clean spark/test
+          TEST_PARALLELISM_COUNT=2 SHARD_ID=${{matrix.shard}} build/sbt -DsparkVersion=master "++ ${{ matrix.scala }}" clean spark/test
```
scottsand-db, line 51: how does this select just a single shard integer? To me this looks like it would select the entire array

build.sbt
scottsand-db: +1

run-tests.py
```diff
@@ -43,16 +43,24 @@ def get_args():
         default=False,
         action="store_true",
         help="Enables test coverage and generates an aggregate report for all subprojects")
+
```
scottsand-db, line 46: nit: delete white space

project/TestParallelization.scala
```diff
@@ -50,17 +52,24 @@ object TestParallelization {
     Test / forkTestJVMCount := {
       sys.env.get("TEST_PARALLELISM_COUNT").map(_.toInt).getOrElse(4)
     },
+    Test / shardId := {
+      sys.env.get("SHARD_ID").map(_.toInt)
```
scottsand-db, line 56: can you use `numShards` here?

.github/workflows/spark_master_test.yaml
amogh-jahagirdar: The GH matrix option will tell GH actions to trigger a job per matrix combination. https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-a-matrix-for-your-jobs

So in this case for spark master there will be a  job for Scala 2.13.13 for shard 0, another one for 1, for 2 etc.

matrix.{value} always references what is the concrete value for that given job. 
For example, It's how the delta spark tests work for testing the different scala versions (2.12.8 and 2.13.13 in spark_test.yaml). Those scala versions ultimately get passed through to `run-tests.py`

build.sbt
amogh-jahagirdar: Sure will fix this

project/TestParallelization.scala
amogh-jahagirdar: No, that wouldn't be correct, this is supposed to be the shard ID assigned to a given action.
`numShards` is just used for the modulo math to determine if a test is assigned to the given shard ID or not. I should probably comment this, or if there's better names, open to suggestion.

amogh-jahagirdar:
`CloneIcebergSuiteByName` is failing after I rebased, I'm retriggering to see if it's flakey

project/TestParallelization.scala
scottsand-db: Oh, sorry, I just misread this. Ignore this comment. Thanks!

project/TestParallelization.scala
```diff
@@ -107,25 +116,38 @@ object TestParallelization {
...
+  class SimpleHashStrategy private(
+      groups: Map[Int, Tests.Group],
+      shard: Option[Int]) extends GroupingStrategy {
 
     lazy val testGroups = groups.values.toList
     val groupCount = groups.size
 
     override def add(testDefinition: TestDefinition): GroupingStrategy = {
       val groupIdx = math.abs(testDefinition.name.hashCode % groupCount)
+      if (shard.isDefined) {
```
scottsand-db, line 128: add some comments here to explain this please. After a few minutes I still don't 100% understand the logic

scottsand-db, line 128: if we wanted to disable this, say 6 months from now due to some crazy bug, the minimal change we would have to do is just disable this if statement, right? e.g. `if (false)`, right?

amogh-jahagirdar: I updated with a comment, please let me know if it still is unclear, I think that's important to make sure everyone's confident on this sharding logic!

amogh-jahagirdar: Correct, that's def a way to do it in the code itself. From the GH configuration perspective, it's possible to also just comment out the 2 lines for matrix shard ID list and the shard ID env variable.

.github/workflows/spark_test.yaml
```diff
@@ -7,8 +7,10 @@ jobs:
       matrix:
         # These Scala versions must match those in the build.sbt
         scala: [2.12.18, 2.13.13]
+        shard: [0, 1, 2]
     env:
       SCALA_VERSION: ${{ matrix.scala }}
+      NUM_SHARDS: 3
```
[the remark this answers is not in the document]
amogh-jahagirdar: https://github.com/delta-io/delta/pull/3517/files#diff-b77047dc65c62d814f7f67eec57c23c4cf9d9796e6a8a018558e0e935739d8d8R10 

The Github matrix strategies itself. Github CI itself will read the matrix fields, and create 1 job per matrix combination. It's important NUM_SHARDS == length (matrix.shards) I've commented that in the Yaml so anyone who tries to change this will be super aware of the assumptions, and I also commented how the sharding logic works in TestParallelization.

We just need the environment variable NUM_SHARDS so that in TestParallelization we can determine if a given test is assigned to a shard based on % NUM_SHARDS == the assigned shard ID.

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-a-matrix-for-your-jobs#about-matrix-strategies

project/TestParallelization.scala
scottsand-db: @amogh-jahagirdar - please include this "how to disable later" in the PR description. That will make it super findable.

scottsand-db (approved):

PR #3524, opened 2024-08-12
[Infra] Increase test parallelism

nastra:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [x] Other (Infra)

## Description

Increase test JVM parallelization to 4.

## How was this patch tested?

CI tests.

No.

### DIFF
--- a/project/TestParallelization.scala
+++ b/project/TestParallelization.scala
@@ -50,7 +50,8 @@ object TestParallelization {
    */
   lazy val simpleGroupingStrategySettings = Seq(
     Test / forkTestJVMCount := {
-      sys.env.get("TEST_PARALLELISM_COUNT").map(_.toInt).getOrElse(4)
+      sys.env.get("TEST_PARALLELISM_COUNT").map(_.toInt)
+        .getOrElse(java.lang.Runtime.getRuntime.availableProcessors)
     },
     Test / shardId := {
       sys.env.get("SHARD_ID").map(_.toInt)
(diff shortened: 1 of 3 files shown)


project/TestParallelization.scala
```diff
@@ -48,7 +48,8 @@ object TestParallelization {
    */
   lazy val simpleGroupingStrategySettings = Seq(
     Test / forkTestJVMCount := {
-      sys.env.get("TEST_PARALLELISM_COUNT").map(_.toInt).getOrElse(4)
+      sys.env.get("TEST_PARALLELISM_COUNT").map(_.toInt)
```
tdas, line 51: we dont need to change this for speeding up github tests.. isnt it?

tdas: if not, lets separate out user-facing defaults from github tests.
each carry a different risk. increasing parallelism too much can lead to flakiness, and so using all available processors may be problematic.

tdas:
how much benefit are you getting?

nastra:
> how much benefit are you getting?

spark tests with Scala 2.13 are down to 3h and the same tests with Scala 2.12 are down to 3:30h

project/TestParallelization.scala
nastra: makes sense, I've reverted this for now

tdas:
With the sharing bringing it down already to 2 hours, how much additional
benefits does this pr provide?

scottsand-db:
Before this PR:
- delta-spark against Spark 3.5.2 -- Scala 2.12 :: 94, 124, 117 -- AVG: 111.67
- delta-spark against Spark 3.5.2 -- Scala 2.13 :: 72, 108, 93 -- AVG 91.00

With this PR:
- delta-spark against Spark 3.5.2 -- Scala 2.12 :: 84, 115, 95 -- AVG: 98.00
- delta-spark against Spark 3.5.2 -- Scala 2.13 ::  65, 93, 78 -- AVG: 78.67

PR #3526, opened 2024-08-12
[SPARK][TEST-ONLY] Add more tests for Identity Column

zhipengmao-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
This PR is part of https://github.com/delta-io/delta/issues/1959 .
It add more tests for Identity Column to test
- logging identity column properties and stats
- reading table should not see identity column properties
- compatibility with table of older protocols
- identity value generation starting at range boundaries of long data type

## How was this patch tested?
Test only change.

No.


spark/src/test/scala/org/apache/spark/sql/delta/IdentityColumnSuite.scala
```diff
@@ -124,6 +130,251 @@ trait IdentityColumnSuiteBase extends IdentityColumnTestUtils {
...
+
+      val commands: Map[Int, () => Dataset[_]] = Map(
+        1 -> (() => spark.table(tblName)),
+        2 -> (() => sql(s"SELECT * FROM $tblName")),
+        3 -> (() => sql(s"SELECT * FROM delta.`$path`")),
+        4 -> (() => spark.read.format("delta").load(path)),
+        5 -> (() => spark.read.format("delta").table(tblName)),
+        6 -> (() => spark.readStream.format("delta").load(path)),
+        7 -> (() => spark.readStream.format("delta").table(tblName))
+      )
```
c27kwan, line 215: We either don't need the indices here or we just make it the parameter (Int) => Dataset[]. They're just used for "test number". Let's simplify this.

spark/src/test/scala/org/apache/spark/sql/delta/IdentityColumnSuite.scala
```diff
@@ -124,6 +130,251 @@ trait IdentityColumnSuiteBase extends IdentityColumnTestUtils {
...
+      step <- steps
+    } {
+      withTable(tblName) {
+        createTableWithIdColAndIntValueCol(
+          tblName, generatedAsIdentityType, Some(start), Some(step))
+        val table = DeltaLog.forTable(spark, TableIdentifier(tblName))
+        val actualSchema =
+          DeltaColumnMapping.dropColumnMappingMetadata(table.snapshot.metadata.schema)
+        assert(actualSchema === expectedSchema(generatedAsIdentityType, start, step))
+        if ((start < 0L) == (step < 0L)) {
```
c27kwan, line 364: This could use a quick comment about Long underflow and overflow. It was not obvious, especially because the exception message isn't checked.

c27kwan (approved):
New tests lgtm, thanks for adding them! Don't forget to reference the parent github issue which tracks all Identity Column PRs.

c27kwan:
The PR description could be more descriptive. The list of tests is pretty short so it's not verbose to do it in this case :)

scottsand-db:
Failed test https://github.com/delta-io/delta/actions/runs/10351172194/job/28649104059?pr=3526 PTAL. Flaky?

felipepessoto:
@zhipengmao-db do we have any tests for Identity + MERGE? I'm not sure if the implementation is already done though

zhipengmao-db:
> Failed test https://github.com/delta-io/delta/actions/runs/10351172194/job/28649104059?pr=3526 PTAL. Flaky?

The failed test is `org.apache.spark.sql.delta.CloneIcebergByNameSuite`, which should not be related to this change as the change is test only.

zhipengmao-db:
> @zhipengmao-db do we have any tests for Identity + MERGE? I'm not sure if the implementation is already done though

MERGE support has not been done yet but it is on the plan

PR #3536, opened 2024-08-13
always include CatalogTable#storage#properties into data source options

cloud-fan:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

`DeltaTableV2` uses the options to create `DeltaLog`, assuming that the options should always be consistent with the one in `CatalogTable#storage#properties`. This may not be true as custom catalogs may add more table storage properties on the fly, like Unity Catalog.

## How was this patch tested?

Tested locally with UC. It's not an issue for HMS.

no

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaTableV2.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaTableV2.scala
@@ -91,7 +91,21 @@ case class DeltaTableV2(
   // in cases where we will fallback to the V1 behavior.
   lazy val deltaLog: DeltaLog = {
     DeltaTableV2.withEnrichedUnsupportedTableException(catalogTable, tableIdentifier) {
-      DeltaLog.forTable(spark, rootPath, options)
+      // Ideally the table storage properties should always be the same as the options load from
+      // the Delta log, as Delta CREATE TABLE command guarantees it. However, custom catalogs such
+      // as Unity Catalog may add more table storage properties on the fly. We should respect it
+      // and merge the table storage properties and Delta options.
+      val dataSourceOptions = if (catalogTable.isDefined) {
+        // To be safe, here we only extra file system options from table storage properties and
+        // the original `options` has higher priority than the table storage properties.
+        val fileSystemOptions = catalogTable.get.storage.properties.filter { case (k, _) =>
+          DeltaTableUtils.validDeltaTableHadoopPrefixes.exists(k.startsWith)
+        }
+        fileSystemOptions ++ options
+      } else {
+        options
+      }
+      DeltaLog.forTable(spark, rootPath, dataSourceOptions)
     }
   }
 
--- a/spark/src/test/scala/org/apache/spark/sql/delta/CustomCatalogSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/CustomCatalogSuite.scala
@@ -32,10 +32,11 @@ import org.apache.spark.sql.catalyst.analysis.ResolvedTable
 import org.apache.spark.sql.catalyst.catalog.{CatalogStorageFormat, CatalogTable, CatalogTableType}
 import org.apache.spark.sql.catalyst.plans.logical.{AppendData, LogicalPlan, SetTableProperties, UnaryNode, UnsetTableProperties}
 import org.apache.spark.sql.catalyst.trees.UnaryLike
-import org.apache.spark.sql.connector.catalog.{Identifier, Table, TableCatalog, TableChange}
+import org.apache.spark.sql.connector.catalog.{DelegatingCatalogExtension, Identifier, InMemoryTable, InMemoryTableCatalog, Table, TableCatalog, TableChange, V1Table}
 import org.apache.spark.sql.connector.expressions.Transform
+import org.apache.spark.sql.delta.catalog.DeltaCatalog
 import org.apache.spark.sql.execution.command.LeafRunnableCommand
-import org.apache.spark.sql.execution.datasources.v2.DataSourceV2Relation
+import org.apache.spark.sql.execution.datasources.v2.{DataSourceV2Relation, V2SessionCatalog}
 import org.apache.spark.sql.test.SharedSparkSession
 import org.apache.spark.sql.types.StructType
 import org.apache.spark.sql.util.CaseInsensitiveStringMap
@@ -281,6 +282,19 @@ class CustomCatalogSuite extends QueryTest with SharedSparkSession
     }
   }
 
+  test("custom catalog that adds additional table storage properties") {
+    withSQLConf("spark.sql.catalog.spark_catalog" -> classOf[DummySessionCatalog].getName) {
+      withTable("t") {
+        withTempPath { path =>
+          spark.range(10).write.format("delta").save(path.getCanonicalPath)
+          sql(s"CREATE TABLE t (id LONG) USING delta LOCATION '${path.getCanonicalPath}'")
+          val t = spark.sessionState.catalogManager.v2SessionCatalog.asInstanceOf[TableCatalog]
+            .loadTable(Identifier.of(Array("default"), "t")).asInstanceOf[DeltaTableV2]
+          assert(t.deltaLog.options("fs.myKey") == "val")
+        }
+      }
+    }
+  }
 }
 
 class DummyCatalog extends TableCatalog {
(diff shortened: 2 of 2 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaTableV2.scala
```diff
@@ -91,7 +91,16 @@ case class DeltaTableV2(
   // in cases where we will fallback to the V1 behavior.
   lazy val deltaLog: DeltaLog = {
     DeltaTableV2.withEnrichedUnsupportedTableException(catalogTable, tableIdentifier) {
-      DeltaLog.forTable(spark, rootPath, options)
+      // Ideally the table storage properties should always be the same as the options load from
+      // the Delta log, as Delta CREATE TABLE command guarantees it. However, custom catalogs such
+      // as Unity Catalog may add more table storage properties on the fly. We should respect it
+      // and merge the table storage properties and Delta options.
+      val dataSourceOptions = if (catalogTable.isDefined) {
+        DataSourceUtils.generateDatasourceOptions(caseInsensitiveOptions, catalogTable.get)
```
prakharjain09, line 99: From where is `options` variable passed to DeltaTableV2? Is that from: `spark.read.options(k1, v1).load(<path>)` ?

If the above is true, should we be respecting options even in case when `catalogTable` is defined?

i.e. 

```
        options ++ DataSourceUtils.generateDatasourceOptions(caseInsensitiveOptions, catalogTable.get)
```

cloud-fan: This is `DeltaTableV2`, so the `options` was loaded from the delta log.

prakharjain09: No - DeltaTableV2 options don't come from DeltaLog. They can come from `.options()` API as well. This change introduced these options: https://github.com/delta-io/delta/commit/4afd4156680ef37c9e8509cebab10ad43a0755f4

cloud-fan: The case here is `catalogTable` is present. This means the table exists and the `options` should be loaded from the Delta log?

prakharjain09: > The case here is catalogTable is present.

Yes I understand that. DeltaLog directory does not store any options.

A user can pass options explcitly as part of command. e.g.

    spark.read.option(k1, v1).load("\<path\>") -- definitely works and options are tunneled to DeltaLog object properly

    spark.read.option(k1, v1).table("c1.s1.t1") -- most like options are tunnel properly today

scovich: > should we be respecting `options` even in case when `catalogTable` is defined?

Isn't `caseInsensitiveOptions` just a different representation of `options`?
```scala
private lazy val caseInsensitiveOptions = new CaseInsensitiveStringMap(options.asJava)  
```
(so I think it's included already by the current code)

Meanwhile tho:
> `spark.read.option(k1, v1).table("c1.s1.t1")` -- most like `options` are tunnel properly today

IIRC, options passed by `option` are _not_ propagated through `table`. At least, not all of them. It prevents option-based time travel on a catalog table, for example.

prakharjain09: > Isn't caseInsensitiveOptions just a different representation of options?

Oh right. I missed this part. So we are already respective passed `options`.

prakharjain09 (approved):

spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaTableV2.scala
tdas: is this thread resolved?

cloud-fan:
To avoid risks, I've made this PR more surgical: only extract file system options from the table storage properties and only apply them when they are not defined in the original `options`. This is the smallest change we need to make for UC.

spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaTableV2.scala
```diff
@@ -91,7 +91,21 @@ case class DeltaTableV2(
...
+        // To be safe, here we only extra file system options from table storage properties and
+        // the original `options` has higher priority than the table storage properties.
+        val fileSystemOptions = catalogTable.get.storage.properties.filter { case (k, _) =>
+          DeltaTableUtils.validDeltaTableHadoopPrefixes.exists(k.startsWith)
+        }
+        fileSystemOptions ++ options
+      } else {
+        options
+      }
+      DeltaLog.forTable(spark, rootPath, dataSourceOptions)
```
tdas, line 108: no unit test?

spark/src/test/scala/org/apache/spark/sql/delta/CustomCatalogSuite.scala
```diff
@@ -366,3 +382,52 @@ class DummyCatalog extends TableCatalog {
     )
   }
 }
+
+class DummySessionCatalogInner extends DelegatingCatalogExtension {
```
tdas, line 386: can you add class docs to explain the purpose of these 2 catalog classes?

spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaTableV2.scala
```diff
@@ -91,7 +91,21 @@ case class DeltaTableV2(
   // in cases where we will fallback to the V1 behavior.
   lazy val deltaLog: DeltaLog = {
     DeltaTableV2.withEnrichedUnsupportedTableException(catalogTable, tableIdentifier) {
-      DeltaLog.forTable(spark, rootPath, options)
+      // Ideally the table storage properties should always be the same as the options load from
+      // the Delta log, as Delta CREATE TABLE command guarantees it. However, custom catalogs such
+      // as Unity Catalog may add more table storage properties on the fly. We should respect it
+      // and merge the table storage properties and Delta options.
+      val dataSourceOptions = if (catalogTable.isDefined) {
+        // To be safe, here we only extra file system options from table storage properties and
```
scovich, line 99:
```suggestion
        // To be safe, here we only consider file system options from table storage properties and
```

spark/src/test/scala/org/apache/spark/sql/delta/CustomCatalogSuite.scala
```diff
@@ -366,3 +382,56 @@ class DummyCatalog extends TableCatalog {
...
 }
+
+// A dummy catalog that adds additional table storage properties after the table is loaded.
+// It's only used inside `DummySessionCatalog`.
+class DummySessionCatalogInner extends DelegatingCatalogExtension {
+  override def loadTable(ident: Identifier): Table = {
+    val t = super.loadTable(ident).asInstanceOf[V1Table]
+    V1Table(t.v1Table.copy(
+      storage = t.v1Table.storage.copy(
+        properties = t.v1Table.storage.properties ++ Map("fs.myKey" -> "val")
```
scovich, line 393: nit: Does this work?
```suggestion
        properties = t.v1Table.storage.properties.updated("fs.myKey" -> "val")
```

scovich (approved):
> To avoid risks, I've made this PR more surgical: only extract file system options from the table storage properties and only apply them when they are not defined in the original `options`. This is the smallest change we need to make for UC.

That seems sensible? For example, given a conflict between`options` and a spark conf, I believe the former prevails?

spark/src/test/scala/org/apache/spark/sql/delta/CustomCatalogSuite.scala
cloud-fan: not sure, we are still on scala 2.12 and I feel it's safer to use common APIs.

cloud-fan:
> For example, given a conflict between options and a spark conf, I believe the former prevails?

Yes, this is to maximize the backward compatibility.

spark/src/test/scala/org/apache/spark/sql/delta/CustomCatalogSuite.scala
scovich: [Map.updated](https://www.scala-lang.org/api/2.12.7/scala/collection/Map.html#updated(key:K,value:V):scala.collection.Map[K,V]) is a scala-2.12 API..

PR #3540, opened 2024-08-13
[Spark] Add RowTrackingBackfillConflictsSuite

longvu-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Add extensive testing that checks various scenario of conflict scenarios between Backfill and other concurrent transactions.

Note: We have not yet covered the case of conflicts between Row Tracking Backfill and Clone, and also the case of multiple Row Tracking Backfills, which will be added in follow-up PRs.

## How was this patch tested?
Added UTs.

No.


tomvanbussel (approved):
Nice! These tests are extremely detailed! LGTM

PR #3558, opened 2024-08-15
[SPARK] Allow non-deterministic expressions in actions of merge

zhipengmao-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

It makes `MergeIntoCommandBase` extend a trait `SupportsNonDeterministicExpression` in Spark that logical plans can extend to check whether it can allow non-deterministic expressions and pass the CheckAnalysis rule.

`MergeIntoCommandBase` extends `SupportsNonDeterministicExpression` to check whether all the conditions in the Merge command are deterministic.

This is harmless and allows more flexible usage of merge. For example, we use a non-deterministic UDF to generate identity values for identity columns, so it is required to allow non-deterministic expressions in updated/inserted column values of merge statements in order to support merge on target tables with identity columns. So this PR is part of https://github.com/delta-io/delta/issues/1959.

## How was this patch tested?
New test cases.

Yes.
We are changing the behavior to allow non-deterministic expressions in updated/inserted column values of merge statements. We still don't allow non-deterministic expressions in conditions of merge statements.

e.g. 
We currently don't allow the merge statement to add a random noise to the value that is inserted in merge

```
MERGE INTO target USING source
ON target.key = source.key
WHEN MATCHED THEN UPDATE SET target.value = source.value + rand()
```

Now we are allowing this as this may be helpful in terms of data privacy to not disclose the actual data while preserving the data properties e.g. mean values etc.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/commands/MergeIntoCommandBase.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/commands/MergeIntoCommandBase.scala
@@ -45,7 +45,8 @@ trait MergeIntoCommandBase extends LeafRunnableCommand
   with PredicateHelper
   with ImplicitMetadataOperation
   with MergeIntoMaterializeSource
-  with UpdateExpressionsSupport {
+  with UpdateExpressionsSupport
+  with SupportsNonDeterministicExpression {
 
   @transient val source: LogicalPlan
   @transient val target: LogicalPlan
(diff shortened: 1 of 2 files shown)


c27kwan (approved):
Overall change LGTM. The PR Description needs more context as it is a significant change. 
- It is a user facing change. 
- we should mention the trait name `SupportsNonDeterministicExpression` instead of generic trait. 
- one of the motivations for this change is Identity Column support in MERGE -- we use a non-deterministic UDF to generate identity values. I would link the parent Github issue.

zhipengmao-db:
Compilation error in Spark Master test seems not related:
```
[error] /home/runner/work/delta/delta/spark/src/main/scala/org/apache/spark/sql/delta/DeltaUDF.scala:84:62: value copy is not a member of org.apache.spark.sql.Encoder[_$1]
[error]       val inputEncoders = template.inputEncoders.map(_.map(_.copy()))
[error]                                                              ^
[error] /home/runner/work/delta/delta/spark/src/main/scala/org/apache/spark/sql/delta/DeltaUDF.scala:85:56: value copy is not a member of org.apache.spark.sql.Encoder[_$2]
[error]       val outputEncoder = template.outputEncoder.map(_.copy())
[error]        
```

PR #3566, opened 2024-08-16
[SPARK] Add MERGE support for tables with Identity Columns

zhipengmao-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)


spark/src/test/scala/org/apache/spark/sql/delta/IdentityColumnDMLSuite.scala
```diff
@@ -0,0 +1,591 @@
...
+        .as[IdentityColumnTestTableRow]
+        .collect()
+
+      assert(result.length === 8)
+      checkGeneratedIdentityValues(
+        sortedRows = result,
+        start = 0,
+        step = 1,
+        expectedLowerBound = 0,
+        expectedUpperBound =
```
c27kwan, line 527: Let's use a hard coded value for expectedUpperBound. When we dynamically compute expected test result, we risk having a testing error where the test is erroneously passing due to the dynamic computation also being wrong.  

We just want it to be in a "reasonable" range. E.g. 0 - 20 or 0 - 100.

spark/src/test/scala/org/apache/spark/sql/delta/IdentityColumnDMLSuite.scala
```diff
@@ -0,0 +1,591 @@
...
+      withTempView(src) {
+        (1 :: 20 :: 30 :: Nil).toDF("value").createOrReplaceTempView(src)
+        sql(
+          s"""MERGE INTO $tgt target
+             |USING $src src ON target.value = src.value
+             |WHEN MATCHED THEN UPDATE SET target.value = src.value
+             |WHEN NOT MATCHED THEN INSERT (target.value) VALUES (src.value)""".stripMargin)
+      }
+
+      val result = sql(
```
c27kwan, line 514: rename so that it's more obvious it's sorted. the later parts don't make sense otherwise.

spark/src/test/scala/org/apache/spark/sql/delta/IdentityColumnDMLSuite.scala
```diff
@@ -0,0 +1,591 @@
...
+
+      checkAnswer(sql(
+        s"""SELECT id, value, ${CDCReader.CDC_TYPE_COLUMN_NAME}
+           |FROM table_changes('$tgt', 8)
+           |ORDER BY id, value, _change_type""".stripMargin),
+        Seq(
+          Row(1, 1, CDCReader.CDC_TYPE_UPDATE_POSTIMAGE),
+          Row(1, 1, CDCReader.CDC_TYPE_UPDATE_PREIMAGE),
+          Row(result(6).id, 20, CDCReader.CDC_TYPE_INSERT),
+          Row(result(7).id, 30, CDCReader.CDC_TYPE_INSERT)))
```
c27kwan, line 539: This is somewhat correct, but not very obvious what it's checking. 

Is it guaranteed that the identity value generated for `20` will be less than the one that is generated for `30`? That seems like an implementation detail that's subject to flakiness. We should be indexing based on what the generated id value was for `20` and `30` instead of assuming.

I suggest making a small helper `getIdForValue(value: long)` which is basically `result.filter(_.value = value).id`. I'd also add a small comment before the checkanswer here that described that it's validating  the ids in the CDCReader matches those in the logical data.

c27kwan:
Can you also make the PR description more detailed to give context to the necessary changes?

spark/src/main/scala/org/apache/spark/sql/delta/PreprocessTableMerge.scala
```diff
@@ -395,7 +409,9 @@ case class PreprocessTableMerge(override val conf: SQLConf)
     allActions: Seq[DeltaMergeAction],
     sourcePlan: LogicalPlan,
     columnWithDefaultExpr: Seq[(StructField, Boolean)],
-    postEvolutionTargetSchema: StructType): (Seq[DeltaMergeAction], Set[String]) = {
+    postEvolutionTargetSchema: StructType,
+    identityColumnExpressionMap: mutable.Map[StructField, Expression])
+      : (Seq[DeltaMergeAction], Set[String]) = {
```
c27kwan, line 414: The scalastyle is off here

spark/src/test/scala/org/apache/spark/sql/delta/IdentityColumnDMLSuite.scala
```diff
@@ -0,0 +1,598 @@
...
+      // Validate the ids in CDCReader match those in logical data.
+      checkAnswer(sql(
+        s"""SELECT id, value, ${CDCReader.CDC_TYPE_COLUMN_NAME}
+           |FROM table_changes('$tgt', 8)
+           |ORDER BY id, value, _change_type""".stripMargin),
+        Seq(
+          Row(1, 1, CDCReader.CDC_TYPE_UPDATE_POSTIMAGE),
+          Row(1, 1, CDCReader.CDC_TYPE_UPDATE_PREIMAGE),
+          Row(getIdForValue(20), 20, CDCReader.CDC_TYPE_INSERT),
+          Row(getIdForValue(30), 30, CDCReader.CDC_TYPE_INSERT)))
```
c27kwan, line 546: named parameters for getIdForValue

spark/src/test/scala/org/apache/spark/sql/delta/IdentityColumnDMLSuite.scala
```diff
@@ -0,0 +1,598 @@
...
+      checkGeneratedIdentityValues(
+        sortedRows = sortedResult,
+        start = 0,
+        step = 1,
+        expectedLowerBound = 0,
+        expectedUpperBound = 20,
+        expectedDistinctCount = 8)
+
+      def getIdForValue(value: Int): Long = {
+        sortedResult.find(_.value == value.toString).get.id
```
c27kwan, line 534: nit: it's better to split this into multiple lines and assert before calling the `.get` so that we never get a None.get error or at the very least it'll be easier to debug. `assert(blah, "expected value $value to be written to the table")` and assert that there is _exactly one_ row with this value.

spark/src/main/scala/org/apache/spark/sql/delta/PreprocessTableMerge.scala
```diff
@@ -127,6 +125,13 @@ case class PreprocessTableMerge(override val conf: SQLConf)
       case m: DeltaMergeIntoNotMatchedBySourceDeleteClause => m // Delete does not need reordering
     }
 
+    // A mapping from the identity column struct field to the GenerateIdentityColumnValues
+    // expression for the target table in the MERGE clause.
+    val identityColumnExpressionMap = mutable.Map[StructField, Expression]()
+
```
c27kwan, line 131: nit: i would delete this newline. These two variables kind of go together.

spark/src/main/scala/org/apache/spark/sql/delta/PreprocessTableMerge.scala
```diff
@@ -98,7 +95,8 @@ case class PreprocessTableMerge(override val conf: SQLConf)
       throw DeltaErrors.operationOnTempViewWithGenerateColsNotSupported("MERGE INTO")
     }
     // Additional columns with default expressions.
-    var additionalColumns = Seq[StructField]()
+    val additionalColumns = IdentityColumn.getIdentityColumns(
+      tahoeFileIndex.snapshotAtAnalysis.metadata.schema)
```
c27kwan, line 99: Can you rename `additionalColumns` to `identityColumns`? It's only ever passed in the code where we need to pass an indentity column and there are no other usage AFAICT.

spark/src/main/scala/org/apache/spark/sql/delta/PreprocessTableMerge.scala
```diff
@@ -119,14 +117,20 @@ case class PreprocessTableMerge(override val conf: SQLConf)
...
         m.copy(m.condition, alignedActions)
       case m: DeltaMergeIntoNotMatchedBySourceDeleteClause => m // Delete does not need reordering
     }
 
+    // A mapping from the identity column struct field to the GenerateIdentityColumnValues
+    // expression for the target table in the MERGE clause.
+    val identityColumnExpressionMap = mutable.Map[StructField, Expression]()
+    // Column names for which we need to track IDENTITY high water marks.
+    var trackHighWaterMarks = Set[String]()
+
```
c27kwan, line 133: I'd either move these up to where `identityColumns` is defined to group them together OR I'd move these down to where they are first used. This is defined in an awkward place right now.

c27kwan (approved):
Lgtm! :D

spark/src/main/scala/org/apache/spark/sql/delta/PreprocessTableMerge.scala
zhipengmao-db: Hmm they were defined here because they need to be outside the `notMatched.map {...}` block. Have moved them up.

PR #3573, opened 2024-08-17
[Delta] Block alter command from overriding or unsetting coordinated commits properties

yumingxuanguo-db:
1. Blocks ALTER TABLE command from setting Coordinated Commits properties if the table already had them. If the table did not have them, checks that the CC property overrides contains exactly Coordinator Name and Coordinator Conf, no Table Conf.
2. Blocks ALTER TABLE command from unsetting Coordinated Commits properties if the table already has them.
3. Added UTs for both cases. Moved the UT to a new suite `CoordinatedCommitsUtilsSuite.scala`.
4. Renamed some methods for clarity.

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## How was this patch tested?

UTs.

No.

### DIFF
--- a/spark/src/main/resources/error/delta-error-classes.json
+++ b/spark/src/main/resources/error/delta-error-classes.json
@@ -279,6 +279,12 @@
     ],
     "sqlState" : "42613"
   },
+  "DELTA_CANNOT_UNSET_COORDINATED_COMMITS_CONFS" : {
+    "message" : [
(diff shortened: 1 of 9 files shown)


prakharjain09 (approved):

prakharjain09 (commented):
Some UTs have been deleted. Maybe an issue in adding new file to git commit.

prakharjain09 (approved):

PR #3576, opened 2024-08-18
[Spark] Remove waiting for a fixed time for Delta Connect Server to be available in Delta Connect testing

longvu-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
For local E2E Delta Connect testing, we also designed an [util class](https://github.com/delta-io/delta/blob/01bf60743b77c47147843e9083129320490f1629/spark-connect/client/src/test/scala-spark-master/io/delta/connect/tables/RemoteSparkSession.scala#L62) to start a local server in a different process similar to [SparkConnect](https://github.com/apache/spark/blob/ba208b9ca99990fa329c36b28d0aa2a5f4d0a77e/connector/connect/client/jvm/src/test/scala/org/apache/spark/sql/test/RemoteSparkSession.scala#L37).

We noticed that the server takes a random amount of seconds to start up, and back then we received the error `INVALID_HANDLE.SESSION_NOT_FOUND] The handle 746e6c86-9fa9-4b08-9572-388c20eaed47 is invalid. Session not found. SQLSTATE: HY000"`, so what we did is to add a 10s `Thread.sleep` before starting the client.

This is not robust, so we are removing the `Thread.sleep`. This should work because:
1. The SparkSession's builder here already uses the default [Configuration](https://github.com/apache/spark/blob/3edc9c23a723a92c5a951cea0436529de65c640a/connector/connect/client/jvm/src/main/scala/org/apache/spark/sql/SparkSession.scala#L891) of the `SparkConnectClient` which includes a default retry policy.
2. Spark patches the error `INVALID_HANDLE.SESSION_NOT_FOUND` in this [PR](https://github.com/apache/spark/pull/46971) at some point, so we should be able to retry even if encountering this error.

## How was this patch tested?
Existing UTs.

Ran `build/sbt` tests locally at least 20 times.

No.

### DIFF
--- a/spark-connect/client/src/test/scala-spark-master/io/delta/connect/tables/DeltaTableSuite.scala
+++ b/spark-connect/client/src/test/scala-spark-master/io/delta/connect/tables/DeltaTableSuite.scala
@@ -16,9 +16,10 @@
 
 package io.delta.tables
 
-import org.apache.spark.sql.delta.tables.RemoteSparkSession
 import java.io.File
 import java.text.SimpleDateFormat
+
+import org.apache.spark.sql.delta.tables.RemoteSparkSession
 
 import org.apache.spark.sql.Row
 import org.apache.spark.sql.test.DeltaQueryTest
--- a/spark-connect/client/src/test/scala-spark-master/io/delta/connect/tables/DeltaTableSuite.scala
+++ b/spark-connect/client/src/test/scala-spark-master/io/delta/connect/tables/DeltaTableSuite.scala
@@ -16,6 +16,7 @@
 
 package io.delta.tables
 
+import org.apache.spark.sql.delta.tables.RemoteSparkSession
 import java.io.File
 import java.text.SimpleDateFormat
 
--- a/spark-connect/client/src/test/scala-spark-master/io/delta/connect/tables/DeltaTableSuite.scala
+++ b/spark-connect/client/src/test/scala-spark-master/io/delta/connect/tables/DeltaTableSuite.scala
@@ -19,8 +19,6 @@ package io.delta.tables
 import java.io.File
 import java.text.SimpleDateFormat
 
-import org.apache.spark.sql.delta.tables.RemoteSparkSession
-
 import org.apache.spark.sql.Row
 import org.apache.spark.sql.test.DeltaQueryTest
 
--- a/spark-connect/client/src/test/scala-spark-master/io/delta/connect/tables/RemoteSparkSession.scala
+++ b/spark-connect/client/src/test/scala-spark-master/io/delta/connect/tables/RemoteSparkSession.scala
@@ -36,12 +36,11 @@
  * limitations under the License.
  */
 
-package org.apache.spark.sql.delta.tables
+package io.delta.tables
 
 import org.scalatest.{BeforeAndAfterAll, Suite}
 
 import org.apache.spark.sql.SparkSession
-import org.apache.spark.sql.connect.client.{RetryPolicy, SparkConnectClient}
 
 /**
  * An util class to start a local Delta Connect server in a different process for local E2E tests.
@@ -98,14 +97,7 @@ trait RemoteSparkSession extends BeforeAndAfterAll { self: Suite =>
   override def beforeAll(): Unit = {
     super.beforeAll()
     server
-
-    val sparkConnectClient = SparkConnectClient
-      .builder()
-      .userId("test")
-      .port(serverPort)
-      .retryPolicy(RetryPolicy.defaultPolicy())
-
-    spark = SparkSession.builder().client(sparkConnectClient.build()).create()
+    spark = SparkSession.builder().remote(s"sc://localhost:$serverPort").create()
   }
 
   override def afterAll(): Unit = {
--- a/spark-connect/client/src/test/scala-spark-master/io/delta/connect/tables/RemoteSparkSession.scala
+++ b/spark-connect/client/src/test/scala-spark-master/io/delta/connect/tables/RemoteSparkSession.scala
@@ -36,11 +36,13 @@
  * limitations under the License.
  */
 
-package io.delta.tables
+package org.apache.spark.sql.delta.tables
 
 import org.scalatest.{BeforeAndAfterAll, Suite}
 
 import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.connect.client.SparkConnectClient
+import org.apache.spark.sql.connect.client.RetryPolicy
 
 /**
  * An util class to start a local Delta Connect server in a different process for local E2E tests.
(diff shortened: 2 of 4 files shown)


spark-connect/client/src/test/scala-spark-master/io/delta/connect/tables/RemoteSparkSession.scala
```diff
@@ -97,10 +98,14 @@ trait RemoteSparkSession extends BeforeAndAfterAll { self: Suite =>
...
-    Thread.sleep(10000)
-    spark = SparkSession.builder().remote(s"sc://localhost:$serverPort").build()
+
+    val sparkConnectClient = SparkConnectClient
+      .builder()
+      .userId("test")
+      .port(serverPort)
+      .retryPolicy(RetryPolicy.defaultPolicy())
+
+    spark = SparkSession.builder().client(sparkConnectClient.build()).create()
```
longvu-db, line 108: `build` is deprecated - https://github.com/apache/spark/blob/3edc9c23a723a92c5a951cea0436529de65c640a/connector/connect/client/jvm/src/main/scala/org/apache/spark/sql/SparkSession.scala#L1015

spark-connect/client/src/test/scala-spark-master/io/delta/connect/tables/DeltaTableSuite.scala
```diff
@@ -19,6 +19,8 @@ package io.delta.tables
 import java.io.File
 import java.text.SimpleDateFormat
 
+import org.apache.spark.sql.delta.tables.RemoteSparkSession
```
tomvanbussel, line 22: Unrelated change?

spark-connect/client/src/test/scala-spark-master/io/delta/connect/tables/RemoteSparkSession.scala
```diff
@@ -97,10 +98,14 @@ trait RemoteSparkSession extends BeforeAndAfterAll { self: Suite =>
...
-    // TODO: Instead of sleeping for a fixed time, which is a bit brittle,
-    // we should repeatedly check when the server is ready.
-    Thread.sleep(10000)
-    spark = SparkSession.builder().remote(s"sc://localhost:$serverPort").build()
+
+    val sparkConnectClient = SparkConnectClient
+      .builder()
+      .userId("test")
+      .port(serverPort)
+      .retryPolicy(RetryPolicy.defaultPolicy())
```
tomvanbussel, line 106: I don't think this is necessary. This is already the default.

spark-connect/client/src/test/scala-spark-master/io/delta/connect/tables/RemoteSparkSession.scala
tomvanbussel, line 108: Why can't we just replace `.build()` with `.create()`? Why do we need to provide our own `SparkConnectClient`?

tomvanbussel (commented):
Not sure why these changes are necessary. The retry policy is already the default. Wouldn't just removing the sleep be enough?

spark-connect/client/src/test/scala-spark-master/io/delta/connect/tables/RemoteSparkSession.scala
longvu-db: @tomvanbussel I put it here because I feel like the fact that is we are using the `defaultPolicy` might not be obvious and the viewers might not know, should I put a comment instead?

spark-connect/client/src/test/scala-spark-master/io/delta/connect/tables/DeltaTableSuite.scala
longvu-db: We changed the package of `RemoteSparkSession`from `io.delta.tables` to ` org.apache.spark.sql.delta.tables` because `SparkConnectClient` is `private[sql]`.

https://github.com/apache/spark/blob/3edc9c23a723a92c5a951cea0436529de65c640a/sql/connect/common/src/main/scala/org/apache/spark/sql/connect/client/SparkConnectClient.scala#L40

So now `DeltaTableSuite` doesn't have the same package name as `RemoteSparkSession`, we have to include explicitly.

spark-connect/client/src/test/scala-spark-master/io/delta/connect/tables/RemoteSparkSession.scala
longvu-db: @tomvanbussel 

https://github.com/apache/spark/blob/3edc9c23a723a92c5a951cea0436529de65c640a/connector/connect/client/jvm/src/main/scala/org/apache/spark/sql/SparkSession.scala#L1028

If we only do like this,

`SparkSession.builder().remote(s"sc://localhost:$serverPort").create()`

The only config it takes into account is `remote`

We need to include `client` to have `RetryPolicy` from `SparkConnectClient`

https://github.com/apache/spark/blob/3edc9c23a723a92c5a951cea0436529de65c640a/sql/connect/common/src/main/scala/org/apache/spark/sql/connect/client/SparkConnectClient.scala#L722

`SparkSession`'s builder doesn't have `RetryPolicy`
https://github.com/apache/spark/blob/3edc9c23a723a92c5a951cea0436529de65c640a/connector/connect/client/jvm/src/main/scala/org/apache/spark/sql/SparkSession.scala#L887

Similar to here
https://github.com/apache/spark/blob/3edc9c23a723a92c5a951cea0436529de65c640a/connector/connect/client/jvm/src/test/scala/org/apache/spark/sql/test/RemoteSparkSession.scala#L181

spark-connect/client/src/test/scala-spark-master/io/delta/connect/tables/RemoteSparkSession.scala
tomvanbussel: Unfortunately it requires accessing a package private interface. I would rather use the public interfaces, even if that means not being explicit about using the default retry policy.

spark-connect/client/src/test/scala-spark-master/io/delta/connect/tables/RemoteSparkSession.scala
tomvanbussel: `remote` still uses the default retry policy, right?

longvu-db: @tomvanbussel You are right, by default the `SparkSession`'s `builder` here already uses the default `Configuration` of the `SparkConnectClient` (https://github.com/apache/spark/blob/3edc9c23a723a92c5a951cea0436529de65c640a/connector/connect/client/jvm/src/main/scala/org/apache/spark/sql/SparkSession.scala#L891).

I thought that the old approach would not work because it was failing back then, with this error
`INVALID_HANDLE.SESSION_NOT_FOUND] The handle 746e6c86-9fa9-4b08-9572-388c20eaed47 is invalid. Session not found. SQLSTATE: HY000"`, but Spark patches this [error](https://github.com/apache/spark/pull/46971) a week later by still retrying even when encountering this exception, so our way back then should work even without the `Thread.sleep`

tomvanbussel (approved):

PR #3584, opened 2024-08-20
[Spark] CDC reader accept case insensitive Boolean option values

juliuszsompolski:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

`DeltaOptions` are equipped to accept case-insensitive values of boolean flags, but CDCReader was not, resulting in not-accepting 'True'. Make it case insensitive. 

A separate bug in Spark Connect was causing "True" to be passed from Python boolean True. That is being fixed by https://github.com/apache/spark/pull/47790

## How was this patch tested?

Tests added.

Datasource option to enable CDC should now accept "True" and other mixed-case variants and not only "true".

### DIFF
--- a/python/delta/tests/test_deltatable.py
+++ b/python/delta/tests/test_deltatable.py
@@ -561,7 +561,7 @@ class DeltaTableTestsMixin:
         for option in [True, "true", "tRuE"]:
             cdf = self.spark.read.format("delta") \
                 .option("readChangeData", option) \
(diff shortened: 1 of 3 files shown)


juliuszsompolski:
@larsk-db

larsk-db (approved):
LGTM

PR #3601, opened 2024-08-25
[Spark] Relax check for generated columns and CHECK constraints on nested struct fields

### LINKED ISSUE #3250: [Feature Request][Spark] Relax check for generated columns and CHECK constraints depending on nested struct fields

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

### Overview
The check to prevent invalid type changes from being applied in [ImplicitMetadataOperation.checkDependentExpression](https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/schema/ImplicitMetadataOperation.scala#L241) is too strict and can be relaxed to allow changing the type of a field in a struct if that field isn't referenced by a CHECK constraint or generated column.

### Motivation

The type of columns and nested fields that are referenced by a CHECK constraint or generated columns can’t be changed during schema evolution as the expressions may depend on the type.

For example `hash(col)` or `typeof(col)` may/will return different results depending on the column type, leading to existing rows that may not satisfy CHECK constraints anymore or cause existing rows in a generated column to not match the actual result of the generated column expression anymore.

The check enforcing that such type changes aren't applied is too strict when changing the type of fields inside structs and can be relaxed to allow more use cses.

### Further details

https://github.com/delta-io/delta/pull/2881 refactored the check to reject changing the type of a column or field during schema evolution in such a case but for simplicity, it doesn’t try to identify if nested struct fields that are modified overlap or not with fields referenced by a CHECK constraint or a generated column and instead consider the struct as a whole as being modified.

By checking individual fields, we could allow the following valid use case which is rejected today:
```
CREATE TABLE t (a struct<x: byte, y: byte>) USING DELTA TBLPROPERTIES('delta.enableTypeWidening' = 'true');
ALTER TABLE t ADD CONSTRAINT ck CHECK (hash(a.x) > 0);
SET "spark.databricks.delta.schema.autoMerge.enabled" = "true";
INSERT INTO t VALUES (CAST(2 AS INT))
```
Changing the type of struct field a.y  during schema evolution will fail, even though only `a.x` is referenced by the CHECK constraint.

### Willingness to contribute

I'm also happy to support anyone willing to implement this improvement.

xzhseh:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

close #3250.

this PR relaxes the check for nested struct fields especially when only some are being referenced by CHECK constraints or generated columns, which allows for more valid use cases in scenarios involving type widening or schema evolution.

the core function, `checkReferencedByCheckConstraintsOrGeneratedColumns`, inspects the nested/inner fields of the provided `StructType` to determine if any are referenced by dependent (CHECK) constraints or generated columns; for column types like `ArrayType` or `MapType`, the function checks these properties directly without inspecting the inner fields.

## How was this patch tested?
through unit tests in `TypeWideningConstraintsSuite` and `TypeWideningGeneratedColumnsSuite`. 

yes, now the following (valid) use case will not be rejected by the check in [ImplicitMetadataOperation.checkDependentExpression](https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/schema/ImplicitMetadataOperation.scala#L241).

```sql
-- with `DELTA_SCHEMA_AUTO_MIGRATE` enabled
create table t (a struct<x: byte, y: byte>) using delta;
alter table t add constraint ck check (hash(a.x) > 0);

-- changing the type of struct field `a.y` when it's not
-- the field referenced by the CHECK constraint is allowed now.
insert into t (a) values (named_struct('x', CAST(2 AS byte), 'y', 1030))
```


spark/src/main/scala/org/apache/spark/sql/delta/schema/ImplicitMetadataOperation.scala
```diff
@@ -236,42 +337,23 @@ object ImplicitMetadataOperation {
...
-      updateField
-    case (_, field, _, _) => field
-  }
+    SchemaMergingUtils.transformColumns(metadata.schema, dataSchema) {
+      case (fieldPath, currentField, Some(updateField), _)
+        if !SchemaMergingUtils.equalsIgnoreCaseAndCompatibleNullability(
+          currentField.dataType,
+          updateField.dataType
+        ) =>
+        checkReferencedByCheckConstraintsOrGeneratedColumns(
```
johanl-db, line 346: `SchemaMergingUtils.transformColumns` will apply the transform to all StructField in the schema, so you will end up calling `checkReferencedByCheckConstraintsOrGeneratedColumns` multiple times for nested struct fields since that method also recurses into the schema.

For example:
`col struct<a: struct<x: int, y: int>>`
`transformColumns` will call `checkReferencedByCheckConstraintsOrGeneratedColumns` on `col` first, which will recurse into the struct and call itself on `a`.
Then `transformColumns` will call `checkReferencedByCheckConstraintsOrGeneratedColumns` again on `a` directly.

Not really an issue, you can either skip recursing into structs in `checkReferencedByCheckConstraintsOrGeneratedColumns` or call out in a comment there that it's redundant

spark/src/main/scala/org/apache/spark/sql/delta/schema/ImplicitMetadataOperation.scala
```diff
@@ -227,6 +226,108 @@ object ImplicitMetadataOperation {
...
+          spark = spark,
+          protocol = protocol,
+          path = path :+ fromField.name,
+          from = fromField.dataType,
+          to = toField.dataType,
+          metadata = metadata
+        )
+      }
+    case (fromDataType, toDataType) =>
```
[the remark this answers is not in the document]
xzhseh: @johanl-db thanks for the review!

after investigating a bit, it seems that we can not directly access the field of a nested `struct` inside a `array`, e.g.,

```sql
create table t (a struct<b: array<struct<x: byte, y: byte>>>) using delta;

-- this will fail due to - "[FIELD_NOT_FOUND] No such struct field `element` in `x`, `y`."
-- same for generated columns.
alter table t add constraint ck check (hash(a.b.element.x) > 0);
```

there could be other way to achieve this, e.g., `alter table t add constraint ck check (hash(a.b[0].x) > 0)`, but this will cause other issues - even if we check for the generated columns/constraints, the path will not be consistent as `a.b[0].x` since we automatically append the `"element"` in [SchemaMergingUtils.transformColumns](https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/schema/SchemaMergingUtils.scala#L311), which will lead to the checks for dependant constraints/generated columns being bypassed even if we've checked the field for the two properties, i.e., `a.b.element.x` (actual) is not consistent with `a.b[0].x` (expected).

I think we could probably block `ArrayType` and `MapType` (due to the appended `"key"` and `"value"`) for now, and only relax the check for `StructType`.

spark/src/main/scala/org/apache/spark/sql/delta/schema/ImplicitMetadataOperation.scala
xzhseh: thanks for pointing this out!

I've skipped the recursing into structs for `checkReferencedByCheckConstraintsOrGeneratedColumns`, will finalize this after the discussion regarding `MapType` and `ArrayType` [here](https://github.com/delta-io/delta/pull/3601#discussion_r1732091240).

spark/src/main/scala/org/apache/spark/sql/delta/schema/ImplicitMetadataOperation.scala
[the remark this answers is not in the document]
johanl-db: Agree, it's better to keep what you have currently and keep being too strict for map and arrays - it's not worth the effort and risk of bugs

xzhseh:
@johanl-db several updates:
- `checkReferencedByCheckConstraintsOrGeneratedColumns` will now skip recursing into nested `StructType` to better integrate with the existing `SchemaMergingUtils.transformColumns` logic.
- we will remain strict for `MapType` and `ArrayType`, particularly concerning potential nested structs, as discussed [here](https://github.com/delta-io/delta/pull/3601#discussion_r1731203107).
- relevant unit tests in `GeneratedColumnSuite`, `TypeWideningGeneratedColumnsSuite`, and `TypeWideningConstraintsSuite` have been updated accordingly.

I'll add relevant unit tests for `MapType` and `ArrayType` later - other than that, I think we are good to go.

spark/src/main/scala/org/apache/spark/sql/delta/schema/ImplicitMetadataOperation.scala
```diff
@@ -227,6 +227,149 @@ object ImplicitMetadataOperation {
...
+   * @param metadata the metadata used for checking constraints and generated columns.
+   */
+  private def checkReferencedByCheckConstraintsOrGeneratedColumns(
+      spark: SparkSession,
+      protocol: Protocol,
+      path: Seq[String],
+      from: DataType,
+      to: DataType,
+      metadata: Metadata): Unit = (from, to) match {
+    case (StructType(fromFields), StructType(toFields)) =>
```
johanl-db, line 347: I think you may be able to simplify a bit by removing this layer (`checkReferencedByCheckConstraintsOrGeneratedColumns`) and call `checkConstraintsOrGeneratedColumnsExceptStruct` directly from `checkDependentExpressions`, since `transformColumns` will call that method on all struct fields in the schema already:


```
  private def checkConstraintsOrGeneratedColumnsExceptStruct(
      spark: SparkSession,
      path: Seq[String],
      protocol: Protocol,
      metadata: Metadata,
      currentDt: DataType,
      updateDt: DataType): Unit = (currentDt, updateDt) match {
    case (StructType(_), StructType(_)) =>
      // we explicitly ignore the check for `StructType` here.
    case (from, to) =>
      if (currentDt != updateDt) {
        checkDependentConstraints(spark, path, metadata, from, to)
        checkDependentGeneratedColumns(spark, path, protocol, metadata, from, to)
      }
  }

    SchemaMergingUtils.transformColumns(metadata.schema, dataSchema) {
      case (fieldPath, currentField, Some(updateField), _)
        if !SchemaMergingUtils.equalsIgnoreCaseAndCompatibleNullability(
          currentField.dataType,
          updateField.dataType
        ) =>
        checkConstraintsOrGeneratedColumnsExceptStruct(
          spark = sparkSession,
          protocol = protocol,
          path = fieldPath :+ currentField.name,
          from = currentField.dataType,
          to = updateField.dataType,
          metadata = metadata
        )
```

nit: I would rename `checkConstraintsOrGeneratedColumnsExceptStruct` to e.g. `checkDependentExpressionsOnStructField`

spark/src/main/scala/org/apache/spark/sql/delta/schema/ImplicitMetadataOperation.scala
```diff
@@ -227,6 +227,149 @@ object ImplicitMetadataOperation {
...
+          protocol = protocol,
+          metadata = metadata,
+          currentDt = fromField.dataType,
+          updateDt = toField.dataType
+        )
+      }
+    // do note that for `MapType` and `ArrayType`, we intentionally keep being too strict when
+    // checking nested fields - i.e., even if the containing nested fields are not
+    // being referenced by any dependant constraints or generated columns.
+    // reference: <https://github.com/delta-io/delta/pull/3601#discussion_r1731203107>.
```
johanl-db, line 361: Personal opinion: I'm rather against putting links in comments, these can get out of sync, be inaccessible to some, and just generally require context switching to get information.
Might be better instead to give a quick explanation inline: the field paths for maps/arrays in constraints/generated expressions aren't consistent with regular field paths: `a.arr[0].x` vs. `a.element.x`. This makes it hard to recurse into arrays and maps here, so we prefer being too strict.

spark/src/test/scala/org/apache/spark/sql/delta/typewidening/TypeWideningConstraintsSuite.scala
```diff
@@ -133,25 +133,58 @@ trait TypeWideningConstraintsTests { self: QueryTest with TypeWideningTestMixin
...
+        // changing the type of struct field `a.y` and `a.x.h` when it's not
+        // the field referenced by the CHECK constraint is allowed.
+        sql(
+          """
+            | INSERT INTO t (a) VALUES (
+            |   named_struct('x', named_struct('z', CAST(2 AS BYTE), 'h', 2002), 'y', 1030)
+            | )
+            |""".stripMargin
+        )
+        checkAnswer(sql("SELECT hash(a.x.z) FROM t"), Seq(Row(1765031574), Row(1765031574)))
```
johanl-db, line 187: Can you add one test with e.g. `a struct<arr: array<struct<x, y>>>` with a constraint defined on `a.arr[0].x` and changing the type of a.element.x and a.element.x both fail?

johanl-db (commented):
The change looks good and correct, I left some follow up comments for a last iteration before merging

spark/src/test/scala/org/apache/spark/sql/delta/typewidening/TypeWideningConstraintsSuite.scala
xzhseh: yes, I did attempt to add the corresponding tests; the issue is that, for both versions of `checkDependentConstraints`, we are unable to block operations involving `constraints` or `generated columns` related to `a.arr[0].x`, i.e., the operations pass without exceptions being thrown.

this seems to be more of a bug rather than something that should be addressed in this PR - I’m considering either adding the tests with a `// FIXME: ...` comment to highlight the issue, or simply skipping this scenario for now.

spark/src/main/scala/org/apache/spark/sql/delta/schema/ImplicitMetadataOperation.scala
xzhseh: agreed, I’ll replace the link with an explanatory comment.

thanks for the suggestion!

spark/src/main/scala/org/apache/spark/sql/delta/schema/ImplicitMetadataOperation.scala
xzhseh: thanks for the suggestion!
- `checkReferencedByCheckConstraintsOrGeneratedColumns` has been removed.
- `checkDependentExpressionsOnStructField` is now directly called in `checkDependentExpressions`.

spark/src/main/scala/org/apache/spark/sql/delta/schema/ImplicitMetadataOperation.scala
```diff
@@ -227,6 +227,104 @@ object ImplicitMetadataOperation {
...
+      spark: SparkSession,
+      path: Seq[String],
+      protocol: Protocol,
+      metadata: Metadata,
+      currentDt: DataType,
+      updateDt: DataType): Unit = (currentDt, updateDt) match {
+    // we explicitly ignore the check for `StructType` here.
+    case (StructType(_), StructType(_)) =>
+
+    // FIXME: we intentionally incorporate the pattern match for `ArrayType` and `MapType`
```
xzhseh, line 314: added a `FIXME` comment here to indicate this potential bug, not sure whether this is necessary or not, @johanl-db.

spark/src/test/scala/org/apache/spark/sql/delta/typewidening/TypeWideningConstraintsSuite.scala
johanl-db: I created a github issue to track this: https://github.com/delta-io/delta/issues/3634, this is indeed a bug. I expect that use case of applying a check constraint or generated column on a specific array  or map element should be fairly rare.

johanl-db (approved):
The change looks good, we can merge it
The CI failed, failure seems random and not related to anything in this PR
@tdas can you restart the CI + merge when it's green?

tdas:
restarted all the tests. lets see

PR #3605, opened 2024-08-26
[Spark] Add tests for implicit cast in INSERT to delta table

johanl-db:
## Description
Add tests covering implicit casting when inserting into a Delta table.
Covers various insert API:
- Dataframe V1, V2, SQL, streaming
- Append vs. Overwrite
- Position-based vs. name-based

Changes:
- Move test abstraction to run insert using various APIs out of `TypeWideningInsertSchemaEvolutionSuite` and into its own trait to allow reusability.
- Add streaming write to the set of insert APIs that are covered by that abstraction.
- Add implicit casting tests for insert.

## How was this patch tested?
Test-only
No


spark/src/test/scala/org/apache/spark/sql/delta/DeltaInsertIntoTest.scala
```diff
@@ -0,0 +1,272 @@
...
+/**
+ * There are **many** different ways to run an insert:
+ * - Using SQL, the dataframe v1 and v2 APIs or the streaming API.
+ * - Append vs. Overwrite / Partition overwrite.
+ * - Position-based vs. name-based resolution.
+ *
+ * Each take a unique path through analysis. The abstractions below captures these different
+ * inserts to allow more easily running tests with all or a subset of them.
+ */
+trait DeltaInsertIntoTest extends QueryTest with DeltaDMLTestUtils with DeltaSQLCommandTest {
```
johanl-db, line 36: This is moved from `TypeWideningInsertSchemaEvolutionSuite`.

`StreamingInsert` was added in the process to cover streaming writes, and the test runner method `testInserts` was updated to make it slightly more expressive

johanl-db:
Tests covering streaming insert assume https://github.com/delta-io/delta/pull/3443 is merged, these will fail until that change is actually merged and will succeed afterwards.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaInsertIntoTest.scala
```diff
@@ -0,0 +1,272 @@
...
+      initialJsonData: Seq[String],
+      partitionBy: Seq[String] = Seq.empty,
+      insertSchemaDDL: String,
+      insertJsonData: Seq[String],
+      overwriteWhere: (String, Int),
+      expectedSchema: StructType = null,
+      checkError: SparkThrowable => Unit = null,
+      includeInserts: Seq[Insert] = allInsertTypes,
+      excludeInserts: Seq[Insert] = Seq.empty,
+      confs: Seq[(String, String)] = Seq.empty): Unit = {
```
c27kwan, line 227: This list is getting long, and it also seems like maybe some data goes together. E.g. (initialSchemaDDL,      initialJsonData) and (insertSchemaDDL, insertJsonData) should have probably been a case class. (includeInserts, excludeInserts) can be 1 variable passed in with the default value of `allInsertTypes`. This PR is already quite big so if you choose to do this additional refactor it'd be better to do it in another PR.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaInsertIntoTest.scala
```diff
@@ -0,0 +1,272 @@
...
+  }
+
+  /** df.write.mode(mode).insertInto() */
+  case class DFv1InsertInto(mode: SaveMode) extends Insert {
+    val name: String = s"DFv1 insertInto() - $mode"
+    def runInsert(columns: Seq[String], whereCol: String, whereValue: Int): Unit =
+      spark.read.table("source").write.mode(mode).insertInto("target")
+  }
+
+  /** df.write.mode(mode).saveAsTable() */
```
c27kwan, line 124: There is also `save()` API fyi

spark/src/test/scala/org/apache/spark/sql/delta/DeltaInsertIntoTest.scala
```diff
@@ -0,0 +1,272 @@
...
+   */
+  def testInserts(name: String)(
+      initialSchemaDDL: String,
+      initialJsonData: Seq[String],
+      partitionBy: Seq[String] = Seq.empty,
+      insertSchemaDDL: String,
+      insertJsonData: Seq[String],
+      overwriteWhere: (String, Int),
+      expectedSchema: StructType = null,
+      checkError: SparkThrowable => Unit = null,
```
c27kwan, line 224: We prefer `Option` over `null`s in Scala. In this case, it sounds like we want `Either`: https://www.scala-lang.org/api/2.13.6/scala/util/Either.html

spark/src/test/scala/org/apache/spark/sql/delta/DeltaInsertIntoImplicitCastSuite.scala
```diff
@@ -0,0 +1,208 @@
...
+ *
+ * The casting behavior is (unfortunately) dependent on the API used to run the INSERT, e.g.
+ * Dataframe V1 insertInto() vs V2 saveAsTable() or using SQL.
+ * This suite intends to exhaustively cover all the ways INSERT can be run on a Delta table. See
+ * [[DeltaInsertIntoTest]] for a list of these INSERT operations covered.
+ */
+class DeltaInsertIntoImplicitCastSuite extends DeltaInsertIntoTest {
+
+  for (schemaEvolution <- BOOLEAN_DOMAIN) {
+    testInserts("insert with implicit up and down cast on top-level fields, " +
```
c27kwan, line 36: nit: isn't there _no_ implicit up and down cast since the schema remains the same?

spark/src/test/scala/org/apache/spark/sql/delta/DeltaInsertIntoImplicitCastSuite.scala
```diff
@@ -0,0 +1,208 @@
...
+        DFv2Append,
+        DFv2Overwrite,
+        DFv2OverwritePartition
+      ),
+      confs = Seq(DeltaSQLConf.DELTA_SCHEMA_AUTO_MIGRATE.key -> schemaEvolution.toString)
+    )
+
+    testInserts("insert with implicit up and down cast on fields nested in map, " +
+      s"schemaEvolution=$schemaEvolution")(
+      initialSchemaDDL = "key int, m map<string, struct<x: long, y: int>>",
```
c27kwan, line 152: Should we test the `map`'s key as well?

c27kwan (approved):
lgtm!

spark/src/test/scala/org/apache/spark/sql/delta/DeltaInsertIntoImplicitCastSuite.scala
johanl-db: This is going to be hard to test here, these tests rely on parsing json to setup the table and ingest data. and I don't believe that supports using anything other than a string for map keys.
That would be a pretty significant rewrite of all the tests

spark/src/test/scala/org/apache/spark/sql/delta/DeltaInsertIntoImplicitCastSuite.scala
johanl-db: The schema of the data inserted isn't the same as the schema of the table:
table: `a long, b int`
data: `a int, b long`

so there' is an upcast for `a` and downcast for `b`

PR #3610, opened 2024-08-27
[3.2] always include CatalogTable#storage#properties into data source options

cloud-fan:
backport https://github.com/delta-io/delta/pull/3536 . It's a clean cherry-pick

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

`DeltaTableV2` uses the options to create `DeltaLog`, assuming that the options should always be consistent with the one in `CatalogTable#storage#properties`. This may not be true as custom catalogs may add more table storage properties on the fly, like Unity Catalog.

## How was this patch tested?

Tested locally with UC. It's not an issue for HMS.

no


cloud-fan:
cc @tdas @scovich

tdas (approved):

PR #3626, opened 2024-08-30
[Spark] Automatic type widening in Delta streaming sink

johanl-db:
## Description
This change introduces automatic type widening during schema evolution in the Delta streaming sink.
Conditions for type widening to trigger:
- Type widening is enabled on the Delta table
- Schema evolution (`mergeSchema`) is enabled on the sink
- The data written to the sink uses a type that is strictly wider than the current type in the table schema, and moving from the narrower to the wider type is eligible for type widening - see `TypeWidening.isTypeChangeSupportedForSchemaEvolution`

When all conditions are satisfied, the table schema is updated to use the wider type before ingesting the data.

## How was this patch tested?
Added test suite `TypeWideningStreamingSinkSuite` covering type widening in the Delta streaming sink

This builds on the user-facing change introduced in https://github.com/delta-io/delta/pull/3443 that allows writing to a delta sink using a different type than the current table type.
Without type widening:
```
spark.readStream
    .table("delta_source")
    # Column 'a' has type INT in 'delta_sink'.
    .select(col("a").cast("long").alias("a"))
    .writeStream
    .format("delta")
    .option("checkpointLocation", "<location>")
    .toTable("delta_sink")
```
The write to the sink succeeds, column `a`  retains its type `INT` and the data is cast from `LONG` to `INT` on write.

With type widening:
```
spark.sql("ALTER TABLE delta_sink SET TBLPROPERTIES ('delta.enableTypeWidening' = 'true')")
spark.readStream
    .table("delta_source")
    # Column 'a' has type INT in 'delta_sink'.
    .select(col("a").cast("long").alias("a"))
    .writeStream
    .format("delta")
    .option("checkpointLocation", "<location>")
    .option("mergeSchema", "true")
    .toTable("delta_sink")
```
The write to sink succeeds, the type of column `a` is changed from `INT` to `LONG`, data is ingested as `LONG`.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSink.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSink.scala
@@ -20,7 +20,7 @@ import java.util.concurrent.ConcurrentHashMap
 
 import org.apache.spark.sql.delta._
 import org.apache.spark.sql.delta.DeltaOperations.StreamingUpdate
-import org.apache.spark.sql.delta.actions.{FileAction, Metadata, SetTransaction}
+import org.apache.spark.sql.delta.actions.{FileAction, Metadata, Protocol, SetTransaction}
 import org.apache.spark.sql.delta.logging.DeltaLogKeys
 import org.apache.spark.sql.delta.metering.DeltaLogging
 import org.apache.spark.sql.delta.schema.{ImplicitMetadataOperation, SchemaMergingUtils, SchemaUtils}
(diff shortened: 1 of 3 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSink.scala
```diff
@@ -168,17 +167,20 @@ case class DeltaSink(
 
   /**
    * Returns the schema to use to write data to this delta table. The write schema includes new
-   * columns to add with schema evolution and reconciles types to match the table types.
+   * columns to add with schema evolution and reconciles types to match the table types when
+   * possible or apply type widening if enabled.
    */
-  private def getWriteSchema(metadata: Metadata, dataSchema: StructType): StructType = {
+  private def getWriteSchema(protocol: Protocol, metadata: Metadata, dataSchema: StructType)
+    : StructType = {
```
tomvanbussel, line 174: Style nit:
```suggestion
  private def getWriteSchema(
      protocol: Protocol, metadata: Metadata, dataSchema: StructType): StructType = {
```

tomvanbussel (approved):
Simple code change and extensive tests, nice! LGTM!

PR #3637, opened 2024-09-04
[Spark] Align supported type changes in type widening preview with Iceberg

johanl-db:
## Description
The following type changes available in the preview of type widening aren't supported by Iceberg and it will be very hard to support them - mainly due to the [Binary single-value serialization](https://iceberg.apache.org/spec/#binary-single-value-serialization) part of the spec:
- (byte,short,int,long) → decimals
- (byte,short,int) → double
- decimal scale increase, e.g. decimal(12,2) → decimal(14,4)

These type changes will get in the way of interoperability with Iceberg. To prevent issues in the future, we won't allow applying these types changes in the stable version of the type widening feature. We can add these back once Iceberg supports them. Reading a table that had such a type change applied will still be allowed though.

## How was this patch tested?
Moved existing test cases for these type changes to the 'unsuported' list.
Added a dedicated test to ensure we can still read back data if such a type change was applied in the past.

Running `ALTER TABLE <table> CHANGE COLUMN <col> TYPE <type>` to apply one of the deprecated type changes will fail with:
```
[DELTA_UNSUPPORTED_ALTER_TABLE_CHANGE_COL_OP] ALTER TABLE CHANGE COLUMN is not supported for changing column <fieldPath> from <oldField> to <newField>
```

### DIFF
--- a/spark/src/main/scala-spark-3.5/shims/TypeWideningShims.scala
+++ b/spark/src/main/scala-spark-3.5/shims/TypeWideningShims.scala
@@ -42,12 +42,7 @@ object TypeWideningShims {
       case _ => false
     }
 
-  /**
-   * Returns whether the given type change can be applied during schema evolution. Only a
-   * subset of supported type changes are considered for schema evolution.
-   */
-  def isTypeChangeSupportedForSchemaEvolution(fromType: AtomicType, toType: AtomicType): Boolean = {
-    // All supported type changes are eligible for schema evolution.
+  /** Whether we support reading the given type change. */
+  def canReadTypeChange(fromType: AtomicType, toType: AtomicType): Boolean =
     isTypeChangeSupported(fromType, toType)
-  }
 }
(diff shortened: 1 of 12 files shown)


spark/src/main/scala-spark-3.5/shims/TypeWideningShims.scala
```diff
@@ -42,12 +42,7 @@ object TypeWideningShims {
       case _ => false
     }
 
-  /**
-   * Returns whether the given type change can be applied during schema evolution. Only a
-   * subset of supported type changes are considered for schema evolution.
-   */
-  def isTypeChangeSupportedForSchemaEvolution(fromType: AtomicType, toType: AtomicType): Boolean = {
```
johanl-db, line 49: There are no types changes anymore that can only be applied via `ALTER TABLE` but not during schema evolution, this method is removed.

PR #3644, opened 2024-09-04
[Infra] Include GitHub workflow name in each task

scottsand-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [X] Other (Infra)

## Description

I'd like to do some data science using the GitHub REST API on our PR runtimes.

Unfortunately, using the GitHub CHECKs API (https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#list-check-runs-for-a-git-reference), I only get the name of the individual job. Getting the name of the parent workflow requires an additional API call.

This is because our job names are the default job names right now for most of our GitHub workflows. This doesn't give us any info as to which workflow it is for. For example, today I'd get result `test (2.13.13, 0)` but I am unable to determine if this is for Delta Spark Latest or Delta Spark Master.

This PR updates the workflow job name to include the workflow name so now we can uniquely identify the jobs.

## How was this patch tested?

CI tests.

No

### DIFF
--- a/.github/workflows/connectors_test.yaml
+++ b/.github/workflows/connectors_test.yaml
@@ -2,7 +2,7 @@ name: "Delta Connectors Tests"
 on: [push, pull_request]
 jobs:
   build:
-    name: "Run tests"
+    name: "Delta Connectors: Scala ${{ matrix.scala }}"
     runs-on: ubuntu-20.04
     strategy:
       matrix:
--- a/.github/workflows/connectors_test.yaml
+++ b/.github/workflows/connectors_test.yaml
@@ -1,8 +1,8 @@
-name: "Delta Connectors Tests"
+name: "Delta Connectors"
 on: [push, pull_request]
 jobs:
   build:
-    name: "Delta Connectors: Scala ${{ matrix.scala }}"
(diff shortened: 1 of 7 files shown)


zachschuermann (approved):
It looks like I can see the new names in the CI that's running on this PR? Unfortunately this feels redundant - many have names like `Delta Kernel Tests / Delta Kernel` - any idea if we could alias somehow to shorten? if not then oh well probably worth the naming redundancy to do your science

scottsand-db:
> It looks like I can see the new names in the CI that's running on this PR? Unfortunately this feels redundant - many have names like `Delta Kernel Tests / Delta Kernel` - any idea if we could alias somehow to shorten? if not then oh well probably worth the naming redundancy to do your science

@zachschuermann - no way that I know of. If this redundancy escalates to something more serious/concerning, we can revert this PR

scottsand-db:
> The names in the UI are getting very very long and cut off. This aint great. Especially this `Delta Spark Local Publishing and Examples Compilation / Delta Spark Local Publishing and Examples Compilation: Scala 2.13.13 (pull_request)`

@tdas - ACK, I pushed a new commit with shortened names

PR #3652, opened 2024-09-06
[Spark] Fix overflow in GeneratedColumnSuite

johanl-db:
## Description
Small test fix, follow up from https://github.com/delta-io/delta/pull/3601
The test contains an overflow and fails when running with ANSI_MODE enabled.

## How was this patch tested?
Test-only


longvu-db (approved):
LGTM, just for learning, ANSI_MODE doesn't like overflow so it complained about that right?

johanl-db:
> LGTM, just for learning, ANSI_MODE doesn't like overflow so it complained about that right?

Right, see https://spark.apache.org/docs/latest/sql-ref-ansi-compliance.html#ansi-compliance

PR #3654, opened 2024-09-09
DeltaCatalog.createTable should respect PROP_IS_MANAGED_LOCATION

cloud-fan:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Even if a table has the location field, it should still be a managed table if `PROP_IS_MANAGED_LOCATION` is present in the table properties.

Note: this case won't happen with Spark integration solely. It's only an issue for third-party catalogs that delegate requests to `DeltaCatalog`, such as Unity Catalog.

## How was this patch tested?

new test

no

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala
@@ -152,8 +152,13 @@ class DeltaCatalog extends DelegatingCatalogExtension
       .getOrElse(spark.sessionState.catalog.defaultTablePath(id))
     val storage = DataSource.buildStorageFormatFromOptions(writeOptions)
       .copy(locationUri = Option(loc))
-    val tableType =
-      if (location.isDefined) CatalogTableType.EXTERNAL else CatalogTableType.MANAGED
+    val isManagedLocation = Option(allTableProperties.get(TableCatalog.PROP_IS_MANAGED_LOCATION))
+      .exists(_.equalsIgnoreCase("true"))
+    val tableType = if (location.isEmpty || isManagedLocation) {
+      CatalogTableType.MANAGED
+    } else {
+      CatalogTableType.EXTERNAL
+    }
     val commentOpt = Option(allTableProperties.get("comment"))
 
 
--- a/spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala
@@ -152,9 +152,14 @@ class DeltaCatalog extends DelegatingCatalogExtension
       .getOrElse(spark.sessionState.catalog.defaultTablePath(id))
     val storage = DataSource.buildStorageFormatFromOptions(writeOptions)
       .copy(locationUri = Option(loc))
+    // PROP_IS_MANAGED_LOCATION indicates that the table location is not user-specified but
+    // system-generated. The table should be created as managed table in this case.
     val isManagedLocation = Option(allTableProperties.get(TableCatalog.PROP_IS_MANAGED_LOCATION))
       .exists(_.equalsIgnoreCase("true"))
-    val tableType = if (location.isEmpty || isManagedLocation) {
+    // Note: Spark generates the table location for managed tables in
+    // `DeltaCatalog#delegate#createTable`, so `isManagedLocation` should never be true if
+    // Unity Catalog is not involved. For safety we also check `isUnityCatalog` here.
+    val tableType = if (location.isEmpty || (isManagedLocation && isUnityCatalog)) {
       CatalogTableType.MANAGED
     } else {
       CatalogTableType.EXTERNAL
--- a/spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala
@@ -159,7 +159,8 @@ class DeltaCatalog extends DelegatingCatalogExtension
     // Note: Spark generates the table location for managed tables in
     // `DeltaCatalog#delegate#createTable`, so `isManagedLocation` should never be true if
     // Unity Catalog is not involved. For safety we also check `isUnityCatalog` here.
-    val tableType = if (location.isEmpty || (isManagedLocation && isUnityCatalog)) {
+    val respectManagedLoc = isUnityCatalog || org.apache.spark.util.Utils.isTesting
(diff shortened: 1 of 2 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala
```diff
@@ -152,8 +152,13 @@ class DeltaCatalog extends DelegatingCatalogExtension
       .getOrElse(spark.sessionState.catalog.defaultTablePath(id))
     val storage = DataSource.buildStorageFormatFromOptions(writeOptions)
       .copy(locationUri = Option(loc))
-    val tableType =
-      if (location.isDefined) CatalogTableType.EXTERNAL else CatalogTableType.MANAGED
+    val isManagedLocation = Option(allTableProperties.get(TableCatalog.PROP_IS_MANAGED_LOCATION))
```
tdas, line 155: can you please add a comment here explaining this logic? The context around this property is not clear at all.

spark/src/test/scala/org/apache/spark/sql/delta/CustomCatalogSuite.scala
```diff
@@ -297,6 +297,22 @@ class CustomCatalogSuite extends QueryTest with SharedSparkSession
...
+    // Reset catalog manager so that the new `spark_catalog` implementation can apply.
+    spark.sessionState.catalogManager.reset()
+    withSQLConf("spark.sql.catalog.spark_catalog" -> classOf[DummySessionCatalog].getName) {
+      withTable("t") {
+        withTempPath { path =>
+          sql(s"CREATE TABLE t (id LONG) USING delta TBLPROPERTIES (fakeLoc='$path')")
+          val t = spark.sessionState.catalogManager.v2SessionCatalog.asInstanceOf[TableCatalog]
+            .loadTable(Identifier.of(Array("default"), "t"))
+          // It should be a managed table.
+          assert(!t.properties().containsKey(TableCatalog.PROP_EXTERNAL))
```
tdas, line 311: can we also test that if the catalog does not support managed tables, then the create table can throw error?

spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala
prakharjain09, line 155: TableCatalog also seems to have `PROP_EXTERNAL` ? Is this also serving the same purpose? What happens when both `PROP_EXTERNAL` and `PROP_IS_MANAGED_LOCATION` are set to true?

      /**
       * A reserved property to specify a table was created with EXTERNAL.
       */
      String PROP_EXTERNAL = "external";

Since we are making DeltaCatalog aware of the `TableCatalog.PROP_IS_MANAGED_LOCATION`, should we also make it aware of `PROP_EXTERNAL` for the sake of completeness in this PR itself?

tdas: +1

I also think we should add asserts and checks here so that UC-Spark catalog never sets these mutually exclusive properly inconsistently. 

same comments i made in the linked PR - https://github.com/unitycatalog/unitycatalog/pull/449/files#r1750747701

spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala
```diff
@@ -152,8 +152,13 @@ class DeltaCatalog extends DelegatingCatalogExtension
...
+      .exists(_.equalsIgnoreCase("true"))
+    val tableType = if (location.isEmpty || isManagedLocation) {
```
tdas, line 157: What will be the impact of this PR with non-UC catalogs like HMS if we backport this to branch 3.2?

tdas: or should we make it safer to backport by using the [isUnityCatalog](https://github.com/delta-io/delta/blob/89c57836e503a957aae0a5ee073cde44219a7491/spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala#L74) flag here as well?

I think its better we use the flag.

spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala
cloud-fan: Spark will never set `PROP_IS_MANAGED_LOCATION` when calling `createTable` of the root catalog. Spark only   sets it during `loadTable` for `V1Table`, so that commands like DESCRIBE TABLE can know it's a managed table even if it has a location.

cloud-fan: Also see https://github.com/unitycatalog/unitycatalog/pull/449#discussion_r1751065632

spark/src/test/scala/org/apache/spark/sql/delta/CustomCatalogSuite.scala
cloud-fan: I can't test this here as the underlying catalog is Spark HMS which we can't customize. I think the current test is OK as we have guaranteed the created table is managed table and if we replace HMS with UC, we can for sure fail correctly.

spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala
prakharjain09: It is upto the catalog impl to decide how they want to proceed in such scenario? Can we add an assertion in Delta to fail in such a scenario?

prakharjain09 (approved):

tdas (approved):

tdas:
Approving and merging the PR to speed up backporting to 3.2.

spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala
cloud-fan: Delta is more like a proxy and it will eventually forward the request to HMS/UC or whatever custom catalog. I think it's better to leave it to the underlying catalog to enforce it.

cloud-fan: Anyway, this is not an issue for now as neither HMS nor UC generate location for external tables.

tdas: It is fine for now. Thats why i merged the PR. But it always good to think of future proofing things so that there are no surprises in the future when somebody also loads DeltaCatalog with other catalogs.

PR #3660, opened 2024-09-10
[Spark] Allow setting the same active Delta transaction

cstavr:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

OptimisticTransaction.setActive() and OptimisticTransaction.withActive() methods will fail if the active transaction is already set, even if the caller tries to set the same transaction. This commit fixes this issue and allows setting the same transaction instance.

## How was this patch tested?

New and existing tests.

No


larsk-db (approved):
LGTM

PR #3661, opened 2024-09-10
[Spark] Report writer offset mismatch in DVStore

xupefei:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

We identified a potential issue with the API used to write DV to files, where the `DataOutputStream.size()` method may not return the correct file size. Our investigation revealed that `DataOutputStream` maintains its own byte count, but its `write(data)` method does not increment this counter `DataOutputStream` has multiple subclasses, which might override the counter or the `write(data)` method to update the counter correctly. We want to find out which class is being used when the issue occurs, thus this PR.

To address this, we introduced our own mechanism to track the number of bytes written, which will be used solely for logging. If there is a discrepancy between the system's reported file size and our own record, a Delta event will be triggered.

## How was this patch tested?

Not needed.

No.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/storage/dv/DeletionVectorStore.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/storage/dv/DeletionVectorStore.scala
@@ -25,15 +25,15 @@ import java.util.zip.CRC32
 import org.apache.spark.sql.delta.DeltaErrors
 import org.apache.spark.sql.delta.actions.DeletionVectorDescriptor
 import org.apache.spark.sql.delta.deletionvectors.{RoaringBitmapArray, StoredBitmap}
+import org.apache.spark.sql.delta.metering.DeltaLogging
 import org.apache.spark.sql.delta.util.PathWithFileSystem
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.{FileSystem, FSDataOutputStream, Path}
 
-import org.apache.spark.internal.Logging
 import org.apache.spark.paths.SparkPath
 import org.apache.spark.util.Utils
 
-trait DeletionVectorStore extends Logging {
+trait DeletionVectorStore extends DeltaLogging {
   /**
    * Read a Deletion Vector and parse it as [[RoaringBitmapArray]].
    */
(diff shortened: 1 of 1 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/storage/dv/DeletionVectorStore.scala
```diff
@@ -205,22 +205,36 @@ class HadoopFileSystemDVStore(hadoopConf: Configuration)
...
         val dvRange = DeletionVectorStore.DVRangeDescriptor(
           offset = outputStream.size(),
           length = data.length,
-          checksum = DeletionVectorStore.calculateChecksum(data)
-          )
+          checksum = DeletionVectorStore.calculateChecksum(data))
+
+        if (writtenBytes != dvRange.offset) {
+          recordDeltaEvent(
+            deltaLog = null,
```
c27kwan, line 224: Do we have a lot of call sites for createWriter? I wonder if we can pass the delta log there.

spark/src/main/scala/org/apache/spark/sql/delta/storage/dv/DeletionVectorStore.scala
```diff
@@ -205,22 +205,36 @@ class HadoopFileSystemDVStore(hadoopConf: Configuration)
...
+          writtenBytes += 1
         }
         val dvRange = DeletionVectorStore.DVRangeDescriptor(
           offset = outputStream.size(),
           length = data.length,
-          checksum = DeletionVectorStore.calculateChecksum(data)
-          )
+          checksum = DeletionVectorStore.calculateChecksum(data))
+
+        if (writtenBytes != dvRange.offset) {
```
c27kwan, line 222: should we throw an exception? also if we pay the overhead of tracking the bytes ourselves, why do we still want `outputStream.size()`?

xupefei: The current logging is a non-invasive addition to the existing logic. We don't want to alter the return value - when a mismatch happen the old "CRC checksum error" will be thrown upon validation, so we're still safe.

spark/src/main/scala/org/apache/spark/sql/delta/storage/dv/DeletionVectorStore.scala
xupefei: That would change more places in the code, which is what I don't want two.
Having the DV path and class name is more than enough for debugging.

c27kwan (approved):
lgtm!

larsk-db (approved):
LGTM

PR #3667, opened 2024-09-11
[Kernel] Add some logging to the getChanges implementation

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Adds some logging to the getChanges implementation.

## How was this patch tested?

N/A

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaLogActionUtils.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaLogActionUtils.java
@@ -266,7 +266,7 @@ public class DeltaLogActionUtils {
    */
   private static List<FileStatus> listCommitFiles(
       Engine engine, Path tablePath, long startVersion, long endVersion) {
-    
+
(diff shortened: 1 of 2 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaLogActionUtils.java
```diff
@@ -261,22 +266,38 @@ private static CloseableIterator<FileStatus> listLogDir(
    */
   private static List<FileStatus> listCommitFiles(
       Engine engine, Path tablePath, long startVersion, long endVersion) {
-
+    
+    // TODO update to support coordinated commits; suggested to load the Snapshot at endVersion
+    //  and get the backfilled/unbackfilled commits from the LogSegment to combine with commit files
+    //  listed from [startVersion, LogSegment.checkpointVersion]
```
allisonport-db, line 272: I've also added this note since we had a discussion on this and I think is relevant context for future implementation. I've already added updating getChanges to the coordinated commits project plan.

vkorukanti (approved):

PR #3672, opened 2024-09-12
[Spark] Table features protocols should handle legacy reader features as reader features

zhipengmao-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
This PR fixes a bug where protocol (2, 7) is not considered to support reader features. This issue could manifest in several places. For example, it would result in hiding (2, x) features in the reader features list when it was the only reader feature present. For example, Protocol(2, 7, None, Set(ColumnMapping, RowTracking).

## How was this patch tested?
Added new test and adapted existing tests.

No.


bart-samwel (approved):
LG! Not a trivial change. Column Mapping is really an odd feature in table features land when you don't have the reader table features protocol v3 enabled.

PR #3673, opened 2024-09-12
[Spark] Fix for pushing down filters referencing attributes with special characters in name into CDF

tomvanbussel:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR fixes the push down of filters into CDF. Previously this would cause a parsing exception to be thrown if the filter referenced an attribute with at least one special character in the name. This is due to having to turn the `Filter`s back into `Expression`s. We avoid this by extending `CatalystScan` instead, which avoids the `Expression` to `Filter` to `Expression` roundtrip.

## How was this patch tested?

Added a new test

No


cstavr (approved):
Good catch! LGTM

PR #3674, opened 2024-09-12
DeltaCatalog#createTable should respect write options

cloud-fan:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

As of today, Delta extracts write options from table properties only for CTAS (in `StagedDeltaTableV2#commitStagedChanges`), but not for CREATE TABLE. In general, this makes sense because CREATE TABLE has no data-writing but CTAS has. However, the write options can be file system configs that we should respect because CREATE TABLE needs to access Delta logs.

This PR makes Delta CREATE TABLE to follow CTAS and also extract write options from table properties.

## How was this patch tested?

Locally tested with Unity Catalog. It's hard to write a test in Delta because the write options are not persisted but only used during table creation.

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala
@@ -347,12 +347,21 @@ class DeltaCatalog extends DelegatingCatalogExtension
       properties: util.Map[String, String]) : Table =
     recordFrameProfile("DeltaCatalog", "createTable") {
       if (DeltaSourceUtils.isDeltaDataSourceName(getProvider(properties))) {
+        val (props, writeOptions) = getTablePropsAndWriteOptions(properties)
+        writeOptions.foreach { case (k, v) =>
+          // Continue putting in Delta prefixed options to avoid breaking workloads
+          if (k.toLowerCase(Locale.ROOT).startsWith("delta.")) {
+            props.put(k, v)
+          }
+        }
         createDeltaTable(
           ident,
           schema,
           partitions,
-          properties,
-          Map.empty,
+          // TODO: we should extract write options from table properties for all the cases. We
+          //       can remove the UC check when we have confidence.
+          if (isUnityCatalog) props else properties,
+          if (isUnityCatalog) writeOptions else Map.empty,
           sourceQuery = None,
           TableCreationModes.Create
         )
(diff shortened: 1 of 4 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala
```diff
@@ -523,6 +532,27 @@ class DeltaCatalog extends DelegatingCatalogExtension
     }
   }
 
+  private def getTablePropsAndWriteOptions(properties: util.Map[String, String])
```
cloud-fan, line 535: This is the same code moved from https://github.com/delta-io/delta/pull/3674/files#diff-e880b2b407605f237c5bb566e3e06f981a4fbbb6ff675d7832b2cf5d2daa6b3dL547

spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala
```diff
@@ -347,12 +347,21 @@ class DeltaCatalog extends DelegatingCatalogExtension
       properties: util.Map[String, String]) : Table =
     recordFrameProfile("DeltaCatalog", "createTable") {
       if (DeltaSourceUtils.isDeltaDataSourceName(getProvider(properties))) {
+        val (props, writeOptions) = getTablePropsAndWriteOptions(properties)
+        writeOptions.foreach { case (k, v) =>
+          // Continue putting in Delta prefixed options to avoid breaking workloads
+          if (k.toLowerCase(Locale.ROOT).startsWith("delta.")) {
+            props.put(k, v)
+          }
```
prakharjain09, line 355: Why is this needed i.e. why do we need to put `options.delta.xyz` to delta table properties? In what scenarios workload can break?

spark/src/main/scala/org/apache/spark/sql/delta/commands/CreateDeltaTableCommand.scala
```diff
@@ -134,7 +134,12 @@ case class CreateDeltaTableCommand(
     }
 
     val tableLocation = getDeltaTablePath(tableWithLocation)
-    val deltaLog = DeltaLog.forTable(sparkSession, tableLocation)
+    // To be safe, here we only extract file system options from table storage properties, to create
+    // the DeltaLog.
+    val fileSystemOptions = table.storage.properties.filter { case (k, _) =>
+      DeltaTableUtils.validDeltaTableHadoopPrefixes.exists(k.startsWith)
+    }
+    val deltaLog = DeltaLog.forTable(sparkSession, tableLocation, fileSystemOptions)
```
prakharjain09, line 142: Can we add a test for this? that the right file system options are passed to DeltaLog.

spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala
cloud-fan: This is copied from the existing code: https://github.com/delta-io/delta/blob/d81ce5ee557ad9741052c9f3402fbef3ff91da87/spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala#L586-L588

I don't have enough context to answer this question... This is just for safety as before this change, all the Spark table properties are kept as the Delta table properties, including `options.xyz` and `options.delta.xyz`. After this change only `options.xyz` is excluded from Delta table properties. This is consistent with CTAS command.

spark/src/main/scala/org/apache/spark/sql/delta/commands/CreateDeltaTableCommand.scala
cloud-fan: I may need some help on it. Do we have a fake file system in the Delta test framework? This is the write options and I can't verify it after the Delta table is created, as write options are not persisted into the Delta log.

spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala
cloud-fan: maybe I should respect that legacy config here as well

spark/src/main/scala/org/apache/spark/sql/delta/commands/CreateDeltaTableCommand.scala
prakharjain09: Yes there is a FakeFileSystem and here is a [reference test suite using it](https://github.com/delta-io/delta/blame/master/spark/src/test/scala/io/delta/tables/DeltaTableSuite.scala#L290).

spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala
prakharjain09: +1 - just for consistency we can continue to do that.

prakharjain09 (approved):

PR #3675, opened 2024-09-13
[Delta Sharing] Enable D2D delta sharing with type widening

johanl-db:
## Description
Adds type widening to the list of supported features for D2D delta sharing and adds client-side tests covering reading a table that had a type change applied using the type widening table feature.

## How was this patch tested?
Added tests.


zhipengmao-db (approved):
LGTM! Thank you for enabling the feature and adding the tests!

linzhou-db (approved):

PR #3678, opened 2024-09-13
Refactor Vacuum code to properly handle path url encoding 

rajeshparangi:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This is intended to handle special characters in the table path. For context, DV team recently made a change to include special characters in all paths used by tests.

As of today, the paths we get from listing are not url encoded -- they get url-encoded later in the logic. However, paths contained in delta log files are already url encoded. To keep these two things compatible, this change makes the file names from listing to be url encoded and changes the later logic to not url encode again.

## How was this patch tested?

Existing tests

NO

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/commands/VacuumCommand.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/commands/VacuumCommand.scala
@@ -178,6 +178,11 @@ object VacuumCommand extends VacuumCommandImpl with Serializable {
             None
           }
       }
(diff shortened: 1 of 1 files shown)


rahulsmahadev (approved):

somani (approved):

rajeshparangi:
jenkins merge

rajeshparangi:
jenkins merge

PR #3680, opened 2024-09-16
[Spark] Use `condition` instead of `errorClass` in `checkError()`

MaxGekk:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

In the PR, I propose to use the `condition` parameter instead of `errorClass` in calls of `checkError` because `errorClass` was renamed in Spark by the PR https://github.com/apache/spark/pull/48027. This PR fixes compilation issues like:
```
[error]       checkError(
[error]       ^
[error] /home/runner/work/delta/delta/spark/src/test/scala/org/apache/spark/sql/delta/rowtracking/RowTrackingReadWriteSuite.scala:304:7: overloaded method checkError with alternatives:
[error]   (exception: org.apache.spark.SparkThrowable,condition: String,sqlState: Option[String],parameters: Map[String,String],context: RowTrackingReadWriteSuite.this.ExpectedContext)Unit <and>
```

## How was this patch tested?

By compiling locally.

No. This makes changes in tests only.

### DIFF
--- a/spark/src/test/scala-spark-master/org/apache/spark/sql/delta/DeltaVariantSuite.scala
+++ b/spark/src/test/scala-spark-master/org/apache/spark/sql/delta/DeltaVariantSuite.scala
@@ -100,7 +100,7 @@ class DeltaVariantSuite
       // check previously thrown error message
       checkError(
         e,
-        errorClass = "DELTA_FEATURES_REQUIRE_MANUAL_ENABLEMENT",
+        condition = "DELTA_FEATURES_REQUIRE_MANUAL_ENABLEMENT",
         parameters = Map(
           "unsupportedFeatures" -> VariantTypeTableFeature.name,
           "supportedFeatures" -> currentFeatures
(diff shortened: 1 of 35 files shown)


MaxGekk:
@cloud-fan Could you have a look at the PR, please. It fixes compilation issues.

cloud-fan:
I think Delta master branch needs to cross-compile with both Spark master and 3.5 branches

MaxGekk:
> We should create a shim layer for checkError

@cloud-fan As an alternative solution, we could pass the first two arguments `exception` and `condition` (`errorClass`) by **positions** not by names. The rest arg could be passed by names. I have double checked all `checkError`, they all starts from those two args:
```scala
  protected def checkError(
      exception: SparkThrowable,
      condition: String,
```

MaxGekk:
@cloud-fan Compilation has been fixed. The failed tests are not related to my changes, I hope.

cloud-fan (approved):

vkorukanti (approved):

PR #3686, opened 2024-09-18
Exclude metadata only updates from DV check

cstavr:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

During commit we validate that `AddFile` actions cannot contain Deletion Vectors when DVs are not enabled for a table (table property). This restriction is incorrect for actions that update metadata of existing files, e.g. `ComputeStatistics` or `RowTrackingBackfill`. The current code skips the check for `ComputeStatistics` operation but not for other operations that perform in-place-metadata updates. The new `isInPlaceFileMetadataUpdate` method is added to Delta operations so that we can easily distinguish such operations. 

The `getAssertDeletionVectorWellFormedFunc` function is slightly refactor to be more readable.

## How was this patch tested?

Existing tests provide coverage.

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaOperations.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaOperations.scala
@@ -91,6 +91,21 @@ object DeltaOperations {
      */
     def checkAddFileWithDeletionVectorStatsAreNotTightBounds: Boolean
 
+    /**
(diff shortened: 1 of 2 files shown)


juliuszsompolski (approved):
LGTM, thanks!

juliuszsompolski (approved):

PR #3692, opened 2024-09-19
Revert "[UNIFORM] Disable cleanup of files in expire snapshots via API"

amogh-jahagirdar:
Uniform

## Description
This reverts commit b51b5b45595e46748339e42dbb0792e8b485a234.
disabling the cleanExpiredFiles API technically prevents removal manifests/manifest lists and users may not be running orphan file removal, so for those users manifests/manifest list may never be cleaned up. For now we can revert this patch to preserve the original behavior of just preventing data file removal so storage can be reclaimed via manifest/manifest list cleanup during uniform commits.

This reverts to cleaning up unreachable manifests/manifest lists during background commits, not really a direct user facing change.


lzlfred (approved):
LGTM

nastra (approved):

PR #3698, opened 2024-09-20
[3.2] DeltaCatalog#createTable should respect write options (#3674)

cloud-fan:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

As of today, Delta extracts write options from table properties only for CTAS (in `StagedDeltaTableV2#commitStagedChanges`), but not for CREATE TABLE. In general, this makes sense because CREATE TABLE has no data-writing but CTAS has. However, the write options can be file system configs that we should respect because CREATE TABLE needs to access Delta logs.

This PR makes Delta CREATE TABLE to follow CTAS and also extract write options from table properties.

## How was this patch tested?

Locally tested with Unity Catalog. It's hard to write a test in Delta because the write options are not persisted but only used during table creation.

No


cloud-fan:
backport https://github.com/delta-io/delta/pull/3674

cloud-fan:
cc @prakharjain09 @tdas

tdas (approved):

PR #3711, opened 2024-09-23
[Storage] Add Javadoc for CoordinatedCommitsUtils::getCoordinatorName

sumeet-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [x] Other (Storage)

## Description
This is a simple cleanup PR for improving documentation of getCoordinatorName

1. Return an Optional<String> to indicate the method may return null if no coordinator is set
2. Rename the method from getCoordinator -> getCoordinatorName

## How was this patch tested?

Build

No

### DIFF
--- a/storage/src/main/java/io/delta/storage/commit/CoordinatedCommitsUtils.java
+++ b/storage/src/main/java/io/delta/storage/commit/CoordinatedCommitsUtils.java
@@ -68,9 +68,9 @@ public class CoordinatedCommitsUtils {
             Long commitVersion,
             UpdatedActions updatedActions) {
         boolean oldMetadataHasCoordinatedCommits =
-                !getCoordinator(updatedActions.getOldMetadata()).isEmpty();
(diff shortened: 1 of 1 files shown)


storage/src/main/java/io/delta/storage/commit/CoordinatedCommitsUtils.java
```diff
@@ -120,10 +120,17 @@ public static Path commitDirPath(Path logPath) {
         return new Path(logPath, COMMIT_SUBDIR);
     }
 
-    public static String getCoordinator(AbstractMetadata metadata) {
+    /**
+     * Retrieves the coordinator name from the provided abstract metadata.
+     * If no coordinator is set, an empty string is returned.
```
LukasRupprecht, line 125: Given that we've changed the signature to an `Optional` now, should this say "an empty optional" instead? Same in line 134 below?

LukasRupprecht (approved):
LGTM!

PR #3694, opened 2024-09-19
[Infra] [WIP] Add test report listener to delta-spark and bucket test suites by estimated runtime

scottsand-db:
#### Which Delta project/connector is this regarding?

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [X] Other (Infra)

## Description

Adds a test report listener to delta-spark to record test runtimes. This listener logs test runtime metrics and writes metrics to a csv file (per CI shard, per attempt, per jvm, per thread) and uploads it as a test artifact.

We can use the output of this job to learn how to better bucket our slowest + largest tests.

https://github.com/delta-io/delta/actions/runs/11002528077?pr=3694

image

## How was this patch tested?

Locally and GitHub CI.

## Does this PR introduce _any_ user-facing changes?

No.

PR #3712, opened 2024-09-23
[Infra] [Spark] Reduce delta-spark CI test runtime by 33 mins (1h46m to 1h13m)

scottsand-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [X] Other (Infra)

## Description

This PR reduces delta-spark CI test runtime by 33 mins. Previously the max shard duration was 1h 46 mins, and now it is 1h 13 mins.

This PR does so by the following

0. We add an extra shard
1. I used https://github.com/delta-io/delta/pull/3694 to collect some metrics about delta-spark test runtime execution.
2. I specifically identified (a) the 50 slowest test suites and (b) the average suite duration excluding those top 50 (it was 0.71 minutes)
3. I used this information to update `TestParallelization` to do smarter test suite assignment. The logic is as follows:
    - For the top 50 slowest test suites, we assign them deterministically by, in sorted descending order, assigning the suites to the shard + group (group means thread) with the lowest duration so far.
    - For the remaining tests that are not in the top 50, we assign them to a random shard, and within that shard we assign it to the group with the lowest duration so far, too
4. We also update the hash function used to me MurmurHash3 which is known to create balanced assignments in scenarios where the input strings (test names) might have similar prefixes or patterns

Note that purely adding another shard and using a better hash function does NOT yield any better results. That was attempted here: https://github.com/delta-io/delta/pull/3715.

## How was this patch tested?

GitHub Ci tests.

https://github.com/delta-io/delta/actions/runs/11004181545?pr=3712

image

No.

### DIFF
--- a/project/TestParallelization.scala
+++ b/project/TestParallelization.scala
@@ -3,8 +3,6 @@ import scala.util.hashing.MurmurHash3
 import sbt.Keys._
 import sbt._
 
-// scalastyle:off println
-
 object TestParallelization {
 
   lazy val numShardsOpt = sys.env.get("NUM_SHARDS").map(_.toInt)
@@ -144,12 +142,14 @@ object TestParallelization {
     override def add(testDefinition: TestDefinition): GroupingStrategy = {
       val testSuiteName = testDefinition.name
       val isHighDurationTest = TOP_50_HIGH_DURATION_TEST_SUITES.exists(_._1 == testSuiteName)
-      val highDurationTestGroupIndex =
-        highDurationTestAssignment.indexWhere(_.contains(testSuiteName))
 
       if (isHighDurationTest) {
+        val highDurationTestGroupIndex =
+          highDurationTestAssignment.indexWhere(_.contains(testSuiteName))
+
         if (highDurationTestGroupIndex >= 0) {
-          // Case 1: this is a high duration test that belongs to this shard. Assign it.
+          // Case 1: this is a high duration test that was pre-computed in the optimal assignment to
+          // belong to this shard. Assign it.
           val duration = TOP_50_HIGH_DURATION_TEST_SUITES.find(_._1 == testSuiteName).get._2
 
           val currentGroup = groups(highDurationTestGroupIndex)
@@ -168,7 +168,6 @@ object TestParallelization {
         // Case 3: this is a normal test that belongs to this shard. Assign it.
 
         val minDurationGroupIndex = groupRuntimes.zipWithIndex.minBy(_._1)._2
-
         val currentGroup = groups(minDurationGroupIndex)
         val updatedGroup = currentGroup.withTests(currentGroup.tests :+ testDefinition)
         groups(minDurationGroupIndex) = updatedGroup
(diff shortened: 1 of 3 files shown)


project/TestParallelization.scala
```diff
@@ -91,98 +85,263 @@ object TestParallelization {
...
+   * the group with the smallest estimated runtime.
+   *
+   * @param groups The initial mapping of group indices to their respective [[sbt.Tests.Group]]
+   *               objects, which hold test definitions.
+   * @param shardId The shard ID that this instance is responsible for.
+   * @param highDurationTestAssignment Precomputed assignments of high-duration test suites to
+   *                                   specific groups within the shard.
+   * @param groupRuntimes Array holding the current total runtime for each group within the shard.
+   */
+  class GreedyHashStrategy private(
```
scottsand-db, line 125: Rename this. It no longer uses Hash

scottsand-db:
2761e90612dd7d25540e766cb165bd7b8783a393

https://github.com/delta-io/delta/actions/runs/11002325672?pr=3712

image

project/TestParallelization.scala
```diff
@@ -1,26 +1,31 @@
+import scala.util.hashing.MurmurHash3
```
scottsand-db, line 1: don't use this; just use scala.util.random.nextInt for a more even distribution

tdas: can we use a fixed seed to make it deterministic? otherwise it might be hard to track stuff when trying to optimize further.

scottsand-db: Upon reflection, I'm actually satisfied with `scala.util.hashing.MurmurHash3`. It's deterministic and has better even distribution properties than normal hashcode

project/TestParallelization.scala
```diff
@@ -91,98 +89,263 @@ object TestParallelization {
...
+        s"Assigned shard ID $shardId is not between 0 and ${NUM_SHARDS - 1} inclusive")
+    }
 
     lazy val testGroups = groups.values.toList
-    val groupCount = groups.size
 
     override def add(testDefinition: TestDefinition): GroupingStrategy = {
+      val testSuiteName = testDefinition.name
+      val isHighDurationTest = TOP_50_HIGH_DURATION_TEST_SUITES.exists(_._1 == testSuiteName)
+      val highDurationTestGroupIndex =
```
scottsand-db, line 147: declare this variable lower

project/TestParallelization.scala
```diff
@@ -91,98 +89,263 @@ object TestParallelization {
...
+      ("org.apache.spark.sql.delta.CheckpointsWithCoordinatedCommitsBatch2Suite", 3.19),
+      ("org.apache.spark.sql.delta.DeltaSourceIdColumnMappingSuite", 3.18),
+      ("org.apache.spark.sql.delta.rowid.RowTrackingMergeCDFDVSuite", 3.18),
+      ("org.apache.spark.sql.delta.rowid.UpdateWithRowTrackingTableFeatureCDCSuite", 3.12),
+      ("org.apache.spark.sql.delta.UpdateSQLWithDeletionVectorsAndPredicatePushdownSuite", 3.01),
+      ("org.apache.spark.sql.delta.rowid.RowTrackingMergeDVSuite", 2.97)
+    )
+
+    /**
+     * Generate the optimal test assignment across shards and groups for high duration test suites.
```
scottsand-db, line 261: add method doc explaining what this does

scottsand-db:
f27edf23bdaa49e151e81304e12e0271e1d0382e

https://github.com/delta-io/delta/actions/runs/11004181545?pr=3712

image

project/TestParallelization.scala
```diff
@@ -91,98 +89,263 @@ object TestParallelization {
...
 
-  object SimpleHashStrategy {
+  object MinShardGroupDurationStrategy {
+
+    val NUM_SHARDS = numShardsOpt.get
+
+    val AVG_TEST_SUITE_DURATION_EXCLUDING_SLOWEST_50 = 0.71
+
+    /** 50 slowest test suites and their durations. */
+    val TOP_50_HIGH_DURATION_TEST_SUITES: List[(String, Double)] = List(
```
scottsand-db, line 207: talk about how to get this, how to generate this

vkorukanti (approved):

PR #3714, opened 2024-09-23
[Spark][ICT] Make CDCReader.changesToDF aware of InCommitTimestamps

dhruvarya-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Context: Currently CDCReader.changesToDF relies on DeltaHistoryManager.getCommits for getting a list of commits and their timestamps. Since DeltaHistoryManager.getCommits is not aware of InCommitTimestamps and Coordinated Commits, it will either return the wrong timestamp or no timestamp at all for certain commits. 

This PR updates CDCReader.changesToDF so that it only relies on DeltaHistoryManager.getCommits for non-ICT commits. The rest of CDCReader.changesToDF relies on the output of deltaLog.getChanges which is already Coordinated Commit-aware. The function also already extracts the `CommitInfo` for all of these commits, which we reuse to get the In-Commit Timestamp for relevant commits. Since the actions were already being read in the function, this PR does not add any additional IO.
This PR also updates `DeltaSource` so that it propagates `CommitInfo` actions to `CDCReader.changesToDF`. These `CommitInfo` actions are only used for InCommitTimestamps are later filtered out.

## How was this patch tested?

Added a Coordinated Commit variant of DeltaCDCScalaSuite with a batch size of 10.
New test cases in InCommitTimestampSuite. More tests coming up.

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
@@ -191,7 +191,7 @@ class DeltaHistoryManager(
     if (end - start > 2 * maxKeysPerList) {
       parallelSearch(time, start, end)
     } else {
-      val commits = getCommits(
+      val commits = getCommitsWithNonIctTimestamps(
         deltaLog.store,
         deltaLog.logPath,
         start,
@@ -599,9 +599,11 @@ object DeltaHistoryManager extends DeltaLogging {
    * Returns the commit version and timestamps of all commits in `[start, end)`. If `end` is not
    * specified, will return all commits that exist after `start`. Will guarantee that the commits
    * returned will have both monotonically increasing versions as well as timestamps.
-   * Exposed for tests.
+   * Note that this function will return non-ICT timestamps even for commits where
+   * InCommitTimestamps are enabled. The caller is responsible for ensuring that the appropriate
+   * timestamps are used.
    */
-  private[delta] def getCommits(
+  private[delta] def getCommitsWithNonIctTimestamps(
       logStore: LogStore,
       logPath: Path,
       start: Long,
@@ -688,7 +690,7 @@ object DeltaHistoryManager extends DeltaLogging {
       val logStore = LogStore(SparkEnv.get.conf, conf.value)
       val basePath = new Path(logPath)
       startVersions.map { startVersion =>
-        val commits = getCommits(
+        val commits = getCommitsWithNonIctTimestamps(
           logStore,
           basePath,
           startVersion,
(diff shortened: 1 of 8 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/commands/cdc/CDCReader.scala
```diff
@@ -635,6 +638,18 @@ trait CDCReaderImpl extends DeltaLogging {
...
           case _ => // do nothing
         }
+        // If the commit has an In-Commit Timestamp, we should use that as the commit timestamp.
+        // Note that it is technically possible for a commit range to begin with ICT commits
+        // followed by non-ICT commits, and end with ICT commits again. Ideally, for these commits
+        // we should use the file modification time for the first two ranges. However, this
+        // scenario is an edge case not worth optimizing for.
+        val ts = commitInfo
+          .flatMap(_.inCommitTimestamp)
+          .map{ ict => new Timestamp(ict) }
```
prakharjain09, line 648:
```suggestion
          .map(ict => new Timestamp(ict))
```

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -599,9 +599,11 @@ object DeltaHistoryManager extends DeltaLogging {
    * Returns the commit version and timestamps of all commits in `[start, end)`. If `end` is not
    * specified, will return all commits that exist after `start`. Will guarantee that the commits
    * returned will have both monotonically increasing versions as well as timestamps.
-   * Exposed for tests.
+   * Note that this function will return non-ICT timestamps even for commits where
+   * InCommitTimestamps are enabled. The caller is responsible for ensuring that the appropriate
+   * timestamps are used.
    */
-  private[delta] def getCommits(
+  private[delta] def getCommitsWithNonICTTimestamps(
```
prakharjain09, line 606:
```suggestion
  private[delta] def getCommitsWithNonICTTimestamps(
```

spark/src/test/scala/org/apache/spark/sql/delta/InCommitTimestampSuite.scala
```diff
@@ -58,6 +61,17 @@ class InCommitTimestampSuite
...
 
+  private def getFileModificationTimesMap(
+      deltaLog: DeltaLog, start: Long, end: Long): Map[Long, Long] = {
+    deltaLog.store.listFrom(
+        FileNames.listingPrefix(deltaLog.logPath, start), deltaLog.newDeltaHadoopConf())
+      .collect {
+        case FileNames.DeltaFile(fs, v) =>
+          v -> fs.getModificationTime
+      }.takeWhile(_._1 <= end)
+      .toMap
```
prakharjain09, line 72:
```suggestion
      .collect { case FileNames.DeltaFile(fs, v) => v -> fs.getModificationTime }
      .takeWhile(_._1 <= end)
```

dhruvarya-db: I have updated the PR with all of the suggested changes

prakharjain09 (approved):

PR #3728, opened 2024-09-26
[Spark] Nicer error when failing to read table due to type change not support…

johanl-db:
## Description
Delta 3.2/3.3 only supports a limited subset of type changes that will become available with Delta 4.0 / Spark 4.0.
This changes improves the error returned when reading a table with an unsupported type change to tell user to upgrade to Delta 4.0 in case the type change will be supported in that version.

## How was this patch tested?
Added tests to cover the error path.

Updates error thrown on unsupported type change when reading a table.

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/delta/typewidening/TypeWideningTableFeatureSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/typewidening/TypeWideningTableFeatureSuite.scala
@@ -45,7 +45,10 @@ class TypeWideningTableFeatureSuite
     with TypeWideningDropFeatureTestMixin
     with TypeWideningTableFeatureTests
 
-trait TypeWideningTableFeatureTests extends RowTrackingTestUtils with TypeWideningTestCases {
+trait TypeWideningTableFeatureTests
+  extends RowTrackingTestUtils
+    with DeltaExcludedBySparkVersionTestMixinShims
+    with TypeWideningTestCases {
   self: QueryTest
     with TypeWideningTestMixin
     with TypeWideningDropFeatureTestMixin =>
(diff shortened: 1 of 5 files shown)


spark/src/test/scala/org/apache/spark/sql/delta/typewidening/TypeWideningTableFeatureSuite.scala
```diff
@@ -454,6 +457,43 @@ trait TypeWideningTableFeatureTests extends RowTrackingTestUtils with TypeWideni
...
+          .putString("toType", "long")
+          .putString("fromType", "int")
+          .putLong("tableVersion", 1)
+          .build()
+      )).build()
+
+    // Delta 3.2/3.3 doesn't support changing type from int->long, we manually commit that type
+    // change to simulate what Delta 4.0 could do.
+    deltaLog.withNewTransaction { txn =>
+      txn.commit(
```
xupefei, line 475: Commit a new protocol with `TypeWideningPreviewTableFeature` supported?

johanl-db: Type widening is enabled automatically by default on all tables created in this test suite: https://github.com/delta-io/delta/blob/b1e4a033ba81c57ba64a3705ded73d6a93d44497/spark/src/test/scala/org/apache/spark/sql/delta/typewidening/TypeWideningTestMixin.scala#L43
This adds the preview table feature to the table

Is that what you were pointing out?

xupefei: Yup.

xupefei (approved):
LGTM!

PR #3736, opened 2024-09-30
[Spark] Add isShallow argument to Clone Scala APIs

longvu-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
- Add the `isShallow` argument to the Clone Scala APIs (`clone`, `cloneAtVersion`, and `cloneAtTimestamp`).
- This enables users to be able to decide the type of clone they would like from the API call.

## How was this patch tested?
Added UTs.

Yes, modified the Clone Scala APIs of `DeltaTable`

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/delta/CloneTableScalaSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/CloneTableScalaSuite.scala
@@ -16,13 +16,20 @@
 
 package org.apache.spark.sql.delta
 
+import org.apache.spark.sql.delta.util.AnalysisHelper
+import org.apache.hadoop.fs.Path
+
 class CloneTableScalaSuite extends CloneTableSuiteBase
-  with DeltaColumnMappingTestUtils {
+    with AnalysisHelper
+    with DeltaColumnMappingTestUtils {
+
+  import testImplicits._
 
   // scalastyle:off argcount
   override protected def cloneTable(
       source: String,
       target: String,
+      isShallow: Boolean,
       sourceIsTable: Boolean = false,
       targetIsTable: Boolean = false,
       targetLocation: Option[String] = None,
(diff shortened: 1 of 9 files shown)


spark/src/test/scala/org/apache/spark/sql/delta/CloneTableScalaSuite.scala
```diff
@@ -38,12 +45,113 @@ class CloneTableScalaSuite extends CloneTableSuiteBase
     }
 
     if (versionAsOf.isDefined) {
-      table.cloneAtVersion(versionAsOf.get, target, isReplace, tableProperties)
+      table.cloneAtVersion(versionAsOf.get,
+        target, isShallow, replace = isReplace, tableProperties)
```
johanl-db, line 49: either use named arguments for both `isShallow` and `replace` or for none

spark/src/test/scala/org/apache/spark/sql/delta/CloneTableScalaSuite.scala
```diff
@@ -38,12 +45,113 @@ class CloneTableScalaSuite extends CloneTableSuiteBase
...
+    runAndValidateClone(source, target, isShallow, sourceVersion = Some(0)) {
+      () => {
+        sourceTbl.cloneAtVersion(0, target, isShallow)
+      }
+    }
+  }
+
+  test("deep clone not supported yet") {
+    withSourceTargetDir { (source, clone) =>
+      assert(intercept[DeltaIllegalArgumentException] {
```
johanl-db, line 76: Use `checkError` whenever you want to cover an exception

spark/src/test/scala/org/apache/spark/sql/delta/CloneTableScalaSuite.scala
```diff
@@ -38,12 +45,113 @@ class CloneTableScalaSuite extends CloneTableSuiteBase
...
+    val format = new java.text.SimpleDateFormat("yyyy-MM-dd")
+    val time = format.parse(desiredTime).getTime
+
+    val path = new Path(source + "/_delta_log/00000000000000000000.json")
+    // scalastyle:off deltahadoopconfiguration
+    val fs = path.getFileSystem(spark.sessionState.newHadoopConf())
+    // scalastyle:on deltahadoopconfiguration
+    fs.setTimes(path, time, 0)
+    if (coordinatedCommitsEnabledInTests) {
+      InCommitTimestampTestUtils.overwriteICTInDeltaFile(
```
johanl-db, line 121: You can provide a clock to the Delta log I believe and that may allow achieving the same effect.
Anyway, if this method works then no need to look further

johanl-db (approved):
Only small comments, the API looks good

PR #3740, opened 2024-09-30
[Protocol] Amend Row Tracking Protocol to explicitly require domainMetadata

c27kwan:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [x] Other (Protocol changes)

## Description
Right now, the Delta protocol for Row tracking references the DomainMetadata table feature, but does not explicitly state domainMetadata as one of its required table feature like e.g. Clusteredtable.  The code in `TableFeature.scala` accurately lists domainMetadata as one of the required table feature. 

This PR amends the Protocol to accurately reflect the state of the system and the remainder of the Row tracking proposition. Row Tracking cannot require writers to write DomainMetadata if it is not listing domainMetadata as one of its required feature. This is also to be consistent with how other table features handle listing their required table features.

## How was this patch tested?
N/A

'No'

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -1063,7 +1063,7 @@ Row Tracking is defined to be **supported** or **enabled** on a table as follows
 
 Enablement:
 - The table must be on Writer Version 7.
-- The feature `rowTracking` must exist in the table `protocol`'s `writerFeatures`.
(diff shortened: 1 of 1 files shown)


c27kwan:
@johanl-db PTAL

johanl-db (approved):
Thanks! Definitely clearer than having to dig through all the requirements to see that `domainMetadata` is needed

PR #3741, opened 2024-09-30
[Protocol RFC] Specify collation versions

c27kwan:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [x] Other (RFC protocol change)

## Description

In this PR, we amend the current RFC proposal for the Collations table feature. 

* We should not have readVersion hints as part of the table, since the choice of collation versions is a decision at the engine level. For e.g. consider a join between two tables with different collation versions.
* We also specify the Reader Requirements and Write Requirements more explicitly for the table feature.

## How was this patch tested?
N/A

'No'

### DIFF
--- a/protocol_rfcs/collated-string-type.md
+++ b/protocol_rfcs/collated-string-type.md
@@ -8,12 +8,33 @@ This protocol change adds support for collated strings. It consists of three cha
 * Domain metadata with active collation version
 
 --------
+> *** Add New Section after the [Clustered Table](#clustered-table) section***
+# Collations Table Feature
+
+To support this feature:
+* The table must have Writer Version 7. A feature name `collations` must exist in the table's `writerFeatures`.
+* The feature `domainMetadata` is required in the table protocol's `writerFeatures`. 
+
+## Reader Requirements for Collations:
+
+When Collations is supported (when the `writerFeatures` field of a table's protocol action contains `collations`), then:
+- Readers could do comparisons and sorting of strings based on the collation specified in the schema. The choice of collation and collation version is up to the reader.
+- If the collation is not specified for a string type, then the reader must use the default utf8 binary collation.
+
+## Writer Requirements for Collations:
+
+When Collations is supported (when the `writerFeatures` field of a table's protocol action contains `collations`), then:
+- Writers must write the collation identifier in the schema metadata for a column with non-default collation (utf8 binary).
+- Writer must not write the collation identifier in the schema metadata for a column with default collation (utf8 binary).
+- Writers could write per-file statistics for string columns with collations other than utf8 binary collation in `statsWithCollation`. See [Per-file Statistics](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#per-file-statistics) for more details.
+- Writers could collect per-file statistics for collation versions that are specified in the `domainMetadata` for the `collations` table feature.
+- Writers could remove a collation version from the `domainMetadata` for the `collations` table feature if the collation version is no longer present in the statistics of any AddFile.
+- If a writer adds per-file statistics for a new version of a collation, the writer should also update the DomainMetadata for the `collations` table feature to include the new collation versions that are used to collect statistics.
 
 > ***Add a new section in front of the [Primitive Types](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#primitive-types) section.***
 
-### Collations table feature
-
-Collations are a set of rules for how strings are compared. They are supported by the `collations` table feature. Collations do not affect how strings are stored. Collations are applied when comparing strings for equality or to determine the sort order of two strings. Case insensitive comparison is one example of a collation where case is ignored when string are compared for equality and the lower cased variant of a string is used to determine its sort order.
+### Collations
+Collations are a set of rules for how strings are compared. Collations do not affect how strings are stored. Collations are applied when comparing strings for equality or to determine the sort order of two strings. Case insensitive comparison is one example of a collation where case is ignored when string are compared for equality and the lower cased variant of a string is used to determine its sort order.
 
 Each string field can have a collation, which is specified in the table schema. It is also possible to store statistics per collation version. This is required because the min and max values of a column can differ based on the used collation or collation version.
 
@@ -109,15 +130,12 @@ Schema with collation information
 
 #### Collation versions
 
-The [Domain Metadata](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#domain-metadata) for the `collations` table feature contains hints which version of a collation should be used to read from the table and for which versions of a collations clients should produce statistics when writing to the table. They allow clients to choose a collation version without having to look at the statistics of all AddFiles first. Clients are allowed to ignore the hints.
+The [Domain Metadata](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#domain-metadata) for the `collations` table feature contains hints for which versions of a collations clients should produce statistics when writing to the table. They allow clients to choose a collation version without having to look at the statistics of all AddFiles first. Clients are allowed to ignore the hints.
 
 `collations` [Domain Metadata](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#domain-metadata)
 
 ```
 {
-  "readVersions": {
-    "ICU.en_US": "73"
-  },
   "writeVersions": {
     "ICU.en_US": ["72", "73"]
   }


protocol_rfcs/collated-string-type.md
```diff
@@ -8,12 +8,33 @@ This protocol change adds support for collated strings. It consists of three cha
...
+
+When Collations is supported (when the `writerFeatures` field of a table's protocol action contains `collations`), then:
+- Readers could do comparisons and sorting of strings based on the collation specified in the schema. The choice of collation and collation version is up to the reader.
+- If the collation is not specified for a string type, then the reader must use the default utf8 binary collation.
 
-### Collations table feature
+## Writer Requirements for Collations:
+
+When Collations is supported (when the `writerFeatures` field of a table's protocol action contains `collations`), then:
+- Writers must write the collation identifier in the schema metadata for a column with non-default collation (utf8 binary).
```
olaky, line 27:
```suggestion
- Writers must write the collation identifier in the schema metadata for a column with non-default collation (not utf8 binary).
```

protocol_rfcs/collated-string-type.md
```diff
@@ -8,12 +8,33 @@ This protocol change adds support for collated strings. It consists of three cha
...
+- Writer must not write the collation identifier in the schema metadata for a column with default collation (utf8 binary).
+- Writers could write per-file statistics for string columns with collations other than utf8 binary collation in `statsWithCollation`. See [Per-file Statistics](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#per-file-statistics) for more details.
+- Writers could collect per-file statistics for collation versions that are specified in the `domainMetadata` for the `collations` table feature.
```
olaky, line 30: I like the idea to keep it flexible, but does it still make sense to add what is a good default behaviour? I guess this is only using the latest version. If the "default" is to collect for all, we can quickly blow up stats size. While this is fine to be allowed, I think a hint on what is a good default to implement makes sense here.

protocol_rfcs/collated-string-type.md
```diff
@@ -8,12 +8,33 @@ This protocol change adds support for collated strings. It consists of three cha
 * Domain metadata with active collation version
 
 --------
+> *** Add New Section after the [Clustered Table](#clustered-table) section***
+# Collations Table Feature
 
-> ***Add a new section in front of the [Primitive Types](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#primitive-types) section.***
+To support this feature:
+* The table must have Writer Version 7. A feature name `collations` must exist in the table's `writerFeatures`.
+* The feature `domainMetadata` is required in the table protocol's `writerFeatures`.
```
cstavr, line 16:
```suggestion
* The table must have Writer Version 7.
* The feature `collations` must exist in the table's `writerFeatures`.
* The feature `domainMetadata` must exist in the table's `writerFeatures`.
```

protocol_rfcs/collated-string-type.md
```diff
@@ -8,12 +8,33 @@ This protocol change adds support for collated strings. It consists of three cha
...
 
-### Collations table feature
+## Writer Requirements for Collations:
+
+When Collations is supported (when the `writerFeatures` field of a table's protocol action contains `collations`), then:
+- Writers must write the collation identifier in the schema metadata for a column with non-default collation (utf8 binary).
+- Writer must not write the collation identifier in the schema metadata for a column with default collation (utf8 binary).
+- Writers could write per-file statistics for string columns with collations other than utf8 binary collation in `statsWithCollation`. See [Per-file Statistics](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#per-file-statistics) for more details.
+- Writers could collect per-file statistics for collation versions that are specified in the `domainMetadata` for the `collations` table feature.
+- Writers could remove a collation version from the `domainMetadata` for the `collations` table feature if the collation version is no longer present in the statistics of any AddFile.
```
cstavr, line 31: Why there should be no `AddFile` actions with statistics for this collation to allow removal of the collation hint from the `domainMetadata` Doesn't this cause a potential chicken-egg problem with concurrent transactions?

c27kwan: On the other hand, if any writer can freely remove the collation version hint from the table, then two different writers can overwrite each other's preferences. Consider this scenario:
* Writer A for engine 1 wants to collect stats for version 120, writer B for engine 2 wants to collect stats for version 121, and writer C for engine 3 can collect stats for version 120 and 121.
* There is no obligation for writer A to collect stats for version 121 nor for writer B to collect stats for version 120 if your writeHints are [120, 121].
* We want writer C to collect stats for both version 120 and 121 because they're both "still in use".
* Writer A should not remove version 121 from the domain metadata and Writer B should not remove version 120, because they are used by other writers.

Among writer A, B, and C, who has more decision power over what collation versions can be dropped from the DomainMetadata? I'd like to avoid a scenario where we are flip-flopping between writeHints for [120] and [121] because writer A and writer B both think the other version is not in use. 
* We could propose tie breaking with preference for newer collation versions, but this may not be reasonable if the user values legacy compatibility. I don't think the Delta protocol should define these rules.
* We could go for semantics where 1) any writer can freely remove collation versions from the domain metadata but 2) any writer _must_ update the collation versions in the domain metadata if it adds an AddFile that has statsWithCollations with a specific version. In this case, the domain metadata serves almost as a LRU cache.
* We could have this be inconsistent (i.e. the version hints can have versions not present in the AddFile stats and the AddFiles can have stats that are not present in the version hints) and not require anything else, but then what is the purpose of the version hints in this case? What guarantees do the version hints provides? What work does it prevent the writers from doing? In this scenario, I would argue that we should just remove the hints from the format. The decision of what collation versions to collect stats for is completely up to the underlying engine.

protocol_rfcs/collated-string-type.md
```diff
@@ -8,12 +8,33 @@ This protocol change adds support for collated strings. It consists of three cha
...
+## Reader Requirements for Collations:
+
+When Collations is supported (when the `writerFeatures` field of a table's protocol action contains `collations`), then:
+- Readers could do comparisons and sorting of strings based on the collation specified in the schema. The choice of collation and collation version is up to the reader.
+- If the collation is not specified for a string type, then the reader must use the default utf8 binary collation.
 
-### Collations table feature
+## Writer Requirements for Collations:
+
+When Collations is supported (when the `writerFeatures` field of a table's protocol action contains `collations`), then:
```
olaky, line 26: Do we require DomainMetadata to have at least one version that writers have to collect stats for and that readers can use?

protocol_rfcs/collated-string-type.md
olaky: I think the scenario that has to be legal is: There is version A, then we add B and collect stats for A and B, after backfilling stats for B we remove version A (also from AddFiles to save space).

Now what we want to forbid is difficult indeed. One thing I can see is that writers must not write stats for a version that is not in the DomainMetadata.

protocol_rfcs/collated-string-type.md
c27kwan: I don't think we need to, because all Delta file level stats are always optional. If stats are not written, then it's not relevant to add it to the domain metadata. The purpose of the domain metadata is to inform what versions we already have stats for, as a shortcut from looking through all the AddFiles stats. It's currently a "should" and not a "must":
>  If a writer adds per-file statistics for a new version of a collation, the writer should also update the DomainMetadata for the `collations` table feature to include the new collation versions that are used to collect statistics.

protocol_rfcs/collated-string-type.md
c27kwan: We don't need to be too prescriptive here, as I can see good reasons to want to collect more than 1 version at a time. For an engine that is considering ICU library upgrade, they may collect statistics for multiple versions at a time so that there is automatically file skipping support upon migration, without losing file skipping for the present system. I think it's up to the engine to decide what is relevant.

I think that realistically there shouldn't be too many versions that are in use at the same time. If there are many versions, it's not necessarily a bad thing to support statistics collection on all of them -- so that there is always file skipping support no matter the reader/writer engine.

protocol_rfcs/collated-string-type.md
c27kwan: Thinking about this more, I don't think there is a chicken and egg problem. The concurrent txn that is adding new AddFiles with the version being removed SHOULD add the version to the DomainMetadata. In ConflictChecker, the old version will get added back due to this clause:
> - If a writer adds per-file statistics for a new version of a collation, the writer should also update the DomainMetadata for the `collations` table feature to include the new collation versions that are used to collect statistics.

I do see the point in removing the restriction that the AddFiles must not contain the stats. We should not require a rewrite of all AddFiles stats, which is expensive. When updating collations version, it's better to just leave the stats as is. I've updated it to:
> - Writers could remove a collation version from the `domainMetadata` for the `collations` table feature if stats collection for the collation version is no longer desired.

protocol_rfcs/collated-string-type.md
```diff
@@ -100,15 +122,12 @@ Schema with collation information
 
 #### Collation versions
 
-The [Domain Metadata](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#domain-metadata) for the `collations` table feature contains hints which version of a collation should be used to read from the table and for which versions of a collations clients should produce statistics when writing to the table. They allow clients to choose a collation version without having to look at the statistics of all AddFiles first. Clients are allowed to ignore the hints.
+The [Domain Metadata](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#domain-metadata) for the `collations` table feature contains hints for which versions of a collations clients should produce statistics when writing to the table. They allow clients to choose a collation version without having to look at the statistics of all AddFiles first. Clients are allowed to ignore the hints.
```
cstavr:
```suggestion
The [Domain Metadata](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#domain-metadata) for the `collations` table feature contains hints for which versions of a collations clients should produce statistics when writing to the table. The hints allow clients to choose a collation version without having to look at the statistics of all AddFiles first. Clients are allowed to ignore the hints.
```

protocol_rfcs/collated-string-type.md
```diff
@@ -8,12 +8,34 @@ This protocol change adds support for collated strings. It consists of three cha
...
-> ***Add a new section in front of the [Primitive Types](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#primitive-types) section.***
+To support this feature:
+* The table must have Writer Version 7. 
+* The feature `collations` must exist in the table's `writerFeatures`.
+* The feature `domainMetadata` must exist in the table's `writerFeatures`. 
+
+## Reader Requirements for Collations:
+
+When Collations is supported (when the `writerFeatures` field of a table's protocol action contains `collations`), then:
+- Readers could do comparisons and sorting of strings based on the collation specified in the schema. The choice of collation and collation version is up to the reader.
```
cstavr: What does "The choice of collation and collation version is up to the reader." mean? Readers should respect the schema's collation, right? Or you mean up to the writer?

Also, should we specify that readers can use statistics only if the version matches?

protocol_rfcs/collated-string-type.md
```diff
@@ -8,12 +8,34 @@ This protocol change adds support for collated strings. It consists of three cha
...
+- If the collation is not specified for a string type, then the reader must use the default utf8 binary collation.
 
-### Collations table feature
+## Writer Requirements for Collations:
+
+When Collations is supported (when the `writerFeatures` field of a table's protocol action contains `collations`), then:
+- Writers must write the collation identifier in the schema metadata for a column with non-default collation, i.e., any collation that is not utf8 binary.
+- Writer must not write the collation identifier in the schema metadata for a column with default collation (utf8 binary).
+- Writers could write per-file statistics for string columns with collations other than utf8 binary collation in `statsWithCollation`. See [Per-file Statistics](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#per-file-statistics) for more details.
+- Writers could collect per-file statistics for collation versions that are specified in the `domainMetadata` for the `collations` table feature.
```
cstavr: What is the difference from the previous point?

protocol_rfcs/collated-string-type.md
olaky: I agree that we could more explicit that the engine can apply collations freely to the data, but the requirements are when the engine wants to use statistics for collated columns. Maybe we say that collations can be freely applied except to the min and max values in the statistics?

protocol_rfcs/collated-string-type.md
```diff
@@ -8,12 +8,34 @@ This protocol change adds support for collated strings. It consists of three cha
...
+## Reader Requirements for Collations:
+
+When Collations is supported (when the `writerFeatures` field of a table's protocol action contains `collations`), then:
+- Readers could do comparisons and sorting of strings based on the collation specified in the schema. The choice of collation and collation version is up to the reader.
+- If the collation is not specified for a string type, then the reader must use the default utf8 binary collation.
 
-### Collations table feature
+## Writer Requirements for Collations:
+
+When Collations is supported (when the `writerFeatures` field of a table's protocol action contains `collations`), then:
```
olaky:
```suggestion
When Collations are supported (when the `writerFeatures` field of a table's protocol action contains `collations`), then:
```

protocol_rfcs/collated-string-type.md
c27kwan: I was trying to clarify that the table does not dictate the collation and that it is actually an engine responsibility. I removed it from the reader requirements and put it under the description of Collations.
> The column level collation defined may not always be applied. Engines are responsible for choosing what collation to apply on a query. An engine may apply a different collation than the schema collation based on the engine's collation precedence rules.

protocol_rfcs/collated-string-type.md
c27kwan: I can remove it. I think it's understood that all versions listed in the domainMetadata are optional since stats are optional in general.

protocol_rfcs/collated-string-type.md
```diff
@@ -8,19 +8,43 @@ This protocol change adds support for collated strings. It consists of three cha
...
 
 Each string field can have a collation, which is specified in the table schema. It is also possible to store statistics per collation version. This is required because the min and max values of a column can differ based on the used collation or collation version.
 
 By default, all strings are collated using binary collation. That means that strings compare equal if their binary UTF8 encoded representations are equal. The binary UTF8 encoded representation is also used to sort them. Note that in Delta all strings are encoded in UTF8.
 
 The `collations` table feature is a writer only feature and allows clients that do not support collations to read the table using UTF8 binary collation. To support the table feature clients must preserve collations when they change the schema. Collecting collated statistics is optional and it is valid to store UTF8 binary collated statistics for fields with a collation other than UTF8 binary.
 
+The column level collation defined may not always be applied. Engines are responsible for choosing what collation to apply on a query. An engine may apply a different collation than the schema collation based on the engine's collation precedence rules.
```
olaky: Here we could mention that this does not apply to file skipping?

protocol_rfcs/collated-string-type.md
```diff
@@ -8,19 +8,43 @@ This protocol change adds support for collated strings. It consists of three cha
...
+## Writer Requirements for Collations:
+
+When Collations is supported (when the `writerFeatures` field of a table's protocol action contains `collations`), then:
+- Writers must write the collation identifier in the schema metadata for a column with non-default collation, i.e., any collation that is not utf8 binary.
+- Writer must not write the collation identifier in the schema metadata for a column with default collation (utf8 binary).
+- Writers could write per-file statistics for string columns with collations other than utf8 binary collation in `statsWithCollation`. See [Per-file Statistics](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#per-file-statistics) for more details.
+- If a writer adds per-file statistics for a new version of a collation, the writer should also update the `domainMetadata` for the `collations` table feature to include the new collation versions that are used to collect statistics.
+- Writers could remove a collation version from the `domainMetadata` for the `collations` table feature if stats collection for the collation version is no longer desired.
```
olaky: We could mention that writers are allowed to remove / replace statistics for example during a version upgrade?

c27kwan: I've added a sentence to clarify. Although I think it might be better in the description instead of the Writers requirement.

protocol_rfcs/collated-string-type.md
```diff
@@ -8,19 +8,44 @@ This protocol change adds support for collated strings. It consists of three cha
...
+* The feature `domainMetadata` must exist in the table's `writerFeatures`. 
+
+## Reader Requirements for Collations:
 
-### Collations table feature
+When Collations are supported (when the `writerFeatures` field of a table's protocol action contains `collations`), then:
+- Readers could do comparisons and sorting of strings based on the collation specified in the schema. 
+- If the collation is not specified for a string type, then the reader must use the default utf8 binary collation.
+- Readers must only use statistics for a collation if the version matches.
+- Readers must only do file skipping based on statistics from a collation if the collation is applied for the read. For example, when reading a string column using the collation ICU.en_US.72, the reader must not use file skipping statistics from the collation spark.UTF8_LCASE.75.
```
bart-samwel: "the version" -- the version of what matches what? I think something like "if the version of the collation used by the reader matches the version of the collation for which the statistics were collected".

This line should probably come after the statistics version. Or maybe the two lines should be combined?

I also think it might be a bit more subtle w.r.t. what it means if "the collation is applied for the read". I think the key point is that the _comparison operator_ that is used to push down the expression is specified to use a specific collation. E.g. myCol = CAST("foo" AS STRING COLLATE "en_US") ends up using the collation "en_US" for the comparison because that's what the language rules decided, and that decision is recorded in the _comparison operator_.  

```suggestion
- Readers must only do file skipping based on column statistics for a collation if the filter operator used for the data skipping is specified to treat the column as having that same collation. For example, when filtering a string column using the string equality comparison operator that is configured with the collation ICU.en_US.72, the reader must not use file skipping statistics from the collation spark.UTF8_LCASE.75. It should also not use ICU.en_US.69 because the collation version number does not match.```

File: protocol_rfcs/collated-string-type.md [hunk 13]
```diff
@@ -8,19 +8,44 @@ This protocol change adds support for collated strings. It consists of three cha
...
 
-Collations are a set of rules for how strings are compared. They are supported by the `collations` table feature. Collations do not affect how strings are stored. Collations are applied when comparing strings for equality or to determine the sort order of two strings. Case insensitive comparison is one example of a collation where case is ignored when string are compared for equality and the lower cased variant of a string is used to determine its sort order.
+## Writer Requirements for Collations:
+
+When Collations are supported (when the `writerFeatures` field of a table's protocol action contains `collations`), then:
+- Writers must write the collation identifier in the schema metadata for a column with non-default collation, i.e., any collation that is not utf8 binary.
```
bart-samwel: "utf8 binary" -> should be use the technical name for this?

protocol_rfcs/collated-string-type.md
```diff
@@ -8,19 +8,44 @@ This protocol change adds support for collated strings. It consists of three cha
...
+- Writer must not write the collation identifier in the schema metadata for a column with default collation (utf8 binary).
```
bart-samwel:
```suggestion
- Writers must not write the collation identifier in the schema metadata for a column with default collation (utf8 binary).
```

protocol_rfcs/collated-string-type.md
```diff
@@ -8,19 +8,44 @@ This protocol change adds support for collated strings. It consists of three cha
...
+- Writers could write per-file statistics for string columns with collations other than utf8 binary collation in `statsWithCollation`. See [Per-file Statistics](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#per-file-statistics) for more details.
```
bart-samwel: Is this in the spec somewhere?

protocol_rfcs/collated-string-type.md
```diff
@@ -8,19 +8,44 @@ This protocol change adds support for collated strings. It consists of three cha
...
 
 By default, all strings are collated using binary collation. That means that strings compare equal if their binary UTF8 encoded representations are equal. The binary UTF8 encoded representation is also used to sort them. Note that in Delta all strings are encoded in UTF8.
 
 The `collations` table feature is a writer only feature and allows clients that do not support collations to read the table using UTF8 binary collation. To support the table feature clients must preserve collations when they change the schema. Collecting collated statistics is optional and it is valid to store UTF8 binary collated statistics for fields with a collation other than UTF8 binary.
 
+The column level collation defined may not always be applied. Engines are responsible for choosing what collation to apply on a query. An engine may apply a different collation than the schema collation based on the engine's collation precedence rules. However, an engine must never use file statistics for file skipping from a different collation than the collation applied for the read.
```
bart-samwel:
```suggestion
The column level collation indicates the default collation that readers should use to operate on a column. However, readers are responsible for choosing what collation to actually apply on operations. An engine may apply a different collation than the schema collation based on the engine's collation precedence rules. However, an engine must take care to only use column statistics for file skipping from a collation that is identical to the one specified in the filtering operation in all aspects, including the collation version.
```

protocol_rfcs/collated-string-type.md
c27kwan: Not yet, it's part of the proposal in this RFC. On line 132 of the old diff, we can see the section where we request amending the Per-file Statistics portion.

protocol_rfcs/collated-string-type.md
c27kwan: Good point. I'll amend it with something along the lines of "comparison of binary representation under utf-8 encoding"?

protocol_rfcs/collated-string-type.md
c27kwan: That's a good rewrite suggestion. Much more clear. :)

PR #3742, opened 2024-09-30
[Spark] Fix DeltaColumnMappingSuite."column mapping batch scan should detect physical name changes" test

juliuszsompolski:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

After https://github.com/apache/spark/pull/48211 this test needs to use a separate Dataframe object for the failing / succeeding test case, because LazyTry will cache the error of the failing case.

## How was this patch tested?

Running the test on Spark master.

No.


allisonport-db (approved):
LGTM

PR #3747, opened 2024-10-02
[Spark] Relax protocol reader feature requirements to allow backward compatibility

andreaschat-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This is a followup of https://github.com/delta-io/delta/commit/920f185a04382ff466e47e75240dfda48efe40d3. We relax the protocol reader feature requirements to allow backward compatibility with older writers. This is because before https://github.com/delta-io/delta/commit/920f185a04382ff466e47e75240dfda48efe40d3 was possible to serialize invalid (2, 7) protocols without the reader features set.

## How was this patch tested?

Added new UTs.

No.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala
@@ -139,10 +139,15 @@ case class Protocol private (
   extends Action
   with AbstractProtocol
   with TableFeatureSupport {
-  // Correctness check
-  // Reader and writer versions must match the status of reader and writer features
+  // Correctness check.
+  // Reader and writer versions must match the status of reader and writer features.
+  // For the reader features requirement we exclude the column mapping reader version (2)
+  // to allow backward compatibility with older writers. The correct behavior is when
+  // column mapping exists it should be included in the reader features. However, there used
+  // to be a bug where the reader features would not be set in that case.
   require(
-    (supportsReaderFeatures || canSupportColumnMappingFeature) == readerFeatures.isDefined,
+    supportsReaderFeatures == readerFeatures.isDefined ||
+      minReaderVersion == ColumnMappingTableFeature.minReaderVersion,
     "Mismatched minReaderVersion and readerFeatures.")
   require(
     supportsWriterFeatures == writerFeatures.isDefined,
--- a/spark/src/test/scala/org/apache/spark/sql/delta/DeltaProtocolVersionSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/DeltaProtocolVersionSuite.scala
@@ -2397,6 +2397,60 @@ trait DeltaProtocolVersionSuiteBase extends QueryTest
     }
   }
 
+  test("Can construct a protocol with column mapping only in writer features") {
+    // Column mapping should appear in both reader and writer features. However,
+    // we allow the construction of invalid protocols such as the one below to allow backward
+    // compatibility. This is to protect against the case a table was created by an older writer
+    // (that somehow serialized the protocol without column mapping in the reader features set)
+    // and then attempted to read the table with a newer reader.
+    Protocol(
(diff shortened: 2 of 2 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala
```diff
@@ -139,10 +139,15 @@ case class Protocol private (
   extends Action
   with AbstractProtocol
   with TableFeatureSupport {
-  // Correctness check
-  // Reader and writer versions must match the status of reader and writer features
+  // Correctness check.
+  // Reader and writer versions must match the status of reader and writer features.
+  // For the reader features requirement we exclude the column mapping reader version (2)
```
tomvanbussel, line 144: This comment describes the case when the reader version is 3, but the reader features are not defined. The code below on the other hand handles the case in which the reader version is 2.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaProtocolVersionSuite.scala
```diff
@@ -2397,6 +2397,60 @@ trait DeltaProtocolVersionSuiteBase extends QueryTest
     }
   }
 
+  test("Can construct a protocol with column mapping only in writer features") {
+    // Column mapping should appear in both reader and writer features. However,
+    // we allow the construction of invalid protocols such as the one below to allow backward
```
tomvanbussel, line 2402: Why would the one below be invalid? Is it because `readerFeatures` should have been `None`?

spark/src/test/scala/org/apache/spark/sql/delta/DeltaProtocolVersionSuite.scala
```diff
@@ -2397,6 +2397,60 @@ trait DeltaProtocolVersionSuiteBase extends QueryTest
...
+      readerFeatures = Some(Set.empty),
+      writerFeatures = Some(Set(ColumnMappingTableFeature.name)))
+
+    Protocol(
+      minReaderVersion = 2,
+      minWriterVersion = 7,
+      readerFeatures = Some(Set.empty),
+      writerFeatures = Some(Set(ColumnMappingTableFeature.name, TestWriterFeature.name)))
+
+    // Valid protocol are also allowed.
```
tomvanbussel, line 2418: Surprised that this is valid. Shouldn't the reader features be missing in this case, because the reader protocol version does not support reader features?

tomvanbussel (approved):
Left a few questions, code changes LGTM otherwise.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaProtocolVersionSuite.scala
tomvanbussel: Yes, but wouldn't reader version 2 already indicate that this table uses column mapping? And on the other hand reader features are not guaranteed to be understood when the reader version is 2.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaProtocolVersionSuite.scala
tomvanbussel: But the reader version below is 2. In that case the readers features should not be set, right?

andreaschat-db: Both (2, 5) and (2, 7, columnMapping, ...) are valid protocols. The first is in legacy form and the later in the table features form.

Once we are in the table features territory all features need to be explicitly set. In a sense, yes (2, 7) implies there is column mapping only because column mapping is the only reader feature with 2. However, the semantics of the table feature protocols is to explicitly state all supported features.

I will add an extra tests to also include the none case.

PR #3751, opened 2024-10-04
make the iceberg conversion on replace/overwrite uniform table asynchronous

lzlfred:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

There is a problem today that the iceberg conversion on replace/overwrite uniform table is sync, so if user do concurrent replace, many will fail because there are very likely conversion of higher versions succeeded/committed. Making the conversion async/one-at-a-time will be a good solution to stick with existing behavior for all other kinds of commits. If there is a conflict, thats because a higher version of conversion committed and we should not worry about the conflict.

## How was this patch tested?

manual test

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/commands/CreateDeltaTableCommand.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/commands/CreateDeltaTableCommand.scala
@@ -240,11 +240,13 @@ case class CreateDeltaTableCommand(
     updateCatalog(sparkSession, tableWithLocation, postCommitSnapshot, didNotChangeMetadata)
 
 
-    if (UniversalFormat.icebergEnabled(postCommitSnapshot.metadata)) {
+    if (UniversalFormat.icebergEnabled(postCommitSnapshot.metadata) &&
+        !txnUsedForCommit.containsPostCommitHook(IcebergConverterHook)) {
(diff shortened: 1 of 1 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/commands/CreateDeltaTableCommand.scala
```diff
@@ -240,11 +240,13 @@ case class CreateDeltaTableCommand(
     updateCatalog(sparkSession, tableWithLocation, postCommitSnapshot, didNotChangeMetadata)
 
 
-    if (UniversalFormat.icebergEnabled(postCommitSnapshot.metadata)) {
+    if (UniversalFormat.icebergEnabled(postCommitSnapshot.metadata) &&
+        !txnUsedForCommit.containsPostCommitHook(IcebergConverterHook)) {
```
harperjiang, line 244: nit: why don't we use `isReplace` as a condition here?

harperjiang (approved):

spark/src/main/scala/org/apache/spark/sql/delta/commands/CreateDeltaTableCommand.scala
lzlfred: I think this way should be a little more future proof if the condition in the other place changed..

PR #1296, opened 2022-07-27
Don't enforce null checks on fields inside nullable structs

Kimahriman:
## Description
Resolves https://github.com/delta-io/delta/issues/860, (and maybe https://github.com/delta-io/delta/issues/1239 too)

Currently a non-nullable field inside a nullable struct is enforced to be not null. However, the way Spark handles nullability, a non-nullable field could be "null" if its parent is null. The non-nullable on the inner field simply means "if my parent is non-null then I will definitely be non-null". This PR turns a non-nullable field inside a nullable struct into an unenforced not null.

## How was this patch tested?
Updated existing unit tests to make sure the parent structs are non-nullable for the same current behavior, and new UTs to test the new behavior.

## Does this PR introduce _any_ user-facing changes?
Yes, non-nullable fields inside nullable structs become unenforced not nulls, thereby failing to create the schema by default without the `allowUnenforcedNotNull` setting enabled, instead of creating runtime null checks for the field.

PR #3752, opened 2024-10-04
[SPARK] Add new config to disable implicit not null invariants

### LINKED ISSUE #860: Not null invariant incorrectly fails on non-nullable field inside a nullable struct
An existing test actually tests for what I think is incorrect behavior: https://github.com/delta-io/delta/blob/master/core/src/test/scala/org/apache/spark/sql/delta/schema/InvariantEnforcementSuite.scala#L186

```
 testQuietly("reject non-nullable nested column") {
    val schema = new StructType()
      .add("top", new StructType()
        .add("key", StringType, nullable = false)
        .add("value", IntegerType))
    testBatchWriteRejection(
      NotNull(Seq("key")),
      schema,
      spark.createDataFrame(Seq(Row(Row("a", 1)), Row(Row(null, 2))).asJava, schema.asNullable),
      "top.key"
    )
    testBatchWriteRejection(
      NotNull(Seq("key")),
      schema,
      spark.createDataFrame(Seq(Row(Row("a", 1)), Row(null)).asJava, schema.asNullable),
      "top.key"
    )
  }
```

Here the `top` struct is nullable but the `key` field is not, and the current invariant checks only care about the fact that `key` is non-nullable therefore selecting that value (through a series of `GetStructField`'s), will always not be null. However, it is valid for `top` to be null, and it's more accurate to say that `key` is never null when `top` is not null I think.

So in this test case, the first one is a valid test case. `top` is not null but `key` is. However, in the second test case, `top` is null, which should be valid behavior and not throw an exception I believe.

After looking through the code I can see a few ways to make it basically skip checking `key` in this case, but it might be more ideal but more complicated to have it check `top` first, and only if that's not null, then check `key` and fail only in that case.

Kimahriman:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Resolves #860 

Adds a new config that can disable implicit not null constraints that are added for non-nullable fields. This is to get around the fact that Delta does not properly respect the struct nullability semantics of Spark. The discussion of why this is true is in the issue. There is currently no workaround for this issue, so this new config at least lets users opt-in to say "I know what I'm doing" and prevent the unwanted default behavior.

An original attempt to address this issue is https://github.com/delta-io/delta/pull/1296. This is an alternative in the hopes that _something_ can be done to fix this issue that more users are reporting, and has frustrated my team for a long time.

## How was this patch tested?

New UT

Allows users to opt-in to skipping potential erroneous not null constraints.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/constraints/Invariants.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/constraints/Invariants.scala
@@ -18,6 +18,7 @@ package org.apache.spark.sql.delta.constraints
 
 import org.apache.spark.sql.delta.DeltaErrors
 import org.apache.spark.sql.delta.schema.SchemaUtils
+import org.apache.spark.sql.delta.sources.DeltaSQLConf
 import org.apache.spark.sql.delta.util.JsonUtils
 
 import org.apache.spark.sql.SparkSession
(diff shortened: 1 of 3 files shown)


foss-contributor:
@Kimahriman, thank you for your contribution! This pull request has been inactive for over 18 months and will be closed in one week. To keep it open, please leave a comment on this pull request or contact us in the `#contributors` channel on [Delta Lake Slack](https://go.delta.io/slack). We appreciate your involvement in Delta Lake!

PR #3759, opened 2024-10-08
[Spark] Allow setting both allowTypeWidening and keepExistingTypes to true in mergeSchemas

aldenlau-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Currently, `SchemaMergingUtils.mergeSchemas` will not widen fields when both `keepExistingTypes` and `allowTypeWidening` are set to true. This change will allow types to be widened when both `keepExistingTypes` and `allowTypeWidening` are true, and in the case of non-widening type changes, the existing type will be kept. This change can be used to fall back to existing types instead of throwing a `DeltaAnalysisException` when we want to merge a schema with widened fields that may also have non-widenable type changes.

## How was this patch tested?

New unit test in `SchemaUtilsSuite`.scala

No


spark/src/test/scala/org/apache/spark/sql/delta/schema/SchemaUtilsSuite.scala
```diff
@@ -2428,6 +2428,40 @@ class SchemaUtilsSuite extends QueryTest
...
+      .add("map", MapType(IntegerType, IntegerType))
+      .add("array", ArrayType(StringType))
+      .add("nonwidened", IntegerType)
+
+    val update = new StructType()
+      .add("widened", LongType)
+      .add("struct", new StructType()
+        .add("b", IntegerType)
+        .add("a", IntegerType))
+      .add("map", MapType(LongType, StringType))
```
johanl-db, line 2448: This test fails in `Delta Spark Latest` test job because `int` -> `long` is only supported when using Spark master.
- `Delta Spark Latest` builds against Spark 3.5 and only supports byte->short->int. See https://github.com/delta-io/delta/blob/8e0b133f46f641941ad15ed8cbe7c2d1cc777a5b/spark/src/main/scala-spark-3.5/shims/TypeWideningShims.scala#L34
- `Delta Spark Master` builds against Spark master and supports all type changes. See https://github.com/delta-io/delta/blob/8e0b133f46f641941ad15ed8cbe7c2d1cc777a5b/spark/src/main/scala-spark-master/shims/TypeWideningShims.scala#L40

You'll have to either use e.g. short->int instead of int->long, or use `testSparkMasterOnly` to run the test - although the spark master build is currently broken and has been for a while so probably best to avoid the test only against spark master

johanl-db (approved):
Thanks, that behavior makes more sense that what I initially introduced

PR #3765, opened 2024-10-10
[Spark] Writing of UUID commits should not use put-if-absent semantics

sumeet-db:
- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR fixes the coordinated commits utils to not write UUID-based commit files with put-if-absent semantics. This is not necessary because we assume that UUID-based commit files are globally unique so we will never have concurrent writers attempting to write the same commit file.

DynamoDBCommitCoordinator also now uses the utils for writing backfilled files.

## How was this patch tested?

Existing tests are sufficient as this only affects how a commit is
written in the underlying storage layer but does not change any logic in
Delta Spark.

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/coordinatedcommits/CoordinatedCommitsUtils.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/coordinatedcommits/CoordinatedCommitsUtils.scala
@@ -149,27 +149,6 @@ object CoordinatedCommitsUtils extends DeltaLogging {
     }
   }
 
-  /**
-   * Write a UUID-based commit file for the specified version to the
-   * table at [[logPath]].
-   */
-  def writeCommitFile(
-      logStore: LogStore,
-      hadoopConf: Configuration,
-      logPath: Path,
-      commitVersion: Long,
-      actions: Iterator[String],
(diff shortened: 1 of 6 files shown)


dhruvarya-db (approved):
Thanks for catching this!

spark/src/main/scala/org/apache/spark/sql/delta/coordinatedcommits/CoordinatedCommitsUtils.scala
```diff
@@ -149,27 +149,6 @@ object CoordinatedCommitsUtils extends DeltaLogging {
     }
   }
 
-  /**
-   * Write a UUID-based commit file for the specified version to the
-   * table at [[logPath]].
-   */
-  def writeCommitFile(
```
prakharjain09, line 156: After this change, is there no duplication of any methods between the Java and the Scala version?

sumeet-db: Yes, no more duplication

prakharjain09 (approved):

PR #3777, opened 2024-10-17
[Protocol] Add a version checksum to the specification

dhruvarya-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [X] Other (PROTOCOL)

## Description

Adds the concept of a Version Checksum to the protocol. This version checksum can be emitted on every commit and stores important bits of information about the snapshot which can later be used to validate the integrity of the delta log.

## How was this patch tested?

N/A

N/A

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -347,7 +347,7 @@ domainMetadata | Array[`domainMetadata`] | Live [Domain Metadata](#domain-metada
 metadata | Metadata | The table [metadata](#change-metadata) at this version. | required
 protocol | Protocol | The table [protocol](#protocol-evolution) at this version. | required
 fileSizeHistogram | FileSizeHistogram | Size distribution information of files remaining after [Action Reconciliation](#action-reconciliation). See [FileSizeHistogram](#file-size-histogram-schema) for more details. | optional
-allFiles | Array[AddFile] | All live [AddFile](#add-file-and-remove-file) actions at this version. | optional
+allFiles | Array[`add`] | All live [Add File](#add-file-and-remove-file) actions at this version. | optional
 numDeletedRecordsOpt | Long | Number of records deleted through Deletion Vectors in this table version. | optional
 numDeletionVectorsOpt | Long | Number of Deletion Vectors active in this table version. | optional
 deletedRecordCountsHistogramOpt | DeletedRecordCountsHistogram | Distribution of deleted record counts across files. See [this](#deleted-record-counts-histogram-schema) section for more details. | optional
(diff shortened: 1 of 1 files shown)


prakharjain09 (approved):

prakharjain09 (approved):

PROTOCOL.md
```diff
@@ -324,19 +325,94 @@ Due to the zero-padded encoding of the files in the log, the version id of this
 
 ### Version Checksum File
 
-The Delta transaction log must remain an append-only log. To enable the detection of non-compliant modifications to Delta files, writers can optionally emit an auxiliary file with every commit, which contains important information about the state of the table as of that version. This file is referred to as the **Version Checksum** and can be used to validate the integrity of the table. A file representing the Version Checksum of version X must be named `X.crc`, where X is zero-padded to 20 digits (e.g., for version 1, it will be `00000000000000000001.crc`), and must be a direct child of the `_delta_log` directory. Additionally, the checksum must be a JSON file with the following fields:
+The Delta transaction log must remain an append-only log. To enable the detection of non-compliant modifications to Delta files, writers can optionally emit an auxiliary file with every commit, which contains important information about the state of the table as of that version. This file is referred to as the **Version Checksum** and can be used to validate the integrity of the table.
```
scovich, line 328: We should make clear: the aux file cannot land before the actual commit

scovich: update: that's covered very well later

scovich (approved):
LGTM

PROTOCOL.md
```diff
@@ -321,6 +323,98 @@ The last checkpoint file can help reduce the cost of constructing the latest sna
...
 Due to the zero-padded encoding of the files in the log, the version id of this recent checkpoint can be used on storage systems that support lexicographically-sorted, paginated directory listing to enumerate any delta files or newer checkpoints that comprise more recent versions of the table.
 
+### Version Checksum File
+
+The Delta transaction log must remain an append-only log. To enable the detection of non-compliant modifications to Delta files, writers can optionally emit an auxiliary file with every commit, which contains important information about the state of the table as of that version. This file is referred to as the **Version Checksum** and can be used to validate the integrity of the table.
+
+### Version Checksum File Schema
+
+A Version Checksum file must have the following properties:
+- Be named `{version}.crc` where `version` is zero-padded to 20 digits (e.g., `00000000000000000001.crc`)
```
Kimahriman, line 333: Love this feature that allows loading the schema without trigger a snapshot loading job, but why is `.crc` used? That means a very specific thing, and is even used by certain Hadoop file systems to represent and actual computed checksum.

image

Seems like it's going to add a lot of confusion since these are actually just JSON files with just metadata about the version

dhruvarya-db: Hey @Kimahriman , that is a valid point. Although the "crc" that we write helps with delta log validation, it is not actual CRC. Unfortunately it will be very hard to change this. It might be possible to change the extension in spec in the future but I feel that it is probably not worth breaking backwards compatibility for.

PR #3786, opened 2024-10-21
[Spark] Add `DeltaTable.addFeatureSupport` API to PySpark

xupefei:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR introduces a `DeltaTable.addFeatureSupport` API which was missing in PySpark. This API is used to add support of a table feature to a given Delta table.

## How was this patch tested?

A new test is added.

Yes. See the above `Description` section.

### DIFF
--- a/python/delta/tables.py
+++ b/python/delta/tables.py
@@ -591,7 +591,7 @@ class DeltaTable(object):
 
         See online documentation and Delta's protocol specification at PROTOCOL.md for more details.
         """
-        DeltaTable._verify_type_str(featureName, "feature")
+        DeltaTable._verify_type_str(featureName, "featureName")
         self._jdt.addFeatureSupport(featureName)
 
     @since(1.2)  # type: ignore[arg-type]
--- a/python/delta/tables.py
+++ b/python/delta/tables.py
@@ -581,7 +581,8 @@ class DeltaTable(object):
                              type(writerVersion))
         jdt.upgradeTableProtocol(readerVersion, writerVersion)
 
-    def addFeatureSupport(self, feature: str) -> None:
+    @since(3.3)  # type: ignore[arg-type]
+    def addFeatureSupport(self, featureName: str) -> None:
         """
         Modify the protocol to add a supported feature, and if the table does not support table
         features, upgrade the protocol automatically. In such a case when the provided feature is
@@ -590,11 +591,8 @@ class DeltaTable(object):
 
         See online documentation and Delta's protocol specification at PROTOCOL.md for more details.
         """
-        jdt = self._jdt
-        if not isinstance(feature, str):
-            raise ValueError("The feature needs to be a string but got '%s'." % type(feature))
-        jdt.addFeatureSupport(feature)
-
+        DeltaTable._verify_type_str(featureName, "feature")
+        self._jdt.addFeatureSupport(featureName)
 
     @since(1.2)  # type: ignore[arg-type]
     def restoreToVersion(self, version: int) -> DataFrame:
--- a/python/delta/tables.py
+++ b/python/delta/tables.py
@@ -581,6 +581,21 @@ class DeltaTable(object):
                              type(writerVersion))
         jdt.upgradeTableProtocol(readerVersion, writerVersion)
 
+    def addFeatureSupport(self, feature: str) -> None:
+        """
+        Modify the protocol to add a supported feature, and if the table does not support table
+        features, upgrade the protocol automatically. In such a case when the provided feature is
+        writer-only, the table's writer version will be upgraded to `7`, and when the provided
+        feature is reader-writer, both reader and writer versions will be upgraded, to `(3, 7)`.
+
+        See online documentation and Delta's protocol specification at PROTOCOL.md for more details.
+        """
+        jdt = self._jdt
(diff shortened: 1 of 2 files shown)


xupefei:
Hey @longvu-db, it is you who found this API missing. Could you review this PR?

python/delta/tables.py
```diff
@@ -581,6 +581,21 @@ def upgradeTableProtocol(self, readerVersion: int, writerVersion: int) -> None:
                              type(writerVersion))
         jdt.upgradeTableProtocol(readerVersion, writerVersion)
 
+    def addFeatureSupport(self, feature: str) -> None:
```
longvu-db, line 584: Could we add `@since(3.3)  # type: ignore[arg-type]` like other Python APIs?

python/delta/tables.py
```diff
@@ -581,6 +581,21 @@ def upgradeTableProtocol(self, readerVersion: int, writerVersion: int) -> None:
...
+        """
+        Modify the protocol to add a supported feature, and if the table does not support table
+        features, upgrade the protocol automatically. In such a case when the provided feature is
+        writer-only, the table's writer version will be upgraded to `7`, and when the provided
+        feature is reader-writer, both reader and writer versions will be upgraded, to `(3, 7)`.
+
+        See online documentation and Delta's protocol specification at PROTOCOL.md for more details.
+        """
+        jdt = self._jdt
+        if not isinstance(feature, str):
```
longvu-db, line 594: Using [_verify_type_str](https://github.com/delta-io/delta/blob/master/python/delta/tables.py#L649) will be strong and cleaner

python/delta/tables.py
longvu-db, line 584:
```suggestion
    def addFeatureSupport(self, featureName: str) -> None:
```

Similar to the case of forName ([Scala](https://github.com/delta-io/delta/blob/8442bc60da5ac45d13e76fb0cd59aacfc0c447ad/spark/src/main/scala/io/delta/tables/DeltaTable.scala#L1033), [Python](https://github.com/delta-io/delta/blob/master/python/delta/tables.py#L393)), we should have the argument's names to be identical as the [Scala counterpart](https://github.com/delta-io/delta/blob/8442bc60da5ac45d13e76fb0cd59aacfc0c447ad/spark/src/main/scala/io/delta/tables/DeltaTable.scala#L555)

python/delta/tables.py
```diff
@@ -581,6 +581,21 @@ def upgradeTableProtocol(self, readerVersion: int, writerVersion: int) -> None:
...
+            raise ValueError("The feature needs to be a string but got '%s'." % type(feature))
+        jdt.addFeatureSupport(feature)
```
longvu-db, line 596: Just a nit, either we do self._jdt.addFeatureSupport(featureName) or move `jdt = self._jdt` as close as possible to the callsite, i.e. after `_verify_type_str`

python/delta/tables.py
```diff
@@ -581,6 +581,21 @@ def upgradeTableProtocol(self, readerVersion: int, writerVersion: int) -> None:
...
+
+
```
longvu-db, line 598:
```suggestion
```

python/delta/tests/test_deltatable.py
```diff
@@ -1187,6 +1188,41 @@ def test_protocolUpgrade(self) -> None:
         with self.assertRaisesRegex(ValueError, "writerVersion"):
             dt.upgradeTableProtocol(1, {})  # type: ignore[arg-type]
 
+    def test_addFeatureSupport(self) -> None:
+        try:
+            self.spark.conf.set('spark.databricks.delta.minWriterVersion', '2')
```
longvu-db, line 1193: Nit: Since we are checking `minReaderVersion` and `minWriterVersion`, we should also set `minReaderVersion` first then `minWriterVersion` instead of other way around.

longvu-db (commented):
Awesome work on ensuring API Parity @xupefei :)!

Btw, just in case you forget, could we add [Spark] to the title?

python/delta/tables.py
xupefei: Oops thanks for pointing out. Fixed.

python/delta/tables.py
```diff
@@ -581,6 +581,19 @@ def upgradeTableProtocol(self, readerVersion: int, writerVersion: int) -> None:
...
+    def addFeatureSupport(self, featureName: str) -> None:
+        """
+        Modify the protocol to add a supported feature, and if the table does not support table
+        features, upgrade the protocol automatically. In such a case when the provided feature is
+        writer-only, the table's writer version will be upgraded to `7`, and when the provided
+        feature is reader-writer, both reader and writer versions will be upgraded, to `(3, 7)`.
+
+        See online documentation and Delta's protocol specification at PROTOCOL.md for more details.
+        """
+        DeltaTable._verify_type_str(featureName, "feature")
```
longvu-db, line 594:
```suggestion
        DeltaTable._verify_type_str(featureName, "featureName")
```

xupefei: Fixed!

longvu-db (approved):
LGTM!

PR #3791, opened 2024-10-22
[Kernel] Add `Snapshot::getTimestamp` public API

scottsand-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Add a new `Snapshot::getTimestamp` public API.

## How was this patch tested?

N/A. Trivial.

Yes. Adds a new `Snapshot::getTimestamp` public API.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/Snapshot.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/Snapshot.java
@@ -30,8 +30,7 @@ import java.util.List;
 public interface Snapshot {
 
   /**
-   * Get the version of this snapshot in the table. Returns -1 if the table is empty and so has no
-   * commits.
+   * Get the version of this snapshot in the table.
    *
    * @param engine {@link Engine} instance to use in Delta Kernel.
    * @return version of this snapshot in the Delta table
@@ -50,14 +49,7 @@ public interface Snapshot {
   List<String> getPartitionColumnNames(Engine engine);
 
   /**
-   * Get the timestamp (in milliseconds since the Unix epoch) of the latest commit of this snapshot.
-   * Returns -1 if the table is empty and so has no commits.
-   *
-   * <p>When InCommitTimestampTableFeature is enabled, the timestamp is retrieved from the
-   * CommitInfo of the latest commit which can result in an IO operation.
-   *
-   * <p>For non-ICT tables, this is the same as the file modification time of the latest commit in
-   * the snapshot.
+   * Get the timestamp (in milliseconds since the Unix epoch) of the latest commit in this snapshot.
    *
    * @param engine the engine to use for IO operations
    * @return the timestamp of the latest commit
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/Snapshot.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/Snapshot.java
@@ -47,6 +47,22 @@ public interface Snapshot {
    * @return a list of partition column names, or an empty list if the table is not partitioned.
    */
   List<String> getPartitionColumnNames(Engine engine);
+
+  /**
+   * Get the timestamp (in milliseconds since the Unix epoch) of the latest commit of this snapshot.
+   * For an uninitialized snapshot, this returns -1.
+   *
+   * <p>When InCommitTimestampTableFeature is enabled, the timestamp is retrieved from the
+   * CommitInfo of the latest commit which can result in an IO operation.
+   *
+   * <p>For non-ICT tables, this is the same as the file modification time of the latest commit in
+   * the snapshot.
+   *
+   * @param engine the engine to use for IO operations
+   * @return the timestamp of the latest commit
+   */
+  long getTimestamp(Engine engine);
+
   /**
    * Get the schema of the table at this snapshot.
    *
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/Snapshot.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/Snapshot.java
@@ -30,7 +30,8 @@ import java.util.List;
 public interface Snapshot {
 
   /**
-   * Get the version of this snapshot in the table.
+   * Get the version of this snapshot in the table. Returns -1 if the table is empty and so has no
+   * commits.
    *
    * @param engine {@link Engine} instance to use in Delta Kernel.
    * @return version of this snapshot in the Delta table
@@ -50,7 +51,7 @@ public interface Snapshot {
 
   /**
    * Get the timestamp (in milliseconds since the Unix epoch) of the latest commit of this snapshot.
-   * For an uninitialized snapshot, this returns -1.
+   * Returns -1 if the table is empty and so has no commits.
    *
    * <p>When InCommitTimestampTableFeature is enabled, the timestamp is retrieved from the
    * CommitInfo of the latest commit which can result in an IO operation.
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java
@@ -75,6 +75,9 @@ public class SnapshotImpl implements Snapshot {
 
   /**
    * Get the timestamp (in milliseconds since the Unix epoch) of the latest commit in this Snapshot.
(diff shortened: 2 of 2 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java
```diff
@@ -75,14 +96,30 @@ public ScanBuilder getScanBuilder(Engine engine) {
     return new ScanBuilderImpl(dataPath, protocol, metadata, getSchema(engine), logReplay, engine);
   }
 
-  public Metadata getMetadata() {
-    return metadata;
+  ///////////////////
+  // Internal APIs //
+  ///////////////////
+
+  public Path getLogPath() {
```
scottsand-db, line 103: 1. put all getters at the top of internal APIs, with more complicated internal APIs below
2. re-ordered these getters to match the order of member field declarations

kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java
```diff
@@ -75,14 +96,30 @@ public ScanBuilder getScanBuilder(Engine engine) {
     return new ScanBuilderImpl(dataPath, protocol, metadata, getSchema(engine), logReplay, engine);
   }
 
-  public Metadata getMetadata() {
-    return metadata;
+  ///////////////////
+  // Internal APIs //
```
ebyhr, line 100: This comment seems misleading. Even if this line says "Internal", 3rd party developers may call the subsequent method as they are public.

kernel/kernel-api/src/main/java/io/delta/kernel/Snapshot.java
```diff
@@ -36,6 +36,21 @@ public interface Snapshot {
    */
   long getVersion(Engine engine);
 
+  /**
+   * Get the timestamp (in milliseconds since the Unix epoch) of the latest commit of this snapshot.
+   * For an uninitialized snapshot, this returns -1.
```
allisonport-db, line 41: Can you specify what "uninitialized snapshot" means?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java
```diff
@@ -60,11 +60,32 @@ public SnapshotImpl(
...
+  /////////////////
+
   @Override
   public long getVersion(Engine engine) {
     return version;
   }
 
+  @Override
+  public long getTimestamp(Engine engine) {
+    if (TableConfig.isICTEnabled(engine, metadata)) {
```
allisonport-db, line 74: I'm wondering if maybe this should be part of the snapshot creation (when we also load protocol & metadata)?

allisonport-db: `getTimestamp` also kind of implies that we don't do any real work to retrieve this

scottsand-db: > I'm wondering if maybe this should be part of the snapshot creation (when we also load protocol & metadata)?

That's a great point! Let me re-think this ...

kernel/kernel-api/src/main/java/io/delta/kernel/Snapshot.java
scottsand-db: This was just met copying + pasting the existing javadocs.

Let me think about this. We should probably just say "for an empty table"

kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java
scottsand-db: This will be a decent amount of work to do this. I wish this had been done with ICT was implemented in Kernel.

@vkorukanti any thoughts here?

scottsand-db: I just spent several hours implementing this, and realized that CommitInfo isn't stored in the checkpoint ... so, loading ICT from logReplay would only work when the latest log segment is a delta file not a checkpoint. darn.

kernel/kernel-api/src/main/java/io/delta/kernel/Snapshot.java
```diff
@@ -29,13 +29,29 @@
...
    * @return version of this snapshot in the Delta table
    */
   long getVersion(Engine engine);
 
+  /**
+   * Get the timestamp (in milliseconds since the Unix epoch) of the latest commit of this snapshot.
+   * Returns -1 if the table is empty and so has no commits.
+   *
+   * <p>When InCommitTimestampTableFeature is enabled, the timestamp is retrieved from the
+   * CommitInfo of the latest commit which can result in an IO operation.
```
allisonport-db, line 45: nit: i'm not sure "latest commit" is the right term here? do we see it as the "latest commit in this snapshot"?

kernel/kernel-api/src/main/java/io/delta/kernel/Snapshot.java
```diff
@@ -29,13 +29,29 @@
...
+   *
+   * <p>For non-ICT tables, this is the same as the file modification time of the latest commit in
+   * the snapshot.
```
allisonport-db, line 48: I actually wonder if we need this much detail here in a top-level API since this seems protocol related.  Maybe we can just say "returns the timestamp of this snapshot/version. This is determined according to the delta protocol ---link etc---"; and move these docs to the internal implementation for developers.

kernel/kernel-api/src/main/java/io/delta/kernel/Snapshot.java
```diff
@@ -29,13 +29,29 @@
...
+   * commits.
    *
    * @param engine {@link Engine} instance to use in Delta Kernel.
    * @return version of this snapshot in the Delta table
    */
   long getVersion(Engine engine);
 
+  /**
+   * Get the timestamp (in milliseconds since the Unix epoch) of the latest commit of this snapshot.
+   * Returns -1 if the table is empty and so has no commits.
```
allisonport-db, line 42: Maybe instead this should just be undefined for an empty table with no commits? Since -1 is technically a valid timestamp

kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java
allisonport-db: Maybe as a future optimization we can get the `inCommitTimestampOpt` as part of log replay if the latest log segment is a delta file, and if not leave it empty. But i'm fine if this PR does not make any changes for that.

kernel/kernel-api/src/main/java/io/delta/kernel/Snapshot.java
scottsand-db: I've changed my mind. Users should never get a Snapshot reference for Delta Table that doesn't exist (i.e. version = -1, or timestamp = -1).

Internal Kernel code may get such a Snapshot reference (e.g. TransactionBuilderImpl creates an InitialSnapshot), but this will never be propagated to users.

So, I can remove this part of the public documentation.

allisonport-db: I wonder if this might change once we support read-->write transactions?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java
```diff
@@ -64,11 +64,41 @@ public SnapshotImpl(
...
 
+  /**
+   * Get the timestamp (in milliseconds since the Unix epoch) of the latest commit in this Snapshot.
+   *
+   * <p>When InCommitTimestampTableFeature is enabled, the timestamp is retrieved from the
+   * CommitInfo of the latest commit in this Snapshot, which can result in an IO operation.
+   *
+   * <p>For non-ICT tables, this is the same as the file modification time of the latest commit in
+   * this Snapshot.
+   */
```
allisonport-db, line 84: If we don't think any external developers will ever see an uninitialized snapshot (empty table) can we at least clarify the behavior here for that case?

allisonport-db (approved):

vkorukanti (approved):

PR #3794, opened 2024-10-22
[Spark] Refactor CommitFailedException to allow passing in an inner exception

ctringdb:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Refactor `CommitFailedException` to allow for passing in a root cause exception so that we don't lose its stack trace

## How was this patch tested?
Simulate exceptions in unit tests and observe the output exception

No


LukasRupprecht (approved):
LGTM!

PR #3797, opened 2024-10-23
[Kernel] [CC Refactor #2] Add `TableDescriptor` and `CommitCoordinatorClient` API

scottsand-db:
This is a stacked PR. Please view this PR's diff here:
- https://github.com/scottsand-db/delta/compare/delta_kernel_cc_1...delta_kernel_cc_2

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Adds new `TableDescriptor` and `CommitCoordinatorClient` API. Adds a new `getCommitCoordinatorClient` API to the `Engine` (with a default implementation that throws an exception).

## How was this patch tested?

N/A trivial.

Yes. See the above.


kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/CommitCoordinatorClient.java
```diff
@@ -0,0 +1,168 @@
...
+   * @param tableDescriptor The descriptor for the table.
+   * @param startVersion The minimum version of the commit that should be returned, or {@code null}
+   *     if there is no minimum.
+   * @param endVersion The maximum version of the commit that should be returned, or {@code null} if
+   *     there is no maximum.
+   * @return {@link GetCommitsResponse} which has a list of {@link
+   *     io.delta.kernel.engine.coordinatedcommits.Commit}s and the latestTableVersion which is
+   *     tracked by the {@link CommitCoordinatorClient}.
+   */
+  GetCommitsResponse getUnbackfilledCommits(
```
scottsand-db, line 136: `getCommits`

kernel/kernel-api/src/main/java/io/delta/kernel/annotation/Nullable.java
```diff
@@ -0,0 +1,27 @@
+/*
```
sumeet-db, line 1: For my knowledge, why don't we use existing javax annotations instead of defining new ones in Kernel?

sumeet-db (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/annotation/Nullable.java
scottsand-db: Don't want to pull in a new dependency!

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/CommitCoordinatorClient.java
```diff
@@ -0,0 +1,168 @@
...
+  /**
+   * Register the table represented by the given {@code logPath} at the provided {@code
+   * currentVersion} with the commit coordinator this commit coordinator client represents.
+   *
+   * <p>This API is called when the table is being converted from an existing file system table to a
+   * coordinated-commit table.
+   *
+   * <p>When a new coordinated-commit table is being created, the {@code currentVersion} will be -1
+   * and the upgrade commit needs to be a file system commit which will write the backfilled file
+   * directly.
```
allisonport-db, line 59: For my understanding you (the client, ie Spark/Kernel etc) call this first for some version N. Then when `commit` is called with version N, the CCC recognizes that this is the same version and thus the commit needs to be immediately backfilled/written to the filesystem?

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/CommitCoordinatorClient.java
```diff
@@ -0,0 +1,168 @@
...
+      AbstractProtocol currentProtocol);
+
+  /**
+   * Commit the given set of actions to the table represented by {@code tableDescriptor}.
+   *
+   * @param engine The {@link Engine} instance to use.
+   * @param tableDescriptor The descriptor for the table.
+   * @param commitVersion The version of the commit that is being committed.
+   * @param actions The set of actions to be committed
+   * @param updatedActions Additional information for the commit, including:
```
allisonport-db, line 93: Sorry a bunch of questions about CC writes not necessarily specific to this PR. What are the `updatedActions` for and why do they need to be separated from the other `actions`?

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/CommitCoordinatorClient.java
```diff
@@ -0,0 +1,168 @@
...
+   *     <ul>
+   *       <li>Commit info
+   *       <li>Metadata changes
+   *       <li>Protocol changes
+   *     </ul>
+   *
+   * @return {@link CommitResponse} containing the file status of the committed file. Note: If the
```
allisonport-db, line 100: this is the unbackfilled file right?

kernel/kernel-api/src/main/java/io/delta/kernel/engine/Engine.java
```diff
@@ -56,6 +57,23 @@ public interface Engine {
...
+   * @param commitCoordinatorConf The configuration settings for the underlying commit coordinator
+   *     client, taken directly from the Delta table property {@link
+   *     io.delta.kernel.internal.TableConfig#COORDINATED_COMMITS_TABLE_CONF}
+   * @return A {@link CommitCoordinatorClient} implementation corresponding to the specified commit
+   *     coordinator name
+   * @since 3.3.0
+   */
+  default CommitCoordinatorClient getCommitCoordinatorClient(
+      String commitCoordinatorName, Map<String, String> commitCoordinatorConf) {
+    throw new UnsupportedOperationException("Not implemented");
```
allisonport-db, line 74: Is this temporarily with a default implementation or will we be keeping it like this?

What will be the expected behavior if an engine interface hasn't implemented this method but some user tries to read a CC table? Will it throw this exception? Or do we want to force all engine impls to override this?

allisonport-db (commented):
Mostly clarifying questions due to my lack of knowledge for CC

kernel/kernel-api/src/main/java/io/delta/kernel/engine/Engine.java
scottsand-db: > temporarily with a default implementation
Yup! This is so we can merge this without having to go and update all implementations of Engine within this PR

allisonport-db (approved):
Changes themselves LGTM

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/CommitCoordinatorClient.java
LukasRupprecht: The updated actions are the CommitInfo and the previous and current Metadata/Protocol. They are also included in the `actions` (Protocol and Metadata only if they changed) but we want to pass them separately for convenience in case commit coordinator client implementations want to do something with them (for example check if the Metadata of a table has changed).

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/CommitCoordinatorClient.java
```diff
@@ -0,0 +1,168 @@
...
+   * currentVersion} with the commit coordinator this commit coordinator client represents.
+   *
+   * <p>This API is called when the table is being converted from an existing file system table to a
+   * coordinated-commit table.
+   *
+   * <p>When a new coordinated-commit table is being created, the {@code currentVersion} will be -1
+   * and the upgrade commit needs to be a file system commit which will write the backfilled file
+   * directly.
+   *
+   * @param engine The {@link Engine} instance to use.
```
LukasRupprecht, line 61: Can we add some information on what the Engine would/should be used for during table registration?

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/CommitCoordinatorClient.java
```diff
@@ -0,0 +1,168 @@
...
+      String logPath,
+      @Nullable TableIdentifier tableIdentifier,
+      long currentVersion,
+      AbstractMetadata currentMetadata,
+      AbstractProtocol currentProtocol);
+
+  /**
+   * Commit the given set of actions to the table represented by {@code tableDescriptor}.
+   *
+   * @param engine The {@link Engine} instance to use.
```
LukasRupprecht, line 89: Again, add some info for what we need the engine (here it'd be mainly for writing the commit file). Same for the other APIs below.

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/CommitCoordinatorClient.java
```diff
@@ -0,0 +1,168 @@
...
+   * @return A map of key-value pairs which is issued by the commit coordinator to uniquely identify
+   *     the table. This should be stored in the table's metadata for table property {@link
+   *     io.delta.kernel.internal.TableConfig#COORDINATED_COMMITS_TABLE_CONF}. This information
+   *     needs to be passed to the {@link #commit}, {@link #getCommits}, and {@link
+   *     #backfillToVersion} APIs to identify the table.
+   */
+  Map<String, String> registerTable(
+      Engine engine,
+      String logPath,
+      @Nullable TableIdentifier tableIdentifier,
```
LukasRupprecht, line 81: Why don't we use `Optional` like in `TableDescriptor`?

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/TableDescriptor.java
```diff
@@ -0,0 +1,66 @@
...
+import static java.util.Objects.requireNonNull;
+
+import io.delta.kernel.TableIdentifier;
+import io.delta.kernel.annotation.Evolving;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * The complete descriptor of a Coordinated Commits (CC) Delta table, including its logPath, table
+ * identifier, and table CC table configuration.
```
LukasRupprecht, line 28:
```suggestion
 * The complete descriptor of a Coordinated Commits (CC) Delta table, including its logPath, table
 * identifier (if access is not path-based), and table CC table configuration.
```

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/CommitCoordinatorClient.java
sumeet-db: Not necessarily. It's acceptable for CC-Client to return the backfilled file too.

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/CommitCoordinatorClient.java
scottsand-db: I've been flip-flopping on this, myself.

Using Optional is not preferred in a java public API ... but I've been considering refactoring and using Optional actually

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/CommitCoordinatorClient.java
allisonport-db: Got it thanks!

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/CommitCoordinatorClient.java
scottsand-db: cc @LukasRupprecht @dhruvarya-db @prakharjain09

LukasRupprecht: Commit N would add the CC configuration to the table so it'll be available in version N+1. It is not in version N so the commit does not go through the newly added commit coordinator client but rather just through the file system, i.e. backfilling is not necessary.

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/CommitCoordinatorClient.java
prakharjain09: Also the `actions` is an iterator i.e. it can only be traversed once. The commit coordinator can't go through it to get these important updates (e.g. schema change / protocol change). So the API explicitly passes such updates.

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/CommitCoordinatorClient.java
scottsand-db: @LukasRupprecht -- we don't want to prescribe what the engine should be used for. It can be used for any json reading or filesystem operations needed. who knows how they implement their Commit Coordinator Client.

scottsand-db: It's also just convention to pass in the engine, so that the client may use any future engine interfaces (think logging or metrics) without having so save a reference of the engine

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/TableDescriptor.java
scottsand-db: Hm... that's not _quite_ correct, right? It's not like you pass either one of the path OR the identifier, but not BOTH, to the CC. We are passing both ...

kernel/kernel-api/src/main/java/io/delta/kernel/engine/Engine.java
dhruvarya-db: Is it right to assume that implementations will still be able to read other dynamic configurations when building the coordinator? e.g. the Delta-spark getCCC interface also takes in a sparkSession allowing for dynamic configuration of the client. Implementations of this method will still be able to read some other configuration source (even though it is not explicitly being passed) right?

scottsand-db: > Is it right to assume that implementations will still be able to read other dynamic configurations when building the coordinator?

Absolutely. We leave it up to the engine to create the CCC. If the engine is aware of any dynamodb configurations, it can use them!

> Implementations of this method will still be able to read some other configuration source (even though it is not explicitly being passed) right?

Yes. I'd encourage you to look at the tracking issue https://github.com/delta-io/delta/issues/3817 to look at future PRs where you can see this being done.

dhruvarya-db (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/TableDescriptor.java
LukasRupprecht: I meant that if the access is path-based, there won't be a table identifier.

scottsand-db: I'll include this change in PR 3 https://github.com/delta-io/delta/pull/3798

PR #3798, opened 2024-10-23
[Kernel] [CC Refactor #3] Add `ConfigurationProvider` and `AbstractCommitCoordinatorBuilder` APIs

scottsand-db:
- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Adds new `ConfigurationProvider` and `AbstractCommitCoordinatorBuilder` APIs

## How was this patch tested?

Tests are added in CC Refactor PR 5: https://github.com/delta-io/delta/pull/3821

Yes see the above


kernel/kernel-api/src/main/java/io/delta/kernel/config/ConfigurationProvider.java
```diff
@@ -0,0 +1,61 @@
...
+@Evolving
+public interface ConfigurationProvider {
+  /**
+   * Retrieves the configuration value for the given key.
+   *
+   * @param key the configuration key
+   * @return the configuration value associated with the key, if it exists
+   * @throws NoSuchElementException if the key is not found
+   */
+  String get(String key) throws NoSuchElementException;
```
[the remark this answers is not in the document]
sumeet-db: HadoopConf returns null but SparkConf throws a `NoSuchElementException`.
The latter seems cleaner from my pov.

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/AbstractCommitCoordinatorBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+  public abstract String getName();
+
+  /**
+   * Build the {@link CommitCoordinatorClient}. This may be a new instance or a cached one.
+   *
+   * @param sessionConfig The session-level configuration that may represent environment-specific
+   *     configurations such as {@code HadoopConf} or {@code SparkConf}. This sessionConfig can be
+   *     used by builders to lookup per-session configuration values unique to that builder.
+   * @param commitCoordinatorConf the parsed value of the Delta table property {@link
+   *     io.delta.kernel.internal.TableConfig#COORDINATED_COMMITS_TABLE_CONF} and represents the
```
sumeet-db, line 55: This should be the COORDINATOR_CONF and not the TABLE_CONF

Specifically, delta.coordinatedCommits.commitCoordinatorConf-preview

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/AbstractCommitCoordinatorBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+   *     configuration properties for describing the Delta table to the commit-coordinator.
```
sumeet-db, line 56: The intention of this conf is to track the configuration properties for the commit coordinator which is needed to build the commit coordinator client. e.g. DynamoDB endpoint https://github.com/delta-io/delta/blob/285ed6bd2434ac760dba61399a8156ecdb4c1c47/spark/src/main/java/io/delta/dynamodbcommitcoordinator/DynamoDBCommitCoordinatorClientBuilder.java#L53

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/CommitCoordinatorClient.java
```diff
@@ -0,0 +1,168 @@
...
+
+  /**
+   * Backfill all commits up to {@code version} and notify the commit coordinator.
+   *
+   * <p>If this API returns successfully, that means the backfill must have been completed, although
+   * the commit coordinator may not be aware of it yet.
+   *
+   * @param engine The {@link Engine} instance to use.
+   * @param tableDescriptor The descriptor for the table.
+   * @param version The version until which the commit coordinator client should backfill.
```
sumeet-db, line 150: `version` -> `version (inclusive)`

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/CommitCoordinatorClient.java
```diff
@@ -0,0 +1,168 @@
...
+   * @param lastKnownBackfilledVersion The last known version that was backfilled before this API
```
sumeet-db, line 151: `version` -> `version (inclusive)`

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/CommitCoordinatorUtils.java
```diff
@@ -0,0 +1,85 @@
+/*
```
sumeet-db, line 1: We also have [CoordinatedCommitUtils.java](https://github.com/delta-io/delta/blob/master/storage/src/main/java/io/delta/storage/commit/CoordinatedCommitsUtils.java)

Would be nice to either merge them in the same file eventually or have better names for them

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/TableDescriptor.java
```diff
@@ -0,0 +1,66 @@
...
+import static java.util.Objects.requireNonNull;
+
+import io.delta.kernel.TableIdentifier;
+import io.delta.kernel.annotation.Evolving;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * The complete descriptor of a Coordinated Commits (CC) Delta table, including its logPath, table
+ * identifier, and table CC table configuration.
```
sumeet-db, line 28: Nit: Fix comment

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/CommitCoordinatorUtils.java
scottsand-db: Eventually, everything Coordinated-Commits-related in delta-storage will be refactored and moved to delta-kernel, and then be deleted.

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/AbstractCommitCoordinatorBuilder.java
```diff
@@ -0,0 +1,133 @@
...
+   *   <li>Determines the specific builder configuration lookup key.
+   *   <li>Grabs the corresponding builder className for that key from the provided sessionConfig.
+   *   <li>Instantiates a new instance of that {@link AbstractCommitCoordinatorBuilder}.
+   *   <li>Invokes the builder's build method, passing along the sessionConfig and commit
+   *       coordinator config map.
+   * </ul>
+   *
+   * @param commitCoordinatorName the commit coordinator name
+   * @param sessionConfig The session-level configuration that may represent environment-specific
+   *     configurations such as {@code HadoopConf} or {@code SparkConf}. This sessionConfig is used
```
LukasRupprecht, line 66: Do we need to make them implement `ConfigurationProvider` or how can we, e.g., pass a SparkConf as a `ConfigurationProvider`?

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/AbstractCommitCoordinatorBuilder.java
```diff
@@ -0,0 +1,133 @@
...
+   *     to look up the right builder className to instantiate for the given commit coordinator
+   *     name. It can also be used by builders to lookup per-session configuration values unique to
+   *     that builder.
+   * @param commitCoordinatorConf the parsed value of the Delta table property {@link
+   *     io.delta.kernel.internal.TableConfig#COORDINATED_COMMITS_TABLE_CONF} and represents the
```
LukasRupprecht, line 71: Should this be
```suggestion
   *     io.delta.kernel.internal.TableConfig#COORDINATED_COMMITS_COORDINATOR_CONF} and represents the
```
?

LukasRupprecht: If so, then L72 below should probably say something like
```
configuration properties for allowing the commit coordinator client to talk to the right commit coordinator.
```

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/AbstractCommitCoordinatorBuilder.java
```diff
@@ -0,0 +1,133 @@
...
+/**
+ * A builder class to create a {@link CommitCoordinatorClient} (CCC).
+ *
+ * <p>Subclasses of this class must provide a no-argument constructor.
+ *
+ * <p>Note: These builder classes are optional. The {@link
+ * io.delta.kernel.engine.Engine#getCommitCoordinatorClient} API does not prescribe how to create
+ * the underlying CCC and does not require a builder.
+ *
+ * <p>The benefit of implementing a builder for your CCC is that your Engine may then invoke {@link
```
sumeet-db, line 35: What's the alternative solution for a CC if a CC builder is not implemented? Can you please specify that too? In general, as an engine owner, how do I decide whether I should implement a builder or not?

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/AbstractCommitCoordinatorBuilder.java
sumeet-db, line 66: What if you need both hadoopConf and sparkConf?

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/AbstractCommitCoordinatorBuilder.java
sumeet-db: I had commented this in one of the PRs 🤔

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/AbstractCommitCoordinatorBuilder.java
scottsand-db: They need to pass in _something_ that is a `ConfigurationProvider` (i.e. that implements the interface). You can easily create a wrapper around a SparkConf or a HadoopConf to implemeent the `ConfigurationProvider` interface

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/AbstractCommitCoordinatorBuilder.java
scottsand-db: The alternative is: build it yourself! Who knows what clients they are instantiating. Maybe those clients have buildings that they want to be instantiated via reflection. Maybe some engine hardcodes what client to use and there's no need to have a fancy builder at all.

It's completely up to engines.

> how do I decide whether I should implement a builder or not?
When you want the CCC instantaited at runtime via reflection. i.e. you want which CCC to use to be decided at runtime. I can add this.

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/AbstractCommitCoordinatorBuilder.java
scottsand-db: Yup! Thanks! Good catch.

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/AbstractCommitCoordinatorBuilder.java
```diff
@@ -0,0 +1,133 @@
...
+   *     configurations such as {@code HadoopConf} or {@code SparkConf}. This sessionConfig is used
+   *     to look up the right builder className to instantiate for the given commit coordinator
+   *     name. It can also be used by builders to lookup per-session configuration values unique to
+   *     that builder.
+   * @param commitCoordinatorConf the parsed value of the Delta table property {@link
+   *     io.delta.kernel.internal.TableConfig#COORDINATED_COMMITS_TABLE_CONF} and represents the
+   *     configuration properties for describing the Delta table to the commit-coordinator.
+   * @return the {@link CommitCoordinatorClient} corresponding to the given commit coordinator name
+   */
+  public static CommitCoordinatorClient buildCommitCoordinatorClient(
```
sumeet-db, line 75: If I understand correctly, the dev experience for using this method will be:
```
CommitCoordinatorClient client = AbstractCommitCoordinatorBuilder
                .buildCommitCoordinatorClient(
                    "dynamodb",
                    sessionConfig,
                    coordinatorConf
                );
```
Somehow, it doesn't feel intuitive that this code will look up the builder in the conf and do stuff

scottsand-db:
@LukasRupprecht and @sumeet-db -- is it every possible or intended/desired that AbstractMetadata and AbstractProtocol would be null? or are these required params? I see `null` being passed in in some existing tests today ...

https://github.com/delta-io/delta/blob/6ae4b62845ed579bb5a19f4646831c4ee2931c02/kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/CommitCoordinatorClient.java#L83

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/AbstractCommitCoordinatorBuilder.java
sumeet-db: How's this method different than `CommitCoordinatorProvider.getBuilder(... /* load builder dynamically from conf */).build(...)`? And why do you prefer this one? 🤔

kernel/kernel-api/src/main/java/io/delta/kernel/config/ConfigurationProvider.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.NoSuchElementException;
+import java.util.Optional;
+
+/**
+ * A generic interface for retrieving configuration values.
+ *
+ * @since 3.3.0
+ */
+@Evolving
+public interface ConfigurationProvider {
```
prakharjain09, line 29: @scottsand-db  This class doesn't provide any method to iterate all configs. Does that work for UC Commit Coordinator? Don't we need to iterate over all configs in that? Will let @sumeet-db confirm.

scottsand-db: @prakharjain09 -- Yup, great callout. We can update APIs when we get there. IMO there's no rush or requirement to add it now.

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/AbstractCommitCoordinatorBuilder.java
scottsand-db: https://github.com/delta-io/delta/blob/master/kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/internal/coordinatedcommits/CommitCoordinatorProvider.java

is the old code that will be deleted when this is refactored. it uses HadoopConf and io.delta.storage APIs

scottsand-db: > Somehow, it doesn't feel intuitive that this code will look up the builder in the conf and do stuff

Really? I quite like it. Feel free to let me know what doesn't feel right about this. This is what we agreed on in the sync on Tuesday Oct 22 @ 1:00pm

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/AbstractCommitCoordinatorBuilder.java
scottsand-db: Great question. You can union them to implement the `get` interface.

When we go to implement the UC CCC using this API, we can adjust methods as needed.

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/AbstractCommitCoordinatorBuilder.java
scottsand-db: Further, this API feels _very_ similar to the previous API of

```
CommitCoordinatorClient client = CommitCoordinatorProvider
  .getCommitCoordinatorClient(
    hadoopConf,
    name,
    commitCoordinatorConf);
```

LukasRupprecht:
> @LukasRupprecht and @sumeet-db -- is it every possible or intended/desired that AbstractMetadata and AbstractProtocol would be null? or are these required params? I see null being passed in in some existing tests today ...

They shouldn't be null as client implementations may rely on it (test code is an exception if we can be sure that the underlying client doesn't require it).

kernel/kernel-api/src/main/java/io/delta/kernel/config/ConfigurationProvider.java
sumeet-db: Will need a `getAll` / `getAllByPrefix` API for UC Commit Coordinator Builder when we get there.

kernel/kernel-api/src/main/java/io/delta/kernel/coordinatedcommits/AbstractCommitCoordinatorBuilder.java
sumeet-db: I was hoping we could prevent patterns like the following by moving `buildCommitCoordinatorClient` to a separate `DefaultCommitCoordinatorClientBuilder` that `DynamoDBCCClientBuilder` doesn't need to extend:
```
CommitCoordinatorClient client = DynamoDBCCClientBuilder.buildCommitCoordinatorClient(
    "dynamodb", 
    config, 
    coordConf);
```

PR #3799, opened 2024-10-24
[Spark] Write a checksum after every commit

dhruvarya-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR adds a `ChecksumHook` which is responsible for a writing a checksum (See https://github.com/delta-io/delta/pull/3777) of the current table state after every commit. This is guarded behind a flag which is `false` by default. Currently, every checksum write will trigger a full state reconstruction, which can be very expensive. An upcoming PR will try to make this checksum computation incremental so that we don't have to pay a performance penalty.

## How was this patch tested?

Added a new suite --- ChecksumSuite.

No

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/delta/ChecksumSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/ChecksumSuite.scala
@@ -1,0 +1,63 @@
+/*
+ * Copyright (2021) The Delta Lake Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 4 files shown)


prakharjain09 (approved):

spark/src/test/scala/org/apache/spark/sql/delta/ChecksumSuite.scala
```diff
@@ -0,0 +1,61 @@
...
+import org.apache.spark.sql.delta.util.FileNames
+import org.apache.hadoop.fs.Path
+
+import org.apache.spark.sql.QueryTest
+import org.apache.spark.sql.test.SharedSparkSession
+
+class ChecksumSuite
+  extends QueryTest
+    with SharedSparkSession {
+
```
felipepessoto, line 33: Missing with DeltaSQLCommandTest

dhruvarya-db: Thanks for pointing that out. Fixed!

dhruvarya-db:
> Could you share an example of CRC file? I tried to run locally but it throws an error:

 >  org.apache.spark.sql.delta.DeltaAnalysisException: [DELTA_CONFIGURE_SPARK_SESSION_WITH_EXTENSION_AND_CATALOG] This Delta operation requires the SparkSession to be configured with the DeltaSparkSessionExtension and the DeltaCatalog. Please set the necessary configurations when creating the SparkSession as shown below.

Hey @felipepessoto , I will try to generate one locally and post here.

felipepessoto:
Don't worry, I already did it locally after the UT fix

PR #3805, opened 2024-10-25
[Spark] Replace the default pattern matching for LogicalRelation to LogicalRelationWithTable

HeartSaVioR:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR proposes to replace the default pattern matching for `LogicalRelation` to newly introduced pattern object `LogicalRelationWithTable` which will be available in upcoming Spark 4.0.

This change helps the project to modify less pieces of code when Spark makes the change to the LogicalRelation; most pattern matchings with LogicalRelation only extract the relation and catalog table, hence they fit with LogicalRelationWithTable.

## How was this patch tested?

Existing tests would suffice.

No.


HeartSaVioR:
I'm planning to propose the change of LogicalRelation (new parameter of constructor) in Spark 4.0, which will break all pattern matches. For creating an instance of LogicalRelation, I'll try to figure out the way to be compatible with existing constructor. But for pattern matching, every addition of parameter require patterns to be updated. It won't be needed after this PR.

vkorukanti (approved):

PR #3831, opened 2024-10-30
[Spark] Support partition-like data filters

chirag-s-db:
- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Given an arbitrary data skipping expression, we can skip a file if:
1. For all of the referenced attributes in the expression, the collected min and max values are equal AND there are 0 nulls on that column.
AND
2. The data skipping expression (when evaluated on the collect min==max values on all referenced attributes) evaluates to false.

This PR adds support for some of these expressions with the following limitations:
1. The table must be >= 100 files (this is to ensure that the added data skipping expressions to avoid regressing the performance for small tables that won't have many files with the same min-max value).
2. The table must be a clustered table and all referenced attributes must be clustering columns (we use this heuristic to avoid adding extra complexity to data skipping for expressions that won't be able to filter out many files).
3. The expression must not reference a Timestamp-type column. Because stats on timestamp columns are truncated to millisecond precision, we can't safely assume that the min and max value for a timestamp column are the same (even if the collected stats are the same). Because timestamp is generally quite high cardinality, it should anyways be relatively rare that the min and max value for a file are equal for the timestamp column.

One more minor nuance: There's one more case where the collected stats differs from the behavior of partitioned tables - a truncated string. However, if a string value is truncated to the first 32 characters, then the collected max value for the string will not be equal to the collected min value (as one or more tiebreaker character(s) will be appended to the collected max value). As a result, it should be sufficient to validate equality, since for any truncated string, the min and max value will not be equal.

## How was this patch tested?
See new test.

No.


chirag-s-db:
@scovich Could you take a look?

spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingReader.scala
```diff
@@ -583,6 +588,191 @@ trait DataSkippingReaderBase
...
+          // Rewrite the children of InSubqueryExec, then replace the subquery with an InSet
+          // containing the materialized values.
+          rewriteDataFiltersAsPartitionLikeInternal(in.child, clusteringColumnPaths).flatMap {
+            case (rewrittenChildren, referencedStats) =>
+              Some(InSet(rewrittenChildren, possiblyNullValues.toSet), referencedStats)
+          }
+        }
+      // Don't allow rewriting UDFs - even if deterministic, UDFs might have some unexpected
+      // side effects when executed twice.
+      case _: UserDefinedExpression => None
```
scovich, line 672: Is the problem that spark UDF are marked "nondeterministic" by default, and their creator has to recognize whether that's true and override as needed? So basically we have to assume that any UDF which claims "nondeterministic" is lying?

spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingReader.scala
```diff
@@ -583,6 +588,191 @@ trait DataSkippingReaderBase
...
+      // Pushdown NOT through OR - we prefer AND to OR because AND can tolerate one branch not being
+      // rewriteable.
+      case Not(Or(e1, e2)) =>
+        rewriteDataFiltersAsPartitionLikeInternal(And(Not(e1), Not(e2)), clusteringColumnPaths)
+      // For AND expressions, we can tolerate one side not being eligible for partition-like
+      // data skipping - simply remove the ineligible side.
+      case And(left, right) =>
```
scovich, line 679: I don't think I understand why we can't support/rewrite `Not(And)`? Because AFAIK it's otherwise just plain ineligible for data skipping (without even looking at the stats). Is that really better than the risk that `Or(Not(a), Not(b))` might have a NULL input for some subset of rows?

spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingReader.scala
```diff
@@ -583,6 +588,191 @@ trait DataSkippingReaderBase
...
+        val rightResult = rewriteDataFiltersAsPartitionLikeInternal(right, clusteringColumnPaths)
+        (leftResult, rightResult) match {
+          case (Some((newLeft, statsLeft)), Some((newRight, statsRight))) =>
+            Some((And(newLeft, newRight), statsLeft ++ statsRight))
+          case _ => leftResult.orElse(rightResult)
+        }
+      // For any other expression, recursively rewrite the expression's children unless the
+      // expression might be very expensive to evaluate (JSON parsing, regex, etc.).
+      case other if !other.containsAnyPattern(INVOKE, JSON_TO_STRUCT, LIKE_FAMLIY,
+        REGEXP_EXTRACT_FAMILY, REGEXP_REPLACE) =>
```
scovich, line 690: Should we add the various UDF types as well? They can be arbitrarily expensive, no?
(the other check only rejects a UDF node that we actually reach, not the parents)

scovich: But actually... wouldn't this approach pessimize predicates that have an AND somewhere deeper, when not all legs of the AND have expensive/bad operations?

spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingReader.scala
```diff
@@ -583,6 +588,191 @@ trait DataSkippingReaderBase
...
+      //    so we don't have to check here.
+      // 2. The min-max stats on a string column might be truncated for an unpartitioned table.
+      //    Note that just validating that the min and max are equal is enough to prevent this case
+      //    - if the string is truncated, the collected max value is guaranteed to be longer than
+      //    the min value due to the tiebreaker character(s) appended at the end of the max.
+      Not(
+        Or(
+          allNulls,
+          And(
+            EqualNullSafe(
```
scovich, line 730: Shouldn't this be `EqualTo` in case the min/max stats are missing and inferred to be NULL?

spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingReader.scala
```diff
@@ -583,6 +588,191 @@ trait DataSkippingReaderBase
...
+          // Create an expression that returns true if a file must be read because it has mismatched
+          // min-max values or partial nulls on any of the referenced columns.
+          val numRecordsStatsCol = StatsColumn(NUM_RECORDS, pathToColumn = Nil, LongType)
+          val numRecordsColOpt = getStatsColumnOpt(numRecordsStatsCol)
+          val (finalExpr, statsCols) =
+            referencedStats.foldLeft((newExpr, Seq(numRecordsStatsCol))) {
+              case ((oldExpr, stats), resolvedReference) =>
+                val updatedExpr = Or(
+                  oldExpr, fileMustBeScanned(resolvedReference, numRecordsColOpt))
+                (updatedExpr, stats ++ resolvedReference.referencedStatsCols)
```
scovich, line 767: Seems like we should be doing `++=` on a mutable structure, to avoid quadratic behavior from copy-on-write the list over and over?

scovich (commented):
Nice. Surprised it's so little code, to be honest.

spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingReader.scala
chirag-s-db: Exactly - we can't trust that UDFs are deterministic (as the creators rarely, if ever, recognize that UDFs are nondeterministic).

The other issue is that UDFs can be arbitrarily slow/expensive to call, so we risk a performance regression when double executing UDFs (some users are also sensitive to double executing UDFs due to having UDFs that call a metered API).

spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingReader.scala
chirag-s-db: I think the key distinction here is that (unlike traditional data skipping) partition-like data skipping can support arbitrary NOT expressions. As a result, we prefer to not push down the NOT through the AND.

For example, say we have NOT(AND(a, b)). `a` can be rewritten as partition-like, but `b` can't be.
If we push through the NOT to create OR(NOT(a), NOT(b)), we attempt rewriting both `NOT(a)` and `NOT(b)`, then we get OR(None, NOT(rewritten b)). Since OR can't tolerate one branch not being rewritten, we end up being unable to rewrite the expression.

However, if we don't push the NOT through the AND, then we attempt rewriting both `a` and `b` and get NOT(AND(None, b)). Since AND can tolerate one branch not being rewritten, we end up with NOT(b) as the partition-like data filter, which we prefer.

spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingReader.scala
chirag-s-db: Yeah, we would prefer to only reject nodes that we reach. Rewrote this to explicitly match against these expression families so that we don't prematurely reject these expressions.

spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingReader.scala
chirag-s-db: I think this _theoretically_ can be EqualNullSafe (if the min/max stats are missing and inferred to be NULL, then the `verifyStatsForFilter` check should trigger and we should read the file anyways). 

However, for clarity, we never expect the min-max stats to be null when in this expression branch (given the check below that there are no nulls on the column), so replaced with EqualTo.

spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingReader.scala
chirag-s-db: At the moment, the quadratic behavior shouldn't be too much of a concern (given that the returned references is a set of clustering columns, this set can have a maximum size of 4). However, in case we do expand coverage in the future, this quadratic behavior would be easy to miss. Updated to use a mutable buffer to avoid this case.

scovich (approved):
LGTM!

felipepessoto:
@chirag-s-db the DELTA_DATASKIPPING_PARTITION_LIKE_FILTERS_ENABLED is disabled by default, is there are concerns to don't enable it?

Thanks.

PR #3832, opened 2024-10-30
[SPARK] Match collated StringType in TypeWideningMetadata

zhipengmao-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Spark master introduced collated `StringType`. In `TypeWideningMetadata.collectTypeChanges`, we used to only match StringType instance, therefore if the type change is from/to `StringType(collationId)`, which is expected and unrelated to type widening, we will throw an error `typeWidening.unexpectedTypeChange`. This PR fixes this by matching both collated and uncollated `StringType` when capturing string type changes. 

## How was this patch tested?
Existent tests.

Tests related to collations cannot be added as delta does not support collations yet.

No.


johanl-db (approved):
Thanks for fixing this!

PR #3846, opened 2024-11-05
[Spark] Validate computed state against checksum on checkpoint

dhruvarya-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Follow up for https://github.com/delta-io/delta/pull/3828.

This PR adds checksum validation logic. On every checkpoint, we will take the computed state of the table as per the deltas and the previous checkpoint and compare it against the checksum that was written at that version. The same methods can potentially be used to validate more frequently (if needed).

## How was this patch tested?

Added a new test case in ChecksumSuite that tests that all logically corrupted fields are being caught by the validation logic.

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/Checkpoints.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/Checkpoints.scala
@@ -608,6 +608,10 @@ object Checkpoints
       deltaLog: DeltaLog,
       snapshot: Snapshot): LastCheckpointInfo = recordFrameProfile(
       "Delta", "Checkpoints.writeCheckpoint") {
+    if (spark.conf.get(DeltaSQLConf.DELTA_WRITE_CHECKSUM_ENABLED)) {
+      snapshot.validateChecksum(Map("context" -> "writeCheckpoint"))
+    }
+
     val hadoopConf = deltaLog.newDeltaHadoopConf()
 
     // The writing of checkpoints doesn't go through log store, so we need to check with the
--- a/spark/src/main/scala/org/apache/spark/sql/delta/Checksum.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/Checksum.scala
@@ -20,6 +20,7 @@ import java.io.FileNotFoundException
 import java.nio.charset.StandardCharsets.UTF_8
 
 // scalastyle:off import.ordering.noEmptyLine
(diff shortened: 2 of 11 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/Checkpoints.scala
```diff
@@ -608,6 +608,10 @@ object Checkpoints
       deltaLog: DeltaLog,
       snapshot: Snapshot): LastCheckpointInfo = recordFrameProfile(
       "Delta", "Checkpoints.writeCheckpoint") {
+    if (spark.conf.get(DeltaSQLConf.DELTA_WRITE_CHECKSUM_ENABLED)) {
+      snapshot.validateChecksum(Map("context" -> "writeCheckpoint"))
+    }
```
prakharjain09, line 613: Should this check be done in the beginning or in the end?

dhruvarya-db: I feel that it would be a waste of time to write a checkpoint only to find out that there is something wrong with the delta log.

dhruvarya-db: @prakharjain09 Since most of this PR is about the validation logic itself, maybe we can address the question of when validation should be triggered in a follow up?

spark/src/main/scala/org/apache/spark/sql/delta/Checksum.scala
```diff
@@ -366,3 +371,163 @@ trait ReadChecksum extends DeltaLogging { self: DeltaLog =>
...
+        "mismatchingFields" -> mismatchErrorMap.keys.toSeq,
+        "detailedErrorMap" -> detailedErrorMapForUsageLogs,
+        "v2CheckpointEnabled" ->
+          CheckpointProvider.isV2CheckpointEnabled(this),
+        "checkpointProviderCheckpointPolicy" ->
+          checkpointProvider.checkpointPolicy.map(_.name).getOrElse("")
+      ) ++ contextInfo)
+
+    val spark = sparkOpt.getOrElse {
+      throw DeltaErrors.sparkSessionNotSetException()
```
tedyu, line 421: This can be lifted to beginning of the method so that the potential exception can be thrown early.

PR #3859, opened 2024-11-08
[Spark] Accept generated columns of nested types that differ in nullability

larsk-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

- Changes the behaviour of generated columns over nested types to allow nullability mismatches like we do for top-level types by switching exact data type equal to using `equalsIgnoreNullability` instead.
- Additionally, improve the error message in the case where the SQL string is identical (which hopefully shouldn't happen anymore after this, but just in case this comes up again later).

## How was this patch tested?

- Added new tests for various type (mis-) matches.

Yes: Generated columns now allow the type of the column definition to differ by nullability from the type of the generating expression, even when the type is a nested type.


johanl-db (approved):
LGTM

PR #3860, opened 2024-11-08
[Spark] Remove use of current_date in MERGE test

johanl-db:
## Description
Fix flaky test that very rarely fail with actual result being off by one day compared to the expected result.

Turns out it calls `current_date()` separately to compute the input data and the expected result.
 
## How was this patch tested?
~Run the test around midnight~  (as a thought experiment)


cstavr (approved):

larsk-db (approved):
LGTM

PR #3863, opened 2024-11-08
[Spark] Pass table catalogs throughout DeltaLog

ctring:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Previously, we passed the table identifier to the DeltaLog's constructor and methods to use. For future-proof purpose, we should pass the whole catalog table instead.

## How was this patch tested?

Unit test
No


LukasRupprecht (approved):
LGTM!

PR #3870, opened 2024-11-12
[Spark] Detect opaque URIs and throw proper exception

cstavr:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

An opaque URI is an absolute URI whose scheme-specific part does not begin with a slash character ('/'), and are not further parsed by `java.net.URI` library (see https://docs.oracle.com/javase/7/docs/api/java/net/URI.html):
```
val uri = new URI("http:example.com")
uri.isOpaque -> true
uri.isAbsolute -> true
uri.getPath -> null
```

This causes issues when we try to call path-related methods in the URIs, e.g.:
```
val filePath = new Path(uri)
filePath.toString -> "http:"
filePath.isAbsolute -> NullPointerException
```

This commit fixes this issue by detecting such URIs in Delta file actions and throwing a proper exception.

## How was this patch tested?

Add new UT.

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala
@@ -590,7 +590,15 @@ sealed trait FileAction extends Action {
   @JsonIgnore
   val tags: Map[String, String]
   @JsonIgnore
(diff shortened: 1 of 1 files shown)


larsk-db (approved):
LGTM

PR #3872, opened 2024-11-12
[Spark] Restrict partition-like data filters to whitelist of known-good expressions

chirag-s-db:
- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Currently, we try to rewrite any arbitrary expression as partition-like. To avoid having to repeatedly remove known-bad expressions, start with a whitelist (to be expanded) of known-good expressions that can safely be rewritten. 

This change will fix an existing issue where partition-like filters are generated for a non-skipping eligible column. This partition-like filter will throw an analysis exception because these referenced columns aren't found in the stats. This issue was originally missed (and is a difference in behavior vs. partition filters) because partitioning isn't allowed on non-atomic types (or string types), so we missed adding this additional match.

## How was this patch tested?
See test changes.

No.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingReader.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingReader.scala
@@ -598,6 +598,93 @@ trait DataSkippingReaderBase
         nullCountExpr: Expression)
 
     /**
+     * Whitelist of expressions that can be rewritten as partition-like.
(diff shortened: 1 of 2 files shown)


chirag-s-db:
@scovich Could you take a look at this PR? Thanks!

scovich (approved):
Good find, glad it first surfaced as a query error instead of an incorrect result!

PR #3882, opened 2024-11-15
[Kernel] Fix RoaringBitmapArray create/add methods. Closes issue #3881

### LINKED ISSUE #3881: [BUG][Kernel] RoaringBitmapArray create/add test methods are broken

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

### Describe the problem

io.delta.kernel.internal.deletionvectors.RoaringBitmapArray create/add methods are broken. Fail with NPE

There are two issues:

1. The `add` and `extendsBitMaps` methods expect the `bitmaps` field to be not null, at least should be an array of 0 length. But the instance created from the `create` method doesn't have the `bitmaps` field initialized.
2. The `extendsBitMaps` methods, copies the current bitmaps to the new bitmaps, to at the next step overwrite them with empty maps. The method should instead set empty bitmaps from the old length to the new length (instead of from 0 to the old length)

#### Steps to reproduce

```
    @Test
    void testCreate() {
        RoaringBitmapArray underTest = RoaringBitmapArray.create(1L);
        Assertions.assertTrue(underTest.contains(1L));
    }
```

#### Observed results

NPE is thrown:
```
java.lang.NullPointerException: Cannot read the array length because "this.bitmaps" is null

	at io.delta.kernel.internal.deletionvectors.RoaringBitmapArray.add(RoaringBitmapArray.java:262)
	at io.delta.kernel.internal.deletionvectors.RoaringBitmapArray.create(RoaringBitmapArray.java:290)
```

#### Expected results

Test pases

### Environment information

* Delta Lake version: 4.0.0rc1 
* Spark version:
* Scala version:

### Willingness to contribute

- [X] Yes. I can contribute a fix for this bug independently.

YotillaAntoni:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Resolves #3881 
 - The bitmaps array should be initialized in the `create` path.
 - The `expandBitMaps` method should set new bitmaps from the old length up to the new length, instead of overwriting the old ones.

Also adds a `toArray` method mimicking the one provided by the scala version which the class is based on.

## How was this patch tested?

Added unit tests.

No

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/deletionvectors/RoaringBitmapArray.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/deletionvectors/RoaringBitmapArray.java
@@ -23,6 +23,7 @@ import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.util.ArrayList;
+import org.roaringbitmap.IntIterator;
 import org.roaringbitmap.RoaringBitmap;
 
 /**
@@ -151,7 +152,7 @@ public final class RoaringBitmapArray {
    */
   static class NativeRoaringBitmapArraySerializationFormat {
     /** Magic number prefix for serialization with this format. */
-    static int MAGIC_NUMBER = 1681511376;
+    static final int MAGIC_NUMBER = 1681511376;
 
     /** Deserialize all bitmaps from the `buffer` into a fresh array. */
     static RoaringBitmap[] deserialize(ByteBuffer buffer) throws IOException {
@@ -186,7 +187,7 @@ public final class RoaringBitmapArray {
    */
   static class PortableRoaringBitmapArraySerializationFormat {
     /** Magic number prefix for serialization with this format. */
-    static int MAGIC_NUMBER = 1681511377;
+    static final int MAGIC_NUMBER = 1681511377;
 
     /** Deserialize all bitmaps from the `buffer` into a fresh array. */
     static RoaringBitmap[] deserialize(ByteBuffer buffer) throws IOException {
@@ -258,7 +259,7 @@ public final class RoaringBitmapArray {
         newBitmaps, // dest
         0, // dest start pos
         bitmaps.length); // number of entries to copy
-    for (int i = 0; i < bitmaps.length; i++) {
+    for (int i = bitmaps.length; i < newBitmaps.length; i++) {
       newBitmaps[i] = new RoaringBitmap();
     }
     bitmaps = newBitmaps;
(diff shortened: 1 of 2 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/deletionvectors/RoaringBitmapArray.java
```diff
@@ -258,17 +259,39 @@ private void extendBitmaps(int newLength) {
...
-    for (int i = 0; i < bitmaps.length; i++) {
+    for (int i = bitmaps.length; i < newBitmaps.length; i++) {
       newBitmaps[i] = new RoaringBitmap();
     }
     bitmaps = newBitmaps;
   }
 
   public static RoaringBitmapArray create(long... values) {
     RoaringBitmapArray bitmap = new RoaringBitmapArray();
+    bitmap.bitmaps = new RoaringBitmap[0];
```
allisonport-db, line 270: Instead of doing this here we should just initialize `bitmaps` to an empty array in the class. Since we don't have a constructor you can just do this on line 103

kernel/kernel-api/src/main/java/io/delta/kernel/internal/deletionvectors/RoaringBitmapArray.java
```diff
@@ -258,17 +259,39 @@ private void extendBitmaps(int newLength) {
...
     for (long value : values) {
       bitmap.add(value);
     }
     return bitmap;
   }
+
+  public long[] toArray() {
```
allisonport-db, line 277: Any particular reason for adding this?

I think I would prefer to either
(1) not add this for now since we don't use it
(2) if we are adding it, it needs more tests

I would lean toward (1) since it's not used but if you have a good reason for adding it lmk

allisonport-db (changes requested):
Thanks for doing this!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/deletionvectors/RoaringBitmapArray.java
YotillaAntoni: 👍  Good point

kernel/kernel-api/src/main/java/io/delta/kernel/internal/deletionvectors/RoaringBitmapArray.java
YotillaAntoni: One reason I think is that with the current methods available is not possible to assert that the contents of the data structure are exactly the expected. 
The unit test I added  tests the `contains` method for some particular values that are expected or not expected to be there, but it cannot completely discard that an unwanted value is present. With this method it can be asserted that the wanted values are there, and only them. 

Another one, is somehow related. Without this method any implementation of the scanning of files is constrained to use a filter approach to read the data, and call the `contains` for each single row. Depending on the number and the relation between total rows and deleted rows, and how the file is processed, I think it can be interesting to also allow the approach of reading the whole rows and then discarding the ones deleted.

allisonport-db: Okay I'm still okay with either (1) or (2). But if we are going with (2) can you please add full test coverage?

i.e. cover things like
- varios number of underlying bitmaps
- empty case
- etc

YotillaAntoni:
Incorporated the feedback. Increased the coverage, now it covers the following cases:
 - empty array
 - values only in the first bitmap
 - values only in the second bitmap
 - values only in the first and third bitmap

allisonport-db (approved):
LGTM

PR #3884, opened 2024-11-15
[SPARK] Tests for streaming use case of coordinated commits

ctring:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Add tests for the streaming use case of coordinated commits.

## How was this patch tested?
Unit tests

No


LukasRupprecht (approved):
Thanks for adding these tests!

Issue #3227, opened 2024-06-06
[BUG][Spark] INSERT INTO struct evolution in map/arrays breaks when a column is renamed


johanl-db:
### Describe the problem
Schema evolution in INSERT doesn't always work properly when the new column is added to a struct nested within an array or map. If another column is renamed, the operation fails when it should succeed.

#### Steps to reproduce
For example with a map, in python, renaming column `key` to `renamed_key` and added a field `comment` in the a struct inside the map:
```
(
  spark.createDataFrame([[1, {"event": (1, 1)}]], "key int, metrics map<string, struct<id: int, value: int>>")
    .write
    .option("overwriteSchema", "true")
    .format("delta")
    .saveAsTable("insert_map_schema_evolution")
)
(
  spark.createDataFrame([[1, {"event": (1, 1, "deprecated")}]], "renamed_key int, metrics map<string, struct<id: int, value: int, comment: string>>")
    .write
    .mode("append")
    .option("mergeSchema", "true")
    .format("delta")
    .insertInto("insert_map_schema_evolution")
)
[[DATATYPE_MISMATCH.CAST_WITHOUT_SUGGESTION](https://docs.databricks.com/error-messages/error-classes.html#datatype_mismatch.cast_without_suggestion)] Cannot resolve "metrics" due to data type mismatch: cannot cast "MAP<STRING, STRUCT<id: INT, value: INT, comment: STRING>>" to "MAP<STRING, STRUCT<id: INT, value: INT>>". 
```

Note that the struct inside the map isn't evolved to add the new field. Without the unrelated column being renamed, this works well:
```
(
  spark.createDataFrame([[1, {"event": (1, 1)}]], "key int, metrics map<string, struct<id: int, value: int>>")
    .write
    .option("overwriteSchema", "true")
    .format("delta")
    .saveAsTable("insert_map_schema_evolution")
)
(
  spark.createDataFrame([[1, {"event": (1, 1, "deprecated")}]], "key int, metrics map<string, struct<id: int, value: int, comment: string>>")
    .write
    .mode("append")
    .option("mergeSchema", "true")
    .format("delta")
    .insertInto("johan_lasperas.playground.insert_renamed_array_map")
)
```
#### Observed results
The operation fails.

#### Expected results
The operation succeeds, the table schema is changed to `key int, metrics map<string, struct<id: int, value: int, comment: string>>` and the data is inserted.

### Willingness to contribute

[...]

PR #3886, opened 2024-11-16
[Spark] Fix schema evolution issue with nested struct (within a map) and column renamed

Richard-code-gig:
This PR fixes an issue with schema evolution in Delta Lake where adding a new field to a struct within a map and renaming an existing top level field caused the operation to fail.

The fix includes logic to handle these transformations properly, ensuring that new fields are added without conflicts.

It also resolved a ToDo of casting map types in the [DeltaAnalysis.scala](https://github.com/Richard-code-gig/delta/blob/feature/schema-evolution-with-map-fix/spark/src/main/scala/org/apache/spark/sql/delta/DeltaAnalysis.scala) module.

### Changes:
- Updated schema evolution logic to support complex map transformations.
- Enabled schema evolution for both map keys, simple and nested values
- Added additional case statements to handle MapTypes in addCastToColumn method in DeltaAnalysis.scala module.
- Modified TypeWideningInsertSchemaEvolutionSuite test to support schema evolution of maps.
- Added an additional method (addCastsToMaps) to DeltaAnalysis.scala module.
- Changed argument type of addCastToColumn from attributes to namedExpression
- Added [EvolutionWithMap](https://github.com/Richard-code-gig/delta/blob/feature/schema-evolution-with-map-fix/examples/scala/src/main/scala/example/EvolutionWithMap.scala) in the example modules to demonstrate use case.
- Modified nested struct type evolution with field upcast test in map in TypeWideningInsertSchemaEvolutionSuite.scala
- Added new tests cases for maps to DeltaInsertIntoTableSuite.scala

### Related Issues:
- Resolves: #3227 

- [✓] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## How was this patch tested?
Tested through:
- Integration Tests: Validated changes with Delta Lake and Spark integration. See [EvolutionWithMap](https://github.com/Richard-code-gig/delta/blob/feature/schema-evolution-with-map-fix/examples/scala/src/main/scala/example/EvolutionWithMap.scala).
- Validated the test suites passed and [TypeWideningInsertSchemaEvolutionSuite](https://github.com/Richard-code-gig/delta/blob/feature/schema-evolution-with-map-fix/spark/src/test/scala/org/apache/spark/sql/delta/typewidening/TypeWideningInsertSchemaEvolutionSuite.scala) to add support for maps.
- Added additional tests cases in [DeltaInsertIntoTableSuite](https://github.com/Richard-code-gig/delta/blob/feature/schema-evolution-with-map-fix/spark/src/test/scala/org/apache/spark/sql/DeltaInsertIntoTableSuite.scala) to cover complex map transformations

No, it doesn't introduce any user-facing changes. It only resolved an issue even in the released versions of Delta Lake.

The previous behaviour was an error message when attempting operations involving adding extra fields to StructField in maps:
[[DATATYPE_MISMATCH.CAST_WITHOUT_SUGGESTION](https://docs.databricks.com/error-messages/error-classes.html#datatype_mismatch.cast_without_suggestion)] Cannot resolve "metrics" due to data type mismatch: cannot cast "MAP<STRING, STRUCT<id: INT, value: INT, comment: STRING>>" to "MAP<STRING, STRUCT<id: INT, value: INT>>".

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaAnalysis.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaAnalysis.scala
@@ -933,7 +933,7 @@ class DeltaAnalysis(session: SparkSession)
         attr
       case (s: MapType, t: MapType)
         if !DataType.equalsStructurally(s, t, ignoreNullability = true) || allowTypeWidening =>
-        // only trigger addCastsToMaps if exits differences like extra fields, renaming
+        // only trigger addCastsToMaps if exists differences like extra fields, renaming
         // Or allowTypeWidening is enabled
         addCastsToMaps(tblName, attr, s, t, allowTypeWidening)
       case _ =>
@@ -1138,24 +1138,23 @@ class DeltaAnalysis(session: SparkSession)
       sourceMapType: MapType,
       targetMapType: MapType,
       allowTypeWidening: Boolean): Expression = {
-
     val transformedKeys =
       if (sourceMapType.keyType != targetMapType.keyType) {
         // Create a transformation for the keys
         ArrayTransform(MapKeys(parent), {
           val key = NamedLambdaVariable(
-                  "key", sourceMapType.keyType, nullable = false)
+            "key", sourceMapType.keyType, nullable = false)
 
           val keyAttr = AttributeReference(
-              "key", targetMapType.keyType, nullable = false)()
+            "key", targetMapType.keyType, nullable = false)()
 
           val castedKey =
-              addCastToColumn(
-                key,
-                keyAttr,
-                tableName,
-                allowTypeWidening
-              )
+            addCastToColumn(
+              key,
+              keyAttr,
+              tableName,
+              allowTypeWidening
+            )
           LambdaFunction(castedKey, Seq(key))
         })
       } else {
@@ -1167,18 +1166,18 @@ class DeltaAnalysis(session: SparkSession)
         // Create a transformation for the values
         ArrayTransform(MapValues(parent), {
           val value = NamedLambdaVariable(
-              "value", sourceMapType.valueType, sourceMapType.valueContainsNull)
+            "value", sourceMapType.valueType, sourceMapType.valueContainsNull)
 
           val valueAttr = AttributeReference(
-              "value", targetMapType.valueType, sourceMapType.valueContainsNull)()
+            "value", targetMapType.valueType, sourceMapType.valueContainsNull)()
 
           val castedValue =
-              addCastToColumn(
-                value,
-                valueAttr,
-                tableName,
-                allowTypeWidening
-              )
+            addCastToColumn(
+              value,
+              valueAttr,
+              tableName,
+              allowTypeWidening
+            )
           LambdaFunction(castedValue, Seq(value))
         })
       } else {
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaAnalysis.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaAnalysis.scala
@@ -69,6 +69,7 @@ import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.types._
 import org.apache.spark.sql.util.CaseInsensitiveStringMap
 
+
 /**
  * Analysis rules for Delta. Currently, these rules enable schema enforcement / evolution with
  * INSERT INTO.
@@ -913,8 +914,8 @@ class DeltaAnalysis(session: SparkSession)
   }
 
   private def addCastToColumn(
-      attr: Attribute,
-      targetAttr: Attribute,
+      attr: NamedExpression,
+      targetAttr: NamedExpression,
       tblName: String,
       allowTypeWidening: Boolean): NamedExpression = {
     val expr = (attr.dataType, targetAttr.dataType) match {
@@ -930,6 +931,11 @@ class DeltaAnalysis(session: SparkSession)
         // Keep the type from the query, the target schema will be updated to widen the existing
         // type to match it.
         attr
+      case (s: MapType, t: MapType)
+        if !DataType.equalsStructurally(s, t, ignoreNullability = true) || allowTypeWidening =>
+        // only trigger addCastsToMaps if exits differences like extra fields, renaming
+        // Or allowTypeWidening is enabled
+        addCastsToMaps(tblName, attr, s, t, allowTypeWidening)
       case _ =>
         getCastFunction(attr, targetAttr.dataType, targetAttr.name)
     }
@@ -1047,8 +1053,7 @@ class DeltaAnalysis(session: SparkSession)
   }
 
   /**
-   * Recursively casts structs in case it contains null types.
-   * TODO: Support other complex types like MapType and ArrayType
+   * Recursively casts struct data types in case the source/target type differs.
    */
   private def addCastsToStructs(
       tableName: String,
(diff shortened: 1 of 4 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/DeltaAnalysis.scala
```diff
@@ -1124,6 +1140,48 @@ class DeltaAnalysis(session: SparkSession)
     DeltaViewHelper.stripTempViewForMerge(plan, conf)
   }
 
+  /**
+   * Recursively casts maps in case it contains null types.
```
johanl-db, line 1144: "Recursively casts map data types in case the key/value type differs"

I'm not entirely sure why the struct method mentions `in case it contains null types` but I don't think that's accurate

spark/src/main/scala/org/apache/spark/sql/delta/DeltaAnalysis.scala
```diff
@@ -1049,6 +1062,7 @@ class DeltaAnalysis(session: SparkSession)
   /**
    * Recursively casts structs in case it contains null types.
    * TODO: Support other complex types like MapType and ArrayType
+   * The case mapType that calls addCastsToMaps addresses the MapType todo
```
johanl-db, line 1065: I would just remove the TODO at this point. Both arrays and maps are now (partially) handled. We can call out their respective limitations on the dedicated methods if needed

spark/src/main/scala/org/apache/spark/sql/delta/DeltaAnalysis.scala
```diff
@@ -1067,6 +1081,8 @@ class DeltaAnalysis(session: SparkSession)
             val subField = Alias(GetStructField(parent, i, Option(name)), target(i).name)(
               explicitMetadata = Option(metadata))
             addCastsToStructs(tableName, subField, nested, t, allowTypeWidening)
+          // We could also handle maptype within struct here but there is restriction
+          // on deep nexted operations that may result in maxIteration error
```
johanl-db, line 1085: I'm curious, do you have the stack trace for the error you're seeing?

It should be possible to handle arbitrary nesting of maps/arrays/structs. although that might require a pretty deep refactor of this code and a high risk of introducing regressions or breaking changes that would likely negate the benefits

spark/src/test/scala/org/apache/spark/sql/delta/typewidening/TypeWideningInsertSchemaEvolutionSuite.scala
```diff
@@ -297,8 +297,7 @@ trait TypeWideningInsertSchemaEvolutionTests
           metadata = typeWideningMetadata(version = 1, from = ShortType, to = IntegerType)))))
   )
 
-  // The next two tests document inconsistencies when handling maps. Using SQL or INSERT by position
-  // doesn't allow type evolution but using dataframe INSERT by name does.
+  // maps now allow type evolution for INSERT by position and name in SQL and dataframe.
```
johanl-db, line 300: It would be good to add more tests, for example in [DeltaInsertIntoSQLSuite](https://github.com/delta-io/delta/blob/3a98b8a33a48416649d0844355acffceca4ae21c/spark/src/test/scala/org/apache/spark/sql/delta/DeltaInsertIntoTableSuite.scala#L44):
- add a test covering the initial issue from https://github.com/delta-io/delta/issues/3227 - essentially a version of the example you provide, except that I don't think we run examples as part of PR tests (I might be wrong)
- add tests covering inserting data into a map type using key/values that have (1) a different type than in the table, (2) more column than in the table with schema evolution enabled/disabled
- if supporting maps/arrays nested inside maps: add a test covering 2 level deep nested maps

spark/src/test/scala/org/apache/spark/sql/delta/typewidening/TypeWideningInsertSchemaEvolutionSuite.scala
```diff
@@ -310,10 +309,11 @@ trait TypeWideningInsertSchemaEvolutionTests
       Seq("""{ "key": 1, "m": { "a": { "x": 3, "y": 4 } } }""")),
     expectedResult = ExpectedResult.Success(new StructType()
       .add("key", IntegerType)
-      // Type evolution wasn't applied in the map.
+      // Type evolution now works for map.
       .add("m", MapType(StringType, new StructType()
         .add("x", IntegerType)
-        .add("y", ShortType)))),
+        .add("y", IntegerType, nullable = true,
+          metadata = typeWideningMetadata(version = 1, from = ShortType, to = IntegerType))))),
```
johanl-db, line 316: You can actually remove this test case altogether and update the one below to remove `includeInserts` so that it will cover all insert types.

johanl-db (commented):
Thanks @Richard-code-gig for taking the time to work on this!

The code looks good, two main points I see to address before we can merge this:
- Support casting map keys and not just map values. It could be surprising for users otherwise to see that one works but not the other.
- Add a few more test cases, I provided some ideas.

nice to have but not required:
- support deeply nested maps
that's arguably a less common use case and I don't know if it's going to be straightforward or not so it's ok if we don't support it.

Also, ignore the Delta Spark Master job failure, it has been broken for quite some time and won't be a blocker to merge this change

Richard-code-gig:
Thanks for the insight @johanl-db.
I will try get the stack trace for maxIteration error too

spark/src/main/scala/org/apache/spark/sql/delta/DeltaAnalysis.scala
```diff
@@ -930,6 +931,18 @@ class DeltaAnalysis(session: SparkSession)
         // Keep the type from the query, the target schema will be updated to widen the existing
         // type to match it.
         attr
+      case (s: MapType, t: MapType) if s != t =>
```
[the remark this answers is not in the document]
Richard-code-gig: Hi @johanl-db,
Should we be calling addCastsToColumn here especially as addCastsToColumn takes an attribute type for both parent and targetAttr arguments instead of NamedLambdaVariable.

Looking forward to your thoughts.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaInsertIntoTableSuite.scala
```diff
@@ -281,6 +281,219 @@ class DeltaInsertIntoSQLSuite
...
+      val evolvedRdd = spark.sparkContext.parallelize(evolvedData)
+      val evolvedDf = spark.createDataFrame(evolvedRdd, evolvedSchema)
+
+      // insert data without schema evolution
+      assert(intercept[AnalysisException] {
+        evolvedDf.write
+          .mode("append")
+          .format("delta")
+          .insertInto(tableName)
+      }.getMessage.contains("A schema mismatch detected when writing to the Delta table")
```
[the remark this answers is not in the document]
Richard-code-gig: Hi Johan,
Yea, I agree with you on the comments and suggestions. Hopefully it is ready now for merging when you're back.
Enjoy your holidays :)

johanl-db (approved):
Nice work @Richard-code-gig!

These aren't blocker, you can address them if you want, otherwise give a heads up to @vkorukanti who can merge the PR on your behalf

spark/src/main/scala/org/apache/spark/sql/delta/DeltaAnalysis.scala
```diff
@@ -1124,6 +1129,65 @@ class DeltaAnalysis(session: SparkSession)
...
+          val key = NamedLambdaVariable(
+                  "key", sourceMapType.keyType, nullable = false)
+
+          val keyAttr = AttributeReference(
+              "key", targetMapType.keyType, nullable = false)()
+
+          val castedKey =
+              addCastToColumn(
+                key,
+                keyAttr,
```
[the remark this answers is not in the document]
Richard-code-gig: Welcome back Johanl-db.
I can address the style issues after work this evening.
The key.toAttribute is causing the maxIteration error I wrote about earlier.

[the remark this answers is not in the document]
johanl-db: Ah right, that's probably because the data type isn't correct with `key.toAttribute` - it's the source type where it should be the target type - and that trips the analyzer

Richard-code-gig:
> Nice work @Richard-code-gig!
> 
> These aren't blocker, you can address them if you want, otherwise give a heads up to @vkorukanti who can merge the PR on your behalf

Hi @johanl-db  @vkorukanti I've resolved the style issues. It can be merged now.

PR #3899, opened 2024-11-23
[Spark][Version Checksum] Incrementally compute allFiles in the checksum

dhruvarya-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Follow up for https://github.com/delta-io/delta/pull/3895. Adds incremental computation of the allFiles field in the VersionChecksum. There are only included until a threshold (defined using a SQL conf) is reached. This field is later used for validation.

## How was this patch tested?

Added DeltaAllFilesInCRCSuite

No

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/delta/ChecksumSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/ChecksumSuite.scala
@@ -17,6 +17,7 @@
 package org.apache.spark.sql.delta
 
 import java.io.File
+import java.util.TimeZone
 
 import com.databricks.spark.util.Log4jUsageLogger
 import org.apache.spark.sql.delta.DeltaTestUtils._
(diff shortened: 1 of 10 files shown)


spark/src/test/scala/org/apache/spark/sql/delta/ChecksumSuite.scala
```diff
@@ -74,7 +74,10 @@ class ChecksumSuite
     for (incrementalCommitEnabled <- BOOLEAN_DOMAIN) {
       withSQLConf(
         DeltaSQLConf.DELTA_WRITE_CHECKSUM_ENABLED.key -> "false",
-        DeltaSQLConf.INCREMENTAL_COMMIT_ENABLED.key -> incrementalCommitEnabled.toString) {
+        DeltaSQLConf.INCREMENTAL_COMMIT_ENABLED.key -> incrementalCommitEnabled.toString,
+        DeltaSQLConf.DELTA_ALL_FILES_IN_CRC_FORCE_VERIFICATION_MODE_FOR_NON_UTC_ENABLED.key ->
```
prakharjain09, line 78: Why this needs to be set here?

prakharjain09 (approved):

spark/src/test/scala/org/apache/spark/sql/delta/ChecksumSuite.scala
dhruvarya-db: If the environment where this test runs has a non-UTC timezone set, the commit will trigger a forced verification of all files without this flag being disabled. To make this more explicit, I can explicitly set the time zone to UTC in the test and remove this flag override.

dhruvarya-db: Done

PR #3903, opened 2024-11-26
[Kernel][Metrics][PR#2] Add SnapshotReport for reporting snapshot construction

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Adds a `SnapshotReport` for reporting snapshot construction.

We record a `SnapshotReport` after successfully constructing a snapshot or if an exception is thrown during construction. We use `SnapshotContext` to propagate and update information about the snapshot construction. For example, we update the "version" as soon as it's resolved for time-travel by timestamp or load latest snapshot queries. This means in the case of the exception we include as much information as is available.

## How was this patch tested?

Adds a test suite `MetricsReportSuite` with unit tests.

No.


kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/SnapshotContext.java
```diff
@@ -0,0 +1,67 @@
...
+  public String getTablePath() {
+    return tablePath;
+  }
+
+  public Optional<Long> getProvidedTimestamp() {
+    return providedTimestamp;
+  }
+
+  public void setVersion(long updatedVersion) {
+    version = Optional.of(updatedVersion);
```
scottsand-db, line 65: Can we please make this class immutable? And just create a new instance with this new value instead? https://cs108.epfl.ch/archive/19/c/i/EffectiveJava_Item17.pdf

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/SnapshotReportImpl.java
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  @Override
+  public String tablePath() {
+    return tablePath;
+  }
+
+  @Override
+  public String operationType() {
+    return OPERATION_TYPE;
```
scottsand-db, line 57: Can you do `SnapshotReport.OPERATION_TYPE`? That's easier to read / reason about where this value comes from

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/SnapshotContext.java
scottsand-db: Makes sense! Thanks!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/SnapshotContext.java
```diff
@@ -0,0 +1,67 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.delta.kernel.internal.metrics;
+
+import io.delta.kernel.metrics.SnapshotReport;
+import java.util.Optional;
+
+/** Stores the context for a given Snapshot query. Used to generate a {@link SnapshotReport} */
+public class SnapshotContext {
```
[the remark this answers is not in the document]
allisonport-db: SGTM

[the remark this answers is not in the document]
allisonport-db: added toString; I don't think we will ever need equality for something like this

kernel/kernel-api/src/main/java/io/delta/kernel/metrics/SnapshotMetricsResult.java
```diff
@@ -0,0 +1,61 @@
...
+   */
+  Optional<Long> timestampToVersionResolutionDuration();
+
+  /**
+   * @return the duration (ns) to load the initial delta actions for the snapshot (such as the table
+   *     protocol and metadata). 0 if snapshot construction fails before log replay.
+   */
+  long loadInitialDeltaActionsDuration();
+
+  static SnapshotMetricsResult fromSnapshotMetrics(SnapshotMetrics snapshotMetrics) {
```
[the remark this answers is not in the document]
allisonport-db: Moved it to `SnapshotMetrics`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -105,8 +107,11 @@ public Transaction build(Engine engine) {
       // Table doesn't exist yet. Create an initial snapshot with the new schema.
       Metadata metadata = getInitialMetadata();
       Protocol protocol = getInitialProtocol();
-      LogReplay logReplay = getEmptyLogReplay(engine, metadata, protocol);
-      snapshot = new InitialSnapshot(table.getDataPath(), logReplay, metadata, protocol);
+      SnapshotContext snapshotContext = SnapshotContext.forVersionSnapshot(tablePath, 0);
```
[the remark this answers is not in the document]
allisonport-db: I updated this, my mistake this should've been -1 (which matches the version returned by the log segment & log replay).

In general it is the intended read version. For this specific case, we treat a non-existent table as read version -1. Note, we don't generate a snapshot report for this case since there is no real snapshot to load!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -532,15 +556,24 @@ private SnapshotImpl createSnapshot(LogSegment initSegment, Engine engine) {
...
-            tablePath, initSegment, logReplay, logReplay.getProtocol(), logReplay.getMetadata());
+            tablePath,
+            initSegment,
+            logReplay,
+            logReplay.getProtocol(),
+            logReplay.getMetadata(),
+            snapshotContext);
+
+    // Push snapshot report to engine
+    engine.getMetricsReporters().forEach(reporter -> reporter.report(snapshot.getSnapshotReport()));
```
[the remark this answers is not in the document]
allisonport-db: I think as soon as the given operation is completed. In this case, as soon as a snapshot is successfully constructed. I viewed this to essentially mean as soon as the metadata/protocol is loaded.

scottsand-db (changes requested):
Just posting what we discussed in a sync! Looks great!

1. Would be great to use factory methods to create `SnapshotReportImpl`s. One for success/normal cases, one for error case. This should take in the `SnapshotQueryContext`.

2. Ensure you have the same exact method / value names between SnapshotMetrics and SnapshotMetricsResult. Also, add docs to both that explain the relationship. i.e. SnapshotMetrics creates SnapshotMetricsResult; SnapshotMetricsResult is created by SnapshotMetrics

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/MetricsReportSuite.scala
```diff
@@ -0,0 +1,379 @@
...
+        expectNonZeroLoadProtocolAndMetadataDuration = true
+      )
+
+      // Test getSnapshotAsOfVersion
+      checkSnapshotReport(
+        (table, engine) => table.getSnapshotAsOfVersion(engine, 0),
+        path,
+        expectException = false,
+        expectedVersion = Optional.of(0),
+        expectedProvidedTimestamp = Optional.empty(), // No time travel
```
scottsand-db, line 173: Didn't we time travel to version 0? Would `// No time travel by timestamp` be a more accurate comment? And I think this would apply to all the other such instances of this comment, right?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/MetricsReportSuite.scala
```diff
@@ -0,0 +1,379 @@
...
+    }
+    if (expectNonZeroLoadProtocolAndMetadataDuration) {
+      assert(snapshotReport.snapshotMetrics.loadInitialDeltaActionsDuration > 0)
+      assert(snapshotReport.snapshotMetrics.loadInitialDeltaActionsDuration < duration)
+    } else {
+      assert(snapshotReport.snapshotMetrics.loadInitialDeltaActionsDuration == 0)
+    }
+  }
+
+  test("SnapshotReport valid queries") {
```
scottsand-db, line 144: Could you also test: try and time travel by a bad version (e.g. version 55) or by a bad timestamp (e.g. the year 1985) but on a valid table?

e.g. you could rename this test to `SnapshotReport queries on valid table`

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/MetricsReportSuite.scala
```diff
@@ -0,0 +1,379 @@
...
+
+/**
+ * Test suite to test the Kernel-API created [[MetricsReport]]s. This suite is in the defaults
+ * package to be able to use real tables and avoid having to mock both file listings AND file
+ * contents.
+ */
+class MetricsReportSuite extends AnyFunSuite with TestUtils {
+
+  ///////////////////////////
+  // SnapshotReport tests //
```
scottsand-db, line 42: How many different types of reports do you anticipate we will have? If test + helper code for just SnapshotReport is ~300 LOC, I wonder if we should split this up into different files?

e.g. `SnapshotMetricsReportSuite extends MetricsReportTestUtils`

scottsand-db (changes requested):

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/MetricsReportSuite.scala
allisonport-db: So originally I did not add these tests because my main goal with the exception cases was to cover the 3 possible places snapshot construction could fail (outlined [here](https://docs.google.com/document/d/1W1NTmHg14qYq-QEIcDtFFSMy3B0GjrV82ll6BtX21fE/edit?tab=t.0#heading=h.wsxjbfksl0zh) in design doc) instead of trying to cover all possible errors/exceptions during snapshot construction.
- Error resolving timestamp --> version (only applicable for time-travel by timestamp)
- Error building the log segment
- Error constructing the snapshot (loading table protocol and metadata)

But I'm happy to add these two cases anyways. I can just add an additional test for them.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -172,32 +176,44 @@ public Snapshot getSnapshotAt(Engine engine, long version) throws TableNotFoundE
...
+                            logPath,
+                            millisSinceEpochUTC,
+                            true /* mustBeRecreatable */,
+                            false /* canReturnLastCommit */,
+                            false /* canReturnEarliestCommit */)
+                        .getVersion());
     logger.info(
         "{}: Took {}ms to fetch version at timestamp {}",
         tablePath,
         System.currentTimeMillis() - startTimeMillis,
```
scottsand-db, line 200: Instead of creating the local variable `startTimeMillis` could we just use the `timestampToVersionResolutionDuration` that we just set? Only if that would improve the cleanliness and reduce the amount of code ...

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/MetricsReportSuite.scala
allisonport-db: Tests have been added

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/MetricsReportSuite.scala
allisonport-db: I think for now this is okay. I think the other report types will have fewer tests (much fewer error cases). If it ends up being a lot in a later PR I can separate the suites then.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
allisonport-db: I updated it to do this

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
```diff
@@ -90,19 +93,39 @@ public String getPath(Engine engine) {
 
   @Override
   public Snapshot getLatestSnapshot(Engine engine) throws TableNotFoundException {
-    return snapshotManager.buildLatestSnapshot(engine);
+    SnapshotQueryContext snapshotContext = SnapshotQueryContext.forLatestSnapshot(tablePath);
```
scottsand-db, line 96: Just curious (sorry if I asked this before the winter holiday's and then just forgot the answer):

Any reason why we do this try and catch inside `TableImpl::getLatestSnapshot` instead of `SnapshotManager::buildLatestSnapshot` ?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
```diff
@@ -316,4 +339,11 @@ private CloseableIterator<ColumnarBatch> getRawChanges(
     logger.info("{}: Reading the commit files with readSchema {}", tablePath, readSchema);
     return DeltaLogActionUtils.readCommitFiles(engine, commitFiles, readSchema);
   }
+
+  /** Creates a {@link SnapshotReport} and pushes it to any {@link MetricsReporter}s. */
+  private void recordSnapshotReport(
```
scottsand-db, line 344: `recordSnapshotReportError`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/SnapshotMetrics.java
```diff
@@ -0,0 +1,64 @@
...
+public class SnapshotMetrics {
+
+  public final Timer timestampToVersionResolutionDuration = new Timer();
+
+  public final Timer loadInitialDeltaActionsDuration = new Timer();
+
+  public SnapshotMetricsResult captureSnapshotMetricsResult() {
+    return new SnapshotMetricsResult() {
+
+      final Optional<Long> timestampToVersionResolutionDurationResult =
```
scottsand-db, line 38: how about we add a method inside `Timer.java` like

```java
/** add javadoc ... */
Optional<Long> totalDurationIfRecorded() {
  return count() > 0 ? Optional.of(totalDuration()) : Optional.empty();
}
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/SnapshotQueryContext.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Optional;
+
+/** Stores the context for a given Snapshot query. Used to generate a {@link SnapshotReport} */
+public class SnapshotQueryContext {
+
+  public static SnapshotQueryContext forLatestSnapshot(String tablePath) {
+    return new SnapshotQueryContext(tablePath, Optional.empty(), Optional.empty());
+  }
+
+  public static SnapshotQueryContext forVersionSnapshot(String tablePath, long version) {
```
scottsand-db, line 28: `/** Creates a SnapshotQueryContext for a Snapshot created by a AS OF VERSION query */`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/SnapshotQueryContext.java
```diff
@@ -0,0 +1,74 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.delta.kernel.internal.metrics;
+
+import io.delta.kernel.metrics.SnapshotReport;
+import java.util.Optional;
+
+/** Stores the context for a given Snapshot query. Used to generate a {@link SnapshotReport} */
```
scottsand-db, line 21: This class is named `SnapshotQueryContext` ... and you say that it "stores the context" ... but can you define a bit what "context" means? You don't need to list all of the information it contains .. but at high level, what does context mean?

```
Stores the context for a given Snapshot query. Used to generate a {@link SnapshotReport}.

Here, "context" means the state and information pertaining to when a Snapshot was constructed, for which table, and its associated metrics
```

scottsand-db: Can you also talk a bit about the expected lifecycle of this class. What is it coupled to?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/SnapshotQueryContext.java
```diff
@@ -0,0 +1,74 @@
...
+
+  public Optional<Long> getProvidedTimestamp() {
+    return providedTimestamp;
+  }
+
+  public SnapshotMetrics getSnapshotMetrics() {
+    return snapshotMetrics;
+  }
+
+  public void setVersion(long updatedVersion) {
```
scottsand-db, line 64: Can you give some context either here or at the top level class docs of when we would change the version of this context? (i.e. does this relate to its lifecycle?)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/SnapshotQueryContext.java
```diff
@@ -0,0 +1,74 @@
...
+  public static SnapshotQueryContext forVersionSnapshot(String tablePath, long version) {
+    return new SnapshotQueryContext(tablePath, Optional.of(version), Optional.empty());
+  }
+
+  public static SnapshotQueryContext forTimestampSnapshot(String tablePath, long timestamp) {
+    return new SnapshotQueryContext(tablePath, Optional.empty(), Optional.of(timestamp));
+  }
+
+  private final String tablePath;
+  private Optional<Long> version;
```
scottsand-db, line 37: So I'm guessing this is not `providedVersion` because we expect this `version` variable to change ....

Could the constructor parameter be called `providedVersion`? Also worth some javadocs here for these variables. i.e. what does "provided" mean? (I know, but might as well add the context for future readers)

kernel/kernel-api/src/main/java/io/delta/kernel/metrics/SnapshotReport.java
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+package io.delta.kernel.metrics;
+
+import java.util.Optional;
+
+/** Defines the metadata and metrics for a snapshot construction {@link MetricsReport} */
+public interface SnapshotReport extends DeltaOperationReport {
+
+  String OPERATION_TYPE = "Snapshot";
```
scottsand-db, line 23: I think all children of `SnapshotReport` will have operationType of `Snapshot`, right?

I think you can just define the default here.

```java
  @Override
  default void operationType() {
    return "Snapshot";
  }
```

scottsand-db: I'm also curious if we would ever want to see a list (to check against, compare against, etc.) of all possible operations?

would we want this to instead be an Enum inside of DeltaOperationReport?

scottsand-db (changes requested):
Looks great! Left some comments and questions. Thanks!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
allisonport-db: I think it's probably just a matter of preference. If we do it in SnapshotManager we still have to do it for all 3 access patterns.

I think I put it here since it makes sense to do it at the outermost layer and catch any exceptions thrown by the nested functions.

kernel/kernel-api/src/main/java/io/delta/kernel/metrics/SnapshotReport.java
allisonport-db: I'm not sure of an exact use case right now but it seems useful enough to me

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/SnapshotQueryContext.java
allisonport-db: Updated the class docs

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/SnapshotQueryContext.java
allisonport-db: Added

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/SnapshotQueryContext.java
allisonport-db: Makes sense updated the param name & added docs

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -154,32 +157,44 @@ public Snapshot getSnapshotAt(Engine engine, long version) throws TableNotFoundE
...
+                            true /* mustBeRecreatable */,
+                            false /* canReturnLastCommit */,
+                            false /* canReturnEarliestCommit */)
+                        .getVersion());
     logger.info(
-        "{}: Took {}ms to fetch version at timestamp {}",
+        "{}: Took {}ns to fetch version at timestamp {}",
         tablePath,
-        System.currentTimeMillis() - startTimeMillis,
+        snapshotContext.getSnapshotMetrics().timestampToVersionResolutionDuration.totalDuration(),
```
scottsand-db, line 180: Offline discussion about consistency of logging nanos vs logging millis ...

kernel/kernel-api/src/main/java/io/delta/kernel/metrics/SnapshotReport.java
```diff
@@ -0,0 +1,48 @@
...
+   *     time-travel query
+   */
+  Optional<Long> providedTimestamp();
+
+  /** @return the metrics for this snapshot construction */
+  SnapshotMetricsResult snapshotMetrics();
+
+  @Override
+  default String operationType() {
+    return DeltaOperationReport.OperationType.SNAPSHOT.toString();
```
scottsand-db, line 46: Hm. If we return `String` still I wonder if the enum isn't as useful? what if we returned the enum value instead?

i.e. how do we expect this `operationType()` method to be used?

I'm fine with whatever you decide, no need to block.

scottsand-db (changes requested):
Looks great! Minor comments mainly on variable/api naming. After this LGTM

kernel/kernel-api/src/main/java/io/delta/kernel/metrics/SnapshotReport.java
allisonport-db: My thought is that it's useful to go to `DeltaOperationType.OperationType` and see all the possible operation types we might encounter. I think it makes sense to be string instead of the enum to try to keep the metric return types to standard built in types.

scottsand-db: > My thought is that it's useful to go to DeltaOperationReport.OperationType and see all the possible operation types we might encounter.

But there's nothing enforcing this, right? It's just declaring an enum -- children of `DeltaOperationReport` don't actually _need_ to add a new type to this enum, they could just return an arbitrary string for `String operationType()`.

If we had `operationType` return an enum, then it's fair to require and expect that children will add new values for the enum.

scottsand-db: > But maybe the answer to that is we don't need the enum and it's possible to just look at all the metrics interfaces we have added.

I think the question is: how will the result of `operationType` be used? is it just for printing? do we want to check for a particular value (in this case, an enum would be nice)

allisonport-db: Hmm. I think we would probably wouldn't check for a specific value. If you want a snapshot report you would check if it's an instance of `SnapshotReport`.

I think i'll remove the enum and keep `operationType` to return a string.

scottsand-db: SGTM!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/Timer.java
```diff
@@ -18,12 +18,17 @@
...
+import java.util.Optional;
 import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.LongAdder;
 import java.util.function.Supplier;
 
 /** A timer class for measuring the duration of operations in nanoseconds */
 public class Timer {
+
+  private static long NANOS_PER_MILLISECOND = 1000000;
```
scottsand-db, line 30: unused variable? (nit: and `final`?)

allisonport-db: oops good catch

scottsand-db (approved):

PR #3904, opened 2024-11-26
[Kernel][Metrics][PR#3] Metrics report JSON serializer and LoggingMetricsReporter for the default engine

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

This PR is based off of https://github.com/delta-io/delta/pull/3903
See the diff for just this PR [here](https://github.com/delta-io/delta/pull/3904/files/aec95cf3dc0086c37f4c45e2b3e192b7b881768c..678ac473f4de65a8f7fd770696aad2d31a15aef7)

Adds a JSON serializer for metrics reports with serialization logic for SnapshotReport. Also adds a `LoggingMetricsReporter` to the default implementation which simply logs the JSON serialized reports using Log4J.

## How was this patch tested?

Adds a test suite.

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/MetricsReportSerializers.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/MetricsReportSerializers.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright (2024) The Delta Lake Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.delta.kernel.internal.metrics;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import com.fasterxml.jackson.databind.ser.std.ToStringSerializer;
+import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
+import io.delta.kernel.metrics.MetricsReport;
+import io.delta.kernel.metrics.SnapshotReport;
+
+/** Defines JSON serializers for {@link MetricsReport} types */
+public final class MetricsReportSerializers {
+
+  /////////////////
+  // Public APIs //
+  /////////////////
+
+  /**
+   * Serializes a {@link SnapshotReport} to a JSON string
+   *
+   * @throws JsonProcessingException
+   */
+  public static String serializeSnapshotReport(SnapshotReport snapshotReport)
+      throws JsonProcessingException {
+    return OBJECT_MAPPER.writeValueAsString(snapshotReport);
+  }
+
+  /////////////////////////////////
+  // Private fields and methods //
+  ////////////////////////////////
+
+  private static final ObjectMapper OBJECT_MAPPER =
+      new ObjectMapper()
+          .registerModule(new Jdk8Module()) // To support Optional
+          .registerModule( // Serialize Exception using toString()
+              new SimpleModule().addSerializer(Exception.class, new ToStringSerializer()));
+
+  private MetricsReportSerializers() {}
+}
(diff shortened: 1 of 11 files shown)


kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/engine/LoggingMetricsReporter.java
```diff
@@ -0,0 +1,51 @@
...
+public class LoggingMetricsReporter implements MetricsReporter {
+
+  private static final Logger logger = LoggerFactory.getLogger(SnapshotManager.class);
+
+  @Override
+  public void report(MetricsReport report) {
+    try {
+      if (report instanceof SnapshotReport) {
+        logger.info(
+            "SnapshotReport = %s",
```
scottsand-db, line 40: aren't we supposed to use {} instead of %s with `logger.info`?

kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/engine/LoggingMetricsReporter.java
```diff
@@ -0,0 +1,51 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * An implementation of {@link MetricsReporter} that logs the reports (as JSON) to Log4J at the info
+ * level.
+ */
+public class LoggingMetricsReporter implements MetricsReporter {
+
+  private static final Logger logger = LoggerFactory.getLogger(SnapshotManager.class);
```
scottsand-db, line 33: `LoggingMetricsReporter.class`

scottsand-db (changes requested):
Looks great! 2 minor comments

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/MetricsReportSerializers.java
```diff
@@ -0,0 +1,54 @@
...
+  /////////////////
+  // Public APIs //
+  /////////////////
+
+  /**
+   * Serializes a {@link SnapshotReport} to a JSON string
+   *
+   * @throws JsonProcessingException
+   */
+  public static String serializeSnapshotReport(SnapshotReport snapshotReport)
```
scottsand-db, line 38: Seems like we would have to do this for every type of report, right?

Any way to avoid that? (Since they would all just call `OBJECT_MAPPER.writeValueAsString(inputVariable)`

Could we just take in a DeltaOperationReport?

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/metrics/MetricsReportSerializerSuite.scala
```diff
@@ -0,0 +1,97 @@
...
+        s""""${option.get()}"""" // For string objects wrap with quotes
+      } else {
+        option.get().toString
+      }
+    } else {
+      "null"
+    }
+  }
+
+  def testSnapshotReport(snapshotReport: SnapshotReport): Unit = {
```
scottsand-db, line 37: nit: private?

scottsand-db (approved):
LGTM with two minor comments!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/MetricsReportSerializers.java
allisonport-db: yeah I wasn't sure about this because I also think there's a use-case where you are only serializing a specific report type. maybe we can have both `serializeSnapshotReport` and `serializeDeltaOperationReport`

But then I need to make `DeltaOperationReport` serializable (which currently I did not). What do you think?

scottsand-db: Every report we create will implement DeltaOperationReport, right?

And every report we create we will by default log in with our default engine, right?

Then it seems fair that DeltaOperationReport be serializable 👍

allisonport-db: Hmm I have some thoughts that I haven't fully fleshed out.

Every operation-type report we create will implement DeltaOperationReport, possibly not every report. But yes we probably plan to log all report types in our default engine (I can't think of a reason not to currently). But that would argue that we should make `MetricsReport` serializable....

If we add a new report type i.e. `XXReport extends DeltaOperationReport` but don't make it serializable, it will still be serialized but will be missing additional information in XXReport.

Do we ever expect an report that only extends DeltaOperationReport? Maybe we just want to report that an operation occurred or success vs failure?

Need to verify that if both DeltaOperationReport and SnapshotReport are serializable jackson will use the lowest ancestor.

allisonport-db: I wonder if this is something we can flesh out later? These are all internal APIs and for now, I think, making sure the main report types are serializable is enough.

scottsand-db: > I wonder if this is something we can flesh out later?

Yup! SGTM!

PR #3907, opened 2024-11-27
[Spark][Version Checksum] Track histogram of file sizes in the checksum

dhruvarya-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Follow up for https://github.com/delta-io/delta/pull/3899. Makes it so that histograms of file sizes are tracked in the version checksum --- these are later used for validation. These are computed incrementally to avoid state reconstruction.

## How was this patch tested?

Added FileSizeHistogramSuite and updated some existing suites.

No


felipepessoto:
@dhruvarya-db do you think we can merge it before the 4.0 release? It could help with #4487

PR #3914, opened 2024-12-02
[SPARK] Rewrite GENERATE SYMLINK MANIFEST command to use Spark table resolution

ctring:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR rewrites the GENERATE SYMLINK MANIFEST command (`DeltaGenerateCommand`) to use Spark's table resolution mechanism. It is replaced to be a `UnaryRunnableCommand` that takes in a child logical plan, which is initially an unresolved table. The table is then resolved by Spark's resolution mechanism, resulting in a `DeltaTableV2` object.

With this object, we also have a catalog table which is passed down into `generateFulllManifest` as part of the effort to pass the catalog table to DeltaLog API's call sites.

## How was this patch tested?

unit tests

No


LukasRupprecht (approved):
Thanks for fixing this command!

PR #3915, opened 2024-12-03
[Spark] Add support for dropping the CheckpointProtection table feature

andreaschat-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Adds support for dropping the `CheckpointProtection` table feature. To drop the feature, we need to truncate all history prior to the CheckpointProtection version. For this cleanup operation we use a shorter retention period of 24 hours. This is configured in `delta.dropFeatureTruncateHistory.retentionDuration` and is the same config we use in the drop feature with History Truncation implementation.

## How was this patch tested?

Added tests in `DeltaFastDropFeatureSuite`.

No.

### DIFF
--- a/spark/src/main/resources/error/delta-error-classes.json
+++ b/spark/src/main/resources/error/delta-error-classes.json
@@ -967,10 +967,32 @@
     ],
     "sqlState" : "0AKDE"
   },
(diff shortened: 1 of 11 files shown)


cstavr (commented):
LGTM! Thanks!

cstavr (approved):

PR #3922, opened 2024-12-04
Increase timeout on first test in DeltaSinkImplicitCastSuite

johanl-db:
## Description
Test `write wider type - long -> int` in suite `DeltaSinkImplicitCastSuiteBase`  is randomly timing out. Specifically, the first write in the test suite times out, due to the initial streaming setup taking a long time.

The timeout on that first write is increased to address the test flakiness

## How was this patch tested?
Test-only change

No


c27kwan (approved):
lgtm! Thanks for fixing this :)

PR #3929, opened 2024-12-06
allow user-specified schema in read if it's consistent

cloud-fan:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

User-specified schema may come from the catalog if the Delta table is stored in an external catalog that syncs the table schema with the Delta log. We should allow it if it's the same as the real Delta table schema.

This is already the case for batch read, see https://github.com/apache/spark/pull/15046

This PR changes the Delta streaming read to allow it as well.

Note: since Delta uses DS v2 (`TableProvider`) and explicitly claims that user-specified schema is not supported (`TableProvider#supportsExternalMetadata` returns false by default), end users still can't specify schema in `spark.read/readStream.schema`. This change is only for advanced Spark plugins that can construct logical plans to triggers Delta v1 source stream scan.

## How was this patch tested?

a new test

No

### DIFF
--- a/spark/src/main/resources/error/delta-error-classes.json
+++ b/spark/src/main/resources/error/delta-error-classes.json
@@ -2859,7 +2859,7 @@
   },
   "DELTA_UNSUPPORTED_SCHEMA_DURING_READ" : {
     "message" : [
-      "Delta does not support specifying the schema at read time."
+      "Delta does not support specifying the schema at read time, unless it's the same as the actual table schema."
     ],
     "sqlState" : "0AKDC"
   },
--- a/spark/src/main/resources/error/delta-error-classes.json
+++ b/spark/src/main/resources/error/delta-error-classes.json
@@ -2859,7 +2859,7 @@
   },
   "DELTA_UNSUPPORTED_SCHEMA_DURING_READ" : {
     "message" : [
-      "Delta does not support specifying the schema at read time, unless it's the same as the actual table schema."
+      "Delta does not support specifying the schema at read time."
     ],
     "sqlState" : "0AKDC"
   },
--- a/spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaDataSource.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaDataSource.scala
@@ -75,9 +75,6 @@ class DeltaDataSource
       schema: Option[StructType],
       providerName: String,
       parameters: Map[String, String]): (String, StructType) = {
-    if (schema.nonEmpty && schema.get.nonEmpty) {
-      throw DeltaErrors.specifySchemaAtReadTimeException
-    }
     val path = parameters.getOrElse("path", {
       throw DeltaErrors.pathNotSpecifiedException
     })
@@ -106,6 +103,10 @@ class DeltaDataSource
           mergeConsecutiveSchemaChanges = shouldMergeConsecutiveSchemas)
         .flatMap(_.getCurrentTrackedMetadata.map(_.dataSchema))
         .getOrElse(snapshot.schema)
+    }
+
+    if (schema.nonEmpty && !DataType.equalsIgnoreCompatibleNullability(readSchema, schema.get)) {
+      throw DeltaErrors.specifySchemaAtReadTimeException
     }
 
     val schemaToUse = DeltaColumnMapping.dropColumnMappingMetadata(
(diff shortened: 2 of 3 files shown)


ni-mi:
@allisonport-db @cloud-fan will this very important fix be in the 3.3.0 release?

ni-mi:
@allisonport-db  @cloud-fan will this very important fix be in the 3.3.0 release?

cloud-fan:
cc @tdas

ni-mi:
Hi,

@tdas any news about this fix? 
It's a blocker for using Delta tables with Structured Streaming using Open Accessibility Databricks - and UC...

Thanks!

CarlEkerot:
Hi,
I think I've experienced issues related to this when streaming from tables Glue catalogs as well. If the schema is properly set in the table property `spark.sql.sources.schema`, any streaming reads will fail. From the look of things, this is handled by setting the schema to `{"type":"struct","fields":[]}`. The Glue schema is subsequently set to `col: array<byte>`, making the table break tools like Athena.

nimrod-doubleverify:
@cloud-fan @tdas any update? Anything?

raveeram-db:
> 
@cloud-fan @tdas any update? Anything?

Hi @nimrod-doubleverify apologies for the delay here, unfortunately this didn't make it to the 3.3.0 release but we'll work on getting out a 3.3.1 release this week.

lzlfred (approved):

tdas (approved):

spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaDataSource.scala
```diff
@@ -108,6 +105,10 @@ class DeltaDataSource
         .getOrElse(snapshot.schema)
     }
 
+    if (schema.nonEmpty && !DataType.equalsIgnoreCompatibleNullability(readSchema, schema.get)) {
```
cloud-fan, line 108:
```suggestion
    if (schema.nonEmpty && schema.get.nonEmpty &&
      !DataType.equalsIgnoreCompatibleNullability(readSchema, schema.get)) {
```

spark/src/main/resources/error/delta-error-classes.json
```diff
@@ -2859,7 +2859,7 @@
   },
   "DELTA_UNSUPPORTED_SCHEMA_DURING_READ" : {
     "message" : [
-      "Delta does not support specifying the schema at read time."
+      "Delta does not support specifying the schema at read time, unless it's the same as the actual table schema."
```
cloud-fan, line 2862:
```suggestion
      "Delta does not support specifying the schema at read time."
```

spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaDataSource.scala
cloud-fan, line 108:
```suggestion
    if (schema.nonEmpty && schema.get.nonEmpty &&
      !DataType.equalsIgnoreCompatibleNullability(readSchema, schema.get)) {
```

spark/src/test/scala/org/apache/spark/sql/delta/DeltaSourceSuite.scala
```diff
@@ -154,6 +156,32 @@ class DeltaSourceSuite extends DeltaSourceSuiteBase
...
         assert(e.getMessage.contains(msg))
       }
+
+      val e = intercept[SparkThrowable] {
+        spark.readStream
+          .schema(StructType.fromDDL("value STRING"))
+          .format("delta")
+          .load(inputDir.getCanonicalPath)
+      }
+      assert(e.getMessage.contains("does not support user-specified schema"))
```
cloud-fan, line 166:
```suggestion
      val e2 = intercept[SparkThrowable] {
        spark.readStream
          .schema(StructType.fromDDL("value STRING"))
          .format("delta")
          .load(inputDir.getCanonicalPath)
      }
      assert(e2.getMessage.contains("does not support user-specified schema"))
```

spark/src/test/scala/org/apache/spark/sql/delta/DeltaSourceSuite.scala
```diff
@@ -154,6 +156,32 @@ class DeltaSourceSuite extends DeltaSourceSuiteBase
       for (msg <- Seq("Delta does not support specifying the schema at read time")) {
         assert(e.getMessage.contains(msg))
       }
+
+      val e2 = intercept[SparkThrowable] {
```
cloud-fan, line 160:
```suggestion
      val e2 = intercept[Exception] {
```

tdas:
Thank you for debugging and fixing this @cloud-fan

PR #3930, opened 2024-12-06
Vacuum lite docs

rajeshparangi:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)


docs/source/delta-utility.md
```diff
@@ -35,6 +35,8 @@ default retention threshold for the files is 7 days. To change this behavior, se
 
     VACUUM '/data/events' -- vacuum files in path-based table
 
+    VACUUM '/data/events' -- vacuum files in path-based table
```
ebyhr, line 38: Why do you duplicate this query? The same query exists at L36.

PR #3938, opened 2024-12-09
[Kernel] Remove Coordinated Commits from public API

scottsand-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

We are re-thinking the design of the Coordinated Commits table feature (currently still in RFC). Thus, we should remove it from the public Kernel API for Delta 3.3 release.

To summarize the changes of this PR

- I remove `getCommitCoordinatorClientHandler` from the `Engine` interface
- I move various previously `public` CC interfaces and classes to be `internal` now
- `SnapshotImpl::getTableCommitCoordinatorClientHandlerOpt` is hardcoded to return an empty optional
- Delete failing test suites and unapplicable utils

## How was this patch tested?

Existing CI tests.

We remove coordinated commits from the public kernel API.


vkorukanti (approved):
LGTM % CI failure

allisonport-db (approved):
LGTM! Please cherry-pick to `branch-3.3` as well

dhruvarya-db (approved):
LGTM

PR #3939, opened 2024-12-09
Fix schema tracking location check condition against checkpoint location

jackierwzhang:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
This PR introduces a better way to check if the schema tracking location is under the checkpoint location that would work with arbitrary file systems and paths.

## How was this patch tested?
New UT.

No


johanl-db (approved):
Nice test coverage!

PR #3947, opened 2024-12-10
[Spark] Reject unsupported type changes with Uniform

johanl-db:
## Description
Iceberg only supports a subset of the type changes that Delta's type widening feature covers.
See https://iceberg.apache.org/spec/#schema-evolution.
Unsupported type changes areL
- integers -> decimal
- integers -> double
- decimal scale increase
- date -> timestamp_ntz

This change makes using Uniform with Iceberg compatibility mutually exclusive with using one of these unsupported type changes, e.p.:
- Fail when trying to apply an unsupported type change if Uniform with Iceberg compatibility enabled.
- Fail when trying to enable Uniform with Iceberg compatibility if an unsupported type change was applied in the past.
- Don't automatically widen type during schema evolution if Uniform with Iceberg compatibility is enabled and the type promotion isn't supported by Iceberg.

## How was this patch tested?
- Added suite `TypeWideningUniformSuite` covering all cases listed above. E.p. for schema evolution, covers MERGE and a sample of INSERTs, including streaming writes.

Note: all the unsupported Iceberg type changes covered here will only be supported in Delta starting with Delta 4.0, so this doesn't impact any released Delta version except for 4.0 preview.

Applying a type change not supported by Iceberg when Uniform with Iceberg compatibility now fails with:
```
CREATE TABLE t (a DECIMAL(10, 2)) TBLPROPERTIES (
  'delta.enableTypeWidening' = 'true',
  'delta.IcebergCompatV2' = 'true',
  'delta.universalFormat.enabledFormats' = 'iceberg'
);

ALTER TABLE t CHANGE COLUMN a TYPE DECIMAL(12, 4);
[DELTA_ICEBERG_COMPAT_VIOLATION.UNSUPPORTED_TYPE_WIDENING] 
IcebergCompatV2 is incompatible with a type change applied to this table: Field a was changed from DECIMAL(10, 2) to DECIMAL(12, 4)."
```

Enabling Uniform with Iceberg compatibility on a table that had an unsupported type change applied in the past now fails with:
```
CREATE TABLE t (a DECIMAL(10, 2)) TBLPROPERTIES ('delta.enableTypeWidening' = 'true');
ALTER TABLE t CHANGE COLUMN a TYPE DECIMAL(12, 4);


ALTER TABLE t SET TBLPROPERTIES (
  'delta.IcebergCompatV2' = 'true',
  'delta.universalFormat.enabledFormats' = 'iceberg'
);
[DELTA_ICEBERG_COMPAT_VIOLATION.UNSUPPORTED_TYPE_WIDENING] 
IcebergCompatV2 is incompatible with a type change applied to this table: Field a was changed from DECIMAL(10, 2) to DECIMAL(12, 4)."
```

Type changes not supported by Iceberg aren't eligible anymore for automatic type widening with schema evolution in MERGE/INSERT operation when Uniform with Iceberg compatibility is enabled.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaAnalysis.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaAnalysis.scala
@@ -82,8 +82,8 @@ class DeltaAnalysis(session: SparkSession)
   override def apply(plan: LogicalPlan): LogicalPlan = plan.resolveOperatorsDown {
     // INSERT INTO by ordinal and df.insertInto()
     case a @ AppendDelta(r, d) if !a.isByName &&
-        needsSchemaAdjustmentByOrdinal(d, a.query, r.schema) =>
-      val projection = resolveQueryColumnsByOrdinal(a.query, r.output, d)
+        needsSchemaAdjustmentByOrdinal(d, a.query, r.schema, a.writeOptions) =>
+      val projection = resolveQueryColumnsByOrdinal(a.query, r.output, d, a.writeOptions)
       if (projection != a.query) {
         a.copy(query = projection)
       } else {
@@ -94,9 +94,9 @@ class DeltaAnalysis(session: SparkSession)
     // INSERT INTO by name
     // AppendData.byName is also used for DataFrame append so we check for the SQL origin text
     // since we only want to up-cast for SQL insert into by name
-    case a @ AppendDelta(r, d) if a.isByName &&
-        a.origin.sqlText.nonEmpty && needsSchemaAdjustmentByName(a.query, r.output, d) =>
-      val projection = resolveQueryColumnsByName(a.query, r.output, d)
+    case a @ AppendDelta(r, d) if a.isByName && a.origin.sqlText.nonEmpty &&
+        needsSchemaAdjustmentByName(a.query, r.output, d, a.writeOptions) =>
+      val projection = resolveQueryColumnsByName(a.query, r.output, d, a.writeOptions)
       if (projection != a.query) {
         a.copy(query = projection)
       } else {
@@ -214,8 +214,8 @@ class DeltaAnalysis(session: SparkSession)
 
     // INSERT OVERWRITE by ordinal and df.insertInto()
     case o @ OverwriteDelta(r, d) if !o.isByName &&
-        needsSchemaAdjustmentByOrdinal(d, o.query, r.schema) =>
-      val projection = resolveQueryColumnsByOrdinal(o.query, r.output, d)
+        needsSchemaAdjustmentByOrdinal(d, o.query, r.schema, o.writeOptions) =>
+      val projection = resolveQueryColumnsByOrdinal(o.query, r.output, d, o.writeOptions)
       if (projection != o.query) {
         val aliases = AttributeMap(o.query.output.zip(projection.output).collect {
           case (l: AttributeReference, r: AttributeReference) if !l.sameRef(r) => (l, r)
@@ -231,9 +231,9 @@ class DeltaAnalysis(session: SparkSession)
     // INSERT OVERWRITE by name
     // OverwriteDelta.byName is also used for DataFrame append so we check for the SQL origin text
     // since we only want to up-cast for SQL insert into by name
-    case o @ OverwriteDelta(r, d) if o.isByName &&
-        o.origin.sqlText.nonEmpty && needsSchemaAdjustmentByName(o.query, r.output, d) =>
-      val projection = resolveQueryColumnsByName(o.query, r.output, d)
+    case o @ OverwriteDelta(r, d) if o.isByName && o.origin.sqlText.nonEmpty &&
+        needsSchemaAdjustmentByName(o.query, r.output, d, o.writeOptions) =>
+      val projection = resolveQueryColumnsByName(o.query, r.output, d, o.writeOptions)
       if (projection != o.query) {
         val aliases = AttributeMap(o.query.output.zip(projection.output).collect {
           case (l: AttributeReference, r: AttributeReference) if !l.sameRef(r) => (l, r)
@@ -251,15 +251,15 @@ class DeltaAnalysis(session: SparkSession)
     case o @ DynamicPartitionOverwriteDelta(r, d) if o.resolved
       =>
       val adjustedQuery = if (!o.isByName &&
-          needsSchemaAdjustmentByOrdinal(d, o.query, r.schema)) {
+          needsSchemaAdjustmentByOrdinal(d, o.query, r.schema, o.writeOptions)) {
         // INSERT OVERWRITE by ordinal and df.insertInto()
-        resolveQueryColumnsByOrdinal(o.query, r.output, d)
+        resolveQueryColumnsByOrdinal(o.query, r.output, d, o.writeOptions)
       } else if (o.isByName && o.origin.sqlText.nonEmpty &&
-          needsSchemaAdjustmentByName(o.query, r.output, d)) {
+          needsSchemaAdjustmentByName(o.query, r.output, d, o.writeOptions)) {
         // INSERT OVERWRITE by name
         // OverwriteDelta.byName is also used for DataFrame append so we check for the SQL origin
         // text since we only want to up-cast for SQL insert into by name
-        resolveQueryColumnsByName(o.query, r.output, d)
+        resolveQueryColumnsByName(o.query, r.output, d, o.writeOptions)
       } else {
         o.query
       }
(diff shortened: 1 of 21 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/DeltaAnalysis.scala
```diff
@@ -917,42 +923,46 @@ class DeltaAnalysis(session: SparkSession)
       attr: NamedExpression,
       targetAttr: NamedExpression,
       tblName: String,
-      allowTypeWidening: Boolean): NamedExpression = {
+      shouldWidenType: (AtomicType, AtomicType) => Boolean): NamedExpression = {
```
larsk-db, line 926: I'm slightly concerned about this signature for the same reason I mentioned above: Both types are the same and it's easy to mess them up. Ofc from -> to is logical for you an me. But that doesn't mean it'll be for everyone. It would be good to be explicit to harden this code. So I just making this a single method trait instead that can be implemented differently at various places. But then at least the method has the correct names and can be invoked with by-name-args.

spark/src/test/scala/org/apache/spark/sql/delta/schema/SchemaUtilsSuite.scala
```diff
@@ -2434,8 +2434,15 @@ class SchemaUtilsSuite extends QueryTest
...
-      base1, update1, false, false, false, false, allowOverride = false) ===
+    assert(
+      mergeDataTypes(
+        current = base1,
+        update = update1,
+        allowImplicitConversions = false,
+        keepExistingType = false,
+        shouldWidenType = (_, _) => false,
+        caseSensitive = false,
+        allowOverride = false) ===
```
larsk-db, line 2445: Might I suggest putting the result mergeDataTypes into a val first? By the time we get to === I have almost forgotten this is in an assert :P

spark/src/test/scala/org/apache/spark/sql/delta/typewidening/TypeWideningUniformTests.scala
```diff
@@ -0,0 +1,247 @@
...
+  }
+
+
+  for (insert <- Set(
+    // Cover only a subset of all INSERTs. There's little value in testing all of them and it
+    // quickly gets expensive.
+    SQLInsertByPosition(SaveMode.Append),
+    SQLInsertByName(SaveMode.Append),
+    DFv1InsertInto(SaveMode.Append),
+    StreamingInsert)) {
```
larsk-db, line 199: more indent imo

spark/src/main/scala/org/apache/spark/sql/delta/DeltaErrors.scala
```diff
@@ -3402,6 +3402,23 @@ trait DeltaErrorsBase
...
+  def icebergCompatUnsupportedTypeWideningException(
+      version: Int,
+      fieldPath: Seq[String],
+      oldType: DataType,
+      newType: DataType): Throwable = {
+    new DeltaUnsupportedOperationException(
+      errorClass = "DELTA_ICEBERG_COMPAT_VIOLATION.UNSUPPORTED_TYPE_WIDENING",
+      messageParameters = Array(
+        version.toString,
+        version.toString,
```
larsk-db, line 3414: double version?

spark/src/main/scala/org/apache/spark/sql/delta/DeltaAnalysis.scala
johanl-db: I introduced `TypeWideningMode` following your suggestion

spark/src/main/scala/org/apache/spark/sql/delta/DeltaErrors.scala
johanl-db: That got me at first too: version is used twice in the error message: once in the main error DELTA_ICEBERG_COMPAT_VIOLATION and once in the sub error class UNSUPPORTED_TYPE_WIDENING

spark/src/test/scala/org/apache/spark/sql/delta/schema/SchemaUtilsSuite.scala
johanl-db: Updated

spark/src/test/scala/org/apache/spark/sql/delta/typewidening/TypeWideningUniformTests.scala
johanl-db: Done

spark/src/main/scala/org/apache/spark/sql/delta/DeltaAnalysis.scala
```diff
@@ -917,42 +923,51 @@ class DeltaAnalysis(session: SparkSession)
...
   }
 
   /**
-   * Whether inserting values that have a wider type than the table has is allowed. In that case,
-   * values are not downcasted to the current table type and the table schema is updated instead to
-   * use the wider type.
+   * Returns a mapping of (fromType, toType) to Boolean indicating whether `fromType` is eligible to
+   * be automatically widened to `toType` when ingesting data. If it is, the table schema is updated
+   * to `toType` before ingestion and values are written using their origin `toType` type.
+   * Otherwise, the table type `fromType` is retained and values are downcasted on write.
```
larsk-db, line 955: Need to update this comment now that it actually returns a mode instance, not a function anymore.

spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMode.scala
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * A type widening mode captures a specific set of type changes that are allowed to be applied.
+ * Currently:
+ *  - NoTypeWidening: No type change is allowed.
+ *  - TypeEvolution(uniformIcebergEnabled = true): Type changes that are eligible to be applied
+ *    automatically during schema evolution and that are supported by Iceberg are allowed.
+ *  - TypeEvolution(uniformIcebergEnabled = false): Type changes that are eligible to be applied
+ *    automatically during schema evolution are allowed, even if they are not supported by Iceberg.
+ */
+trait TypeWideningMode {
```
larsk-db, line 29:
```suggestion
sealed trait TypeWideningMode {
```
I think it makes sense to enforce exhaustiveness here. There shouldn't be other places that need to specify customs rules, I think.

spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMode.scala
```diff
@@ -0,0 +1,50 @@
...
+  def shouldWidenType(fromType: AtomicType, toType: AtomicType): Boolean
+}
+
+object TypeWideningMode {
+  /**
+   * No type change allowed. Typically because type widening and/or schema evolution isn't enabled.
+   */
+  object NoTypeWidening extends TypeWideningMode {
```
larsk-db, line 37:
```suggestion
  case object NoTypeWidening extends TypeWideningMode {
```

Just so it looks like a nice enum-like thingy ;)

spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMode.scala
```diff
@@ -0,0 +1,50 @@
...
+    override def shouldWidenType(fromType: AtomicType, toType: AtomicType): Boolean = false
+  }
+
+  /**
+   * Type changes that are eligible to be applied automatically during schema evolution are allowed.
+   * Can be restricted to only type changes supported by Iceberg.
+   */
+  case class TypeEvolution(uniformIcebergEnabled: Boolean) extends TypeWideningMode {
```
larsk-db, line 45: Why not call this `uniformIcebergCompatibleOnly` instead? That way we can also override later in case people want this anyway without having a weird mental mismatch where `uniformIcebergEnabled=false` on tables where it actually *is* enabled.

larsk-db (commented):
Very nice, thank you. Couple more small things and it's good to go.

spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMode.scala
johanl-db: Agree, that fits better

spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMode.scala
johanl-db: :+1:

spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMode.scala
johanl-db: :+1:

spark/src/main/scala/org/apache/spark/sql/delta/DeltaAnalysis.scala
johanl-db: Done

larsk-db (approved):
LGTM

vkorukanti (approved):

PR #3953, opened 2024-12-11
[Spark][Version Checksum] Add DV-related metrics

dhruvarya-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

1. Adds Deletion Vector-related metrics to Version Checksum. 
2. Adds full and incremental computation of these metrics
3. Updates checksum verification so that it takes these new metrics into account 

## How was this patch tested?

ChecksumDVMetricsSuite

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/Checksum.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/Checksum.scala
@@ -27,9 +27,11 @@ import scala.collection.mutable.ArrayBuffer
 import scala.util.control.NonFatal
 
 import org.apache.spark.sql.delta.actions._
+import org.apache.spark.sql.delta.commands.DeletionVectorUtils
 import org.apache.spark.sql.delta.logging.DeltaLogKeys
 import org.apache.spark.sql.delta.metering.DeltaLogging
 import org.apache.spark.sql.delta.sources.DeltaSQLConf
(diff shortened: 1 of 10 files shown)


andreaschat-db (approved):
LGTM.

prakharjain09 (approved):

spark/src/main/scala/org/apache/spark/sql/delta/Checksum.scala
```diff
@@ -794,6 +867,15 @@ trait ValidateChecksum extends DeltaLogging { self: Snapshot =>
         detailedErrorMapForUsageLogs += ("domainMetadata" -> JsonUtils.toJson(eventData))
       }
     }
+    // Deletion vectors metrics.
+    if (DeletionVectorUtils.deletionVectorsReadable(self)) {
+      (checksum.numDeletedRecordsOpt zip computedState.numDeletedRecordsOpt).foreach {
```
felipepessoto, line 872: @dhruvarya-db, @andreaschat-db. I accidentally found this difference for these two validations, I'm planning to send a fix but first I'd like to check if there is any non-obvious reason to use computedState instead of computedStateToCheckAgainst.

Thanks.

\+@openinx, this could be a 4.3 PR if confirmed to be a bug

PR #3964, opened 2024-12-11
[Kernel] Fix incorrect load protocol and metadata time log

### LINKED ISSUE #3948: [BUG] [Kernel] Incorrect load protocol and metadata time log
[Here](https://github.com/delta-io/delta/blob/b39d5b328ffa8e1071fc6aab78cfb345c8f2d8f7/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java#L546) we intend to log the duration of how long it took LogReplay to load the protocol and metadata.

However, the `startTimeMillis` is declared _after_ the LogReplay object has been created, and the P & M are loading in the constructor of LogReplay. Thus, we are not correctly measuring the time it took to load the P & M.

The ask: please fix this so that we correctly log the actual logReplay time.

anton-erofeev:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description
Resolves #3948 
Fixed incorrect load protocol and metadata time log

## How was this patch tested?
Unit tests

No

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
@@ -525,6 +525,8 @@ public class SnapshotManager {
             .orElse(".");
     logger.info("{}: Loading version {} {}", tablePath, initSegment.version, startingFromStr);
 
+    long startTimeMillis = System.currentTimeMillis();
+
     LogReplay logReplay =
         new LogReplay(
             logPath,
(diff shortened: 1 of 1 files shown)


vkorukanti (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -534,8 +536,6 @@ private SnapshotImpl createSnapshot(LogSegment initSegment, Engine engine) {
             initSegment,
             Optional.ofNullable(latestSnapshotHint.get()));
 
-    long startTimeMillis = System.currentTimeMillis();
```
scottsand-db, line 537: btw, it would have been more accurate to move the logger.info here, too. Now we are also including `assertLogFilesBelongToTable` in our time calculation

anton-erofeev: @scottsand-db good point, please see this pull request https://github.com/delta-io/delta/pull/3974

PR #3965, opened 2024-12-12
[Spark] Avoid unnecessarily calling update and some minor clean up in tests

ctring:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
+ Avoid calling `DeltaLog.update()` in `getFilesFromDeltaLog`, which is already taking in a `snapshot`.
+ Add a new `DeltaTableV2ObjectTestHelper::apply` implicit method to simplify some tests

## How was this patch tested?
unit tests

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/commands/VacuumCommand.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/commands/VacuumCommand.scala
@@ -535,7 +535,7 @@ object VacuumCommand extends VacuumCommandImpl with Serializable {
     // The start and the end commit versions give the range of commit files we want to look into
     // to get the list of eligible files for deletion.
     val eligibleStartCommitVersion = math.min(
-      deltaLog.update().version,
+      snapshot.version,
       latestCommitVersionOutsideOfRetentionWindowAsOfLastVacuumOpt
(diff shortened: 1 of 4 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/commands/VacuumCommand.scala
```diff
@@ -535,7 +535,7 @@ object VacuumCommand extends VacuumCommandImpl with Serializable {
     // The start and the end commit versions give the range of commit files we want to look into
     // to get the list of eligible files for deletion.
     val eligibleStartCommitVersion = math.min(
-      deltaLog.update().version,
+      snapshot.version,
```
c27kwan, line 538: Intentional? This isn't a test. It's a bit misleading from the title.

ctring: Thank you for catching this. This is also a clean up change since this method already takes in a snapshot so there is no need to do `update()` again. I will update the title and description.

LukasRupprecht (approved):
LGTM!

PR #3972, opened 2024-12-13
[Spark] Fix NoSuchElementException in DeltaHistoryManager parallel search with empty commits

sumeet-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

When performing time travel operations that trigger parallel search in DeltaHistoryManager (when searching through >2000 commits), a NoSuchElementException can occur if some partitions return empty commit lists. This happens because the code attempts to call .head on potentially empty collections without proper checks.

## How was this patch tested?

UTs

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
@@ -211,6 +211,9 @@ class DeltaHistoryManager(
         start,
         Some(end),
         deltaLog.newDeltaHadoopConf())
+      if (commits.isEmpty) {
+        throw DeltaErrors.noHistoryFound(deltaLog.logPath)
+      }
       lastCommitBeforeTimestamp(commits, time).getOrElse(commits.head)
     }
   }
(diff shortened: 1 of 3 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -211,6 +211,9 @@ class DeltaHistoryManager(
         start,
         Some(end),
         deltaLog.newDeltaHadoopConf())
+      if (commits.isEmpty) {
+        throw DeltaErrors.noHistoryFound(deltaLog.logPath)
```
dhruvarya-db, line 215: Is this the right exception to throw here when [start, end) don't cover the entire history? All the other changes look good though.

prakharjain09 (approved):

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
sumeet-db: That's what we throw in other such scenarios in the parent function

https://github.com/delta-io/delta/blob/82e940f17f51a0ebeaac0b03441b13875da3c439/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala#L313-L316

PR #3974, opened 2024-12-14
Fixed construct the snapshot time

anton-erofeev:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Resolves https://github.com/delta-io/delta/pull/3964#discussion_r1884569458

## How was this patch tested?
Unit tests

No

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
@@ -536,8 +536,6 @@ public class SnapshotManager {
             initSegment,
             Optional.ofNullable(latestSnapshotHint.get()));
 
-    assertLogFilesBelongToTable(logPath, initSegment.allLogFilesUnsorted());
-
     final SnapshotImpl snapshot =
         new SnapshotImpl(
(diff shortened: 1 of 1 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -549,6 +547,8 @@ private SnapshotImpl createSnapshot(LogSegment initSegment, Engine engine) {
         initSegment.version,
         startingFromStr);
 
+    assertLogFilesBelongToTable(logPath, initSegment.allLogFilesUnsorted());
```
scottsand-db, line 550: I believe the main goal of the log statement above is to print out how long it took to load the protocol and metadata objects during LogReplay.

So I think the easiest thing to do would be to update the log statement (to not directly mention constructing the snapshot) and move it to right after the LogReplay instance is created. what do you think?

foss-contributor:
@anton-erofeev, thank you for your contribution! This pull request has been inactive for over 18 months and will be closed in one week. To keep it open, please leave a comment on this pull request or contact us in the `#contributors` channel on [Delta Lake Slack](https://go.delta.io/slack). We appreciate your involvement in Delta Lake!

PR #3981, opened 2024-12-17
[Spark] Fix auto-conflict handling logic in Optimize to handle DVs

rahulsmahadev:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Bug: There was an existing long standing bug where the custom conflict detection logic in Optimize does not catch concurrent transactions that add DVs. e.g. AddFile(path='a') -> AddFile(path='a', dv='dv1').

Fix: Updated the conflict resolution to consider a composite key of (path, dvId) instead of just depending on path.

## How was this patch tested?
- unit tests

no

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/delta/optimize/OptimizeConflictSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/optimize/OptimizeConflictSuite.scala
@@ -1,0 +1,98 @@
+/*
+ * Copyright (2021) The Delta Lake Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


spark/src/test/scala/org/apache/spark/sql/delta/optimize/OptimizeConflictSuite.scala
```diff
@@ -0,0 +1,98 @@
...
+  with PhaseLockingTestMixin
+  with TransactionExecutionTestMixin
+  with DeltaSQLCommandTest {
+
+  protected def appendRows(dir: File, numRows: Int, numFiles: Int): Unit = {
+    spark.range(start = 0, end = numRows, step = 1, numPartitions = numFiles)
+      .write.format("delta").mode("append").save(dir.getAbsolutePath)
+  }
+
+  test("conflict handling between Optimize and Business Txn") {
```
rahulsmahadev, line 43: this test fails without the fix. It does not conflict and produces 100 rows

brkyvz (approved):
LGTM! Thank you for the fix

PR #3986, opened 2024-12-18
[Kernel] Remove CC code from SnapshotManager

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

We are re-thinking the design of the Coordinated Commits table feature and much of this snapshot code will be refactored. Remove it for now as it greatly complicates our snapshot construction, and hopefully we can be more intentional in our code design/organization when re-implementing it.

https://github.com/delta-io/delta/commit/fc81d1247d66cc32e454e985f0cfc81447f897b6 already removed the public interfaces and made it such that `SnapshotImpl::getTableCommitCoordinatorClientHandlerOpt` never returned a handler.

## How was this patch tested?

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java
@@ -29,7 +29,6 @@ import io.delta.kernel.internal.fs.Path;
 import io.delta.kernel.internal.replay.CreateCheckpointIterator;
 import io.delta.kernel.internal.replay.LogReplay;
 import io.delta.kernel.internal.snapshot.LogSegment;
-import io.delta.kernel.internal.snapshot.TableCommitCoordinatorClientHandler;
 import io.delta.kernel.internal.util.VectorUtils;
 import io.delta.kernel.types.StructType;
 import java.util.List;
@@ -168,17 +167,5 @@ public class SnapshotImpl implements Snapshot {
    */
   public Optional<Long> getLatestTransactionVersion(Engine engine, String applicationId) {
     return logReplay.getLatestTransactionIdentifier(engine, applicationId);
-  }
-
-  /**
-   * Returns the commit coordinator client handler based on the table metadata in this snapshot.
-   *
-   * @param engine the engine to use for IO operations
-   * @return the commit coordinator client handler for this snapshot or empty if the metadata is not
-   *     configured to use the commit coordinator.
-   */
-  public Optional<TableCommitCoordinatorClientHandler> getTableCommitCoordinatorClientHandlerOpt(
-      Engine engine) {
-    return Optional.empty();
   }
 }
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
@@ -35,7 +35,6 @@ import io.delta.kernel.exceptions.TableNotFoundException;
 import io.delta.kernel.internal.*;
 import io.delta.kernel.internal.actions.Metadata;
 import io.delta.kernel.internal.checkpoints.*;
-import io.delta.kernel.internal.coordinatedcommits.Commit;
 import io.delta.kernel.internal.fs.Path;
 import io.delta.kernel.internal.lang.ListUtils;
 import io.delta.kernel.internal.replay.CreateCheckpointIterator;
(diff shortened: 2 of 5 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -797,7 +654,7 @@ protected Optional<LogSegment> getLogSegmentAtOrBeforeVersion(
     }
 
     versionToLoadOpt
-        .filter(v -> v < newVersion)
+        .filter(v -> v != newVersion)
```
scottsand-db, line 657: why this change?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -566,31 +431,26 @@ private SnapshotImpl createSnapshot(LogSegment initSegment, Engine engine) {
...
+   *     streaming source. If not provided, we will try to load the latest version of the table.
    * @return Some LogSegment to build a Snapshot if files do exist after the given startCheckpoint.
    *     None, if the delta log directory was missing or empty.
    */
-  protected Optional<LogSegment> getLogSegmentAtOrBeforeVersion(
-      Engine engine,
-      Optional<Long> startCheckpoint,
-      Optional<Long> versionToLoad,
-      Optional<TableCommitCoordinatorClientHandler> tableCommitHandlerOpt) {
+  public Optional<LogSegment> getLogSegmentForVersion(
```
scottsand-db, line 452: why the change from `AtOrBefore` to `For`? also why `protected` -> `public`?

allisonport-db: With CC, it needed to get the log segment _at or before_ the version provided. Without CC we get the log segment for the version provided, and throw an error if it does not exist.

I.e. if we ask for version 3, and only version 2 is available
- with CC we return the log segment up to version 2, since there may be unbackfilled commits for version 3
- without CC we throw an error since version 3 is not available

allisonport-db: `protected` to `public` is because it needs to be `public` for testing. In the previous PR that added this code the package in `SnapshotManagerSuite` was modified, but without changing the filepath. I reverted that change since package should match filepath.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
allisonport-db: Basically same reason as below. We want to throw an error if the version provided is after the latest version found.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java
```diff
@@ -169,16 +168,4 @@ public CreateCheckpointIterator getCreateCheckpointIterator(Engine engine) {
...
   }
-
-  /**
-   * Returns the commit coordinator client handler based on the table metadata in this snapshot.
-   *
-   * @param engine the engine to use for IO operations
-   * @return the commit coordinator client handler for this snapshot or empty if the metadata is not
-   *     configured to use the commit coordinator.
-   */
-  public Optional<TableCommitCoordinatorClientHandler> getTableCommitCoordinatorClientHandlerOpt(
```
scottsand-db, line 180: Can you delete `TableCommitCoordinatorClientHandler`? Or is that part of the planned followup PR?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -566,31 +431,26 @@ private SnapshotImpl createSnapshot(LogSegment initSegment, Engine engine) {
...
+    return getLogSegmentForVersion(
+        engine, startingCheckpoint.map(x -> x.version), Optional.empty());
   }
 
   /**
-   * Get a list of files that can be used to compute a Snapshot at or before version
-   * `versionToLoad`, If `versionToLoad` is not provided, will generate the list of files that are
-   * needed to load the latest version of the Delta table. This method also performs checks to
-   * ensure that the delta files are contiguous.
+   * Get a list of files that can be used to compute a Snapshot at version `versionToLoad`, if
```
scottsand-db, line 439:
```
version `versionToLoad`. If
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
scottsand-db: Makes sense, thanks!

scottsand-db (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java
allisonport-db: Will be in followup

PR #3993, opened 2024-12-19
[Kernel] Implement AddFile class using DelegateRow

qiyuandong-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This PR implements the `AddFile` action class, with the help of `DelegateRow` and `RowBackedAction`, which are also newly introduced in this PR. The core changes include:

`RowBackedAction`
- A new abstract base class for Delta Log actions that are backed by a `Row`. It can be extended by actions like `AddFile`, where we want to avoid materializing of all fields from action rows when creating action instances within Kernel. 

`DelegateRow`:
- A new `Row` implementation that wraps an existing `Row` and allows overriding values for some particular ordinals.
- Currently this is used to update certain fields of an action by creating a modified view of its underlying row without mutating the original row.
- The implementation ensures we only keep one layer of delegation.

`AddFile`
- This PR implements its constructor, field getters, necessary overrides, and APIs for creating a new instance with updated fields.
- It adds two fields `baseRowId` and `defaultRowCommitVersion`, which will be used for Row Tracking later.

## How was this patch tested?

Unit tests were added in `AddFileSuite.scala`.

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/AddFile.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/AddFile.java
@@ -19,15 +19,18 @@ import static io.delta.kernel.internal.util.InternalUtils.relativizePath;
 import static io.delta.kernel.internal.util.PartitionUtils.serializePartitionMap;
 import static java.util.stream.Collectors.toMap;
 
+import io.delta.kernel.data.MapValue;
 import io.delta.kernel.data.Row;
 import io.delta.kernel.expressions.Literal;
+import io.delta.kernel.internal.data.DelegateRow;
 import io.delta.kernel.internal.data.GenericRow;
 import io.delta.kernel.internal.fs.Path;
+import io.delta.kernel.internal.util.VectorUtils;
 import io.delta.kernel.types.*;
+import io.delta.kernel.utils.DataFileStatistics;
 import io.delta.kernel.utils.DataFileStatus;
 import java.net.URI;
-import java.util.HashMap;
-import java.util.Map;
+import java.util.*;
 import java.util.stream.IntStream;
 
 /** Delta log action representing an `AddFile` */
@@ -55,12 +58,15 @@ public class AddFile {
           .add(
               "tags",
               new MapType(StringType.STRING, StringType.STRING, true /* valueContainsNull */),
-              true /* nullable */);
+              true /* nullable */)
+          .add("baseRowId", LongType.LONG, true /* nullable */)
+          .add("defaultRowCommitVersion", LongType.LONG, true /* nullable */);
 
   public static final StructType SCHEMA_WITH_STATS = SCHEMA_WITHOUT_STATS.add(JSON_STATS_FIELD);
 
   /** Full schema of the {@code add} action in the Delta Log. */
   public static final StructType FULL_SCHEMA = SCHEMA_WITH_STATS;
+
   // There are more fields which are added when row-id tracking and clustering is enabled.
   // When Kernel starts supporting row-ids and clustering, we should add those fields here.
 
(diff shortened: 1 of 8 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/AddFile.java
```diff
@@ -55,12 +58,15 @@ public class AddFile {
...
-              true /* nullable */);
+              true /* nullable */)
+          .add("baseRowId", LongType.LONG, true /* nullable */)
+          .add("defaultRowCommitVersion", LongType.LONG, true /* nullable */);
 
   public static final StructType SCHEMA_WITH_STATS = SCHEMA_WITHOUT_STATS.add(JSON_STATS_FIELD);
 
   /** Full schema of the {@code add} action in the Delta Log. */
   public static final StructType FULL_SCHEMA = SCHEMA_WITH_STATS;
+
```
johanl-db, line 69: Can't leave a comment on line 81 so putting it here:
```
  public static Row convertDataFileStatus(
```
This is a good example of what I mentioned in the slack thread:
> - All public kernel APIs should use Row as this is what engines understand
- All internal kernel APIs should use concrete, internal types such as AddFile
- ~(Optional?)~ We should convert from Row to concrete type at the boundary with engines so that all the internal kernel code doesn't need to use Row

This is an internal API, so it should return an `AddFile`.
It's then used in `generateAppendActions` which is a public API, which is where we'll want to turn into a row:
```
  static CloseableIterator<Row> generateAppendActions(...) {
          AddFile addFile =
              AddFile.convertDataFileStatus(
                  tableRoot,
                  dataFileStatus,
                  true /* dataChange */);
          return SingleAction.createAddFileSingleAction(addFile.toRow());
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/AddFile.java
```diff
@@ -93,4 +99,149 @@ public static Row convertDataFileStatus(
...
+  /**
+   * The underlying {@link Row} that represents an 'AddFile' action and contains all its field
+   * values. Can be either a {@link GenericRow} or a {@link DelegateRow}.
+   */
+  private final Row row;
+
+  public AddFile(Row row) {
+    this.row = row;
+  }
```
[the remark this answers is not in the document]
qiyuandong-db: Thanks for the suggestion. This makes sense.

I tried to implement a `RowBacked` with generic implementation for common methods (e.g. `toString`, `equals`, `hashCode`, and more). The difficulty was handling certain data types used by `Row`, like `MapValue`, `ArrayValue`, and nested `Row`. They don't have generic implementations for these overrides, and in actual use cases, they often need special handling and conversions depending on the context. This seems non-trivial and might be better addressed in a separate PR.

I am still looking into this, but for now I just created a abstract class `RowBackedAction`, and having `AddFile` extends it. It’s pretty minimal, just representing the idea that the action is backed by a `Row`. There’s not much shared functionality yet. We can explore adding more common functionality when extending it to other cases like `RemoveFile`.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/AddFile.java
```diff
@@ -55,12 +60,15 @@ public class AddFile {
...
+          .add("baseRowId", LongType.LONG, true /* nullable */)
+          .add("defaultRowCommitVersion", LongType.LONG, true /* nullable */);
 
   public static final StructType SCHEMA_WITH_STATS = SCHEMA_WITHOUT_STATS.add(JSON_STATS_FIELD);
 
   /** Full schema of the {@code add} action in the Delta Log. */
   public static final StructType FULL_SCHEMA = SCHEMA_WITH_STATS;
+
   // There are more fields which are added when row-id tracking and clustering is enabled.
   // When Kernel starts supporting row-ids and clustering, we should add those fields here.
```
johanl-db, line 74: If we have a way to access `COL_NAME_TO_ORDINAL` from `RowBackedAction`, we should be able to move the common logic to access and update fields there and make it easier to implement child classes without as much repetition.

We can make it a non-static member variable of  the action though as that would take up space for every instance. We can't make it static either, as it depends on the concrete action schema.

We could stop pre-computing it and just implement this as a method `getColumnIndex()` that looks up the name in the schema. That's less efficient though.. unless we keep using a map to quickly lookup fields, which `StructType` already contains internally:
https://github.com/delta-io/delta/blob/5a39f7047e37d05d61aef39ba93a5c6b1b3ebd6c/kernel/kernel-api/src/main/java/io/delta/kernel/types/StructType.java#L32
```
  private final Map<String, Tuple2<StructField, Integer>> nameToFieldAndOrdinal;
```


So the following should be as efficient as today:
```
protected Integer indexOf(String fieldName) {
  return getFullSchema().indexOf(fieldName);
}
```
and implement indexOf as a lookup in `nameToFieldAndOrdinal` instead of an array lookup as it's done currently.

You can then have in `RowBackedAction`, e.g.:

```
protected String getString(String fieldName) {
  return row.getString(indexOf(fieldName));
}
```
```
  public String getPath() {
    return getString("path");
  }
```

Or:
```
  protected Row withOverridenValue(String fieldName, Object newValue) {
    Map<Integer, Object> overrides =
        Collections.singletonMap(indexOf(fieldName), newValue);
    return new AddFile(new DelegateRow(row, overrides));
  }
```
```
  public AddFile withNewBaseRowId(long baseRowId) {
    return new AddFile(withOverridenValue("baseRowId", baseRowId));
  }
```

With this, you should actually be able to get to a point where `row` can be private and is only accessed through the base class

johanl-db (commented):
The change looks good, main comment is mostly to see if we can go a bit further with moving some of the repetitive logic to `RowBackedAction` but  this is more a nice to have than a requirement

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/AddFile.java
qiyuandong-db: Thanks for the suggestion! I’ve implemented the changes. `row` is now private to `RowBackedAction`.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/RowBackedAction.java
```diff
@@ -0,0 +1,128 @@
...
+ * </ul>
+ */
+public abstract class RowBackedAction {
+
+  /** The underlying {@link Row} that represents an action and contains all its field values. */
+  private final Row row;
+
+  protected RowBackedAction(Row row) {
+    checkArgument(
+        row.getSchema().equals(getFullSchema()),
```
scottsand-db, line 45: I don't think that `getFullSchema()` needs to be an abstract protected method.

Why can't you just pass in the expected/desired schema to the constructor here to assert against? And then use `row.getSchema()` elsewhere.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/RowBackedAction.java
```diff
@@ -0,0 +1,128 @@
...
+    return index;
+  }
+
+  /**
+   * Gets the value with the given field name from the row. The type of the Object returned depends
+   * on the data type of the field. If the field is nullable and the value is null, this method will
+   * return null. If the field is not nullable and the value is null, this method will throw an
+   * {@link IllegalArgumentException}.
+   */
+  protected Object getValueAsObject(String fieldName) {
```
scottsand-db, line 72: I'd rather that the underlying class just accesses the row directly, like for AddFile

```java
public String getPath() {
  return row.getString(getFieldIndex("path"));
}
```

This avoids unnecessary unboxing

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/RowBackedAction.java
```diff
@@ -0,0 +1,128 @@
...
+      throw new UnsupportedOperationException(
+          "Unsupported data type: " + dataType.getClass().getName());
+    }
+  }
+
+  /**
+   * Creates a new {@link Row} with the same schema and values as the row backing this action, but
+   * with the value of the field with the given name overridden by the given value.
+   */
+  protected Row createRowWithOverriddenValue(String fieldName, Object value) {
```
scottsand-db, line 119: `toRowWithOverriddenValue`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/AddFile.java
```diff
@@ -55,42 +61,205 @@ public class AddFile {
...
+    fieldMap.put(FULL_SCHEMA.indexOf("dataChange"), dataChange);
+    tags.ifPresent(tag -> fieldMap.put(FULL_SCHEMA.indexOf("tags"), tag));
+    baseRowId.ifPresent(id -> fieldMap.put(FULL_SCHEMA.indexOf("baseRowId"), id));
+    defaultRowCommitVersion.ifPresent(
+        version -> fieldMap.put(FULL_SCHEMA.indexOf("defaultRowCommitVersion"), version));
+    stats.ifPresent(stat -> fieldMap.put(FULL_SCHEMA.indexOf("stats"), stat.serializeAsJson()));
+
+    return new GenericRow(FULL_SCHEMA, fieldMap);
+  }
+
```
scottsand-db, line 133:
```java
////////////////////
// Member methods //
////////////////////
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/AddFile.java
```diff
@@ -55,42 +61,205 @@ public class AddFile {
...
+  /** Returns a new {@link AddFile} with the provided defaultRowCommitVersion. */
+  public AddFile withNewDefaultRowCommitVersion(long defaultRowCommitVersion) {
+    return new AddFile(
+        createRowWithOverriddenValue("defaultRowCommitVersion", defaultRowCommitVersion));
+  }
+
+  @Override
+  public String toString() {
+    // Convert the partitionValues and tags to Java Maps and make them sorted by key.
+    Map<String, String> partitionValuesJavaMap = toJavaMap(getPartitionValues());
```
scottsand-db, line 204: This doesn't seem right to me. MapValue doesn't guarantee order, as far as I can tell. Turning this into a Java Map and then a TreeMap certainly also doesn't guarantee they are in the right order.

if the partition columns are, in order, `zzzPart1`, `yyyPart2` ... then tree map will sort them in their natural order, which will be yyyPart2, zzzPart1, which is wrong.

you need to lookup the order as defined in the schema.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/data/DelegateRow.java
```diff
@@ -0,0 +1,212 @@
...
+      //  together with the GenericRow.
+      throwIfUnsafeAccess(ordinal, MapType.class, "map");
+      return (MapValue) overrides.get(ordinal);
+    }
+    return row.getMap(ordinal);
+  }
+
+  private void throwIfUnsafeAccess(
+      int ordinal, Class<? extends DataType> expDataType, String accessType) {
+
```
scottsand-db, line 195: nit: you can avoid empty whitespace lines like this at the beginning of functions

kernel/kernel-api/src/main/java/io/delta/kernel/internal/data/DelegateRow.java
```diff
@@ -0,0 +1,212 @@
...
+    DataType actualDataType = dataType(ordinal);
+    if (!expDataType.isAssignableFrom(actualDataType.getClass())) {
+      String msg =
+          String.format(
+              "Fail to access a '%s' value from a field of type '%s'", accessType, actualDataType);
+      throw new UnsupportedOperationException(msg);
+    }
+  }
+
+  private DataType dataType(int ordinal) {
```
scottsand-db, line 205: you can move this logic into the method above

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/actions/AddFileSuite.scala
```diff
@@ -0,0 +1,234 @@
...
+      baseRowId: Option[Long] = Option.empty,
+      defaultRowCommitVersion: Option[Long] = Option.empty,
+      stats: Option[String] = Option.empty
+  ): Row = {
+    def toJavaOptional[T](option: Option[T]): Optional[T] = option match {
+      case Some(value) => Optional.of(value)
+      case None => Optional.empty()
+    }
+
+    createAddFileRow(
```
scottsand-db, line 53: `AddFile.createAddFileRow`

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/actions/AddFileSuite.scala
```diff
@@ -0,0 +1,234 @@
...
+      dataChange = false,
+      tags = Option(Map("tag1" -> "value1", "tag2" -> "value2")),
+      baseRowId = Option(12345L),
+      defaultRowCommitVersion = Option(67890L),
+      stats = Option("{\"numRecords\":10000}")
+    )
+    val addFile = new AddFile(addFileRow)
+    val expectedString = "AddFile{" +
+      "path='test/path', " +
+      "partitionValues={a=1, b=2, c=3}, " +
```
scottsand-db, line 142: ah, I see the confusion here. We want these printed in the order they are defined in the schema (this is how they will appear in the file system, too)

scottsand-db (changes requested):
Looks great! Requested a few minor changes. Thanks for making this!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/RowBackedAction.java
qiyuandong-db: This also works and makes things simpler. Updated, thanks!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/RowBackedAction.java
qiyuandong-db: Sure, I’ve updated the code to access the row directly in the `AddFile` getters.

In this way, the row is no longer private to `RowBackedAction` and is exposed it to child classes. If we want to keep it private and enforce access through base class APIs, we could implement all possible `get{DataType}` methods (e.g., `getString`, `getInt`, etc.) in `RowBackedAction`. This would introduce quite a lot boilerplate code, so I’m not sure if it’s worth it.

scottsand-db: It's perfectly acceptable for the `row` to be visible to the AddFile -- we pass it into its constructor, after all.

No need for any of that boilerplate code.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/AddFile.java
scottsand-db, line 204: Thanks for the offline discussion. Let's just simplify things and print out the partition values as is. No sorting or order needed.

scottsand-db (approved):

PR #3998, opened 2024-12-22
[SPARK][SHARING][VARIANT] Create GA variant table feature

richardc-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [x] Other (sharing)

## Description

creates the variant GA table feature. 

New variant tables are still created with the preview table feature. However, a future PR will switch this so new tables then are created with the GA table feature. At this point, we will keep the ability to read the `preview` table feature because there were no changes to the spec between preview and GA.

## How was this patch tested?

added UTs

no


charlenelyu-db (approved):
LGTM for `/sharing` changes

johanl-db (approved):

PR #4005, opened 2024-12-28
Improve the performance of Converting Iceberg to Delta

harperjiang:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
We rewrite IcebergFileManifest to allow it to obtain list of Iceberg DataFile directly from Iceberg ManifestFile

## How was this patch tested?
Existing UT

No


lzlfred (approved):
lgtm

PR #4006, opened 2024-12-29
[Spark] Add support for sorting within partitions when Z-ordering

### LINKED ISSUE #4000: [Feature Request] [Spark] Optionally sort within partitions when Z-ordering

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

### Overview

Z-ordering tables doesn’t sort data within partitions (files) and consequently data skipping on the Parquet level, based on row group metadata, is inefficient.

### Motivation

To increase read efficiency by leveraging mdc on the row group level. Global sort is considered in the [design details](https://docs.google.com/document/d/1TYFxAUvhtYqQ6IHAZXjliVuitA5D1u793PMnzsH_3vs/edit?usp=sharing), but deemed too slow. Sorting within partitions, on the other hand, is relatively fast because it does not introduce a shuffle. It can be optionally applied after the current `repartitionByRange` step. To the best of my knowledge, this approach has not been considered.

### Further details

I originally discussed this problem in the Slack channel with @Kimahriman, who suggested I raise an issue here.

I've implemented the feature by adding configuration property `spark.databricks.io.skipping.mdc.sortWithinPartitions` defaulting to `false`. When the property is enabled, the partitions are sorted on `repartitionKeyColName` after `repartitionByRange`.

I ran a comparison based on the [Delta Lake Z Order](https://delta.io/blog/2023-06-03-delta-lake-z-order/) blog post and [notebook](https://github.com/delta-io/delta-examples/blob/master/notebooks/pyspark/delta-z-order.ipynb) by @MrPowers. I don't have local disk for the large data set (`G1_1e9_1e2_0_0.csv`), so I used a medium-sized one instead (`G1_1e8_1e8_100_0.csv`) and timed `query_c` on four table versions:

- **version 0**: unoptimized
- **version 1**: compacted
- **version 2**: z-ordered on `id1` and `id2`
- **version 3**: z-ordered on `id1` and `id2`, and sorted within partitions

On a 2021 MBP with 16 GB RAM. The results were
```
version 0
[id052,id45689,1.0]
Time taken: 4524 ms
version 1
[id052,id45689,1.0]
Time taken: 3137 ms
version 2
[id052,id45689,1.0]
Time taken: 1280 ms
version 3
[id052,id45689,1.0]
Time taken: 112 ms
```
The `id` column values queried are different because the original combination did not exist in my data set.
**Update**:  I ran the experiment on the larger data set (`G1_1e9_1e9_100_0.csv`) using cloud storage and the results are
```
version 0
[id038,id8508161,4.0]
Time taken: 667717 ms
version 1
[id038,id8508161,4.0]
Time taken: 589716 ms
version 2
[id038,id8508161,4.0]
Time taken: 48994 ms
version 3
[id038,id8508161,4.0]
Time taken: 6386 ms
```

### Willingness to contribute

I have opened PR #4006.

maltevelin:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Resolves #4000 by introducing a new configuration property `spark.databricks.io.skipping.mdc.sortWithinPartitions` that  clusters records in row groups, within Parquet files, based on Z-order or Hilbert curve values. This improves data skipping on the Parquet level. Benchmarks included in the issue demonstrate speedups of approximately 8× and 11× on two different datasets. Please refer to the issue for more details.

## How was this patch tested?

Added test cases in `MultiDimClusteringSuite.scala` for Hilbert and Z-order curves.

Yes. This PR introduces a new configuration property `spark.databricks.io.skipping.mdc.sortWithinPartitions`. The property defaults to `false`, ensuring that existing users remain unaffected unless they opt-in by setting it to `true`.

**Previous Behavior**
Z-ordering did not sort data within partitions.

**New Behavior**
When the property is enabled, `sortWithinPartitions` is applied after `repartitionByRange` in `MultiDimClustering.scala`.


maltevelin:
Hi @vkorukanti , @tdas , @scottsand-db , could you kindly review this PR when you have a moment? It introduces a configurable enhancement for Z-ordering that significantly improves read performance. I'm tagging you since you've reviewed and/or implemented PRs related to issue #1134. Your feedback would be highly appreciated!

spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala
```diff
@@ -1434,6 +1434,15 @@ trait DeltaSQLConfBase {
       .booleanConf
       .createWithDefault(true)
 
+  val MDC_SORT_WITHIN_PARTITIONS =
+    SQLConf.buildConf("spark.databricks.io.skipping.mdc.sortWithinPartitions")
+      .internal()
+      .doc("If enabled, partitions are sorted on Z-order values for MDC. " +
+         "This co-locates records with the same Z-order values in row groups, " +
+         "which enables data skipping on the Parquet level.")
```
zedtang, line 1442:
```suggestion
    SQLConf.buildConf("spark.databricks.io.skipping.mdc.sortWithinFiles")
      .internal()
      .doc("If enabled, sort within files by the specified MDC curve. " +
         "This might improve row-group skipping and data compression, at " +
         "the cost of additional overhead for sorting.")
```
two nits:
1. can we rename this to sortWithinFiles, since partitions are so overloaded (in this case it means files, but can also mean hive partition, etc).
2. MDC also includes Hilbert curve, so best to not mention Z-order here.

spark/src/main/scala/org/apache/spark/sql/delta/skipping/MultiDimClustering.scala
```diff
@@ -90,7 +91,15 @@ trait SpaceFillingCurveClustering extends MultiDimClustering {
...
-    repartitionedDf.drop(repartitionKeyColName)
+    val optionallySortedDf = if (sort) {
+      repartitionedDf.sortWithinPartitions(repartitionKeyColName)
+    }
+    else {
+      repartitionedDf
+    }
+
+    optionallySortedDf.drop(repartitionKeyColName)
+
```
zedtang, line 102:
```suggestion
    if (sortWithinFiles) {
      repartitionDf = repartitionedDf.sortWithinPartitions(repartitionKeyColName)
    }

    repartitionDf.drop(repartitionKeyColName)

```

Nit: keep the `var` and add the if block here. This avoids introducing a new variable.

spark/src/main/scala/org/apache/spark/sql/delta/skipping/MultiDimClustering.scala
```diff
@@ -74,12 +74,13 @@ trait SpaceFillingCurveClustering extends MultiDimClustering {
     val conf = df.sparkSession.sessionState.conf
     val numRanges = conf.getConf(DeltaSQLConf.MDC_NUM_RANGE_IDS)
     val addNoise = conf.getConf(DeltaSQLConf.MDC_ADD_NOISE)
+    val sort = conf.getConf(DeltaSQLConf.MDC_SORT_WITHIN_PARTITIONS)
```
zedtang, line 77:
```suggestion
    val sortWithinFiles = conf.getConf(DeltaSQLConf.MDC_SORT_WITHIN_FILES)
```

zedtang (commented):
thanks for contributing

maltevelin:
Thanks for reviewing.

spark/src/main/scala/org/apache/spark/sql/delta/skipping/MultiDimClustering.scala
```diff
@@ -74,12 +74,13 @@ trait SpaceFillingCurveClustering extends MultiDimClustering {
     val conf = df.sparkSession.sessionState.conf
     val numRanges = conf.getConf(DeltaSQLConf.MDC_NUM_RANGE_IDS)
```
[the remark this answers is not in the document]
maltevelin: Good point! Conversely, sorting becomes less effective as file size approaches row group size (which won't happen under the default settings). Did you have a method in mind? Alternatively, these observations could be added to documentation.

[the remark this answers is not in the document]
chirag-s-db: Yeah, there's a couple of domains:
- Row group size << file size: User should scale up num ranges (or sorting within partitions won't be as effective as it could be).
- Row group size < file size: The default configs should be OK when sorting within partitions.
- Row group size approx. equal to file size: No reason to sort within partitions.

I think we can probably leave it up to the user to determine which range their use case falls into (especially given that this conf isn't enabled by default).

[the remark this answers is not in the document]
chirag-s-db: This might be good to add to docs

spark/src/main/scala/org/apache/spark/sql/delta/skipping/MultiDimClustering.scala
```diff
@@ -90,7 +91,12 @@ trait SpaceFillingCurveClustering extends MultiDimClustering {
         .repartitionByRange(approxNumPartitions, col(repartitionKeyColName))
     }
 
+    if (sortWithinFiles) {
+      repartitionedDf = repartitionedDf.sortWithinPartitions(repartitionKeyColName)
+    }
+
     repartitionedDf.drop(repartitionKeyColName)
+
```
zedtang, line 99:
```suggestion
```

zedtang (approved):
LGTM!

PR #4012, opened 2025-01-03
[UNIFORM] Fix converting non-ISO timestamp partition values to Iceberg

amogh-jahagirdar:

- [X ] Spark (Uniform)
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [] Other 

## Description

This change fixes support for timestamp partition values in Uniform specifically when timestamp partition values which are not UTC instants are stored in the Delta log (Delta protocol supports both TS partition values where there is no timezone and  ISO-8601 instant timestamps).

In the fix these partition values will be interpreted in the system time of the system doing the metadata conversion and then UTC adjusted for the Iceberg metadata translation. Previously the fix assumed UTC instants but for compatibility reasons, we do want to be able to support the older form of the metadata for Iceberg conversion. 

Example:

Prior to this fix, the code would fail with a parsing exception for the non-instant case since a string like "2021-01-01 10:00:00.123" would fail to be parsed by Instant.parse(str) since it's not a valid instant.

After this fix: "2021-01-01 10:00:00.123", when conversion is run on a Spark cluster with UTC-8 session timezone, will be interpreted as ""2021-01-01 10:00:00.123UTC-08" and then converted to "2021-01-01 16:00:00.123UTC", which will be stored as microseconds since epoch in Iceberg metadata.

## How was this patch tested?
Added unit tests

No

### DIFF
--- a/iceberg/src/main/scala/org/apache/spark/sql/delta/icebergShaded/IcebergTransactionUtils.scala
+++ b/iceberg/src/main/scala/org/apache/spark/sql/delta/icebergShaded/IcebergTransactionUtils.scala
@@ -18,6 +18,7 @@ package org.apache.spark.sql.delta.icebergShaded
 
 import java.nio.ByteBuffer
 import java.time.Instant
+import java.time.format.DateTimeParseException
 
 import scala.collection.JavaConverters._
 import scala.util.control.NonFatal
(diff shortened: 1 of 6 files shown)


lzlfred (approved):
LGTM !

PR #4018, opened 2025-01-06
[Flink] Fix mapping between Delta's BinaryType and Flink's VarBinaryType

### LINKED ISSUE #3977: [BUG][Flink] Delta Flink Connector: Incorrect Mapping of BYTES to BINARY(1) in Sink Table Schema

- [x ] Flink

### Describe the problem

The Delta Flink connector maps a BYTES type to BINARY(1) in the sink schema, causing schema mismatch errors during an INSERT operation. This behavior persists despite explicitly defining the sink schema or casting the binary_data column in the query. I noticed this when trying to get the raw payload from a kafka topic but I ended up debuging and found that this bug is happening due to something that flink delta connector is doing while creating the mapping at execution plan so I created this sample to show that the schema is not working  

I tried with both datastream api and table api with same results.

#### Steps to reproduce

```python
from pyflink.table import EnvironmentSettings, TableEnvironment

# Initialize the streaming table environment
env_settings = EnvironmentSettings.in_streaming_mode()
table_env = TableEnvironment.create(env_settings)

# Create a source table using the DataGen connector
table_env.execute_sql("""
    CREATE TABLE random_source (
        id BIGINT,
        binary_data BYTES
    ) WITH (
        'connector' = 'datagen',
        'fields.id.kind' = 'sequence',
        'fields.id.start' = '1',
        'fields.id.end' = '8',
        'fields.binary_data.kind' = 'random',
        'fields.binary_data.length' = '16'
    )
""")

# Create a sink table using the Delta connector
table_env.execute_sql("""
    CREATE TABLE print_sink (
        id BIGINT,
        binary_data BYTES
    ) WITH (
        'connector' = 'delta',
        'table-path' = '/path/to/delta/table'
    )
""")

# Insert data from the source into the sink
table_env.execute_sql("""
    INSERT INTO print_sink
    SELECT id, binary_data
    FROM random_source
""").wait()

```

#### Observed results

The schema for the print_sink table is created incorrectly as:

```python
<Row('id', 'BIGINT', True, None, None, None)>
<Row('binary_data', 'BINARY(1)', True, None, None, None)>
```

This causes a schema mismatch error during the INSERT operation:
```sql
org.apache.flink.table.api.ValidationException: Column types of query result and sink for 'print_sink' do not match.
Cause: Incompatible types for sink column 'binary_data' at position 1.
Query schema: [id: BIGINT, binary_data: BYTES]
Sink schema: [id: BIGINT, binary_data: BINARY(1)]
```
#### Expected results

The binary_data column should be mapped correctly to a type compatible with BYTES (e.g., VARBINARY), avoiding schema mismatch errors.

#### Further details

The issue persists even after:

- Explicitly casting binary_data as BYTES in the query.
- Manually defining the sink schema with binary_data VARBINARY.
- Using the configuration option 'delta.ignore-schema-check' = 'true'.

When using a temporary table (e.g., memory or blackhole), the schema for binary_data is correctly identified as BYTES. The issue seems to be specific to the Delta connector.

### Environment information

Tested with 
- Flink 1.19 and 1.20

    <dependencies>
        
        <dependency>
            <groupId>org.apache.flink</groupId>
            <artifactId>flink-connector-kafka</artifactId>
            <version>3.4.0-1.20</version>
        </dependency>

        
        <dependency>
            <groupId>io.delta</groupId>
            <artifactId>delta-flink</artifactId>
            <version>3.2.1</version>
        </dependency>

        
        <dependency>
            <groupId>org.apache.flink</groupId>
            <artifactId>flink-sql-connector-kafka</artifactId>
            <version>3.4.0-1.20</version>            
        </dependency>
        
        
        <dependency>
            <groupId>org.apache.flink</groupId>
            <artifactId>flink-sql-parquet</artifactId>
            <version>1.20.0</version>
        </dependency>

        
        <dependency>
            <groupId>org.apache.flink</groupId>
            <artifactId>flink-table-planner_2.12</artifactId>
            <version>1.20.0</version>
        </dependency>

        
        <dependency>
            <groupId>org.apache.flink</groupId>
            <artifactId>flink-clients</artifactId>
            <version>1.20.0</version>
        </dependency>

        
        <dependency>
            <groupId>io.delta</groupId>
            <artifactId>delta-standalone_2.12</artifactId>
            <version>3.2.1</version>
        </dependency>

### Willingness to contribute

- [x] No. I cannot contribute a bug fix at this time.

scottsand-db:

- [ ] Spark
- [ ] Standalone
- [X] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Resolves #3977

This PR fixes a bug where the Delta-Flink connector would incorrectly map Delta's `BinaryType` (which is variable length) type to Flink's `BinaryType` (which is fixed length). Instead, this PR fixes it so that Delta's `BinaryType` is mapped to Flink's `VarBinaryType(MAX_LENGTH)`. As a comparison, Iceberg does the same [here](https://github.com/apache/iceberg/blob/8e2ffb35da2d4c5059e96cb78a30fd8c54cfbedf/flink/v1.18/flink/src/main/java/org/apache/iceberg/flink/TypeToFlinkType.java#L125).

This incorrect logic caused the behaviour seen in #3977 because:
- First, we had a source (datagen) table with Flink's BYTES (which is a synonym for VarBinaryType of MAX_LENGTH)
- Second, we had a target (delta) table. Flink created it with schema BYTES and the Delta schema had delta type `BinaryType`.
- All good so far. No problems yet.
- When we tried to `INSERT INTO <target> SELECT * FROM <source>`, the DeltaCatalog would lookup the "flink schema" of the target Delta table. It would see Delta's BinaryType and map it incorrectly to flink's BinaryType (not Flink's BYTES or VarBinaryType)
- Hence it would throw the error below

```
org.apache.flink.table.api.ValidationException: Column types of query result and sink for 'print_sink' do not match.
Cause: Incompatible types for sink column 'binary_data' at position 1.
Query schema: [id: BIGINT, binary_data: BYTES]
Sink schema: [id: BIGINT, binary_data: BINARY(1)]
```

## How was this patch tested?

Updated schema conversion tests.

New e2e test which directly tests the problematic scenario brought up in #3977.

Yes. We fix the mapping of Delta's BinaryType to flinks VarBinaryTYpe

### DIFF
--- a/connectors/flink/src/main/java/io/delta/flink/source/internal/SchemaConverter.java
+++ b/connectors/flink/src/main/java/io/delta/flink/source/internal/SchemaConverter.java
@@ -2,7 +2,6 @@ package io.delta.flink.source.internal;
 
 import org.apache.flink.table.types.logical.ArrayType;
 import org.apache.flink.table.types.logical.BigIntType;
-import org.apache.flink.table.types.logical.BinaryType;
 import org.apache.flink.table.types.logical.BooleanType;
 import org.apache.flink.table.types.logical.DateType;
 import org.apache.flink.table.types.logical.DecimalType;
@@ -16,6 +15,7 @@ import org.apache.flink.table.types.logical.RowType;
 import org.apache.flink.table.types.logical.SmallIntType;
 import org.apache.flink.table.types.logical.TimestampType;
 import org.apache.flink.table.types.logical.TinyIntType;
+import org.apache.flink.table.types.logical.VarBinaryType;
 import org.apache.flink.table.types.logical.VarCharType;
 
 import io.delta.standalone.types.DataType;
@@ -64,7 +64,7 @@ public class SchemaConverter {
             case LONG:
                 return new BigIntType(nullable);
             case BINARY:
-                return new BinaryType(nullable, BinaryType.DEFAULT_LENGTH);
+                return new VarBinaryType(nullable, VarBinaryType.MAX_LENGTH);
             case BOOLEAN:
                 return new BooleanType(nullable);
             case BYTE:
--- a/connectors/flink/src/test/java/io/delta/flink/source/internal/SchemaConverterTest.java
+++ b/connectors/flink/src/test/java/io/delta/flink/source/internal/SchemaConverterTest.java
@@ -5,7 +5,6 @@ import java.util.stream.Stream;
 
 import org.apache.flink.table.types.logical.ArrayType;
 import org.apache.flink.table.types.logical.BigIntType;
-import org.apache.flink.table.types.logical.BinaryType;
 import org.apache.flink.table.types.logical.BooleanType;
 import org.apache.flink.table.types.logical.DateType;
 import org.apache.flink.table.types.logical.DecimalType;
@@ -18,6 +17,7 @@ import org.apache.flink.table.types.logical.RowType;
 import org.apache.flink.table.types.logical.SmallIntType;
 import org.apache.flink.table.types.logical.TimestampType;
 import org.apache.flink.table.types.logical.TinyIntType;
+import org.apache.flink.table.types.logical.VarBinaryType;
 import org.apache.flink.table.types.logical.VarCharType;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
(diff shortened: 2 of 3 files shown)


newfront:
The only check failing is the lint `/home/runner/work/delta/delta/connectors/flink/src/test/java/io/delta/flink/source/internal/SchemaConverterTest.java, Line: 74, Message: Line is longer than 100 characters (found 114)`.

allisonport-db (approved):
LGTM

PR #4026, opened 2025-01-08
[Spark] Throw user-facing exception when the value specified for delta.columnMapping.mode in DF's option causes NullPointerException

longvu-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Currently, in DataFrame's `option` API, if users specified a value that causes `NullPointerException` for the `delta.columnMapping.mode`, then a `NullPointerException` exception is thrown.

This exception is rather unclear for users, so for this case, we are throwing the user-facing `DeltaIllegalArgumentException` instead.

I checked all the `DeltaConfigs`, it seems that only Column Mapping mode faces this problems, all other DeltaConfigs are

1. Accepts `null` as a valid value.
2. Throws `IllegalArgumentException`.
3. Throws an user-facing exception for null cases.

## How was this patch tested?
Added UT.

Specifying `option("delta.columnMapping.mode", null)` will now throw a `DeltaIllegalArgumentException` instead of `NullPointerException`.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaColumnMapping.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaColumnMapping.scala
@@ -872,12 +872,9 @@ case object NameMapping extends DeltaColumnMappingMode {
 
 object DeltaColumnMappingMode {
   def apply(columnMappingModeString: String): DeltaColumnMappingMode = {
-    val columnMappingModeLowerCaseString = try {
-      columnMappingModeString.toLowerCase(Locale.ROOT)
-    } catch {
-      case ex: NullPointerException =>
-        throw DeltaErrors.unsupportedColumnMappingModeException(columnMappingModeString)
-    }
+    val columnMappingModeLowerCaseString =
+      Option(columnMappingModeString).map(_.toLowerCase(Locale.ROOT)).
+        getOrElse(throw DeltaErrors.unsupportedColumnMappingModeException(columnMappingModeString))
     columnMappingModeLowerCaseString match {
       case NoMapping.name => NoMapping
       case IdMapping.name => IdMapping
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaColumnMapping.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaColumnMapping.scala
@@ -871,8 +871,14 @@ case object NameMapping extends DeltaColumnMappingMode {
 }
 
 object DeltaColumnMappingMode {
-  def apply(name: String): DeltaColumnMappingMode = {
-    name.toLowerCase(Locale.ROOT) match {
+  def apply(columnMappingModeString: String): DeltaColumnMappingMode = {
+    val columnMappingModeLowerCaseString = try {
+      columnMappingModeString.toLowerCase(Locale.ROOT)
+    } catch {
+      case ex: NullPointerException =>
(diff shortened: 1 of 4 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/DeltaColumnMapping.scala
```diff
@@ -871,8 +871,14 @@ case object NameMapping extends DeltaColumnMappingMode {
 }
 
 object DeltaColumnMappingMode {
-  def apply(name: String): DeltaColumnMappingMode = {
-    name.toLowerCase(Locale.ROOT) match {
+  def apply(columnMappingModeString: String): DeltaColumnMappingMode = {
+    val columnMappingModeLowerCaseString = try {
+      columnMappingModeString.toLowerCase(Locale.ROOT)
+    } catch {
+      case ex: NullPointerException =>
```
longvu-db, line 878: I also have the option to throw

https://github.com/delta-io/delta/blob/633214e5143b6397a2190a45b5f3f5670b5b8ce4/spark/src/main/scala/org/apache/spark/sql/delta/DeltaErrors.scala#L2050

but `"The column mapping mode $mode is " + s"not supported for this Delta version. Please upgrade if you want to use this mode."` could cause confusion, upgrade won't solve the problem and we will probably never support `null` as a valid value (we already have `none`)

`unsupportedModeException` sounds like a good fit so there is no need to introduce a new Exception just for column mapping.

spark/src/main/scala/org/apache/spark/sql/delta/DeltaColumnMapping.scala
```diff
@@ -871,8 +871,14 @@ case object NameMapping extends DeltaColumnMappingMode {
 }
 
 object DeltaColumnMappingMode {
-  def apply(name: String): DeltaColumnMappingMode = {
-    name.toLowerCase(Locale.ROOT) match {
+  def apply(columnMappingModeString: String): DeltaColumnMappingMode = {
+    val columnMappingModeLowerCaseString = try {
+      columnMappingModeString.toLowerCase(Locale.ROOT)
```
longvu-db, line 876: `toLowerCase` throws NullPointerException when string is `null`, probably only `null` value for `columnMappingModeString` throws that exception here, but I do a try catch just to be more safe.

spark/src/main/scala/org/apache/spark/sql/delta/DeltaColumnMapping.scala
```diff
@@ -871,8 +871,14 @@ case object NameMapping extends DeltaColumnMappingMode {
 }
 
 object DeltaColumnMappingMode {
-  def apply(name: String): DeltaColumnMappingMode = {
-    name.toLowerCase(Locale.ROOT) match {
+  def apply(columnMappingModeString: String): DeltaColumnMappingMode = {
+    val columnMappingModeLowerCaseString = try {
```
xupefei, line 875: Imo this is too much defensive programming. A cleaner and more Scalatic way of doing this is
`Option(columnMappingModeString).map(case ...).getOrElse(throw ...)`
because `Option(null)` will be `None`.

spark/src/main/scala/org/apache/spark/sql/delta/DeltaColumnMapping.scala
xupefei: Not needed. No other case will throw NPE.

xupefei (approved):

PR #4036, opened 2025-01-10
[Spark] Remove protocol and metadata from incremental checksum computation function

sumeet-db:
- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This is a small refactoring PR that removes protocol and metadata from incremental checksum computation method since they can be easily fetched from the actions.

## How was this patch tested?

Existing UTs

No


spark/src/main/scala/org/apache/spark/sql/delta/Checksum.scala
```diff
@@ -213,6 +209,10 @@ trait RecordChecksum extends DeltaLogging {
     // Incrementally compute the new version checksum, if the old one is available.
     val ignoreAddFilesInOperation =
       RecordChecksum.operationNamesWhereAddFilesIgnoredForIncrementalCrc.contains(operationName)
+    val protocol =
+      actions.collectFirst { case p: Protocol => p }.getOrElse(oldVersionChecksum.protocol)
```
prakharjain09, line 213: If the CRC file is missing Protocol, can `oldVersionChecksum.protocol` be null?

Same below for Metadata?

prakharjain09 (approved):

PR #4037, opened 2025-01-10
[Kernel][Metrics][PR#6] Support TransactionReport to log metrics for a Transaction operation

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Adds `TransactionReport` for reporting an attempted or successful transaction.

We record `TransactionReport` either after successfully committing the transaction or if an exception is thrown during the commit attempt. We only record a report for failed _commit_ attempts. If there are failures elsewhere, such as while writing the data, they will not be reported.

We also add support for serializing `TransactionReport`s in this PR.

## How was this patch tested?

Adds unit tests.

No.


kernel/kernel-api/src/main/java/io/delta/kernel/metrics/SnapshotMetricsResult.java
```diff
@@ -15,9 +15,11 @@
  */
 package io.delta.kernel.metrics;
 
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import java.util.Optional;
 
 /** Stores the metrics results for a {@link SnapshotReport} */
+@JsonPropertyOrder({"timestampToVersionResolutionDurationNs", "loadInitialDeltaActionsDurationNs"})
```
scottsand-db, line 22: Could you tell me about this? (As a reply to this comment; don't need you to add anything to the code).

kernel/kernel-api/src/main/java/io/delta/kernel/metrics/TransactionReport.java
```diff
@@ -0,0 +1,79 @@
...
+
+  /**
+   * The version of the table the transaction was created from. For example, if the latest table
+   * version is 4 when the transaction is created, the transaction is based off of the snapshot of
+   * the table at version 4. For a new table (e.g. a transaction that is creating a table) this is
+   * -1.
+   *
+   * @return the table version of the snapshot the transaction was started from
+   */
+  long getSnapshotVersion();
```
scottsand-db, line 59: Just some ideas, since I had to read the method docs to understand this method name. `getBaseSnapshotVersion`? `getStartVersion`?

Not blockers and I'm fine with as is, too.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/TransactionMetrics.java
```diff
@@ -0,0 +1,96 @@
...
+  public final Counter commitAttemptsCounter = new Counter();
+
+  public final Counter addFilesCounter = new Counter();
+
+  public final Counter removeFilesCounter = new Counter();
+
+  public final Counter totalActionsCounter = new Counter();
+
+  /**
+   * Resets the action counters (addFilesCounter, removeFilesCounter and totalActionsCounter) to 0
```
scottsand-db, line 41: Could you add a bit of context _why_ someone would want to reset them?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/metrics/MetricsReportTestUtils.scala
```diff
@@ -0,0 +1,87 @@
...
+
+  def resolvePath(path: String): String = {
+    defaultEngine.getFileSystemClient.resolvePath(path)
+  }
+
+  /**
+   * Wraps an {@link Engine} to implement the metrics reporter such that it appends any reports
+   * to the provided in memory buffer.
+   */
+  class EngineWithInMemoryMetricsReporter(buf: ArrayBuffer[MetricsReport], baseEngine: Engine)
```
scottsand-db, line 63: why does it take in the buffer to which we want to append the metrics?

would it ever take in a buffer that is non-empty? all the usages I see use `val reports = ArrayBuffer.empty[MetricsReport]`

do our engine APIs not make it simple enough to get the metrics reports?

scottsand-db (changes requested):
Looks great! Left some comments and asked some questions. Will review the rest of the tests later

kernel/kernel-api/src/main/java/io/delta/kernel/metrics/SnapshotMetricsResult.java
allisonport-db: I found in some test runs the order the fields were serialized in was not deterministic so added this.

kernel/kernel-api/src/main/java/io/delta/kernel/metrics/TransactionReport.java
allisonport-db: I'm good with `getBaseSnapshotVersion`. I feel like `startVersion` means something else

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/TransactionReportImpl.java
```diff
@@ -0,0 +1,95 @@
...
+public class TransactionReportImpl extends DeltaOperationReportImpl implements TransactionReport {
+
+  private final String operation;
+  private final String engineInfo;
+  private final long snapshotVersion;
+  private final Optional<UUID> snapshotReportUUID;
+  private final Optional<Long> committedVersion;
+  private final TransactionMetricsResult transactionMetrics;
+
+  public TransactionReportImpl(
```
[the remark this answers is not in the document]
allisonport-db: I can make this change, it just means there will be slightly more code duplication in `TransactionImpl` when building the reports since I won't be able to unify the construction across the two cases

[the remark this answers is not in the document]
allisonport-db: It's also more places to update the param list if we add information to `TransactionReport` (since we don't have a need for a context storing sort of object yet)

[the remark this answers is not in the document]
allisonport-db: I don't feel strongly about this happy to do anything here

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/metrics/MetricsReportTestUtils.scala
```diff
@@ -0,0 +1,87 @@
...
+        f(new EngineWithInMemoryMetricsReporter(reports, defaultEngine))
+      )
+      (reports, Some(e))
+    } else {
+      f(new EngineWithInMemoryMetricsReporter(reports, defaultEngine))
+      (reports, Option.empty)
+    }
+  }
+
+  def resolvePath(path: String): String = {
```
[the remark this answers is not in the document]
allisonport-db: It's used once in both test suites, and presumably will be used by any future metric report suites

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/TransactionReportImpl.java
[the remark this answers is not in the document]
scottsand-db: To be clear, I wasn't asking you to make the change. I just wanted to know the reason to do it or _not_ to do it.

You reasons to _not_ do it make sense to me! Can leave as is 👍

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/metrics/MetricsReportTestUtils.scala
allisonport-db: I guess this implementation could create the empty buffer itself, but we would still need it to be externally accessible so we can access the reports. I don't think there's much value added by doing this and I think it would actually make `collectMetricsReports` a bit more complex.

> do our engine APIs not make it simple enough to get the metrics reports?

I think this is pretty simple :) we don't provide any out-of-the-box functionality to buffer the reports in memory thus we need to implement an "in memory metrics reporter". We add this to the existing defaultEngine reporter such that we have both this one, and the default logging reporter so we do log the reports in the test suite as well.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/TransactionMetrics.java
allisonport-db: Added

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/metrics/MetricsReportTestUtils.scala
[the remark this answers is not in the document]
scottsand-db: I see. This is still just a passthrough method -- `resolvePath(path)` or `defaultEngine.getFileSystemClient.resolvePath(path)` both fit on one line. What utility does this method really add?

I'm also concerned with _coupling_. I could have

```java
val engine = ... setup some engine ...
resolvePath(path) // this doesn't use the engine above, it uses the defaultEngine
```

whereas if we just used `<engine>.getFileSystemClient.resolvePath(path)` -- this is more obvious and less prone to errors / mixups

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/metrics/MetricsReportTestUtils.scala
scottsand-db: Sounds good to me. Thanks for explaining

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -342,6 +382,23 @@ private Map<String, String> getOperationParameters() {
...
+      TransactionMetrics transactionMetrics,
+      Optional<Exception> exception) {
+    TransactionReport transactionReport =
+        new TransactionReportImpl(
+            dataPath.toString() /* tablePath */,
+            operation.toString(),
+            engineInfo,
+            committedVersion,
+            transactionMetrics,
+            readSnapshot.getSnapshotReport(),
```
scottsand-db, line 397: what is the readSnapshot of a new table creation? is it just `InitialSnapshot`? with version = -1?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/TransactionReportImpl.java
```diff
@@ -0,0 +1,95 @@
...
+    requireNonNull(snapshotReport);
+    checkArgument(
+        !snapshotReport.getException().isPresent(),
+        "Expected a successful SnapshotReport provided report has exception");
+    checkArgument(
+        snapshotReport.getVersion().isPresent(),
+        "Expected a successful SnapshotReport but missing version");
+    this.snapshotVersion = requireNonNull(snapshotReport).getVersion().get();
+    if (snapshotVersion < 0) {
+      // For a new table, no Snapshot is actually loaded and thus no SnapshotReport is emitted
```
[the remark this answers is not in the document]
scottsand-db: Ok. I think we should add a short comment/javadocs to this constructor detailing this special case (i.e. table creation vs existing-table-write)

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/metrics/TransactionReportSuite.scala
```diff
@@ -0,0 +1,378 @@
...
+   * duration it takes for [[Transaction#commit]] to finish.
+   *
+   * @param getTransaction given an engine return a started [[Transaction]]
+   * @param generateCommitActions given a [[Transaction]] and engine generates actions to commit
+   * @param expectException whether we expect committing to throw an exception, which if so, is
+   *                        caught and returned with the other results
+   * @return (TransactionReport, durationToCommit, SnapshotReportIfPresent, ExceptionIfThrown)
+   */
+  def getTransactionAndSnapshotReport(
+    getTransaction: Engine => Transaction,
```
scottsand-db, line 51: nit: `createTransaction`?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/metrics/TransactionReportSuite.scala
```diff
@@ -0,0 +1,378 @@
...
+        generateCommitActions = actionsIterableWithConcurrentAppend,
+        path,
+        expectException = true,
+        expectedSnapshotVersion = 0,
+        expectedNumAttempts = 200
+      )
+    }
+  }
+
+  ///////////////////////////
```
scottsand-db, line 364: nit: un-aligned end of comment block

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/metrics/TransactionReportSuite.scala
```diff
@@ -0,0 +1,378 @@
...
+   * @param buildTransaction function to build a transaction from a transaction builder
+   * @param engineInfo engine info to create the transaction with
+   * @param operation operation to create the transaction with
+   */
+  // scalastyle:off
+  def checkTransactionReport(
+    generateCommitActions: (Transaction, Engine) => CloseableIterable[Row],
+    path: String,
+    expectException: Boolean,
+    expectedSnapshotVersion: Long,
```
scottsand-db, line 97: `expectedBaseSnapshotVersion` -- makes it clearer that this is the base/start version, not the end version after the transaction completes. also this aligns better with `transactionReport.getBaseSnapshotVersion`

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/metrics/TransactionReportSuite.scala
```diff
@@ -0,0 +1,378 @@
...
+          override def close(): Unit = ()
+        }
+      }
+
+      checkTransactionReport(
+        generateCommitActions = actionsIterableWithConcurrentAppend,
+        path,
+        expectException = true,
+        expectedSnapshotVersion = 0,
+        expectedNumAttempts = 200
```
scottsand-db, line 359: would be great if you could rebase this PR off of https://github.com/delta-io/delta/pull/4054 (once it is merged)

so that we don't retry 200 times but maybe more like 5. 200 is probably very slow.

scottsand-db (changes requested):
Looks great! Tests are very thorough. Added some minor feedback and a question or two. Will approve once this minor feedback is implemented.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/metrics/TransactionReportSuite.scala
allisonport-db: Lol I also was very confused when I wrote this test because I did not expect it to be 200!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
allisonport-db: Yeah we create it here https://github.com/delta-io/delta/blob/master/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java#L114

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/metrics/MetricsReportTestUtils.scala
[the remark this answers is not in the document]
allisonport-db: We are already using the `defaultEngine` everywhere that uses `MetricsReportTestUtils`; we don't reference `defaultEngine` anywhere in the child suites (I mention this [here](https://github.com/delta-io/delta/pull/4037/files#diff-1beefd1d9eebcd7c315764bed29cea7a21dd466f412bab21e6e5ac8e7173bd03R34-R35)) so I personally think this is more clear for the current setup, since it consistently abstracts the underlying engine implementation from the suite.

But I don't feel that strongly about it so happy to change it.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/TransactionReportImpl.java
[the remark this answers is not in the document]
allisonport-db: Added java docs!

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/metrics/MetricsReportTestUtils.scala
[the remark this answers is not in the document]
scottsand-db:
```java
// For now this just uses the default engine since we have no need to override it, if we would
// like to use a specific engine in the future for other tests we can simply add another arg here
```

Suppose we add a new engine there as a new parameter. Well, this is a separate method from `resolvePath`. So `collectMetricsReports` is using some new `CustomEngine`, and there's no way (as the code is right now) to tie that same engine to `resolvePath`.

Overall, `resolvePath` is just a passthrough method that isn't coupled to the same engine implementation as other methods in the class. (We could solve this by having the trait define a def currEngine .. lots of different ways to solve this). And we should tend to avoid passthrough methods. I think we should remove it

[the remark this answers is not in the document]
allisonport-db: Updated

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/metrics/TransactionReportSuite.scala
scottsand-db: Could you work with @nicklan to get #4054 merged? Perhaps take over his PR if he is busy? After that, will approve this PR :)

allisonport-db: Merged & updated

scottsand-db (approved):
LGTM!

PR #4041, opened 2025-01-13
[Spark] Refactor out Delta read path

vladimirg-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Refactored out the following read path functionality:
- V2 -> V1 relation conversions.
- Table with partition filters resolution.

This is a refactor-only change to support the single-pass Analyzer project in Spark: https://issues.apache.org/jira/browse/SPARK-49834.

It will be used in single-pass resolver extensions: https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/resolver/ResolverExtension.scala.

## How was this patch tested?

Existing tests.

No.


c27kwan (approved):
LGTM! Nice cleanup :)

PR #4042, opened 2025-01-13
[Spark] Handle type widening in Delta streaming source

johanl-db:
## Description
Streaming reads from a Delta source currently don't allow any form of type change to be propagated: the stream fails with a non-retryable error. Type widening introduced the ability to change the type of an existing column or field in a Delta table.

This change allows widening type changes to be propagated during streaming reads from a Delta source. The same mechanism as non-additive schema changes from column mapping (Drop/Rename) is applied.
To allow widening type changes to propagate, the user must:
- Provide a schema tracking location via `.option("schemaTrackingLocation", ..)`
- Review and accept the type change by setting one of the available SQL confs (detailed in the error message)

Note that the current check for column mapping had a loophole: as long as a schema tracking location was provided, it only gated column drop/rename but allowed *any* type changes to get through. This is fixed here by properly checking and rejecting non-widening type changes.

## How was this patch tested?
- Added unit tests in `DeltaSourceMetadataEvolutionSupportSuite` covering the logic to detect and gate non-additive schema changes.
- Added test suite `TypeWideningStreamingSourceSuite` to cover stream reads from a delta source when a column is widened. 

When reading using a streaming query from a Delta table that had a column type widened:

### Before this change:
The stream fails with a non-retryable error:
```
[DELTA_SCHEMA_CHANGED_WITH_VERSION]
Detected schema change in version <version>:
streaming source schema: <readSchema>
data file schema: <dataSchema>
Please try restarting the query. If this issue repeats across query restarts without
making progress, you have made an incompatible schema change and need to start your
query from scratch using a new checkpoint directory.
```
Note: users could get around this by providing a schema tracking location and applying a column drop or rename to their source. This allowed *arbitrary* type changes to go through unchecked.

### After this change:
Users must provide a schema tracking location via `.option("schemaTrackingLocation")`, otherwise the stream fails with:
```
[DELTA_STREAMING_INCOMPATIBLE_SCHEMA_CHANGE_USE_SCHEMA_LOG]
Streaming read is not supported on tables with read-incompatible schema changes (e.g. rename or drop or datatype changes).
Please provide a 'schemaTrackingLocation' to enable non-additive schema evolution for Delta stream processing.
See <docLink> for more details.
Read schema: <readSchema>. Incompatible data schema: <incompatibleSchema>.
```
When a schema tracking location is provided, non-widening type changes are now properly rejected and fail with error`[DELTA_SCHEMA_CHANGED_WITH_VERSION]` 

When reading the batch that contains the type change, the stream first fail and records the tracked schema:
```
[DELTA_STREAMING_METADATA_EVOLUTION]
The schema, table configuration or protocol of your Delta table has changed during streaming.
The schema or metadata tracking log has been updated.
Please restart the stream to continue processing using the updated metadata.
Updated schema: <schema>.
Updated table configurations: <config>.
Updated table protocol: <protocol>
```
On retry, the stream fails and the user is prompted with a call to action:
```
[DELTA_STREAMING_CANNOT_CONTINUE_PROCESSING_TYPE_WIDENING]
We've detected one or more data types being widened between Delta version 2 and 3:
  col_a: INT -> BIGINT

Your streaming query contains operations that may fail or produce different results after the type change(s).
Please check if you want to update your streaming query before we proceed with stream processing using the finalized schema at version 2.
Once you have updated your streaming query or have decided there is no need to update it, you can set (one of) the following SQL configurations to unblock the type change(s) and continue stream processing.
To unblock for this particular stream just for this series of type change(s): set `spark.databricks.delta.streaming.allowSourceTypeWidening.ckpt_123456 = 2`.
To unblock for this particular stream: set `spark.databricks.delta.streaming.allowSourceTypeWidening.ckpt_123456 = always`
To unblock for all streams: set `spark.databricks.delta.streaming.allowSourceTypeWidening = always`.
```

Users can set one of the proposed config to resume processing.

### DIFF
--- a/spark/src/main/resources/error/delta-error-classes.json
+++ b/spark/src/main/resources/error/delta-error-classes.json
@@ -2310,11 +2310,15 @@
     "message" : [
       "We've detected one or more non-additive schema change(s) (<opType>) between Delta version <previousSchemaChangeVersion> and <currentSchemaChangeVersion> in the Delta streaming source.",
       "Please check if you want to manually propagate the schema change(s) to the sink table before we proceed with stream processing using the finalized schema at version <currentSchemaChangeVersion>.",
-      "Once you have fixed the schema of the sink table or have decided there is no need to fix, you can set (one of) the following SQL configurations to unblock the non-additive schema change(s) and continue stream processing.",
-      "To unblock for this particular stream just for this series of schema change(s): set `<allowCkptVerKey> = <allowCkptVerValue>`.",
-      "To unblock for this particular stream: set `<allowCkptKey> = <allowCkptValue>`",
-      "To unblock for all streams: set `<allowAllKey> = <allowAllValue>`.",
-      "Alternatively if applicable, you may replace the `<allowAllMode>` with `<opSpecificMode>` in the SQL conf to unblock stream for just this schema change type."
+      "Once you have fixed the schema of the sink table or have decided there is no need to fix, you can set the following configuration(s) to unblock the non-additive schema change(s) and continue stream processing.",
+      "",
+      "To unblock for this particular stream just for this series of schema change(s):",
+      "<unblockChangeConfs>",
+      "To unblock for this particular stream:",
+      "<unblockStreamConfs>",
+      "To unblock for all streams:",
+      "<unblockAllConfs>",
+      ""
     ],
     "sqlState" : "KD002"
   },
@@ -2324,11 +2328,16 @@
       "<wideningTypeChanges>",
       "",
       "Your streaming query contains operations that may fail or produce different results after the type change(s).",
-      "Please check if you want to update your streaming query before we proceed with stream processing using the finalized schema at <currentSchemaChangeVersion>.",
-      "Once you have updated your streaming query or have decided there is no need to update it, you can set (one of) the following SQL configurations to unblock the type change(s) and continue stream processing.",
-      "To unblock for this particular stream just for this series of type change(s): set `<allowCkptVerKey> = <allowCkptVerValue>`.",
-      "To unblock for this particular stream: set `<allowCkptKey> = <allowCkptValue>`",
-      "To unblock for all streams: set `<allowAllKey> = <allowAllValue>`."
+      "Please check if you want to update your streaming query before we proceed with stream processing using the finalized schema at version <currentSchemaChangeVersion>.",
+      "Once you have updated your streaming query or have decided there is no need to update it, you can set the following configuration to unblock the type change(s) and continue stream processing.",
+      "",
+      "To unblock for this particular stream just for this series of schema change(s):",
+      "<unblockChangeConfs>",
+      "To unblock for this particular stream:",
+      "<unblockStreamConfs>",
+      "To unblock for all streams:",
+      "<unblockAllConfs>",
+      ""
     ],
     "sqlState" : "KD002"
   },
--- a/spark/src/main/resources/error/delta-error-classes.json
+++ b/spark/src/main/resources/error/delta-error-classes.json
@@ -2309,12 +2309,26 @@
   "DELTA_STREAMING_CANNOT_CONTINUE_PROCESSING_POST_SCHEMA_EVOLUTION" : {
     "message" : [
       "We've detected one or more non-additive schema change(s) (<opType>) between Delta version <previousSchemaChangeVersion> and <currentSchemaChangeVersion> in the Delta streaming source.",
-      "Please check if you want to manually propagate the schema change(s) to the sink table before we proceed with stream processing using the finalized schema at <currentSchemaChangeVersion>.",
+      "Please check if you want to manually propagate the schema change(s) to the sink table before we proceed with stream processing using the finalized schema at version <currentSchemaChangeVersion>.",
       "Once you have fixed the schema of the sink table or have decided there is no need to fix, you can set (one of) the following SQL configurations to unblock the non-additive schema change(s) and continue stream processing.",
       "To unblock for this particular stream just for this series of schema change(s): set `<allowCkptVerKey> = <allowCkptVerValue>`.",
       "To unblock for this particular stream: set `<allowCkptKey> = <allowCkptValue>`",
       "To unblock for all streams: set `<allowAllKey> = <allowAllValue>`.",
       "Alternatively if applicable, you may replace the `<allowAllMode>` with `<opSpecificMode>` in the SQL conf to unblock stream for just this schema change type."
+    ],
+    "sqlState" : "KD002"
+  },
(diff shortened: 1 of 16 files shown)


spark/src/main/resources/error/delta-error-classes.json
```diff
@@ -2318,6 +2318,20 @@
     ],
     "sqlState" : "KD002"
   },
+  "DELTA_STREAMING_CANNOT_CONTINUE_PROCESSING_TYPE_WIDENING" : {
```
tomvanbussel, line 2321: Nit: Do we need a separate error message for this? Type widening is also a form of schema evolution. If we do, should we change it to say "data types being changed" instead of "data types being widened". I don't think this is specific to type widening (expect for those being the only type changes that we support right now).

spark/src/main/scala/org/apache/spark/sql/delta/schema/SchemaUtils.scala
```diff
@@ -403,6 +406,7 @@ def normalizeColumnNamesInDataType(
       readSchema: StructType,
       forbidTightenNullability: Boolean = false,
       allowMissingColumns: Boolean = false,
+      allowMissingStructFields: Boolean = false,
```
tomvanbussel, line 409: This change seems unrelated. Should we separate it out?

spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMetadata.scala
```diff
@@ -171,23 +175,48 @@ private[delta] object TypeWideningMetadata extends DeltaLogging {
     schemaWithMetadata
   }
 
+  /**
+   * Recursively compare `from` and `to` to collect all primitive type widening changes, including
+   * in nested structs, maps and arrays.
+   */
+  def collectTypeChanges(from: StructType, to: StructType): Seq[TypeChange] = {
```
tomvanbussel, line 182: Nit: It's a bit confusing to have two methods called `collectTypeChanges`. Why couldn't we add a `case (from: StructType, to: StructType)` to the other one?

spark/src/test/scala/org/apache/spark/sql/delta/sources/DeltaSourceMetadataEvolutionSupportSuite.scala
```diff
@@ -0,0 +1,570 @@
...
+  )
+
+  testSchemaChange(
+    name = "column rename with widening type change",
+    fromDDL = "a byte",
+    fromPhysicalNames = Map(Seq("a") -> "x"),
+    toDDL = "b int",
+    toPhysicalNames = Map(Seq("b") -> "x"),
+    expectedResult = expectColumnMappingChangeBlocked("RENAME COLUMN"),
+    // We don't block the type change itself: either the user treats 'b' as a nwe column and the
```
tomvanbussel, line 188: Typo
```suggestion
    // We don't block the type change itself: either the user treats 'b' as a new column and the
```

spark/src/test/scala/org/apache/spark/sql/delta/sources/DeltaSourceMetadataEvolutionSupportSuite.scala
```diff
@@ -0,0 +1,570 @@
...
+      "allowSourceColumnRenameAndDropAndTypeWidening")
+  )
+
+  testSchemaChange(
+    name = "column rename with non-widening type change",
+    fromDDL = "a int",
+    fromPhysicalNames = Map(Seq("a") -> "x"),
+    toDDL = "b string",
+    toPhysicalNames = Map(Seq("b") -> "x"),
+    // We don't block the type change itself: either the user treats 'b' as a nwe column and the
```
tomvanbussel, line 204: Typo
```suggestion
    // We don't block the type change itself: either the user treats 'b' as a new column and the
```

tomvanbussel (commented):
Nice! Left a few comments.

spark/src/main/resources/error/delta-error-classes.json
johanl-db: That helps with identifying how the schema change happened, I've leaned more on identifying it as 'type widening'
We can start with being specific about the change being a widening type change since these are the only ones we capture with this error, and update the wording in the future if we want to add more.

I'm changing the SQL conf from allowSourceTypeWidening to allowSourceTypeChange though, since changing that later would be a breaking change.

spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMetadata.scala
johanl-db: I renamed the internal collectTypeChanges to collectTypeChangesInStructField to better distinguish them.

We do need collectTypeChangesInStructField that doesn't recurse into structs to be able to map all type changes to their nearest struct field ancestor, which is where we attach the metadata recording the history of type changes within that struct field (to go a bit deeper: this is due to the fact that we can't attach metadata to map key/values and array elements, so we need collect type changes inside these while ignoring structs)

spark/src/main/scala/org/apache/spark/sql/delta/schema/SchemaUtils.scala
johanl-db: Used in checkIncompatibleSchemaChange: we want to detect incompatible type changes while ignoring dropped/renamed columns and nested fields.

I believe that when this flag was introduced, it wasn't properly passed down to the recursive isReadCompatible call. As a result, we ignore missing top-level columns but not missing struct fields, although there's no real reason to distinguish the two.

I actually see allowMissingColumns is only used in DeltaSource
Since it is only set when allowUnsafeStreamingReadOnColumnMappingSchemaChanges is true (which is an internal flag defaulting to false meant as an escape hatch), and given that code should really also ignore changes struct fields instead of failing, it seems better (and safe) to actually fix passing down allowMissingColumns to also apply to struct fields.

spark/src/test/scala/org/apache/spark/sql/delta/typewidening/TypeWideningStreamingSourceSuite.scala
```diff
@@ -0,0 +1,476 @@
...
+    query = _.selectExpr("scala_udf(widened)"),
+    expectedResult = ExpectedResult.Success(Seq(Row(123456790)))
+  )
+
+  testStreamTypeWidening("widen aggregation grouping key",
+    query = _.groupBy("widened").agg(count(col("other"))),
+    expectedResult = ExpectedResult.Failure { ex =>
+      assert(ex.getErrorClass === "STATE_STORE_KEY_SCHEMA_NOT_COMPATIBLE")
+    },
+    /*
```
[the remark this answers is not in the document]
johanl-db: Yes

vkorukanti (approved):

PR #4047, opened 2025-01-14
[Spark] Add extra metrics for insert overwrite operation

Alexvsalexvsalex:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
When `INSERT OVERWRITE` is used it deletes existing data and write new one. We have statistics about new wrote data, but it is unknown how much data we deleted.
Changes in the PR suggests to enrich statistics by two extra metrics `numRemovedFiles` and `numRemovedBytes` to have basic understanding of deleted data.

## How was this patch tested?
Added new unit tests.

New metrics `numRemovedFiles` and `numRemovedBytes` will appear in table's history for insert overwrite.
If anybody wants to return back old behavior then they need to disable `spark.databricks.delta.insertOverwrite.removeMetrics.enabled`.


larsk-db (approved):
LGTM

PR #4053, opened 2025-01-14
[Kernel] Implement row tracking conflict resolution

qiyuandong-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This PR implements transaction conflict resolution for the row tracking feature. Key changes include:

1. Add logic for resolving conflicts for the row tracking metadata domain
2. Reassign `baseRowId` and `defaultRowCommitVersion` for `AddFile` actions during conflict resolution
3. Update the row tracking domain metadata action during conflict resolution

## How was this patch tested?

Tests were added in `RowTrackingSuite.scala`.

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
@@ -26,14 +26,18 @@ import static java.lang.String.format;
 
 import io.delta.kernel.data.ColumnVector;
 import io.delta.kernel.data.ColumnarBatch;
+import io.delta.kernel.data.Row;
 import io.delta.kernel.engine.Engine;
 import io.delta.kernel.exceptions.ConcurrentWriteException;
 import io.delta.kernel.internal.*;
 import io.delta.kernel.internal.actions.CommitInfo;
 import io.delta.kernel.internal.actions.DomainMetadata;
 import io.delta.kernel.internal.actions.SetTransaction;
+import io.delta.kernel.internal.rowtracking.RowTracking;
+import io.delta.kernel.internal.rowtracking.RowTrackingMetadataDomain;
 import io.delta.kernel.internal.util.DomainMetadataUtils;
 import io.delta.kernel.internal.util.FileNames;
+import io.delta.kernel.utils.CloseableIterable;
 import io.delta.kernel.utils.CloseableIterator;
 import io.delta.kernel.utils.FileStatus;
 import java.io.*;
@@ -61,11 +65,20 @@ public class ConflictChecker {
   // Losing transaction
   private final TransactionImpl transaction;
   private final long attemptVersion;
+  private final CloseableIterable<Row> attemptDataActions;
 
-  private ConflictChecker(SnapshotImpl snapshot, TransactionImpl transaction, long attemptVersion) {
+  // Helper states during conflict resolution
+  private Optional<Long> lastWinningRowIdHighWatermark = Optional.empty();
+
+  private ConflictChecker(
+      SnapshotImpl snapshot,
+      TransactionImpl transaction,
+      long attemptVersion,
+      CloseableIterable<Row> dataActions) {
     this.snapshot = snapshot;
     this.transaction = transaction;
     this.attemptVersion = attemptVersion;
+    this.attemptDataActions = dataActions;
   }
 
   /**
@@ -76,15 +89,22 @@ public class ConflictChecker {
    * @param snapshot {@link SnapshotImpl} of the table when the losing transaction has started
    * @param transaction {@link TransactionImpl} that encountered the conflict (a.k.a the losing
    *     transaction)
+   * @param dataActions {@link CloseableIterable} of data actions that the losing transaction is
+   *     trying to commit
    * @return {@link TransactionRebaseState} that the losing transaction needs to rebase against
    * @throws ConcurrentWriteException if there are logical conflicts between the losing transaction
    *     and the winning transactions that cannot be resolved.
    */
   public static TransactionRebaseState resolveConflicts(
-      Engine engine, SnapshotImpl snapshot, long attemptVersion, TransactionImpl transaction)
+      Engine engine,
+      SnapshotImpl snapshot,
+      long attemptVersion,
+      TransactionImpl transaction,
+      CloseableIterable<Row> dataActions)
       throws ConcurrentWriteException {
     checkArgument(transaction.isBlindAppend(), "Current support is for blind appends only.");
-    return new ConflictChecker(snapshot, transaction, attemptVersion).resolveConflicts(engine);
+    return new ConflictChecker(snapshot, transaction, attemptVersion, dataActions)
+        .resolveConflicts(engine);
   }
 
   public TransactionRebaseState resolveConflicts(Engine engine) throws ConcurrentWriteException {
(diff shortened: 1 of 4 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/RowTracking.java
```diff
@@ -55,11 +56,6 @@ private RowTracking() {
    */
   public static CloseableIterable<Row> assignBaseRowIdAndDefaultRowCommitVersion(
       SnapshotImpl snapshot, long commitVersion, CloseableIterable<Row> dataActions) {
-    checkArgument(
-        TableFeatures.isRowTrackingSupported(snapshot.getProtocol()),
-        "Base row ID and default row commit version are assigned "
-            + "only when feature 'rowTracking' is supported.");
-
```
qiyuandong-db, line 62: This check is already handled at all caller sites and I think it is kind of redundant to do it again within the method.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/RowTracking.java
```diff
@@ -135,6 +127,131 @@ public static Optional<DomainMetadata> createNewHighWaterMarkIfNeeded(
...
+          if (!row.isNullAt(SingleAction.ADD_FILE_ORDINAL)) {
+            AddFile addFile = new AddFile(row.getStruct(SingleAction.ADD_FILE_ORDINAL));
+            checkArgument(addFile.getBaseRowId().isPresent(), "baseRowId must be present");
+
+            // baseRowIds higher than prevRowIdHighWatermark are assigned by this txn
+            if (addFile.getBaseRowId().get() > prevRowIdHighWatermark) {
+              currRowIdHighWatermark.addAndGet(getNumRecordsOrThrow(addFile));
+            }
+          }
+        });
```
qiyuandong-db, line 245: We could possibly cache the total number of new records in the transaction (i.e., the increment to the row ID high watermark) and just reuse it here. This requires no `AddFile` actions are added or removed from `dataActions` during conflict resolution. This is currently true in Kernel, but I'm uncertain whether this will remain the case in the future.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/RowTracking.java
```diff
@@ -135,6 +127,131 @@ public static Optional<DomainMetadata> createNewHighWaterMarkIfNeeded(
...
+   * since protocol changes between losing and winning transactions are not permitted.
+   *
+   * @param snapshot the snapshot of the table that this transaction is reading at
+   * @param winningRowIdHighWatermark the latest row ID high watermark from the winning transactions
+   * @param attemptCommitVersion the commit version of this losing transaction
+   * @param lastWinningVersion the last winning commit version
+   * @param dataActions the data actions that this losing transaction was trying to commit
+   * @return an iterable of data actions with reassigned baseRowIds and defaultRowCommitVersions
+   */
+  public static CloseableIterable<Row> reassignBaseRowIdAndDefaultRowCommitVersion(
```
johanl-db, line 151: It might make sense to merge the logic to reassign row tracking information + update domain metadata with the logic to assign them initially during commit.
These are essentially the same except that:
- Initial assignment: only assign and increment high-water mark if no baseRowId / default row commit versions 
- Reassignment: only reassign and increment high-water mark if baseRowId > previous high-water mark, default commit version = previous commit version

You could achieve that by passing an optional previous high water mark / commit version: None during initial commit, set for reassignment during conflict resolution

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
```diff
@@ -223,9 +282,19 @@ private void handleDomainMetadata(ColumnVector domainMetadataVector) {
...
-        throw concurrentDomainMetadataAction(currentTxnDM, winningTxnDM);
+        // Domain-specific ways of resolving the conflict can be added here.
+        switch (domainName) {
+          case RowTrackingMetadataDomain.DOMAIN_NAME:
+            // We keep updating the new row ID high watermark we have seen from all winning txns.
+            // The latest one will be used to reassign base row IDs later
+            this.winningRowIdHighWatermark =
+                Optional.of(
+                    RowTrackingMetadataDomain.fromJsonConfiguration(winningTxnDM.getConfiguration())
+                        .getRowIdHighWaterMark());
```
johanl-db, line 293: Directly get `winningRowIdHighWatermark` in `resolveConflicts` instead of having this level of indirection, extracting it here and passing it around as the member field.
One issue besides making it more complex to see where it comes from is that you create a dependency between `handleDomainMetadata` and reassigning row ids: `handleDomainMetadata` must be called first

The switch case for row tracking is then empty, that's fine: we know we will handle that case.

Alternatively you can move the logic to compute the new domain metadata here.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/RowTracking.java
```diff
@@ -135,6 +127,131 @@ public static Optional<DomainMetadata> createNewHighWaterMarkIfNeeded(
...
+      SnapshotImpl snapshot,
+      Optional<Long> winningRowIdHighWatermark,
```
johanl-db, line 153: If there's no row id high watermark in the winning transaction, should we actually reassign row ids ?
the previous row id highwatermark is anyway going to be the we used initially to assign row ids right?

We do need to update default row commit versions though

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/RowTracking.java
johanl-db: I wouldn't rely on that assumption either, the current approach is good enough and robust

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
qiyuandong-db: Makes sense! I've updated this, please see the latest code.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/RowTracking.java
qiyuandong-db: This row id reassignment isn’t necessary if there's no row id high watermark in the winning transaction.

In my latest code, I’ve merged the reassignment logic with the initial assignment logic. It still reassigns row IDs even when the winning txns didn't change the row ID high watermark. This makes the code and logic a bit cleaner, and ensures it is still correct in case in the future `AddFile` actions can be added/removed from the `dataActions` during conflict resolution.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
```diff
@@ -114,36 +140,66 @@ public TransactionRebaseState resolveConflicts(Engine engine) throws ConcurrentW
             handleProtocol(batch.getColumnVector(PROTOCOL_ORDINAL));
             handleMetadata(batch.getColumnVector(METADATA_ORDINAL));
             handleTxn(batch.getColumnVector(TXN_ORDINAL));
-            handleDomainMetadata(batch.getColumnVector(DOMAIN_METADATA_ORDINAL));
+            Map<String, DomainMetadata> winningTxnDomainMetadataMap =
+                handleDomainMetadata(batch.getColumnVector(DOMAIN_METADATA_ORDINAL));
+            getRowIdHighWatermark(winningTxnDomainMetadataMap)
+                .ifPresent(hwm -> lastWinningRowIdHighWatermark.set(Optional.of(hwm)));
```
johanl-db, line 146: I'm sorry, I missed that you had to go over all actions anyway to extract the high-water mark,  somehow thought that it would be readily available from the winning transaction.
 (+ also we allow resolving against multiple commits.

What you had initially we collecting the winning highwater mark was actually cleaner. Maybe also make sure that we collect the highest in case we process multiple commits out of order (which we can't today afaict but better safe than sorry)

johanl-db (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
qiyuandong-db: Sure, I've changed to what we had initially and added a check to ensure we observe monotonically increasing row ID high watermarks from winning transactions

scottsand-db (changes requested):
Looks great! Only had a few comments.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/RowTracking.java
qiyuandong-db: Makes sense! I've added this back. I also updated it to check against the `TransactionImpl`‘s protocol instead of the `readSnapshot`’s protocol, as we should validate the (potentially updated) protocol resulting from this transaction.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/RowTracking.java
```diff
@@ -68,9 +71,15 @@ public void close() throws IOException {
...
-        // Use an AtomicLong to allow for updating the high watermark in the lambda.
-        final AtomicLong currRowIdHighWatermark = new AtomicLong(readRowIdHighWaterMark(snapshot));
+        // The row ID high watermark from the snapshot of the table that this txn is reading at.
+        // Any baseRowIds higher than this watermark are assigned by this txn.
+        final long prevRowIdHighWatermark = readRowIdHighWaterMark(snapshot);
+
+        // Used to track the current high watermark as we iterate through the data actions and
+        // assign baseRowIds. Use an AtomicLong to allow for updating in the lambda.
+        final AtomicLong currRowIdHighWatermark =
+            new AtomicLong(winningTxnRowIdHighWatermark.orElse(prevRowIdHighWatermark));
```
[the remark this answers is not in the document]
qiyuandong-db: `currRowIdHighWatermark` should always be greater than or equal to `prevRowIdHighWatermark`.
I've added a check and comments to explain this here and also in the `updateRowIdHighWatermarkIfNeeded` method. Thanks!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/RowTracking.java
```diff
@@ -101,38 +114,58 @@ public CloseableIterator<Row> iterator() {
...
-    checkArgument(
-        TableFeatures.isRowTrackingSupported(snapshot.getProtocol()),
-        "Row ID high watermark is updated only when feature 'rowTracking' is supported.");
-
+  public static List<DomainMetadata> updateRowIdHighWatermarkIfNeeded(
+      SnapshotImpl snapshot,
+      Optional<Long> winningTxnRowIdHighWatermark,
+      CloseableIterable<Row> dataActions,
+      List<DomainMetadata> domainMetadatas) {
+
```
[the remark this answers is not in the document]
qiyuandong-db: Deleted! Apologies for overlooking this again!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/RowTracking.java
```diff
@@ -101,38 +132,71 @@ public CloseableIterator<Row> iterator() {
...
+   * Inserts or updates the {@link DomainMetadata} action reflecting the new row ID high watermark
+   * when this transaction adds rows and pushed it higher.
    *
-   * <p>This method should be called during commit preparation to prepare the domain metadata
-   * actions for commit. It should be called only when the 'rowTracking' feature is supported.
+   * <p>This method should only be called when the 'rowTracking' feature is supported. Similar to
+   * {@link #assignBaseRowIdAndDefaultRowCommitVersion}, it should be called during the initial row
+   * ID assignment or conflict resolution to reflect the change to the row ID high watermark.
    *
    * @param snapshot the snapshot of the table that this transaction is reading at
```
scottsand-db, line 142: can we improve these variable names? would be great if they were self-describing

e.g. `readSnapshot`
e.g. `txnProtocol`, `txnDataActions`, `txnDomainMetadatas` etc

qiyuandong-db: Sure! I've updated these names to be more self-descriptive in my latest change.

scottsand-db (approved):
LGTM!

PR #4058, opened 2025-01-15
Protocol: add restriction on path appearing at more once in a snapshot list of `add` files

zeevm:
add restriction on path appearing at more once in a snapshot list of `add` files per discussion at [4021](https://github.com/delta-io/delta/discussions/4021)


zeevm:
@larsk-db

PROTOCOL.md
```diff
@@ -488,6 +488,7 @@ That means specifically that for any commit…
  - it is **legal** for the same `path` to occur in an `add` action and a `remove` action, but with two different `dvId`s.
  - it is **legal** for the same `path` to be added and/or removed and also occur in a `cdc` action.
  - it is **illegal** for the same `path` to be occur twice with different `dvId`s within each set of `add` or `remove` actions.
+ - it is **illegal** for the same `path` to occur more than once in a snapshot list of `add` actions.
```
larsk-db, line 491:
```suggestion
 - it is **illegal** for a `path` to occur in an `add` action that already occurs with a different `dvId` in the list of `add` actions from the snapshot of the version immediately preceeding the commit, unless the commit also contains a remove for the later combination.
 - it is **legal** to commit an existing `path` and `dvId` combination again (this allows metadata updates).
```

larsk-db (commented):
@bart-samwel @ryan-johnson-databricks @tdas Could you take a look. Having conditions across commit and snapshot is kind of tricky.

PROTOCOL.md
```diff
@@ -488,6 +488,8 @@ That means specifically that for any commit…
  - it is **legal** for the same `path` to occur in an `add` action and a `remove` action, but with two different `dvId`s.
```
scovich, line 488: BTW, this requirement from L485 above seems incorrect?
> for simplicity it is required that there is at most one file action of the same type for any `path` (regardless of `dvId`).

Wouldn't a "normal" remove/add pair that updates the `dvId` violate that requirement?

The requirement might make sense for action reconciliation... but this is the spec for commit file content. Also, I believe CDC actions are also considered file actions, and it's legal to have an add and a cdc for the same path?

PROTOCOL.md
```diff
@@ -826,7 +828,7 @@ A given snapshot of the table can be computed by replaying the events committed
...
+ - A collection of `remove` actions with unique `(path, deletionVector.uniqueId)` keys. The intersection of the primary keys in the `add` collection and `remove` collection must be empty. That means a logical file cannot exist in both the `remove` and `add` collections at the same time; however, the same *data file* can exist with *different* DVs in the `remove` collection, as logically they represent different content. The `remove` actions act as _tombstones_, and only exist for the benefit of the VACUUM command. Snapshot reads only return `add` actions on the read path. In any given snapshot, it is illegal for the same path to occur more than once in `add` actions.
```
scovich, line 831: Seems like it would be simpler/cleaner to just update the "collection of `add` actions" bullet?
> - A collection of `add` actions with unique `path` keys, corresponding to the newest `(path, deletionVector.uniqueId)` pair encountered for each path.

zeevm: sounds good

zeevm:
Is the gate process broken? why should anything fail in checks if only a markdown file is changed?

zeevm:
@larsk-db  @scovich 
What's the plan going forward with this change?

bart-samwel (approved):
I think this is OK. FWIW, in practice it will be very expensive to sanity check this correctness in actual clients. They will have to guarantee this "by construction", i.e., you can add a file that you just created yourself with a random UUID in the name and you can be sure that it won't be a duplicate, and you can add a file with a DV if you remove the same file with a different DV. And I think it's OK to rely on logic to get this right. I think the main point of the change is that _clients are allowed to assume that this holds_.

zeevm:
>  I think the main point of the change is that _clients are allowed to assume that this holds_.

 Exactly the point, thanks.

PROTOCOL.md
scovich: Right 🤦 -- reading comprehension fail on my part, sorry!

PROTOCOL.md
```diff
@@ -827,8 +827,8 @@ A given snapshot of the table can be computed by replaying the events committed
  - A single `metaData` action
  - A collection of `txn` actions with unique `appId`s
  - A collection of `domainMetadata` actions with unique `domain`s.
- - A collection of `add` actions with unique `(path, deletionVector.uniqueId)` keys.
- - A collection of `remove` actions with unique `(path, deletionVector.uniqueId)` keys. The intersection of the primary keys in the `add` collection and `remove` collection must be empty. That means a logical file cannot exist in both the `remove` and `add` collections at the same time; however, the same *data file* can exist with *different* DVs in the `remove` collection, as logically they represent different content. The `remove` actions act as _tombstones_, and only exist for the benefit of the VACUUM command. Snapshot reads only return `add` actions on the read path. In any given snapshot, it is illegal for the same path to occur more than once in `add` actions.
+ - A collection of `add` actions with unique path keys, corresponding to the newest (path, deletionVector.uniqueId) pair encountered for each path.
```
scovich, line 830: I just realized that the second half of this sentence I suggested is redundant with L839 below. Maybe it's still helpful here, to distinguish clearly between the path-only part of the rules vs. normal path+DV pairs?

scovich (approved):

PR #4062, opened 2025-01-16
[Spark] Create DV tombstones when dropping Deletion Vectors #117451

andreaschat-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Drop feature with checkpoint protection allows dropped features to live in the table history. This has consequences in the vacuum command. This is because vacuum only reads the latest snapshot, and uses that to decide what can be deleted or not. This could lead clients that do not support past feature(s) to incorrectly vacuum auxiliary files, unless the latest snapshot can prevent this from happening using only features from the reduced protocol. The way that the table snapshot prevents deletion of files is by referencing files that were recently removed from the table (in the last 7 days) in RemoveFile actions or "tombstones".

This is problematic when a newer feature extends the functionality of these tombstones. For instance, the Deletion Vectors feature has `RemoveFile` actions that contain a reference both to the main file as well as the auxiliary Deletion Vector file. But after the downgrade, clients will not see that auxiliary DV file reference anymore, and will not preserve Deletion Vectors that were only removed from the table within the past 7 days.

This PR resolves this issue by ensuring that the auxiliary files are referenced by adding RemoveFile actions during the pre-downgrade process of Deletion Vectors.

## How was this patch tested?

Added new tests in `DeltaFastDropFeatureSuite`.

No.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala
@@ -470,7 +470,7 @@ trait DeltaSQLConfBase {
       .createWithDefaultFunction(() => SQLConf.get.getConf(DeltaSQLConf.FAST_DROP_FEATURE_ENABLED))
 
   val FAST_DROP_FEATURE_DV_TOMBSTONE_COUNT_THRESHOLD =
-    buildConf("tableFeatures.dev.fastDropFeature.DVTombstoneCountThreshold")
+    buildConf("tableFeatures.dev.fastDropFeature.dvTombstoneCountThreshold")
       .doc(
         """The maximum number of DV tombstones we are allowed store to memory when dropping
           |deletion vectors. When the resulting number of DV tombstones is higher, we use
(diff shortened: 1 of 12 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala
```diff
@@ -441,6 +441,36 @@ trait DeltaSQLConfBase {
...
+  val FAST_DROP_FEATURE_DV_TOMBSTONE_COUNT_THRESHOLD =
+    buildConf("tableFeatures.dev.fastDropFeature.DVTombstoneCountThreshold")
+      .doc(
+        """The maximum number of DV tombstones we are allowed store to memory when dropping
+          |deletion vectors. When the resulting number of DV tombstones is higher, we use
+          |a special commit for large outputs. This does not materialize results to memory
+          |but does not retry in case of a conflict.""".stripMargin)
+      .intConf
+      .checkValue(_ >= 0, "DVTombstoneCountThreshold must not be negative.")
+      .createWithDefault(10000)
```
larsk-db, line 472: This should be internal, no? And start with dv instead of DV?

andreaschat-db: If someone wants to experiment with that is fine I guess?

larsk-db: Fair enough.

andreaschat-db: Renamed to `dvTombstoneCountThreshold`.

PR #4064, opened 2025-01-16
[SPARK] Properly handle streams in `normalizeColumnNames`

stefankandic:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
`normalizeColumnNames` was always doing a select on top of its data frame which would fail for data frames with incremental execution (streams) which lead to its callers (`normalizeData`) to hack around this by not returning the normalized data frame and instead rely on propagating the fixed names through output attributes.

Fix for this is trivial as there is already a method which does select on top of data frames with incremental execution.

## How was this patch tested?
Since we are not changing the behavior this change should be covered by the existing tests.

No.


johanl-db (approved):
That makes me think we should actually go and look around to see if there are other cases where we should use `DeltaStreamUtils.selectFromStreamingDataFrame`

spark/src/main/scala/org/apache/spark/sql/delta/schema/SchemaUtils.scala
```diff
@@ -362,7 +363,12 @@ def normalizeColumnNamesInDataType(
         }
         expression
       }
-      data.select(aliasExpressions: _*)
+      data.queryExecution match {
+        case incrementalExecution: IncrementalExecution =>
```
zsxwing, line 367: Thanks for fixing this! Could you also add a streaming test to verify the change?

cc @HeartSaVioR to review as well. He found some limitations about `selectFromStreamingDataFrame` in the past. I don't know whether these are fixed or not.

PR #4065, opened 2025-01-16
[VARIANT] Upgrade variantType-preview to variantType

richardc-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

upgrades the variant type preview feature to GA by adding a new table feature VariantType

Note there were no changes to the spec between preview and GA, so we will continue allow reading -preview features.

New tables going forward with variants are created with the `variantType` stable table feature.

A new config `"variant.forceUsePreviewTableFeature"` is also added to revert to the old behavior (create variant tables with `variantType-preview`) if need be

## How was this patch tested?

added UTs.


johanl-db (approved):
LGTM

PR #4068, opened 2025-01-17
[Kernel][Metrics][PR#7] Support ScanReport to log metrics for a Scan operation

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Adds `ScanReport` for reporting a Scan.

We record `ScanReport` either after all the scan files have successfully been consumed (and the iterator closed), or if an exception is thrown while reading/filtering/preparing the scan files to be returned to the connector. This is done within the `hasNext` and `next` methods on the returned iterator since that is when we do all of the kernel work/eval (since the iterator is lazily loaded). We only record a report for failures that happen within Kernel, if there are failures from within the connector code, no report will be emitted.

We also add support for serializing `ScanReport` in this PR.

## How was this patch tested?

Adds unit tests.

No.


kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ActiveAddFilesLogReplayMetricsSuite.scala
allisonport-db, line 1: These tests are all moved to `ScanReportSuite`

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/metrics/ScanReportSuite.scala
```diff
@@ -0,0 +1,450 @@
...
+        expectedNumAddFiles = 0,
+        expectedNumAddFilesFromDeltaFiles = 0,
+        expectedNumActiveAddFiles = 0,
+      )
+    }
+  }
+
+  ///////////////////////////////
+  // Log replay metrics tests ///
+  ///////////////////////////////
```
allisonport-db, line 326: The below tests were all copied from `ActiveAddFilesLogReplayMetricsSuite`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -309,4 +346,79 @@ private CloseableIterator<FilteredColumnarBatch> applyDataSkipping(
...
+        try {
+          if (!scanIter.hasNext()) {
+            // The entire scan file iterator has been successfully consumed, report successful Scan
+            reporter.reportSuccess();
+          } else {
+            // TODO what should we do if the iterator is not fully consumed?
+            //  - Do nothing (we don't know why the iterator is being closed early or what is going
+            //    on in the connector)
+            //  - Report a failure report with some sort of placeholder exception
+            //  - Something else?
```
allisonport-db, line 379: TODO - decide on this

For example, this is possible if there is an error somewhere within the connector code; they close the iterator early but Kernel will never see an exception. However, it's possible this happens for other reasons, we have no way of knowing why the connector is closing the iterator before consuming all the actions.

allisonport-db: Additional option would be reporting whether a Scan is "complete" or not

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -118,30 +129,56 @@ public CloseableIterator<FilteredColumnarBatch> getScanFiles(
...
-    // Apply partition pruning
-    scanFileIter = applyPartitionPruning(engine, scanFileIter);
-
-    // Apply data skipping
-    if (hasDataSkippingFilter) {
-      // there was a usable data skipping filter --> apply data skipping
-      scanFileIter = applyDataSkipping(engine, scanFileIter, dataSkippingFilter.get());
-    }
+    Timer.Timed planningDuration = scanMetrics.totalPlanningTimer.start();
+    ScanReportReporter reportReporter =
```
scottsand-db, line 133: so this is like a ... context wrapper? you can invoke the reporter with .reportError or .reportSuccess, and it goes and executes this code, like stopping the planning duration and pushing the scanReport to all the engine's metrics reporters?

makes sense!

maybe a comment here that like we are just building this context wrapper? ... it itself isn't actually invoking anything yet. (just cuz it took me > 30 seconds to look at this and figure out what was going on)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -309,4 +346,79 @@ private CloseableIterator<FilteredColumnarBatch> applyDataSkipping(
               filteredScanFileBatch.getData(), Optional.of(newSelectionVector));
         });
   }
+
+  /**
+   * Wraps a scan file iterator such that we emit {@link ScanReport} to the engine upon success and
+   * failure. Since most of our scan building code is lazily executed (since it occurs as
+   * maps/filters over an iterator) potential errors don't occur just within `getScanFile`s
+   * execution, but rather may occur as the returned iterator is consumed. Similarly, we cannot
```
scottsand-db, line 354: so the fact that exceptions may come from within `getScanFile` and from external (i.e. as it is consumed) is why we are wrapping the iteration, catching exceptions, reporting them to reporters, and _then_ throwing them, right?

I wonder if you could make this a bit more clear? about the _why_ we are doing this, as opposed to _what_ this method is doing?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -309,4 +346,79 @@ private CloseableIterator<FilteredColumnarBatch> applyDataSkipping(
...
+    return new CloseableIterator<FilteredColumnarBatch>() {
+
+      @Override
+      public void close() throws IOException {
+        try {
+          if (!scanIter.hasNext()) {
+            // The entire scan file iterator has been successfully consumed, report successful Scan
+            reporter.reportSuccess();
+          } else {
+            // TODO what should we do if the iterator is not fully consumed?
```
scottsand-db, line 375: Will we ever be invoked/informed if the iterator is not fully consumed? I don't think there's anything we _can_ even do, right?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
scottsand-db: but when would we ever be able to invoke
```java
// okay, now I know for sure that the iterator was NOT fully consumed
scanReport = ... something ... maybe that says it wasn't fully consumed ...
engine.getMetricsReporters().forEach(reporter -> reporter.report(scanReport));
```

I don't think we can?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/MetricsReportSerializers.java
```diff
@@ -59,7 +71,14 @@ public static String serializeTransactionReport(TransactionReport transactionRep
       new ObjectMapper()
           .registerModule(new Jdk8Module()) // To support Optional
           .registerModule( // Serialize Exception using toString()
-              new SimpleModule().addSerializer(Exception.class, new ToStringSerializer()));
+              new SimpleModule().addSerializer(Exception.class, new ToStringSerializer()))
+          .registerModule( // Serialize StructType using toString
+              // TODO Decide how we want to serialize StructType
+              // We can serialize as a JSON string (more verbose) or using toString (less verbose)
+              // Note - JSON string isn't much more parsable? no fixed schema
```
scottsand-db, line 78: Great question. I think .toString is fine? Maybe ask @nicklan -- any quick decision here is fine with me.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/metrics/ScanReportSuite.scala
```diff
@@ -0,0 +1,450 @@
...
+class ScanReportSuite extends AnyFunSuite with MetricsReportTestUtils {
+
+  /**
+   * Creates a [[Scan]] using `getScan` and then requests and consumes the scan files. Uses a custom
+   * engine to collect emitted metrics reports (exactly 1 [[ScanReport]] and 1 [[SnapshotReport]] is
+   * expected). Also times and returns the duration it takes to consume the scan files.
+   *
+   * @param getScan function to generate a [[Scan]] given an engine
+   * @param expectException whether we expect consuming the scan files to throw an exception, which
+   *                        if so, is caught and returend with the other results
```
scottsand-db, line 46: nit: `returned` (typo)

scottsand-db (changes requested):
Looks great! Tests are super solid. Left some comments.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
allisonport-db: When the iterator is closed (inside the close method) but `hasNext = true`. This is for when the iterator is not fully consumed but the connector is correctly closing it.

Nothing we can do if it is not closed.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
allisonport-db: I think answered below, but if the iterator is closed but not fully consumed (i.e. `hasNext = true`)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
allisonport-db: sure will add a comment

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
scottsand-db: great point, I missed this! Thanks!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
allisonport-db: Not exactly. If an exception originates externally (i.e. within connector code) there is nothing we can do/report. We wrap these methods since most of the work in `getScanFiles` is executed lazily (i.e. when you call `hasNext` or `next`). Since these methods are where the bulk of the execution lies this is also where exceptions may originate (for example, IO errors occur when reading files, which we do lazily in log replay as part of the `hasNext` check). When exceptions occur then, lazily, within these methods, `getScanFiles` has already executed.

allisonport-db: I think this addresses the why (at least for the error case) but let me know if there's a way to make it more clear

>  Since most of our scan building code is lazily executed (since it occurs as maps/filters over an iterator) potential errors don't occur just within `getScanFile`s execution, but rather may occur as the returned iterator is consumed.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
scottsand-db: Thanks for explaining! Is it much work to report a success but also indicate that it was not fully consumed? A usecase here could be LIMIT, right? An engine would consume the AddFiles until some LIMIT number of rows, and then close the iterator without fully consuming it. Would be nice to capture + support this.

allisonport-db: I think that's a valid example of a successful case where this might happen. However that's likely also indistinguishable from the case where an error is thrown within the connector, the scan is interrupted/not completed and the iterator is closed. I'm not sure how we want to address these possibilities.

I think maybe this would be this option I called out
> Additional option would be reporting whether a Scan is "complete" or not

We can add a field such as `boolean isFullyConsumed`; how this is interpreted can be left for later (e.g. wrt to the two cases outlined above).

scottsand-db: I think using the boolean is good enough! Connectors are responsible to know whether or not an exception was thrown inside their own connector.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/MetricsReportSerializers.java
nicklan: I think, given that this is most likely to land in a log, let's go with `toString` to not bloat logs more than we need to

allisonport-db: SGTM

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
scottsand-db: I still don't think I get it :) could we have a 5 min sync for you to explain?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -309,4 +349,79 @@ private CloseableIterator<FilteredColumnarBatch> applyDataSkipping(
...
+  /**
+   * Wraps a scan file iterator such that we emit {@link ScanReport} to the engine upon success and
+   * failure. Since most of our scan building code is lazily executed (since it occurs as
+   * maps/filters over an iterator) potential errors don't occur just within `getScanFile`s
+   * execution, but rather may occur as the returned iterator is consumed. Similarly, we cannot
+   * report a successful scan until the iterator has been fully consumed and the log read/filtered
+   * etc. This means we cannot report the successful scan within `getScanFiles` but rather must
+   * report after the iterator has been consumed.
+   *
+   * <p>This method takes a scan file iterator. It wraps the methods `next` and `hasNext` such that
```
allisonport-db, line 362: Mention inner vs outer implementations - this is an outer most wrapper

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
allisonport-db: Updated the docs based on our discussion

kernel/kernel-api/src/main/java/io/delta/kernel/metrics/ScanReport.java
```diff
@@ -0,0 +1,83 @@
...
+  Optional<Predicate> getDataSkippingFilter();
+
+  /**
+   * Whether the scan file iterator had been fully consumed when it was closed. The iterator may be
+   * closed early (before being fully consumed) either due to an exception originating within
+   * connector code or intentionally (such as for a LIMIT query).
+   *
+   * @return whether the scan file iterator had been fully consumed when it was closed
+   */
+  boolean getIsFullyConsumed();
```
scottsand-db, line 74: nit: you can just do `isFullyConsumed`

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/metrics/ScanReportSuite.scala
```diff
@@ -0,0 +1,481 @@
...
+      // No filter - 2 add files one for each partition
+      checkScanReport(
+        path,
+        expectException = false,
+        expectedNumAddFiles = 2,
+        expectedNumAddFilesFromDeltaFiles = 2,
+        expectedNumActiveAddFiles = 2
+      )
+
+      // With partition filter
```
scottsand-db, line 233: are we expecting any of these filters to impact any of the scan metrics? e.g. expectedNumAddFiles expectedNumAddFilesFromDeltaFiles expectedNumActiveAddFiles ?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/metrics/ScanReportSuite.scala
```diff
@@ -0,0 +1,481 @@
...
+        expectedPartitionPredicate = Some(partFilter)
+      )
+    }
+  }
+
+  test("ScanReport error case - error reading the log files") {
+    withTempDir { tempDir =>
+      val path = tempDir.getCanonicalPath
+
+      // We set up table with a giberish json file that falls before a file containing the P&M
```
scottsand-db, line 323: How about `We set up a table with a giberish json file at version 0 and a valid json file at version 1 that contains the P&M. .... ... ...`

scottsand-db (changes requested):
Looks great! 2 minor comments, 1 question, and then LGTM!

kernel/kernel-api/src/main/java/io/delta/kernel/metrics/ScanReport.java
allisonport-db: I can't if we want jackson to automatically serialize it (it looks for getters)

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/metrics/ScanReportSuite.scala
allisonport-db: no we only increment the file counts before filtering; we can add additional metrics for filtering in the future (but this is where it would be really helpful for something like a SV to have an implicit count field or something)

scottsand-db: Okay! It would be nice if you made it easier for the reader to know that we do _not_ expect scans with partition/data filters to have different scan metrics than scans without the filters.

You could do this via a comment or even save `noFilterScanReport = .....` (idk if that method returns a scan report, but tit could) and then do `checkScanReport(..., expectedNumAddFiles = noFilterScanReport.numAddFiles)` etc.

You choose which you think is best

scottsand-db: Will LGTM after this!

scottsand-db (approved):
LGTM

PR #4069, opened 2025-01-17
Fix a correctness issue around referenceless expressions being evaluated as partition filters

brkyvz:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Fixes a data correctness issue, when non-deterministic expressions without any reference columns are used, such as rand() as a filter on a Delta table. These filters were being evaluated as partition filters and getting double evaluated. This caused a filter such as `rand() < 0.5` to filter ~75% of the data (due to being double evaluated) instead of just 50%.

Added a feature flag just in case for old behavior

## How was this patch tested?

Added a unit test and tested the old behavior as well with a feature flag

Filters such as rand() will not be double evaluated anymore

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaTable.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaTable.scala
@@ -309,9 +309,10 @@ object DeltaTableUtils extends PredicateHelper
       partitionColumns: Seq[String],
       spark: SparkSession): Boolean = {
     val nameEquality = spark.sessionState.analyzer.resolver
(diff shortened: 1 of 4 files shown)


foss-contributor:
@brkyvz, thank you for your contribution! This pull request has been inactive for over 18 months and will be closed in one week. To keep it open, please leave a comment on this pull request or contact us in the `#contributors` channel on [Delta Lake Slack](https://go.delta.io/slack). We appreciate your involvement in Delta Lake!

PR #4070, opened 2025-01-17
Verify spark context during documentation link generation

Alexvsalexvsalex:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
If spark context is null when exception is thrown, then we don't need to generate documentation link. That's because it requires non-null context and it will throw another exception otherwise. But it's almost impossible case because a lot of code verifies that context is not null.

## How was this patch tested?
Added new test.

No overlapping exception during generation of documentation link.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaErrors.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaErrors.scala
@@ -87,7 +87,7 @@ trait DocsPath {
       conf: SparkConf,
       relativePath: String,
       skipValidation: Boolean = false): String = {
-    if (conf == null) return "-"
+    require(conf != null)
     if (!skipValidation) assertValidCallingFunction()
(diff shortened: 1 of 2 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/DeltaErrors.scala
```diff
@@ -84,10 +87,19 @@ trait DocsPath {
       conf: SparkConf,
       relativePath: String,
       skipValidation: Boolean = false): String = {
+    if (conf == null) return "-"
```
c27kwan, line 90: Do we really need this? This should be an assertion or a `require` instead. Otherwise, incorrect callers will pass without knowing why. I think it's the caller's responsibility to not pass in a `null` conf.

c27kwan (approved):
lgtm, thanks for fixing

PR #4073, opened 2025-01-18
[Kernel] Implement basic CRC writer 

huan233usc:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This PR adds the utility class for writing CRC files for delta table after commit. Integration will be done in the following PR.

## How was this patch tested?

Added a new test suite

No


kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/ChecksumWriter.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Writers for writing checksum files from a snapshot */
+public class ChecksumWriter {
+
+  private static final Logger log = LoggerFactory.getLogger(ChecksumWriter.class);
+  public static StructType CRC_FILE_SCHEMA =
```
scottsand-db, line 44: Shouldn't this CRC_FILE_SCHEMA be stored in https://github.com/delta-io/delta/pull/4077/files#diff-897131313222a4b4add04e6c677dc523ce53481fb5e784d7058e9d550a12159a `VersionStats`

(although I'd like VersionStats to be renamed)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotHint.java
```diff
@@ -18,17 +18,35 @@
...
 /** Contains summary information of a {@link io.delta.kernel.Snapshot}. */
 public class SnapshotHint {
   private final long version;
   private final Protocol protocol;
   private final Metadata metadata;
+  private final OptionalLong tableSizeBytes;
+  private final OptionalLong numFiles;
 
   public SnapshotHint(long version, Protocol protocol, Metadata metadata) {
     this.version = version;
```
scottsand-db, line 32: it would be better to just call `this(...)` and pass in all the args + the two OptionalLong.empty()

scottsand-db (changes requested):
Looks great! Left some comments.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/ChecksumWriter.java
allisonport-db: you can have a `FULL_SCHEMA` (that we use for writing) and a `READ_SCHEMA` that we use for reading

kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/ChecksumWriter.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+    Path newChecksumPath = FileNames.checksumFile(logPath, postCommitSnapshot.getVersion());
+    try {
+      return wrapEngineExceptionThrowsIO(
+          () -> {
+            engine
+                .getJsonHandler()
+                .writeJsonFileAtomically(
+                    newChecksumPath.toString(),
+                    toCloseableIterator(
```
allisonport-db, line 82: I think you should be able to use `Utils.singletonCloseableIterator` here instead (cleaner since that's exactly what we want)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotHint.java
```diff
@@ -18,17 +18,35 @@
...
+      long version,
+      Protocol protocol,
+      Metadata metadata,
+      OptionalLong tableSizeBytes,
+      OptionalLong numFiles) {
+    this.version = version;
+    this.protocol = protocol;
+    this.metadata = metadata;
+    this.tableSizeBytes = tableSizeBytes;
+    this.numFiles = numFiles;
```
allisonport-db, line 49: Is `metadata` and `protocol` required non-null here? let's add `requireNonNull` to all these non-primitive args

allisonport-db: otherwise we need null checks up in `ChecksumWriter. buildCheckSumRow`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/ChecksumWriter.java
```diff
@@ -0,0 +1,108 @@
...
+    this.logPath = logPath;
+  }
+
+  /**
+   * Writes a checksum file in a best-effort manner from a post commit snapshot, write will only
+   * process if all required fields, including NumFiles and TableSizeBytes, are collected.
+   *
+   * @return true if checksum file is successfully written, false otherwise.
+   */
+  public boolean maybeWriteCheckSum(SnapshotHint postCommitSnapshot, String tnxId) {
```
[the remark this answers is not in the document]
allisonport-db: yeah pretty sure this is optional (I think this is setTxnId?...)

https://github.com/delta-io/delta/blob/master/PROTOCOL.md#version-checksum-file-schema

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/checksum/ChecksumReadWriteSuite.scala
```diff
@@ -0,0 +1,143 @@
...
+    val protocol = createTestProtocol()
+    val metadata = createTestMetadata()
+    val snapshotHint = new SnapshotHint(
+      1,
+      protocol,
+      metadata,
+      OptionalLong.of(100),
+      OptionalLong.of(1))
+    checksumWriter.maybeWriteCheckSum(snapshotHint, "tnx")
+
```
allisonport-db, line 51: validate the schema first

allisonport-db, line 51: validate the checksum file path that it was written to

allisonport-db, line 51: validate that no additional rows were written to the file (only 1 row was written and it has the schema we expect)

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/checksum/ChecksumReadWriteSuite.scala
```diff
@@ -0,0 +1,143 @@
...
+import io.delta.kernel.internal.util.VectorUtils
+import io.delta.kernel.test.{BaseMockJsonHandler, MockEngineUtils}
+import io.delta.kernel.types.{StringType, StructType}
+import io.delta.kernel.utils.CloseableIterator
+import org.scalatest.funsuite.AnyFunSuite
+
+import java.util
+import java.util.{Collections, HashMap, Optional, OptionalLong}
+
+class ChecksumReadWriteSuite extends AnyFunSuite with MockEngineUtils {
```
allisonport-db, line 33: add tests for numFiles or getTableSizeBytes aren't present ==> no checksum file is written

allisonport-db: we also probably want to test that the entire write does not fail if the checksum write fails, but I think we can probably do this in the PR with the e2e integration

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/checksum/ChecksumReadWriteSuite.scala
```diff
@@ -0,0 +1,143 @@
...
+      Optional.of("name"),
+      Optional.of("description"),
+      new Format("parquet", Collections.emptyMap()),
+      "sss",
+      new StructType(),
+      new ArrayValue() { // partitionColumns
+        override def getSize = 1
+
+        override def getElements: ColumnVector = singletonStringColumnVector("c3")
+      },
```
allisonport-db, line 80: You can use `VectorUtils.stringArrayValue` here

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/checksum/ChecksumReadWriteSuite.scala
```diff
@@ -0,0 +1,143 @@
...
+      Optional.of(123),
+      new MapValue() { // conf
+        override def getSize = 1
+
+        override def getKeys: ColumnVector = singletonStringColumnVector("delta.appendOnly")
+
+        override def getValues: ColumnVector =
+          singletonStringColumnVector("true")
+      }
```
allisonport-db, line 89: Same thing here with `VectorUtils.stringStringMapValue`

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/checksum/ChecksumReadWriteSuite.scala
allisonport-db: also tests for when txnId is not present (let's make it optional as in the protocol)

allisonport-db (changes requested):
Left some comments. For the tests (especially in `kernel-api` when we're testing inner utils try to make sure we cover all the cases based on the code added/written).

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/checksum/ChecksumReadWriteSuite.scala
huan233usc: Done for numFiles or getTableSizeBytes aren't present ==> no checksum file is written and when txnId is not present.

for the second one I will make sure to add it in e2e tests.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/ChecksumWriter.java
huan233usc: I create a [ChecksumUtils.java](https://github.com/delta-io/delta/pull/4073/files#diff-2648e3d4c0d32cfaeee8872f126e33dc94927b0fba3463e040de8779ef8d18ee) for holding shared code path like FULL_SCHEMA. In the end we may also want to make read schema also the same as FULL_SCHEMA. I will do a refactor later.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/ChecksumWriter.java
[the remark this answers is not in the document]
huan233usc: Yeah it is optional, changed the signature.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/checksum/ChecksumReadWriteSuite.scala
```diff
@@ -0,0 +1,143 @@
...
+      protocolRow.getInt(Protocol.FULL_SCHEMA.indexOf("minReaderVersion")))
+    assert(protocol.getMinWriterVersion ==
+      protocolRow.getInt(Protocol.FULL_SCHEMA.indexOf("minWriterVersion")))
+  }
+}
+
+class MockCheckSumFileJsonWriter extends BaseMockJsonHandler {
+  var capturedCrcRow: Row = new GenericRow(new StructType(), new util.HashMap[Integer, AnyRef]);
+
+  override def writeJsonFileAtomically(filePath: String,
```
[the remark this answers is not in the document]
huan233usc: fixed scala format plugin locally and re-run

PR #4074, opened 2025-01-18
[Spark] Fix Utils.isTesting calls to use Delta implementation

felipepessoto:
- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Some classes were using the `org.apache.spark.util.Utils` instead of the `org.apache.spark.sql.delta.util.Utils`

The Spark implementation uses the SPARK_TESTING environment variable, while the Delta implementation uses the DELTA_TESTING key.

Spark:
https://github.com/apache/spark/blob/51fb84a54982719209c19136b1d72d2ef44726ee/core/src/main/scala/org/apache/spark/util/Utils.scala#L1878

Delta:
https://github.com/delta-io/delta/blob/221d95cd69fdb9ff3f69cdd842c5c13ed47fd687/spark/src/main/scala/org/apache/spark/sql/delta/util/Utils.scala#L58

It means the unit tests are currently running non test code path, because we only set the DELTA_TESTING:

https://github.com/delta-io/delta/blob/221d95cd69fdb9ff3f69cdd842c5c13ed47fd687/build.sbt#L466

https://github.com/delta-io/delta/blob/221d95cd69fdb9ff3f69cdd842c5c13ed47fd687/run-tests.py#L93

## How was this patch tested?
Unit tests

No


felipepessoto:
@scottsand-db, @allisonport-db could you take a look, please? It is a small fix

ctring:
@felipepessoto I think this looks good and complements https://github.com/delta-io/delta/pull/4119. However, @scottsand-db or @allisonport-db will have the final say on this since I'm not a committer.

felipepessoto:
@vkorukanti the test failures doesn't look related to this change

vkorukanti (approved):

PR #4077, opened 2025-01-21
[Kernel] Load the protocol and metadata from the CRC files when available

huan233usc:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

CP crc loading code to master branch from "kernel-20250115-crc-optimization"
PR created using `git cherry-pick 7d32f66d9efd1dca41ed2a45cf259525cbdd8952`

No


kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ChecksumReader.java
```diff
@@ -0,0 +1,126 @@
...
+        ||
+        // we don't expect any more checksum files as it is the first version
+        readVersion == 0) {
+      return versionStatsOpt;
+    }
+
+    // Try to list the last 100 CRC files and see if we can find a CRC that we can use
+    long lowerBound = Math.max(lowerBoundOpt.orElse(0L) + 1, Math.max(0, readVersion - 100));
+
+    Path listFrom = checksumFile(logPath, lowerBound);
```
scottsand-db, line 67: `lowerBoundFilePath`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ChecksumReader.java
```diff
@@ -0,0 +1,126 @@
...
+   * @param engine the engine to use for reading the checksum file
+   * @param logPath the path to the Delta log
+   * @param readVersion the version to read the checksum file from
+   * @param lowerBoundOpt the lower bound version to search for the checksum file
+   * @return Optional {@link VersionStats} containing the protocol and metadata, and the version of
+   *     the checksum file. If the checksum file is not found, it will return an empty
+   */
+  public static Optional<VersionStats> getVersionStats(
+      Engine engine, Path logPath, long readVersion, Optional<Long> lowerBoundOpt) {
+
```
scottsand-db, line 51: add some better logging please -- i.e. what we are looking for, if we found it, and the lowerbound listfrom, etc.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -208,6 +208,31 @@ protected Tuple2<Protocol, Metadata> loadTableProtocolAndMetadata(
...
 
+    // Compute the lower bound for the CRC search
+    // If the snapshot hint is present, we can use it as the lower bound for the CRC search.
+    // TODO: this can be further improved to make the lower bound until the checkpoint version
+    Optional<Long> crcSearchLowerBound = snapshotHint.map(SnapshotHint::getVersion);
+    Optional<VersionStats> versionStatsOpt =
+        ChecksumReader.getVersionStats(
+            engine, logSegment.logPath, snapshotVersion, crcSearchLowerBound);
+    if (versionStatsOpt.isPresent()) {
+      // We found the protocol and metadata for the version we are looking for
```
scottsand-db, line 219: move this comment to inside the `if` stmt below

scottsand-db, line 219: can we add a check that versionStateOpt.get.version is > (or >=, whatever is corect) than the snapshot hint?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -208,6 +208,31 @@ protected Tuple2<Protocol, Metadata> loadTableProtocolAndMetadata(
...
+      VersionStats versionStats = versionStatsOpt.get();
+      if (versionStats.getVersion() == snapshotVersion) {
+        return new Tuple2<>(versionStats.getProtocol(), versionStats.getMetadata());
+      }
+      // We found the protocol and metadata in a version older than the one we are looking
```
scottsand-db, line 224: Let's make this a bit clearer. Please mention that the info from the CRC is newer than the snapshot hint, that's why we update it

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/LogReplayEngineMetricsSuite.scala
```diff
@@ -342,6 +381,110 @@ class LogReplayEngineMetricsSuite extends QueryTest
       }
     }
   }
+
+  /////////////////////////////////////////////////////////////////////////////////////////////////
+  // Tests for loading P & M  through checksums files                                            //
```
scottsand-db, line 386: nit: extra space after `P & M`

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/LogReplayEngineMetricsSuite.scala
```diff
@@ -342,6 +381,110 @@ class LogReplayEngineMetricsSuite extends QueryTest
...
+  /////////////////////////////////////////////////////////////////////////////////////////////////
+
+  Seq(-1L, 3L, 4L).foreach { version => // -1 means latest version
+    test(s"checksum found at the read version: ${if (version == 1) "latest" else version}") {
```
scottsand-db, line 390: `${if (version == 1` ---> this should be -1, no?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableReadsSuite.scala
```diff
@@ -38,6 +38,15 @@ import scala.collection.JavaConverters._
 
 class DeltaTableReadsSuite extends AnyFunSuite with TestUtils {
 
+  test("CRC metadata & protocol fetching") {
+    val path = getTestResourceFilePath("stream_table_optimize")
```
scottsand-db, line 42: cc @allisonport-db -- does kernel support golden tables? i.e. defining the actual scala/spark code to create the resource table?

For example, how was this table created? What was the code? I can't see that from looking at the resources folder.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/VersionStats.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+package io.delta.kernel.internal.replay;
+
+import io.delta.kernel.data.ColumnarBatch;
+import io.delta.kernel.engine.Engine;
+import io.delta.kernel.internal.actions.Metadata;
+import io.delta.kernel.internal.actions.Protocol;
+import io.delta.kernel.types.StructType;
+
+public class VersionStats {
```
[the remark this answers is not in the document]
huan233usc: Done, renamed VersionStats to CRCInfo.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ChecksumReader.java
```diff
@@ -0,0 +1,126 @@
...
+import org.slf4j.LoggerFactory;
+
+/** Utility method to load protocol and metadata from the Delta log checksum files. */
+public class ChecksumReader {
+  private static final Logger logger = LoggerFactory.getLogger(ChecksumReader.class);
+
+  /**
+   * Load the protocol and metadata from the checksum file at the given version. If the checksum
+   * file is not found at the given version, it will try to find the latest checksum file that is
+   * created after the lower bound version or within the last 100 versions.
```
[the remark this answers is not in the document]
huan233usc: Updated the comments

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ChecksumReader.java
```diff
@@ -0,0 +1,126 @@
...
+    Optional<VersionStats> versionStatsOpt = readChecksumFile(engine, crcFilePath);
+    if (versionStatsOpt.isPresent()
+        ||
+        // we don't expect any more checksum files as it is the first version
+        readVersion == 0) {
+      return versionStatsOpt;
+    }
+
+    // Try to list the last 100 CRC files and see if we can find a CRC that we can use
+    long lowerBound = Math.max(lowerBoundOpt.orElse(0L) + 1, Math.max(0, readVersion - 100));
```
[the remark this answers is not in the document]
huan233usc: @allisonport-db please correct me if I am wrong.

My reading from the code that +1 or not is for the semantic of open or close interval of the lowerBoundOpt. It is just a very slight behavior difference -- whether we should look at (expected - lowerBound).crc. I removed 1 as it brings more confusion as both case (look at (expected - lowerBound).crc or not) seems to be valid to me.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ChecksumReader.java
```diff
@@ -0,0 +1,126 @@
...
+
+  /**
+   * Load the protocol and metadata from the checksum file at the given version. If the checksum
+   * file is not found at the given version, it will try to find the latest checksum file that is
+   * created after the lower bound version or within the last 100 versions.
+   *
+   * @param engine the engine to use for reading the checksum file
+   * @param logPath the path to the Delta log
+   * @param readVersion the version to read the checksum file from
+   * @param lowerBoundOpt the lower bound version to search for the checksum file
```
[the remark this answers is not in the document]
huan233usc: Add a comment to indicate "inclusive", it is also related to the other comment of "+1"

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableReadsSuite.scala
huan233usc: I believe it is the static files in the folder `kernel/kernel-defaults/src/test/resources/stream_table_optimize` I think I could make a follow up PR for switching to golden tables considering many added test cases depends on this added table.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
huan233usc: I wondering if following is the valid case -- for time travel, we want to look at the CRC earlier than the snapshot hint, if snapshot hint points to the latest version.

scottsand-db: That's a great observation! Why don't we do this: in the constructor of LogReplay, let's check if the snapshotHint's version is > our target snapshotVersion. If it is, then set the snapshotHint to Optional.empty

scottsand-db: `<<< LogReplay doesn't own snapshotHint, but just uses it for a fast path of loading P&M >>>`

I agree with this statement. What does this have to do with `<<< let's check if the snapshotHint's version is > our target snapshotVersion. If it is, then set the snapshotHint to Optional.empty >>>` ?

If we set the variable `snapshotHint` to `Optional.empty`, this doesn't impact the original value of snapshotHint from the caller of the method.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableReadsSuite.scala
allisonport-db: Now that Delta 3.3 writes crc files we can either use golden tables, or write the test table in the tests (if it's reused a lot maybe using a golden table is preferable)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ChecksumReader.java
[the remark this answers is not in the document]
allisonport-db: I think it's just a semantic difference let's just be clear as to whether lowerBoundOpt is inclusive or exclusive.

None of this is particularly interesting, I'm sure it was just copied from the Spark code and it's that way because that's how it is in Spark...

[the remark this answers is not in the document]
allisonport-db: _However_ in this PR based on how it is used, it makes sense for it to be _EXCLUSIVE_, since we set the lowerBoundOpt = snapshotHintVersion. There is no reason to read the crc file for the snapshot hint version since we already have the P&M for it.

But alternatively, we can make it exclusive, and call it with `lowerBoundOpt = snapshotHintVersion + 1`.

Between the two of these options I have no preference as long as the documentation is clear as to whether the bound is inclusive or exclusive.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ChecksumReader.java
```diff
@@ -0,0 +1,129 @@
...
+        readVersion,
+        lowerBound);
+
+    Path lowerBoundFilePath = checksumFile(logPath, lowerBound);
+    try (CloseableIterator<FileStatus> crcFiles =
+        engine.getFileSystemClient().listFrom(lowerBoundFilePath.toString())) {
+      List<FileStatus> crcFilesList = new ArrayList<>();
+      crcFiles
+          .filter(file -> isChecksumFile(new Path(file.getPath())))
+          .forEachRemaining(crcFilesList::add);
```
allisonport-db, line 77: @scottsand-db I wonder if this can be refactored to use any of the log listing methods in your PR?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/CRCInfo.java
```diff
@@ -0,0 +1,65 @@
...
+import io.delta.kernel.engine.Engine;
+import io.delta.kernel.internal.actions.Metadata;
+import io.delta.kernel.internal.actions.Protocol;
+import io.delta.kernel.types.StructType;
+
+public class CRCInfo {
+
+  public static CRCInfo fromColumnarBatch(
+      Engine engine, long version, ColumnarBatch batch, int rowId) {
+    // fromColumnVector already takes care of nulls
```
scottsand-db, line 28: metadata and protocol are required in the CRC. https://github.com/delta-io/delta/blob/master/PROTOCOL.md#version-checksum-file-schema

if they aren't there, then we should not yield a CRC info. futher, let's not just propagate nulls around our code base everwhere. that's what Optional is for in java.

why don't we have this `CRCInfo::fromColumnarBatch` return an Optional then, since it needs to do some complex poking around in the underlying vectors.

`fromColumnVector already takes care of nulls` and this is tech debt that I don't like at all :(

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/CRCInfo.java
```diff
@@ -0,0 +1,65 @@
...
+
+  private static final int PROTOCOL_ORDINAL = 0;
+  private static final int METADATA_ORDINAL = 1;
+
+  private final long version;
+  private final Metadata metadata;
+  private final Protocol protocol;
+
+  protected CRCInfo(long version, Metadata metadata, Protocol protocol) {
+    this.version = version;
```
scottsand-db, line 46: let's add a`java.util.Objects.requireNonNull`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -203,11 +203,34 @@ public CloseableIterator<FilteredColumnarBatch> getAddFilesAsColumnarBatches(
   protected Tuple2<Protocol, Metadata> loadTableProtocolAndMetadata(
       Engine engine, Optional<SnapshotHint> snapshotHint, long snapshotVersion) {
 
-    // Exit early if the hint already has the info we need
+    // If the snapshot hint is provided and matches the expected version, load the P&M from version hint.
+    // If not, e.g. for the case of time travel, skip use version hint and fallback to CRC file or log replay.
```
scottsand-db, line 207: Err, I don't want this to imply that time travel is the _only_ case where the snapshot hint isn't used. Also, we aren't really "skipping " the version hint, we are just performing log replay, which _still_ may end up using the version hint.

Let's just revert this back to `// Exit early if the hint already has the info we need`.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -203,11 +203,34 @@ public CloseableIterator<FilteredColumnarBatch> getAddFilesAsColumnarBatches(
...
     if (snapshotHint.isPresent() && snapshotHint.get().getVersion() == snapshotVersion) {
       return new Tuple2<>(snapshotHint.get().getProtocol(), snapshotHint.get().getMetadata());
     }
 
+    // Compute the lower bound for the CRC search
+    // If the snapshot hint is present, we can use it as the lower bound for the CRC search.
+    // TODO: this can be further improved to make the lower bound until the checkpoint version
+    Optional<Long> crcSearchLowerBound = snapshotHint.map(SnapshotHint::getVersion);
```
scottsand-db, line 215: This doesn't seem right to me. If the target version is 13 and the snapshot hint is at version 16, then we are setting the crcSearchLowerBound to 16?

What was wrong with my suggestion of checking in the constructor of LogReplay if the snapshotHint is greater than the target version, and just setting the snapshotHint to empty?

allisonport-db, line 217: As mentioned in other convo, if we are making the bound inclusive, this should be snapshotHintVersion + 1

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -203,11 +203,34 @@ public CloseableIterator<FilteredColumnarBatch> getAddFilesAsColumnarBatches(
...
+      if (crcInfo.getVersion() == snapshotVersion) {
+        // CRC is related to the desired snapshot version. Load protocol and metadata from CRC.
+        return new Tuple2<>(crcInfo.getProtocol(), crcInfo.getMetadata());
+      }
+      // We found the protocol and metadata in a version older than the one we are looking
+      // for. We need to replay the actions to get the latest protocol and metadata, but
+      // update the hint to read the actions from the version we found to check if the
+      // protocol and metadata are updated in the versions after the one we found.
+      // Building a new snapshotHit for determining the time to end the log replay loop.
+      snapshotHint =
```
scottsand-db, line 229: what if, somehow, the crcInfo version is greater than the target version? we shouldn't set the hint then, since that would be useless. I think we should check for this and log a warning if somehow this happened.

scottsand-db (changes requested):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -203,11 +203,34 @@ public CloseableIterator<FilteredColumnarBatch> getAddFilesAsColumnarBatches(
...
+          Optional.of(
+              new SnapshotHint(crcInfo.getVersion(), crcInfo.getProtocol(), crcInfo.getMetadata()));
```
allisonport-db, line 233: nit: maybe `SnapshotHint.fromCRCInfo`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/FileNames.java
```diff
@@ -150,6 +161,10 @@ public static boolean isCommitFile(String fileName) {
         || UUID_DELTA_FILE_REGEX.matcher(filename).matches();
   }
 
+  public static boolean isChecksumFile(Path checksumFilePath) {
```
allisonport-db, line 164: It seems like all our other `isXXFile` methods take in a `String fileName` can we standardize on that? Unless there's a specific reason not to then feel free to object

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableReadsSuite.scala
```diff
@@ -38,6 +39,20 @@ import scala.collection.JavaConverters._
...
 
+  test("CRC metadata & protocol fetching") {
+    withTempDir { dir =>
+      val path = dir.getCanonicalPath + "/table"
+      spark.conf.set(DeltaSQLConf.DELTA_WRITE_CHECKSUM_ENABLED.key, true)
+      spark.sql(s"CREATE TABLE delta.`$path` USING DELTA AS " +
+      s"SELECT 0 as value")
+
+      val snapshot = Table.forPath(defaultEngine, path)
+      .getSnapshotAsOfVersion(defaultEngine, 0)
```
allisonport-db, line 50: nit indention

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableReadsSuite.scala
```diff
@@ -38,6 +39,20 @@ import scala.collection.JavaConverters._
 
 class DeltaTableReadsSuite extends AnyFunSuite with TestUtils {
 
+  test("CRC metadata & protocol fetching") {
+    withTempDir { dir =>
+      val path = dir.getCanonicalPath + "/table"
+      spark.conf.set(DeltaSQLConf.DELTA_WRITE_CHECKSUM_ENABLED.key, true)
```
allisonport-db, line 45: you can use `withSQLConf` for tests instead of setting it (this takes care of unsetting any confs after)

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableReadsSuite.scala
```diff
@@ -38,6 +39,20 @@ import scala.collection.JavaConverters._
 
 class DeltaTableReadsSuite extends AnyFunSuite with TestUtils {
 
+  test("CRC metadata & protocol fetching") {
+    withTempDir { dir =>
+      val path = dir.getCanonicalPath + "/table"
```
allisonport-db, line 44: you can just use the directory path for the table, we don't need to add a "/table"

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
allisonport-db: @scottsand-db I feel like this brings up for question how we should assign `snapshotHint` (and what kind of query patterns we might expect). As the snapshotHint for a time-travel query might be very different from the one that is useful for the latest snapshot query. So when should we be updating it and to what?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/CRCInfo.java
allisonport-db: > why don't we have this CRCInfo::fromColumnarBatch return an Optional then, since it needs to do some complex poking around in the underlying vectors.

+1 this would be an improvement (and we don't need to do the check for their presence elsewhere in the code)

allisonport-db: But let's still log if we find a crc file that's missing one of the two!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
scottsand-db: From the snapshot manager's perspective, `snapshotHint` is the latest hint for a snapshot for the table, where latest means it has the largest version.

For the LogReplay, I think it is smart to keep the current replay-the-log-until-the-hint code, and to intelligently set the hint to be a value from a CRC, if for example (a) we could not find the CRC for our target version and (b) the CRC we did find is newer (greater) than the hint (and, of course, older (less) than the target version)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
scottsand-db: I'm okay with how it is for now. Im starting to see lots of coupling between SnapshotHint and CRCInfo. Let's wait to introduce static helpers like that, and perhaps instead re-evaluate the relationship between these two classes (in a later PR)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ChecksumReader.java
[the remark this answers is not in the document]
huan233usc: I keep the +1 (impl for the private branch) and update the comment to say it is exclusive

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableReadsSuite.scala
huan233usc: Update the tests using spark to generate tables

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/CRCInfo.java
huan233usc: Done

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
huan233usc: Done

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/CRCInfo.java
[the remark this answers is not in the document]
huan233usc: Rewritten the comment and move it above the newly introduced null checks.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/CRCInfo.java
```diff
@@ -0,0 +1,65 @@
...
+    this.metadata = metadata;
+    this.protocol = protocol;
+  }
```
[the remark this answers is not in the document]
huan233usc: Added a null check as both are required per CRC's spec.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ChecksumReader.java
```diff
@@ -0,0 +1,129 @@
...
+    Optional<CRCInfo> crcInfoOpt = readChecksumFile(engine, crcFilePath);
+    if (crcInfoOpt.isPresent()
+        ||
+        // we don't expect any more checksum files as it is the first version
+        readVersion == 0) {
+      return crcInfoOpt;
+    }
+
+    // Try to list the last 100 CRC files and see if we can find a CRC that we can use
+    long lowerBound = Math.max(lowerBoundOpt.orElse(0L), Math.max(0, readVersion - 100));
```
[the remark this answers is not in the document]
huan233usc: Added +1 back per the preference for exclusive bound

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ChecksumReader.java
```diff
@@ -0,0 +1,129 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(ChecksumReader.class);
+
+  /**
+   * Load the protocol and metadata from the checksum file at the given version. If the checksum
+   * file is not found at the given version, it will try to find the latest checksum file that is
+   * created after the lower bound version or within the last 100 versions.
+   *
+   * @param engine the engine to use for reading the checksum file
+   * @param logPath the path to the Delta log
+   * @param readVersion the version to read the checksum file from
```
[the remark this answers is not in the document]
huan233usc: Updated with "targetedVersion"

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/CRCInfo.java
```diff
@@ -0,0 +1,75 @@
...
+  public long getVersion() {
+    return version;
+  }
+
+  /** The {@link Metadata} stored in this VersionStats. May be null. */
+  public Metadata getMetadata() {
+    return metadata;
+  }
+
+  /** The {@link Protocol} stored in this VersionStats. May be null. */
```
scottsand-db, line 71: Can no longer be null, btw.

scottsand-db, line 71: `VersionStats` is out of date class name, too

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ChecksumReader.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Utility method to load protocol and metadata from the Delta log checksum files. */
+public class ChecksumReader {
+  private static final Logger logger = LoggerFactory.getLogger(ChecksumReader.class);
+
+  /**
+   * Load the protocol and metadata from the checksum file at the given version. If the checksum
```
scottsand-db, line 38: the information is returned from this method will probably change over time, right? Might be more future-proof to have this comment be `Load the CRCInfo from the ...`.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/CRCInfo.java
```diff
@@ -0,0 +1,75 @@
...
+  private final Metadata metadata;
+  private final Protocol protocol;
+
+  protected CRCInfo(long version, Metadata metadata, Protocol protocol) {
+    this.version = version;
+    this.metadata = requireNonNull(metadata);
+    this.protocol = requireNonNull(protocol);
+  }
+
+  /** The version of the Delta table that this VersionStats represents. */
```
scottsand-db, line 61: Sorry, still referencing VersionStats

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -203,11 +203,38 @@ public CloseableIterator<FilteredColumnarBatch> getAddFilesAsColumnarBatches(
...
     }
 
+    // Snapshot hit is not use-able in this case for determine the lower bound.
+    if (snapshotHint.isPresent() && snapshotHint.get().getVersion() > snapshotVersion) {
+      snapshotHint = Optional.empty();
+    }
+
+    // Compute the lower bound for the CRC search
+    // If the snapshot hint is present, we can use it as the lower bound for the CRC search.
+    // TODO: this can be further improved to make the lower bound until the checkpoint version
```
scottsand-db, line 218: why not fix this TODO now? `logSegment.checkpointVersionOpt` should resolve this, no?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
scottsand-db: @huan233usc -- sorry, my comment from above still remains. << what if, somehow, the crcInfo version is greater than the target version? we shouldn't set the hint then, since that would be useless. I think we should check for this and log a warning if somehow this happened. >>

let's add a `io.delta.kernel.internal.util.Preconditions.checkArgument` call that crcInfoOpt, if present, has a version >= crcSearchLowerBound && version <= target`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -203,11 +203,38 @@ public CloseableIterator<FilteredColumnarBatch> getAddFilesAsColumnarBatches(
...
+    Optional<Long> crcSearchLowerBound = snapshotHint.map(SnapshotHint::getVersion);
+    Optional<CRCInfo> crcInfoOpt =
+        ChecksumReader.getCRCInfo(engine, logSegment.logPath, snapshotVersion, crcSearchLowerBound);
+    if (crcInfoOpt.isPresent()) {
+      CRCInfo crcInfo = crcInfoOpt.get();
+      if (crcInfo.getVersion() == snapshotVersion) {
+        // CRC is related to the desired snapshot version. Load protocol and metadata from CRC.
+        return new Tuple2<>(crcInfo.getProtocol(), crcInfo.getMetadata());
+      }
+      // We found the protocol and metadata in a version older than the one we are looking
```
scottsand-db, line 228: IMO this comment is getting a bit too long and it could be shorter and simpler.

```
// We found a CRCInfo of a version (a) older than the one we are looking for (snapshotVersion)
// but (b) newer than the current hint. Use this CRCInfo to create a new hint.
```

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableReadsSuite.scala
```diff
@@ -38,6 +39,21 @@ import scala.collection.JavaConverters._
...
+      val path = dir.getCanonicalPath
+      withSQLConf(DeltaSQLConf.DELTA_WRITE_CHECKSUM_ENABLED.key -> "true") {
+        spark.sql(s"CREATE TABLE delta.`$path` USING DELTA AS " +
+        s"SELECT 0 as value")
+      }
+
+      val snapshot = Table.forPath(defaultEngine, path)
+        .getSnapshotAsOfVersion(defaultEngine, 0)
+
+      readSnapshot(snapshot)
```
scottsand-db, line 53: this doesn't really prove that we got the value from the CRC file. if the CRC file was not written, we would still have been able to load version 0. are there any metrics we could collect to help indicate this? perhaps sync with @allisonport-db

scottsand-db (changes requested):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
huan233usc: Updated using max(use-ablesnapshot-hint, checkpoint, snapshotversion - 100(101 actually for exclusive bound))

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
huan233usc: Added a check.  Initially I believe that ChecksumReader.getCRCInfo may help enforce version >= crcSearchLowerBound && version <= target. you made a good point that we should check that before setting the snapshot hint as things may go wrong for arbitrary reason.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableReadsSuite.scala
allisonport-db: I honestly don't know if we need this test -- I think this is tested in `LogReplayEngineMetricsSuite`

allisonport-db: A possible follow-up task is to add crc-specific info/metrics to `SnapshotReport`

scottsand-db: Sounds good. Let's remove this test for now. @huan233usc can you please track this as an item in your project plan to add metrics for CRC reading so we can properly test this aspect later?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ChecksumReader.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Utility method to load protocol and metadata from the Delta log checksum files. */
+public class ChecksumReader {
+  private static final Logger logger = LoggerFactory.getLogger(ChecksumReader.class);
+
+  /**
+   * Load the CRCInfo from the from the checksum file at the given version. If the checksum file is
```
scottsand-db, line 38: nit: ` from the from the`

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableReadsSuite.scala
huan233usc: SG and revert this test for now and will add metric and tests back seperately

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ChecksumReader.java
```diff
@@ -0,0 +1,119 @@
...
+   * not found at the given version, it will try to find the latest checksum file that is created
+   * after the lower bound version or within the last 100 versions.
+   *
+   * @param engine the engine to use for reading the checksum file
+   * @param logPath the path to the Delta log
+   * @param targetedVersion the target version to read the checksum file from
+   * @param lowerBound the exclusive lower bound version to search for the checksum file
```
scottsand-db, line 45: what's the expected behaviour if lowerBound == targetedVersion?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ChecksumReader.java
```diff
@@ -0,0 +1,119 @@
...
+   * @return Optional {@link CRCInfo} containing the protocol and metadata, and the version of the
+   *     checksum file. If the checksum file is not found, it will return an empty
+   */
+  public static Optional<CRCInfo> getCRCInfo(
+      Engine engine, Path logPath, long targetedVersion, long lowerBound) {
+    logger.info("Loading CRC file for version {}", targetedVersion);
```
scottsand-db, line 51: please log the logPath and lowerBound

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ChecksumReader.java
```diff
@@ -0,0 +1,119 @@
...
+
+    Path lowerBoundFilePath = checksumFile(logPath, lowerBound + 1);
+    try (CloseableIterator<FileStatus> crcFiles =
+        engine.getFileSystemClient().listFrom(lowerBoundFilePath.toString())) {
+      List<FileStatus> crcFilesList = new ArrayList<>();
+      crcFiles.filter(file -> isChecksumFile(file.getPath())).forEachRemaining(crcFilesList::add);
+
+      // pick the last file which is the latest version that has the CRC file
+      if (crcFilesList.isEmpty()) {
+        logger.warn("No checksum files found in the range {} to {}", lowerBound, targetedVersion);
```
scottsand-db, line 76: printing the lower bound here will be misleading since this is the exclusive lowerBound.

what was the reasoning behind making lowerbound exclusive? was it worth the tiny little mistakes like this I wonder?

scottsand-db (approved):
LGTM!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ChecksumReader.java
huan233usc: This is a good catch, the past implementation actually had two bug in this case.

(1)16.checkpoint is there but 16.crc is missing, but 17.crc is there. The read will just fail.
(2)15.checkpoint, 15.crc is there but 16.crc is missing. reading 15.crc will be skipped.

I made the following changes -  

-make lowerBound inclusive for getCRCInfo and move the "+1" outside so that the behavior keeps consistent for using old snapshot hint, usable snapshot hint is always preferred with no IO.
-add a check for targetedVersion >= lowerBound and early returns if targetedVersion == lowerBound.
-update the caller in LogReplay to ensure targetedVersion >= lowerBound by add a min.

So in this case,
for 1 the lower bound will be 16, 16.crc is missing so no crc file will be read, p&m will be loaded from 16.checkpoint
for 2 the 15.crc will be read.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ChecksumReader.java
huan233usc: I updated to make in inclusive, pervious exclusive implementation has two issues mentioned above when checkpoint came into picture.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ChecksumReader.java
```diff
@@ -0,0 +1,122 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Utility method to load protocol and metadata from the Delta log checksum files. */
+public class ChecksumReader {
+  private static final Logger logger = LoggerFactory.getLogger(ChecksumReader.class);
+
+  /**
+   * Load the CRCInfo from the checksum file at the given version. If the checksum file is not found
+   * at the given version, it will try to find the latest checksum file that is created after the
```
allisonport-db, line 40: nit: "created at or after"

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ChecksumReader.java
```diff
@@ -0,0 +1,122 @@
...
+    logger.info(
+        "CRC file for version {} not found, attempt to loading version up to {}",
+        targetedVersion,
+        lowerBound);
+
+    Path lowerBoundFilePath = checksumFile(logPath, lowerBound);
+    try (CloseableIterator<FileStatus> crcFiles =
+        engine.getFileSystemClient().listFrom(lowerBoundFilePath.toString())) {
+      List<FileStatus> crcFilesList = new ArrayList<>();
+      crcFiles.filter(file -> isChecksumFile(file.getPath())).forEachRemaining(crcFilesList::add);
```
allisonport-db, line 75: Do you want to maybe create a method in `InternalUtils` to turn an iterator to a list? I think we do this pattern elsewhere (no need to change those instances now) but we can probably reuse this

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ChecksumReader.java
```diff
@@ -0,0 +1,122 @@
...
+
+      // pick the last file which is the latest version that has the CRC file
+      if (crcFilesList.isEmpty()) {
+        logger.warn("No checksum files found in the range {} to {}", lowerBound, targetedVersion);
+        return Optional.empty();
+      }
+
+      FileStatus latestCRCFile = crcFilesList.get(crcFilesList.size() - 1);
+      return readChecksumFile(engine, new Path(latestCRCFile.getPath()));
+    } catch (Exception e) {
```
allisonport-db, line 85: Is the intention to catch unchecked exceptions here as well? just wondering

Otherwise wondering if we should be more specific and just target IOExceptions or something

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ChecksumReader.java
allisonport-db: please update the docs since this no longer uses the 100 version range

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -201,13 +200,50 @@ public CloseableIterator<FilteredColumnarBatch> getAddFilesAsColumnarBatches(
...
+    // the lower bound for the CRC search.
+    long crcSearchLowerBound =
+        Math.min(
+            snapshotVersion,
+            Collections.max(
+                Arrays.asList(
+                    snapshotHint.map(SnapshotHint::getVersion).orElse(0L) + 1,
+                    logSegment.checkpointVersionOpt.orElse(0L),
+                    // Only find the CRC within 100 versions.
+                    Math.max(0, snapshotVersion - 100))));
```
allisonport-db, line 229: I find this confusing to read.

Didn't we already eliminate the `snapshotHint` if it's greater than `snapshotVersion`? So I think the `min` here might be unnecessary?

Also I think we don't want to use `checkpointVersionOpt` but rather the version from `checkpoints`? which will be guaranteed to be less than or equal to `snapshotVersion` -- so also no need for this min operation.

I think all that needs to happen here is
```
Math.max(
   checkpointVersionFromLogSegment,
   snapshotVersion - 100,
   snapshotHint.orElse(0) + 1,
   0
)
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ChecksumReader.java
huan233usc: It was added in the patched PR. I thinks IOExceptions is better but keeping Exceptions is not a deal broker, so didn't change it.
Updated

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
huan233usc: corner case is snapshotVersion == 0 :-( so using min is to ensure snapshotVersion>= lowerbound for corner case like this

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/FileNames.java
```diff
@@ -152,7 +171,7 @@ public static Path topLevelV2CheckpointFile(
 
   /** Returns the path for a V2 sidecar file with a given UUID. */
   public static Path v2CheckpointSidecarFile(Path path, String uuid) {
-    return new Path(String.format("%s/%s/%s.parquet", path.toString(), SIDECAR_DIRECTORY, uuid));
```
scottsand-db, line 162: use the code that is in master please

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
scottsand-db: Good catch Allison! The `checkpointVersionOpt` was my idea, and as Allison pointed out this is indeed incorrect. The version from the LogSegments.checkpoints is correct

+1 on we should make this easier to understand

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/LogReplayEngineMetricsSuite.scala
```diff
@@ -342,6 +379,230 @@ class LogReplayEngineMetricsSuite extends QueryTest
...
+        engine,
+        expJsonVersionsRead = Seq(11),
+        expParquetVersionsRead = Nil,
+        expParquetReadSetSizes = Nil,
+        expChecksumReadSet = Seq(11, 10)
+      )
+    }
+  }
+
+  def buildTableWithCrc(path: String): Unit = {
```
scottsand-db, line 597: what are the assumptions of what delta talbe state this method produces? 0.json to 11.json, 0.crc to 11.crc, 10.checkpoint.parquet?

Can we (a) state these assumptions, (b) assert that they are true, and (c) inline in the test succinctly what his has yieled e.g. `buildTableWithCrc(path) // 0 to 11 .json, 0 to 11.crc, 10.checkpoint.parquet`

This will make understanding the tests much easier.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/LogReplayEngineMetricsSuite.scala
```diff
@@ -342,6 +379,230 @@ class LogReplayEngineMetricsSuite extends QueryTest
...
+        expParquetReadSetSizes = Nil,
+        expChecksumReadSet = Seq(3))
+    }
+  }
+
+  test("checksum not found at the read version, but uses snapshot hint lower bound") {
+    withTempDirAndMetricsEngine { (path, engine) =>
+      buildTableWithCrc(path)
+      (3 to 6).foreach { version =>
+        val crcFile = new File(path, f"_delta_log/$version%020d.crc")
```
scottsand-db, line 460: when creating crc file names, can we please use `FileNames. checksumFile`

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/LogReplayEngineMetricsSuite.scala
huan233usc: >0.json to 11.json, 0.crc to 11.crc, 10.checkpoint.parquet
Yes correct.
Added comments

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
huan233usc: Updated, removed min() and use LogSegments.checkpoints. Thanks both.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/LogReplayEngineMetricsSuite.scala
scottsand-db: Please verbatim include this comment in the `buildTableWithCrc(path)` usages above

`buildTableWithCrc(path) // 0 to 11 .json, 0 to 11.crc, 10.checkpoint.parquet`

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/LogReplayEngineMetricsSuite.scala
```diff
@@ -342,6 +379,271 @@ class LogReplayEngineMetricsSuite extends QueryTest
...
+        expParquetReadSetSizes = Nil,
+        // First we attempt to read at version 6, then we do a listing of last 100 crc files
+        // bound by the snapshot hint which is at version 4 and we don't try to read checksums
+        // beyond version 4
+        expChecksumReadSet = Seq(6)
+      )
+    }
+  }
+
+  test(
```
scottsand-db, line 506:
```java
test("checksum not found at read version and checkpoint exists at read version => use checkpoint") {
```

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/LogReplayEngineMetricsSuite.scala
```diff
@@ -342,6 +379,271 @@ class LogReplayEngineMetricsSuite extends QueryTest
...
+          // shouldn't need to read commit or checkpoint files as P&M are found through checksum
+          expJsonVersionsRead = Nil,
+          expParquetVersionsRead = Nil,
+          expParquetReadSetSizes = Nil,
+          expChecksumReadSet = Seq(if (version == -1) 11 else version))
+      }
+    }
+  }
+
+  test("checksum not found at the read version, but found at a previous version") {
```
scottsand-db, line 408: I think you're missing cases for:
1. snapshot hint found for read verson AND crc found at read version ==> use hint
2. crc found at read version AND checkpoint at read version => use crc

huan233usc: added both

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ChecksumReader.java
```diff
@@ -0,0 +1,128 @@
...
+      ColumnarBatch batch = iter.next();
+      if (batch.getSize() != 1) {
+        String msg = "Expected exactly one row in the checksum file {}, found {} rows";
+        logger.warn(msg, filePath, batch.getSize());
+        return Optional.empty();
+      }
+
+      long crcVersion = FileNames.checksumVersion(filePath);
+
+      return fromColumnarBatch(engine, crcVersion, batch, 0 /* rowId */, filePath.toString());
```
scottsand-db, line 121: `CRCInfo.fromColumnarBatch`.

`fromColumnarBatch` is used in lots of places in the code --> lets include the classname so it's clear to the reader which class and method you are calling

scottsand-db (changes requested):
Looking good! A few more minor changes to make the code a bit clearer

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ChecksumReader.java
```diff
@@ -0,0 +1,128 @@
...
+   * @param logPath the path to the Delta log
+   * @param targetedVersion the target version to read the checksum file from
+   * @param lowerBound the inclusive lower bound version to search for the checksum file
+   * @return Optional {@link CRCInfo} containing the protocol and metadata, and the version of the
+   *     checksum file. If the checksum file is not found, it will return an empty
+   */
+  public static Optional<CRCInfo> getCRCInfo(
+      Engine engine, Path logPath, long targetedVersion, long lowerBound) {
+    // If targetedVersion is 0, lower bound will be ignored as snapshot 0 is the first snapshot.
+    checkArgument(targetedVersion == 0 || targetedVersion >= lowerBound);
```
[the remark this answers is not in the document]
huan233usc: There is an edge case caused by the caller's call which makes lowerBound at least 1 

Math.max(
   checkpointVersionFromLogSegment,
   snapshotVersion - 100,
   snapshotHint.orElse(0) + 1,
   0
)
targetVersion=0 needs to be handle separately.  

I move the check to `checkArgument(targetVersion >= lowerBound)` to line 64 before lower bound is actually used.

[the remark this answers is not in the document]
scottsand-db: > There is an edge case caused by the caller's call which makes lowerBound at least 1
Another option would be to, in getCrcInfo, just set lowerBound to min(lowerBound, targetVersion), right?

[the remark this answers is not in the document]
huan233usc: yes, I think this option is better but was flagged previous due to readability concern. 
https://github.com/delta-io/delta/pull/4077#discussion_r1927602117

Or maybe I could repurpose it in a more readability way (avoid inline)

scottsand-db (approved):
LGTM!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/InternalUtils.java
```diff
@@ -169,4 +168,11 @@ public static Path relativizePath(Path child, URI root) {
   public static Set<String> toLowerCaseSet(Collection<String> set) {
     return set.stream().map(String::toLowerCase).collect(Collectors.toSet());
   }
+
+  public static <T> List<T> toFilteredList(
+      CloseableIterator<T> iterator, Function<T, Boolean> filter) {
+    List<T> result = new ArrayList<>();
+    iterator.filter(filter).forEachRemaining(result::add);
+    return result;
+  }
```
allisonport-db, line 177: Sorry nit, I think it's easier if this is just `toList` and we can do the iterator filtering in the outer call. This way we can re-use this even when we don't want to filter

i.e. `toList(iterator.filter(filter))`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/CRCInfo.java
```diff
@@ -0,0 +1,75 @@
...
+import io.delta.kernel.types.StructType;
+import java.util.Optional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CRCInfo {
+  private static final Logger logger = LoggerFactory.getLogger(CRCInfo.class);
+
+  public static Optional<CRCInfo> fromColumnarBatch(
+      Engine engine, long version, ColumnarBatch batch, int rowId, String crcFilePath) {
```
allisonport-db, line 33: remove `engine` parameter? it's unused

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -201,13 +204,54 @@ public CloseableIterator<FilteredColumnarBatch> getAddFilesAsColumnarBatches(
...
+    // the lower bound for the CRC search.
+    List<Long> eligibleCheckpointVersions =
+        logSegment.checkpoints.stream()
+            .map(cp -> checkpointVersion(cp.getPath()))
+            .filter(v -> v <= snapshotVersion)
+            .collect(Collectors.toList());
+    long crcSearchLowerBound =
+        max(
+            asList(
+                snapshotHint.map(SnapshotHint::getVersion).orElse(0L) + 1,
```
allisonport-db, line 233: nit - maybe add an inline comment why we add +1? might not be clear to a future reader

allisonport-db (changes requested):
Just a few minor remaining comments then LGTM

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ChecksumReader.java
```diff
@@ -0,0 +1,127 @@
...
+
+    Path lowerBoundFilePath = checksumFile(logPath, lowerBound);
+    try (CloseableIterator<FileStatus> crcFiles =
+        engine.getFileSystemClient().listFrom(lowerBoundFilePath.toString())) {
+      List<FileStatus> crcFilesList =
+          InternalUtils.toList(
+              crcFiles.filter(
+                  file ->
+                      isChecksumFile(file.getPath())
+                          && checksumVersion(new Path(file.getPath())) <= targetedVersion));
```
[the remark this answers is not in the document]
huan233usc: For this case, I am afraid that we need to use filter. 

engine.getFileSystemClient().listFrom may return lists in alphabetical order 1.crc 10.crc 2.crc 3.crc.... In this case  takeWhile may end up missing the snapshot like 2,3,4,....9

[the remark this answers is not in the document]
scottsand-db: To be clear, these file names are really:

00000000000000000001.crc
00000000000000000002.crc
00000000000000000010.crc

correct?

so we do expect 002.crc to come before 010.crc, right?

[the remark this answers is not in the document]
huan233usc: Yeah you are correct...let me double check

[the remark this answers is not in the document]
huan233usc: Sorry for the confusion.. I eventually found that my local implementaion had a bug on instructing the take while. 
>I really sure hope this is in order
Confirmed that it is in order and the ordering is just like
00000000000000000001.crc
00000000000000000002.crc
00000000000000000010.crc

Updated by adding a take while in the iterator.
Thanks both

kernel/kernel-api/src/main/java/io/delta/kernel/utils/CloseableIterator.java
```diff
@@ -129,6 +129,49 @@ public void close() throws IOException {
     };
   }
 
+  default CloseableIterator<T> takeWhile(Function<T, Boolean> mapper) {
```
scottsand-db, line 132: A few thoughts on this:
1. this PR is getting long enough, we should refactor this out to another PR and merge it first.
2. the `takeWhile` implementation looks very similar to the `filter` implementation. We should consolidate this into 2 public APIs and then 1 helper method. Unforunately, Java 8 doesn't allow private methods in interfaces, so this underlying implementation should probably go into some `CloseableIteratorUtils` class. I suggest something like the below:

```java
// CloseableIterator.java
default CloseableIterator<T> filter(Function<T, Boolean> mapper) {
    return CloseableIteratorUtils.conditionalFilter(this, mapper, false /* stopOnFalse */);
}

default CloseableIterator<T> takeWhile(Function<T, Boolean> mapper) {
    return CloseableIteratorUtils.conditionalFilter(this, mapper, true /* stopOnFalse */);
}

// /.../internal/CloseableIteratorUtils.java
/** A helper method to handle both filtering and takeWhile behavior. */
public CloseableIterator<T> conditionalFilter(Function<T, Boolean> mapper, boolean stopOnFalse) {
```

kernel/kernel-api/src/test/scala/io/delta/kernel/utils/CloseableIteratorSuite.scala
```diff
@@ -0,0 +1,64 @@
...
+package io.delta.kernel.utils
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatest.matchers.must.Matchers.contain
+import org.scalatest.matchers.should.Matchers.convertToAnyShouldWrapper
+
+import java.util
+
+class CloseableIteratorSuite extends AnyFunSuite {
+
+  test("test take while with all elements matching filter") {
```
scottsand-db, line 25: `test("test ` ---> don't need to say "test" at the start of a test name

`take while` --> `takeWhile`

kernel/kernel-api/src/test/scala/io/delta/kernel/utils/CloseableIteratorSuite.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  test("test take while with no elements matching filter") {
+    val res = new util.ArrayList[Int]();
+    new IntArrayIterator(List(1, 2, 3, 2, 2))
+      .takeWhile((input: Int) => input < 0)
+      .forEachRemaining((input: Int) => res.add(input))
+    assert(res.isEmpty)
+  }
+
+  test("test take while with some elements matching filter") {
```
scottsand-db, line 41: we want to check that the last two `2`s are not even visited. not only are they excluded from the returned output, but they aren't even checked.

you can do this with a custom iterator that throws an exception, e.g. imagine i have values 1 2 3 4 5 6.

first, i have an iterator that throws if it sees a values >= 5.

then i have an iterator that takesWhile value is <= 4

the value = 5 should never be seen and the exception should never be thrown.

scottsand-db (changes requested):
Looks great! Unfortunately, `takeWhile` definitely needs to be its own PR.

kernel/kernel-api/src/main/java/io/delta/kernel/utils/CloseableIterator.java
scottsand-db: There's actually a more expressive syntax we might want to implement, at least locally:
- https://github.com/delta-io/delta/blob/2ffdbbf2748a2355fa45b9795787091ca4795567/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaLogActionUtils.java#L236
- https://github.com/delta-io/delta/blob/2ffdbbf2748a2355fa45b9795787091ca4795567/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaLogActionUtils.java#L262

We might want the result of the `mapper` to be: INCLUDE, EXCLUDE, BREAK.
- filter would, internally, only need INCLUDE or EXCLUDE
- takeWhile would, internally, only need INCLUDE or BREAK
- the DeltaLogActionUtils would want all 3: INCLUDE, EXCLUDE, and BREAK

In this case: we would have something like

```java
enum BreakableFilterResult {
    INCLUDE,
    EXCLUDE,
    BREAK
  }

default CloseableIterator<T> filter(Function<T, Boolean> mapper) {
    return breakableFilter(
        t -> {
          if (mapper.apply(t)) {
            return BreakableFilterResult.INCLUDE;
          } else {
            return BreakableFilterResult.EXCLUDE;
          }
        });
  }

default CloseableIterator<T> takeWhile(Function<T, Boolean> mapper) {
    return breakableFilter(
        t -> {
          if (mapper.apply(t)) {
            return BreakableFilterResult.INCLUDE;
          } else {
            return BreakableFilterResult.BREAK;
          }
        });
  }

default CloseableIterator<T> breakableFilter(Function<T, BreakableFilterResult> mapper) {
```

scottsand-db: I can make a PR for this

scottsand-db: https://github.com/delta-io/delta/pull/4097

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
scottsand-db: Update: We do actually want to use `checkpointVersionOpt`. It should have the save version as `checkpoints`

scottsand-db (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/InternalUtils.java
```diff
@@ -169,4 +167,10 @@ public static Path relativizePath(Path child, URI root) {
   public static Set<String> toLowerCaseSet(Collection<String> set) {
     return set.stream().map(String::toLowerCase).collect(Collectors.toSet());
   }
+
+  public static <T> List<T> toList(CloseableIterator<T> iterator) {
+    List<T> result = new ArrayList<>();
+    iterator.forEachRemaining(result::add);
+    return result;
+  }
```
allisonport-db, line 175: Ugh sorry hope this is the last thing, but @scottsand-db added `CloseableIterator::toInMemoryList` can we use that and remove this method?

allisonport-db (changes requested):

allisonport-db (approved):

PR #4086, opened 2025-01-23
[Kernel] Clean up `FileNames.java` and add more tests

scottsand-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Clean up `FileNames.java` and add more tests

## How was this patch tested?

New UTs.

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/FileNames.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/FileNames.java
@@ -25,28 +25,71 @@ public final class FileNames {
 
   private FileNames() {}
 
+  ////////////////////////////////////////////////
+  // File name patterns and other static values //
+  ////////////////////////////////////////////////
+
+  /** Example: 00000000000000000001.json */
   private static final Pattern DELTA_FILE_PATTERN = Pattern.compile("\\d+\\.json");
 
-  // Example: 00000000000000000001.dc0f9f58-a1a0-46fd-971a-bd8b2e9dbb81.json
+  /** Example: 00000000000000000001.dc0f9f58-a1a0-46fd-971a-bd8b2e9dbb81.json */
   private static final Pattern UUID_DELTA_FILE_REGEX = Pattern.compile("(\\d+)\\.([^\\.]+)\\.json");
 
+  /**
+   * Examples:
+   *
+   * <ul>
+   *   <li>Classic V1 - 00000000000000000001.checkpoint.parquet
+   *   <li>Multi-part V1 - 00000000000000000001.checkpoint.0000000001.0000000010.parquet
+   *   <li>V2 JSON - 00000000000000000001.checkpoint.uuid-1234abcd.json
+   *   <li>V2 Parquet - 00000000000000000001.checkpoint.uuid-1234abcd.parquet
+   * </ul>
+   */
   private static final Pattern CHECKPOINT_FILE_PATTERN =
       Pattern.compile("(\\d+)\\.checkpoint((\\.\\d+\\.\\d+)?\\.parquet|\\.[^.]+\\.(json|parquet))");
 
+  /** Example: 00000000000000000001.checkpoint.parquet */
   private static final Pattern CLASSIC_CHECKPOINT_FILE_PATTERN =
       Pattern.compile("\\d+\\.checkpoint\\.parquet");
 
+  /**
+   * Examples:
+   *
+   * <ul>
+   *   <li>00000000000000000001.checkpoint.dc0f9f58-a1a0-46fd-971a-bd8b2e9dbb81.json
+   *   <li>00000000000000000001.checkpoint.dc0f9f58-a1a0-46fd-971a-bd8b2e9dbb81.parquet
+   * </ul>
+   */
   private static final Pattern V2_CHECKPOINT_FILE_PATTERN =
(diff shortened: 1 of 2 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/FileNames.java
```diff
@@ -25,28 +25,71 @@ public final class FileNames {
...
+  // Version extractors //
+  ////////////////////////
+
+  /**
+   * Get the version of the checkpoint, checksum or delta file. Throws an error if an unexpected
+   * file type is seen. These unexpected files should be filtered out to ensure forward
+   * compatibility in cases where new file types are added, but without an explicit protocol
+   * upgrade.
+   */
+  public static long getFileVersion(Path path) {
```
scottsand-db, line 82: copied verbatim from below

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/FileNames.java
```diff
@@ -128,44 +181,29 @@ public static List<Path> checkpointFileWithParts(Path path, long version, int nu
...
+    return CHECKPOINT_FILE_PATTERN.matcher(new Path(path).getName()).matches();
   }
 
-  public static boolean isMultiPartCheckpointFile(String fileName) {
-    return MULTI_PART_CHECKPOINT_FILE_PATTERN.matcher(fileName).matches();
+  public static boolean isClassicCheckpointFile(String path) {
+    return CLASSIC_CHECKPOINT_FILE_PATTERN.matcher(new Path(path).getName()).matches();
   }
 
-  public static boolean isV2CheckpointFile(String fileName) {
```
scottsand-db, line 143: unfortunately, GitHub is showing this diff terrible :( 

all I did was change the input param to `path` from `fileName`, and now we use `.matcher(new Path(path).getName())` instead of `.matcher(fileName`).

Previously, this would not work with `isMultiPartCheckpointFile("/a/b/c/_delta_log/<multi-part-checkpoint-file>")`

However, `isCheckpointFile("/a/b/c/_delta_log/<multi-part-checkpoint-file>")` WOULD work -- this was an inconsistency

allisonport-db: why is this? what does `new Path(path).getName()` change from just `path`?

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/util/FileNamesSuite.scala
```diff
@@ -59,4 +105,33 @@ class FileNamesSuite extends AnyFunSuite {
       new Path("/a/00000000000000000001.checkpoint.0000000004.0000000005.parquet"),
       new Path("/a/00000000000000000001.checkpoint.0000000005.0000000005.parquet")))
   }
+
+  ///////////////////////////////////
+  // Is <type> file checkers tests //
+  ///////////////////////////////////
+
+  test("is checkpoint file") {
```
allisonport-db, line 113: should we check that the specific checkpoint types don't match the others?
i.e. `assert(!isClassicCheckpointFile(checkpointMultiPart)`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/FileNames.java
scottsand-db: This is the original code for isCheckpointFile:

```java
public static boolean isCheckpointFile(String fileName) {
    return CHECKPOINT_FILE_PATTERN.matcher(new Path(fileName).getName()).matches();
  }
```

Really, it let you input a _path_ and then would get the name from that path, and then check if that name matches the CHECKPOINT_FILE_PATTERN.

This is the original code for, for example, isClassicCheckpointFile:

```java
public static boolean isClassicCheckpointFile(String fileName) {
    return CLASSIC_CHECKPOINT_FILE_PATTERN.matcher(fileName).matches();
}
```

Note how it only lets you input a file name -- not a path. This is less useful than letting it take in a path.

Also note how there is a difference of semantics between `isCheckpointFile` (which works with paths) and `isClassicCheckpointFile` (which only works for file names).

This PR changes it so that all of these methods work with the paths, not just the file names.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/util/FileNamesSuite.scala
scottsand-db: sure!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/FileNames.java
scottsand-db: The new tests should show you why this is nicer/better

allisonport-db: Gotcha nevermind I was confused; the change makes sense to me.

allisonport-db (approved):
LGTM!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/FileNames.java
scottsand-db: No problem!

PR #4089, opened 2025-01-24
[Kernel] Minor SnapshotManager refactor followup

scottsand-db:
- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Followup to #4035.

Our SnapshotManager LogSegment construction logic is ... a bit messy, to say the least. This PR makes `getLogSegmentForVersion` _not_ return an Optional LogSegment.

I also start documenting some of the key steps in constructing a LogSegment. More steps, refactors, and comments will come in future PRs.

## How was this patch tested?

Mainly just a refactor. Existing UTs.

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
@@ -20,7 +20,6 @@ import static io.delta.kernel.internal.DeltaErrors.wrapEngineExceptionThrowsIO;
 import static io.delta.kernel.internal.TableConfig.EXPIRED_LOG_CLEANUP_ENABLED;
 import static io.delta.kernel.internal.TableConfig.LOG_RETENTION;
 import static io.delta.kernel.internal.TableFeatures.validateWriteSupportedTable;
-import static io.delta.kernel.internal.checkpoints.Checkpointer.findLastCompleteCheckpointBefore;
 import static io.delta.kernel.internal.replay.LogReplayUtils.assertLogFilesBelongToTable;
 import static io.delta.kernel.internal.snapshot.MetadataCleanup.cleanupExpiredLogs;
 import static io.delta.kernel.internal.util.Preconditions.checkArgument;
@@ -60,7 +59,7 @@ public class SnapshotManager {
    * The latest {@link SnapshotHint} for this table. The initial value inside the AtomicReference is
    * `null`.
    */
-  private AtomicReference<SnapshotHint> latestSnapshotHint;
+  private final AtomicReference<SnapshotHint> latestSnapshotHint;
 
   private final Path logPath;
   private final Path tablePath;
@@ -81,12 +80,18 @@ public class SnapshotManager {
    * Construct the latest snapshot for given table.
    *
    * @param engine Instance of {@link Engine} to use.
-   * @return
-   * @throws TableNotFoundException
+   * @return the latest {@link Snapshot} of the table
+   * @throws TableNotFoundException if the table does not exist
+   * @throws InvalidTableException if the table is in an invalid state
    */
   public Snapshot buildLatestSnapshot(Engine engine, SnapshotQueryContext snapshotContext)
       throws TableNotFoundException {
-    return getSnapshotAtInit(engine, snapshotContext);
+    final LogSegment logSegment =
+        getLogSegmentForVersion(engine, Optional.empty() /* versionToLoad */);
+
+    snapshotContext.setVersion(logSegment.version);
+
+    return createSnapshot(logSegment, engine, snapshotContext);
   }
 
   /**
@@ -95,20 +100,15 @@ public class SnapshotManager {
    * @param engine Instance of {@link Engine} to use.
    * @param version The snapshot version to construct
    * @return a {@link Snapshot} of the table at version {@code version}
-   * @throws TableNotFoundException
+   * @throws TableNotFoundException if the table does not exist
+   * @throws InvalidTableException if the table is in an invalid state
    */
   public Snapshot getSnapshotAt(Engine engine, long version, SnapshotQueryContext snapshotContext)
       throws TableNotFoundException {
+    final LogSegment logSegment =
+        getLogSegmentForVersion(engine, Optional.of(version) /* versionToLoadOpt */);
 
-    Optional<LogSegment> logSegmentOpt =
-        getLogSegmentForVersion(
-            engine,
-            Optional.empty(), /* startCheckpointOpt */
-            Optional.of(version) /* versionToLoadOpt */);
-
-    return logSegmentOpt
-        .map(logSegment -> createSnapshot(logSegment, engine, snapshotContext))
-        .orElseThrow(() -> new TableNotFoundException(tablePath.toString()));
+    return createSnapshot(logSegment, engine, snapshotContext);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
```diff
@@ -744,12 +735,6 @@ class SnapshotManagerSuite extends AnyFunSuite with MockFileSystemClientUtils {
       versionToLoad = Optional.of(17),
       expectedErrorMessageContains = "missing log file for version 0"
     )
-    testExpectedError[InvalidTableException](
-      deltaFileStatuses(15L until 25L) ++ singularCheckpointFileStatuses(Seq(20L)),
-      startCheckpoint = Optional.of(20),
```
scottsand-db, line 749: this just throws an IllegalArgumentException when we invoke the log listing util ... which is expected.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
```diff
@@ -524,15 +522,8 @@ class SnapshotManagerSuite extends AnyFunSuite with MockFileSystemClientUtils {
...
   test("getLogSegmentForVersion: versionToLoad not constructable from history") {
-    val files = deltaFileStatuses(20L until 25L) ++ singularCheckpointFileStatuses(Seq(20L))
     testExpectedError[RuntimeException](
-      files,
-      versionToLoad = Optional.of(15),
-      expectedErrorMessageContains = "Cannot load table version 15"
-    )
-    testExpectedError[RuntimeException](
-      files,
-      startCheckpoint = Optional.of(20),
```
scottsand-db, line 535: this just throws an IllegalArgumentException when we invoke the log listing util ... which is expected.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -341,78 +317,108 @@ private SnapshotImpl createSnapshot(
...
+                            return checkpointVersion;
+                          });
+                })
+            .orElseGet(
+                () -> {
+                  logger.info("Loading last checkpoint from the _last_checkpoint file");
+                  return new Checkpointer(logPath)
+                      .readLastCheckpointFile(engine)
+                      .map(x -> x.version);
+                });
```
allisonport-db, line 378: Q: in this case (where `versionToLoadOpt` is empty), didn't we previously default to using `findLastCompleteCheckpointBefore` if `_last_checkpoint` didn't exist. Wondering if this is a meaningful change.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -425,8 +431,7 @@ public Optional<LogSegment> getLogSegmentForVersion(
    * Helper function for the getLogSegmentForVersion above. Called with a provided files list, and
    * will then try to construct a new LogSegment using that.
    */
-  protected Optional<LogSegment> getLogSegmentForVersion(
-      Engine engine,
+  private LogSegment getLogSegmentForVersionHelper(
```
allisonport-db, line 434: I don't really like all these "getLogSegmentForVersionHelper" and "getLogSegmentForVersion" naming maybe we can update these to be more specific.

Like maybe this one can be `buildLogSegmentForVersion` or `constructLogSegment` since it isn't doing the listing?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
allisonport-db, line 378: Wondering if maybe all this logic should just be a helper function like `findStartCheckpointVersion` or something?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
scottsand-db: I totally agree. I'm just not sure if this is the PR to do that. I'm trying to do it bit by bit. Today, in master, we have

1. `getLogSegmentFrom(Engine engine, Optional<CheckpointMetaData> startingCheckpoint)`
2. `getLogSegmentForVersion(Engine engine, Optional<Long> startCheckpoint, Optional<Long> versionToLoad)`
3. `getLogSegmentForVersion(Engine engine,Optional<Long> startCheckpointOpt,Optional<Long> versionToLoadOpt, List<FileStatus> newFiles)`

Ideally, we would have only one method: `getLogSegmentForVersion(Engine engine, Optional<Long> versionToLoad)`

This PR gets rid of number 1. Number two is extensively used for tests.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
scottsand-db: Good question. We did not do that. `findLastCompleteCheckpointBefore` is only useful if you can give it a version to search for _before_. If versionToLoad is empty, then we don't have an end version to compare against.

https://github.com/delta-io/delta/blob/107b2d3c3279093ab6481503c399b828579a031a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java#L381

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
scottsand-db: And i'm not sure if I love that number two is used for tests. We only have 2 entry points here: get snapshot for latest, get snapshot for version (from a public API perspective. note that get snapshot for timestamp really just used get snapshot for version).

So IMO our tests should be testing _that_ entry point, not testing some custom method where you can input a starting checkpoint, too. Ideally we just setup our tests better where we define what the starting checkpoint is via setting the table state, as opposed to having that as input to the method ....

scottsand-db: But that's a big refator 😅

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
scottsand-db: I like this suggestion, yup!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
scottsand-db: So I think in this PR i'd like to merge `getLogSegmentForVersion(Engine engine, Optional<Long> versionToLoad)` and in a followup I can go and remove the 33 test instances of `getLogSegmentForVersion(Engine engine, Optional<Long> startCheckpoint, Optional<Long> versionToLoad)`

scottsand-db: constructLogSegmentFromFileList

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -341,78 +317,108 @@ private SnapshotImpl createSnapshot(
...
+    final String versionToLoadStr = versionToLoadOpt.map(String::valueOf).orElse("latest");
+    logger.info("Loading log segment for version {}", versionToLoadStr);
+
+    ////////////////////////////////////////////////////////////////////////////////////////////////
+    // Step 1: Find the latest checkpoint version. If $versionToLoadOpt is empty, use the version //
+    //         referenced by the _LAST_CHECKPOINT file. If $versionToLoad is present, search for  //
+    //         the previous latest complete checkpoint at or before $versionToLoad.               //
+    ////////////////////////////////////////////////////////////////////////////////////////////////
+
+    final Optional<Long> startCheckpointVersionOpt =
```
scottsand-db, line 344: refactor to util

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -625,14 +599,56 @@ protected Optional<LogSegment> getLogSegmentForVersion(
...
+  /////////////////////////
+  // getLogSegment utils //
+  /////////////////////////
+
+  /**
+   * Determine the starting checkpoint version that is at or before `versionToLoadOpt`. If no
+   * `versionToLoadOpt` is provided, will use the checkpoint pointed to by the _last_checkpoint
+   * file.
+   */
+  private Optional<Long> getStartCheckpointVersion(Engine engine, Optional<Long> versionToLoadOpt) {
```
scottsand-db, line 620: In the future we can also add nice tests for each of these individual getLogSegment util helper methods.

I invision a deep getLogSegment method (as opposed to, basically, getLogSegmentA calls getLogSegmentB which calls getLogSegmentC).

Instead, getLogSegment is 1 large method, and it has little utilities that poke and help out.

allisonport-db (approved):
awesome lgtm

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
scottsand-db: done!

PR #4090, opened 2025-01-26
[Kernel] Further refactor and cleanup SnapshotManager

scottsand-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

This PR further refactors and cleans up SnapshotManager. We consolidate all of our `getLogSegment` methods to just one. It has a deep implementation. This PR adds nice, clean header blocks to describe the steps during log segment construction.

## How was this patch tested?

Refactor. Update existing UTs.

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
@@ -48,7 +48,6 @@ import java.io.*;
 import java.nio.file.FileAlreadyExistsException;
 import java.util.*;
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -118,7 +117,8 @@ public class SnapshotManager {
    * @param millisSinceEpochUTC timestamp to fetch the snapshot for in milliseconds since the unix
    *     epoch
    * @return a {@link Snapshot} of the table at the provided timestamp
-   * @throws TableNotFoundException
+   * @throws TableNotFoundException if the table does not exist
+   * @throws InvalidTableException if the table is in an invalid state
    */
   public Snapshot getSnapshotForTimestamp(
       Engine engine, long millisSinceEpochUTC, SnapshotQueryContext snapshotContext)
@@ -236,17 +236,19 @@ public class SnapshotManager {
     }
     expectedStartVersion.ifPresent(
         v -> {
-          checkArgument(
-              !versions.isEmpty() && Objects.equals(versions.get(0), v),
-              "Did not get the first delta file version %s to compute Snapshot",
-              v);
+          if (versions.isEmpty() || !Objects.equals(versions.get(0), v)) {
+            throw new InvalidTableException(
+                tablePath.toString(),
+                String.format("Cannot compute snapshot. Missing delta file version %d.", v));
+          }
         });
     expectedEndVersion.ifPresent(
         v -> {
-          checkArgument(
-              !versions.isEmpty() && Objects.equals(ListUtils.getLast(versions), v),
-              "Did not get the last delta file version %s to compute Snapshot",
-              v);
+          if (versions.isEmpty() || !Objects.equals(ListUtils.getLast(versions), v)) {
+            throw new InvalidTableException(
+                tablePath.toString(),
+                String.format("Cannot compute snapshot. Missing delta file version %d.", v));
+          }
         });
   }
 
@@ -331,7 +333,8 @@ public class SnapshotManager {
    *   <li>Third, process and validate this list of _delta_log files to yield a {@code LogSegment}.
    * </ol>
    */
-  private LogSegment getLogSegmentForVersion(Engine engine, Optional<Long> versionToLoadOpt) {
+  @VisibleForTesting
+  public LogSegment getLogSegmentForVersion(Engine engine, Optional<Long> versionToLoadOpt) {
     final String versionToLoadStr = versionToLoadOpt.map(String::valueOf).orElse("latest");
     logger.info("Loading log segment for version {}", versionToLoadStr);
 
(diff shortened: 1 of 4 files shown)


kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
```diff
@@ -836,28 +805,17 @@ class SnapshotManagerSuite extends AnyFunSuite with MockFileSystemClientUtils {
...
+          fileSystemClient = new MockListFromFileSystemClient(_ => Seq.empty)
+        ),
         Optional.empty()
       )
     }.getMessage
 
     assert(exMsg.contains("Missing checkpoint at version 1"))
   }
-
-  test("getLogSegmentForVersion: startCheckpoint is greater than versionToLoad") {
```
scottsand-db, line 848: not really applicable anymore; we don't directly input the startCheckpoint

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -236,17 +236,19 @@ public static void verifyDeltaVersions(
...
         v -> {
-          checkArgument(
-              !versions.isEmpty() && Objects.equals(ListUtils.getLast(versions), v),
-              "Did not get the last delta file version %s to compute Snapshot",
-              v);
+          if (versions.isEmpty() || !Objects.equals(ListUtils.getLast(versions), v)) {
+            throw new InvalidTableException(
+                tablePath.toString(),
+                String.format("Cannot compute snapshot. Missing delta file version %d.", v));
+          }
```
allisonport-db, line 251: I found this logic somewhere a little confusing, because I think we don't actually hit this right?

We calculate `newVersion` using `deltasAfterCheckpoint` and then we provide it as the endVersion to this method. I feel like maybe `IllegalArgumentException` is the more appropriate error type here?

Can you explain to me what kind of invalid table might hit this exception?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -419,192 +401,180 @@ private LogSegment constructLogSegmentFromFileList(
...
-                Arrays.toString(
-                    checkpoints.stream().map(x -> new Path(x.getPath()).getName()).toArray()),
-                Arrays.toString(
-                    deltas.stream().map(x -> new Path(x.getPath()).getName()).toArray())));
-
-    // Find the latest checkpoint in the listing that is not older than the versionToLoad
-    final CheckpointInstance maxCheckpoint =
-        versionToLoadOpt.map(CheckpointInstance::new).orElse(CheckpointInstance.MAX_VALUE);
-    logger.debug("lastCheckpoint: {}", maxCheckpoint);
+    final List<FileStatus> listedCheckpointFileStatuses = listedCheckpointAndDeltaFileStatuses._1;
```
allisonport-db, line 414: nit: I don't think we need the `listed` prefix for all these variables (these plus those above); seems clear enough and can shorten them all a bit

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -419,192 +401,180 @@ private LogSegment constructLogSegmentFromFileList(
...
     final List<FileStatus> deltasAfterCheckpoint =
-        deltas.stream()
+        listedDeltaFileStatuses.stream()
             .filter(
-                fileStatus ->
-                    FileNames.deltaVersion(new Path(fileStatus.getPath())) > newCheckpointVersion)
+                fs -> {
+                  final long deltaVersion = FileNames.deltaVersion(new Path(fs.getPath()));
+                  return latestCompleteCheckpointVersion + 1 <= deltaVersion
+                      && deltaVersion <= versionToLoadOpt.orElse(Long.MAX_VALUE);
```
allisonport-db, line 459: this is redundant right? since we only list up to versionToLoad? (for my understanding, it's fine to stay)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -419,192 +401,180 @@ private LogSegment constructLogSegmentFromFileList(
...
-      // If we have deltas after the checkpoint, the first file should be 1 greater than our
-      // last checkpoint version. If no checkpoint is present, this means the first delta file
-      // should be version 0.
-      if (deltaVersionsAfterCheckpoint.get(0) != newCheckpointVersion + 1) {
-        throw new InvalidTableException(
-            tablePath.toString(),
-            String.format(
-                "Unable to reconstruct table state: missing log file for version %s",
-                newCheckpointVersion + 1));
-      }
```
allisonport-db, line 558: This check is now done as part of verifyDeltaVersions I think?

allisonport-db: I think if we're throwing `InvalidTableExceptions` inside of `verifyDeltaVersions` the docs for that method need to be very specific to explain the input/output (what they are in the flow of the segment construction) and why they are invalid tables

kernel/kernel-api/src/test/scala/io/delta/kernel/test/MockFileSystemClientUtils.scala
```diff
@@ -107,6 +109,21 @@ trait MockFileSystemClientUtils extends MockEngineUtils {
...
+        new MockReadLastCheckpointFileJsonHandler(
+          s"$logPath/_last_checkpoint",
+          lastCheckpointVersion.get()
+        )
+      } else {
+        null
+      }
+    )
+  }
+
```
allisonport-db, line 126: Any reason to put this here instead of with the `MockReadLastCheckpointFileJsonHandler`?

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
```diff
@@ -836,28 +805,17 @@ class SnapshotManagerSuite extends AnyFunSuite with MockFileSystemClientUtils {
   test("getLogSegmentForVersion: corrupt _last_checkpoint with empty delta log") {
     val exMsg = intercept[InvalidTableException] {
       snapshotManager.getLogSegmentForVersion(
-        createMockFSListFromEngine(Seq.empty),
-        Optional.of(1),
+        mockEngine(
+          jsonHandler = new MockReadLastCheckpointFileJsonHandler(
+            s"$logPath/_last_checkpoint", 1),
+          fileSystemClient = new MockListFromFileSystemClient(_ => Seq.empty)
+        ),
```
allisonport-db, line 812: Sorry why can't this use `createMockFSAndJsonEngineForLastCheckpoint`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
scottsand-db: I actually disagree. There are the checkpoint file statuses we have `listed` and then there's the latest+complete checkpoint file statuses that we end up selecting. I think emphasizing that these are the listed ones -- but that doesn't mean they are the ones we are selecting to be in the log segmnt, is important.

See below:

```java
////////////////////////////////////////////////////////////////////////////////////////////
// Step 10: Grab the actual checkpoint file statuses for latestCompleteCheckpointVersion. //
////////////////////////////////////////////////////////////////////////////////////////////

final List<FileStatus> latestCompleteCheckpointFileStatuses =
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
scottsand-db: yup! "redundant" assuming that the LIST call returns what we expect.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
scottsand-db: > This check is now done as part of verifyDeltaVersions I think?

Exactly!

kernel/kernel-api/src/test/scala/io/delta/kernel/test/MockFileSystemClientUtils.scala
scottsand-db: This is next to all the other `createMockFS____` utilities, thought it would be best here

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
scottsand-db: Great catch! My sunday brain missed this!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
scottsand-db: > because I think we don't actually hit this right?

That's right. This is here as a redundancy. I'll use IllegalArgumentException !

scottsand-db: Synced offline. Will implement the desired approach!

allisonport-db (approved):
LGTM

Issue #2623, opened 2024-02-09
[PROTOCOL RFC] Type Widening Table Feature


johanl-db:
### Description of the protocol change
This protocol change is part of the proposed Type Widening table feature, see feature request: https://github.com/delta-io/delta/issues/2622

Design doc: https://docs.google.com/document/d/1KIqf6o6JMD7e8aMrGlUROSwTfzYeW4NCIZVAUMW_-Tc

The type widening table feature covers changing the type of existing columns or nested fields in a Delta table without having to rewrite the data.

### Willingness to contribute

The Delta Lake Community encourages protocol innovations. Would you or another member of your organization be willing to contribute this feature to the Delta Lake code base?

- [x] Yes. I can contribute.
- [ ] Yes. I would be willing to contribute with guidance from the Delta Lake community.
- [ ] No. I cannot contribute at this time.

emkornfield:
Should this be closed given https://github.com/delta-io/delta/pull/4094 was merged?

johanl-db:
Yes, closing.

PR #4094, opened 2025-01-28
Accept Type Widening RFC

johanl-db:
## Description
The type widening table feature has a stable implementation in delta-spark since Delta 3.2 an in Kernel Java/Rust.

This PR proposes to accept the RFC specification of the feature and add it to the Delta protocol.

Type Widening feature request: https://github.com/delta-io/delta/issues/2623
Related: transitioning table feature from preview to stable: https://github.com/delta-io/delta/pull/4094#issue-2815134778


PROTOCOL.md
```diff
@@ -1387,6 +1387,130 @@ Furthermore, when attempting timestamp-based time travel where table state must
...
+  - `Byte` -> `Short` -> `Int` -> `Long`
+- Floating-point widening:
+  - `Float` -> `Double`
+  - `Byte`, `Short` or `Int` -> `Double`
+- Date widening:
+  - `Date` -> `Timestamp without timezone`
+- Decimal widening - `p` and `s` denote the decimal precision and scale respectively.
+  - `Decimal(p, s)` -> `Decimal(p + k1, s + k2)` where `k1 >= k2 >= 0`.
+  - `Byte`, `Short` or `Int` -> `Decimal(10 + k1, k2)` where `k1 >= k2 >= 0`.
+  - `Long` -> `Decimal(20 + k1, k2)` where `k1 >= k2 >= 0`.
```
scovich, line 1405: I don't remember seeing this `10 + k1` bit before? 
Whenever it was added, I think it invalidated the `k1 >= k2` inequality? 
(maybe just need `10 + k1 >= k2 >= 0`?)

johanl-db: You do need `k1 >= k2` here, otherwise `Decimal(10, 5)` would be an allowed promotion, even though it can't store all possible `int` values: 100000 -> 100000.00000 doesn't fit already

A slightly different way of saying it is that `int`->`Decimal(10,0)` is a supported type widening. Then you can further widen to larger decimals `Decimal(p, s)` -> `Decimal(p + k1, s + k2)`

scovich: Ah, I didn't catch that detail before, thanks for clarifying.

scovich (approved):

PROTOCOL.md
```diff
@@ -1387,6 +1387,130 @@ Furthermore, when attempting timestamp-based time travel where table state must
...
+- Writers must preserve the `delta.typeChanges` field in the metadata fields in the schema when the table schema is updated.
+- Writers may remove the `delta.typeChanges` metadata in the table schema if all data files use the same column and field types as the table schema. 
+
+When Type Widening is enabled (when the table property `delta.enableTypeWidening` is set to `true`), then:
+- Writers should allow updating the table schema to apply a supported type change to a column, struct field, map key/value or array element.
+
+## Reader Requirements for Type Widening
+When Type Widening is supported (when the `readerFeatures` field of a table's `protocol` action contains `typeWidening`), then:
+- Readers must allow reading data files written before the table underwent any supported type change, and must convert such values to the current, wider type.
+- Readers must validate that they support all type changes in the `delta.typeChanges` field in the table schema for the table version they are reading and fail when finding any unsupported type change.
```
tdas, line 1513: what is the impact on other table features that have cross table feature dependencies. For example, IcebergCompatV1/V2. Since Iceberg's Type widening is only going to be in IcebergV3+, Delta's IcebergCompatV1/V2 will not be compatible to type widening feature. I think we should update their protocols to clarify this. 

what do you think @scovich @lzlfred ?

tdas (changes requested):
my main comment is about iceberg compatibility

PROTOCOL.md
johanl-db: For context, what we're currently doing in Delta spark is:
1. Block applying type changes unsupported by Iceberg if Uniform with Iceberg compat is enabled.
2. Block enabling Uniform with Iceberg compat if type changes not supported by Iceberg have been applied.

Users can always DROP the type widening table feature to make the table compatible with Iceberg again, at the cost of having to rewrite the data.

This would translate into the following requirements:
```
## Writer Requirements for Type Widening
- Writers must reject applying type changes not supported by [Iceberg V2](https://iceberg.apache.org/spec/#schema-evolution)
  when either the `IcebergCompatV1` or `IcebergCompatV2` table feature is supported:
  - `Byte`, `Short` or `Int` -> `Double`
  - `Date`  -> `Timestamp without timezone`
  - Decimal scale increase
  - `Byte`, `Short`, `Int` or `Long` -> `Decimal`
```

```
## Writer Requirements for IcebergCompatV1 (/IcebergCompatV2)
- If Type Widening is supported, require that no type changes not supported by
  [Iceberg V2](https://iceberg.apache.org/spec/#schema-evolution) were applied on the table,
  based on the [Type Change Metadata](https://github.com/delta-io/delta/pull/4094/files#diff-62d56fa1c25dda8d49ccdbf9ea31df4576f84625706d54613dff0acc8f974d97R1415)
  recorded in the table schema.
```

tdas: this looks better. but i would like iceberg compat experts to sign off on this. @fred?

furthermore, should we start writing down in the protocol about how to drop a feature? what do writers need to do drop a feature?

lzlfred: lgtm.

johanl-db: I'll update the specs to call out the cross-requirements between type widening and uniform + add requirements for dropping the type widening table feature

johanl-db:
I updated the spec to add:
- cross-requirements with Iceberg Compat V1/V2 for type changes not supported by Iceberg
- Requirements when dropping type widening

@tdas, @lzlfred, @scovich PTAL

PROTOCOL.md
```diff
@@ -1022,6 +1022,7 @@ When supported and active, writers must:
 - Block replacing partitioned tables with a differently-named partition spec
   - e.g. replacing a table partitioned by `part_a INT` with partition spec `part_b INT` must be blocked
   - e.g. replacing a table partitioned by `part_a INT` with partition spec `part_a LONG` is allowed
+- When the [Type Widening](#type-widening) table feature is supported, require that no type changes not supported by [Iceberg V2](https://iceberg.apache.org/spec/#schema-evolution) were applied on the table, based on the [Type Change Metadata](#type-change-metadata) recorded in the table schema.
```
lzlfred, line 1025: nit: double no is a little confusing. how about

require that all type changes applied on the table are supported by [Iceberg V2](https://iceberg.apache.org/spec/#schema-evolution)

lzlfred (approved):

tdas (approved):
LGTM! Thank you for doing this! This is a great addition to the Delta protocol.

scovich (approved):
Good catch about the interaction w/ Iceberg compat. LGTM now.

PR #4096, opened 2025-01-28
[VARIANT] Accept `variantType` RFC

richardc-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

moves the variant type RFC into the accepted folder and inlines it into `PROTOCOL.md`.

Variant has been used in production systems for many months now and has also been added as a logical type in parquet [here](https://github.com/apache/parquet-format/commit/dff0b3e6f02ed28e6b0753e921d53de53e63f506). Additionally, Spark has a robust implementation of the variant type 

## How was this patch tested?

manually tested that links work

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -62,6 +62,11 @@
   - [Reader Requirements for Vacuum Protocol Check](#reader-requirements-for-vacuum-protocol-check)
 - [Clustered Table](#clustered-table)
   - [Writer Requirements for Clustered Table](#writer-requirements-for-clustered-table)
+- [Variant Data Type](#variant-data-type)
+  - [Variant data in Parquet](#variant-data-in-parquet)
+  - [Writer Requirements for Variant Type](#writer-requirements-for-variant-type)
+  - [Reader Requirements for Variant Data Type](#reader-requirements-for-variant-data-type)
+  - [Compatibility with other Delta Features](#compatibility-with-other-delta-features)
 - [Requirements for Writers](#requirements-for-writers)
   - [Creation of New Log Entries](#creation-of-new-log-entries)
   - [Consistency Between Table Metadata and Data Files](#consistency-between-table-metadata-and-data-files)
@@ -100,6 +105,7 @@
     - [Struct Field](#struct-field)
     - [Array Type](#array-type)
     - [Map Type](#map-type)
+    - [Variant Type](#variant-type)
     - [Column Metadata](#column-metadata)
     - [Example](#example)
   - [Checkpoint Schema](#checkpoint-schema)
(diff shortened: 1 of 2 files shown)


PROTOCOL.md
```diff
@@ -1408,13 +1408,14 @@ metadata | binary | The binary-encoded Variant metadata, as described in [Varian
 
 The parquet struct must include the two struct fields `value` and `metadata`.
 Supported writers must write the two binary fields, and supported readers must read the two binary fields.
-Struct fields which start with `_` (underscore) can be safely ignored.
+
+Variant shredding will be introduced in a separate `variantShredding` table feature.
```
scovich, line 1412:
```suggestion
Variant shredding will be introduced later, as a separate `variantShredding` table feature.
```
(again below)

PROTOCOL.md
```diff
@@ -1408,13 +1408,14 @@ metadata | binary | The binary-encoded Variant metadata, as described in [Varian
...
 
 ## Writer Requirements for Variant Data Type
 
 When Variant type is supported (`writerFeatures` field of a table's `protocol` action contains `variantType`), writers:
 - must write a column of type `variant` to parquet as a struct containing the fields `value` and `metadata` and storing values that conform to the [Variant binary encoding specification](https://github.com/apache/spark/blob/master/common/variant/README.md)
-- must not write additional, non-ignorable parquet struct fields. Writing additional struct fields with names starting with `_` (underscore) is allowed.
+- must not write additional parquet struct fields.
```
scovich, line 1418: Is there a particular reason we need to forbid extra fields in this specific case?

Normally the Delta spec just says readers should ignore any unknown fields they might encounter, because the lack of a table feature to protect such fields means they do not affect correctness.

PROTOCOL.md
```diff
@@ -1353,6 +1359,86 @@ The example above converts `configuration` field into JSON format, including esc
 }
 ```
 
+
+# Variant Data Type
+
+This feature enables support for the Variant data type, for storing semi-structured data.
+The schema serialization method is described in [Schema Serialization Format](#schema-serialization-format).
```
scovich, line 1366: It seems a bit odd to specifically mention the schema serialization format here? AFAIK it didn't change with this table feature?

Meanwhile, it seems like we should specifically state that the new data type is called `variant`, rather than relying on the example below to show it?

scovich: Ah -- the schema serialization section adds a new entry for variant, that explains the type name is `variant`.

PROTOCOL.md
```diff
@@ -1353,6 +1359,86 @@ The example above converts `configuration` field into JSON format, including esc
 }
 ```
 
+
+# Variant Data Type
+
+This feature enables support for the Variant data type, for storing semi-structured data.
```
scovich, line 1365: Related to the confusion below:
```suggestion
This feature enables support for the `variant` data type, which stores semi-structured data.
```
(hopefully makes clear what the data type is called, without needing a lot of extra text)

PROTOCOL.md
```diff
@@ -1353,6 +1359,86 @@ The example above converts `configuration` field into JSON format, including esc
...
+    "nullable" : true,
+    "metadata" : { }
+  }, {
+    "name" : "variant_array",
+    "type" : {
+      "type" : "array",
+      "elementType" : {
+        "type" : "variant"
+      },
+      "containsNull" : false
```
scovich, line 1389: According to the [variant spec](https://github.com/apache/parquet-format/blob/master/VariantEncoding.md#encoding-types), a variant column could store a JSON `null` value even if the variant field is non-nullable as in this example. 

How is that handled? Does the Delta spec need to say anything about it or is that the query engine's problem, independent of Delta?

PROTOCOL.md
```diff
@@ -1353,6 +1359,86 @@ The example above converts `configuration` field into JSON format, including esc
...
+
+Struct field name | Parquet primitive type | Description
+-|-|-
+value | binary | The binary-encoded Variant value, as described in [Variant binary encoding](https://github.com/apache/spark/blob/master/common/variant/README.md)
+metadata | binary | The binary-encoded Variant metadata, as described in [Variant binary encoding](https://github.com/apache/spark/blob/master/common/variant/README.md)
+
+The parquet struct must include the two struct fields `value` and `metadata`.
+Supported writers must write the two binary fields, and supported readers must read the two binary fields.
+
+Variant shredding will be introduced in a separate `variantShredding` table feature.
```
scovich, line 1412: Should this reference to variant shredding link to the parquet variant shredding spec? Or is that overkill?

scovich (commented):
Is it safe to finalize this RFC when the [parquet spec](https://github.com/apache/parquet-format/blob/master/VariantEncoding.md) is not finalized yet?
> Important
> This specification is still under active development, and has not been formally adopted.

I guess this RFC actually references the _spark_ variant spec, not the parquet variant spec -- even tho it has a whole section about "Variant data in Parquet"?

Do we need commentary in the Delta spec about it relates to those two different specifications?

PROTOCOL.md
richardc-db: I don't think theres any particular reason, I think we can follow other features, i.e. modify this to say that readers can ignore fields that aren't `metadata` or `value`.

@gene-db or @cashmand do you see any issues with this? I figure this should be ok - the shredding table feature later will specify the "important" columns for shredding

PROTOCOL.md
richardc-db: hmm, yeah I think this is less a storage layer issue and more of an engine issue. That should be addressed in the actual binary layout of variant (which is considered in the parquet spec)

PROTOCOL.md
richardc-db: I think it makes sense to keep it separate for now because i imagine the parquet variant shredding spec may contain more information than is necessary for delta (i.e. the parquet binary format currently contains information for the binary encoding, which we don't include here).

I'd let @gene-db make the call here, though. Gene, do you have an opinion?

PROTOCOL.md
richardc-db: @cashmad, in this case, the `VariantShredding` table feature is enabled on the table though, right? So a shredding-unaware reader won't be able to read the table in the first place.

maybe i'm missing something...

cashmand: That's true. Is it possible that a customer could create a delta table using external parquet files, and we wouldn't know that we nee the shredding feature? It seems safer to me for a reader to fail.

scovich: Does the spec have to require that tho? Seems like an implementation detail, how paranoid to be in validating various aspects of the spec? An implementation could still choose to warn/error if it finds obvious shredding-related columns in a parquet file when the shredding feature is not supported?

scovich: A similar situation could arise if an engine produced deletion vectors when DV feature is not supported. 

Is that a scenario we worry about?

scovich: Actually, we should be fine. The [parquet variant spec](https://github.com/apache/parquet-format/blob/master/VariantEncoding.md#variant-in-parquet) says:
> Both fields `value` and `metadata` are of type binary, and cannot be null

... while the [parquet variant shredding spec](https://github.com/apache/parquet-format/blob/master/VariantShredding.md) says:
> At a high level, we replace the `value` field of the Variant Parquet group with one or more fields called `object`, `array`, `typed_value`, and `variant_value`. These represent a fixed schema suitable for constructing the full Variant value for each row.

Result: If we did ever encounter a shredded parquet field for a table that doesn't support shredding, the reader should blow up because variant shredding forbids the `value` field to exist (**) while binary _requires_ it to exist.

It's a separate question what should happen if the parquet reader _does_ understand shredding and simply chooses to unshred the data if that's what the Delta client asks for?

(**) The current shredding spec doesn't come right out and say the the `value` field is _forbidden_, only that it is "replaced". Maybe stronger wording would be helpful there? Or maybe it's clear enough overall?

Aside: 

The [spark shredding spec](https://github.com/apache/spark/tree/master/common/variant) does _not_ say the `value` and `metadata` fields are required and/or must be non-NULL. That seems bad?

The [spark variant shredding spec](https://github.com/apache/spark/blob/master/common/variant/shredding.md) also seems to have a bug, claiming that the `metadata` column is not present under shredding:
> At a high level, we replace the `value` and `metadata` of the Variant Parquet group with one or more fields called `object`, `array`, `typed_value` and `untyped_value`. These represent a fixed schema suitable for constructing the full Variant value for each row.

scovich: Actually -- 

Both the [spark](https://github.com/apache/spark/blob/master/common/variant/shredding.md#backward-and-forward-compatibility) and [parquet](https://github.com/apache/parquet-format/blob/master/VariantShredding.md#backward-and-forward-compatibility) shredding specs specifically address compatibility of extra fields:

> Any fields in the same group as `typed_value`/`variant_value` that start with _ (underscore) can be ignored. This is intended to allow future backwards-compatible extensions. ... Any extra field names that do not start with an underscore should be assumed to be backwards incompatible, and readers should fail when reading such a schema.

Neither top-level binary variant spec has that provision for `value`, but the intent seems clear enough and so Delta should probably follow the same convention.

cashmand: @scovich There's a large PR to change the shredding spec that is removing both of these: https://github.com/apache/parquet-format/pull/461.

The spec has simplified the fields to rename `untyped_value` to just `value`, in order to make unshredded variant a special case of shredded variant; the PR also removes the wording about underscore.

@bart-samwel's suggestion to specifically call out `typed_value` sounds reasonable to me.

scovich: Agree -- if we know of a common failure mode where invalid data could be silently accepted, it's probably worth calling that out here. Especially if `value` no longer disappears to make it obvious when shredding is in use.

Alternatively -- should we take it up a level of abstraction? Require the Delta client to check for shredded files, and to reject them unless the `variantShredding` feature is _also_ enabled?

cashmand: > Alternatively -- should we take it up a level of abstraction? Require the Delta client to check for shredded files, and to reject them unless the variantShredding feature is also enabled?

I'll leave that to the Delta experts. I don't know how much of a burden this will impose on clients, or if there's other precedent for inspecting the Parquet footer schema for validation.

PROTOCOL.md
scovich: Maybe a better way to put it -- should we give some indication here of what "shredding" even is? (a hyperlink to the spec being perhaps the least intrusive approach)?

PROTOCOL.md
gene-db: I think the simplest way to clarify this spec is to call out the `typed_value` field.

As for checking for shredded files, that would require opening every file footer. That is too high of an overhead for this requirement, right?

PROTOCOL.md
gene-db: I think it would make sense to link to the parquet shredding spec.
```suggestion
[Variant shredding](https://github.com/apache/parquet-format/blob/master/VariantShredding.md) will be introduced in a separate `variantShredding` table feature.
```

With the link, people can refer to it to get more details about the shredding, but we don't have to put those details in the text here.

PROTOCOL.md
richardc-db: makes sense - i updated the wording to particularly callout the `typed_value` field in light of shredding.

"- must not write a parquet struct field named `typed_value` to avoid confusion with the field required by [Variant shredding](https://github.com/apache/parquet-format/blob/master/VariantShredding.md) with the same name."

I chose `must` rather than `should` because I figure its safer to have more strong wording here. But lmk if anyone thinks there should be any tweaks!

scovich (approved):
LGTM

PR #4097, opened 2025-01-28
[Kernel] Minor refactor to DeltaLogActionUtils; add CloseableIterator takeWhile and other helpful methods

scottsand-db:
- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

This PR does the following:
- changes our DeltaLogActionUtils `listDeltaLogFiles` method to return an iterator. This will eventually let it be used in a followup PR to further consolidate and clean up code in DeltaHistoryManager (which lists the delta log and expects an iterator)
- adds CloseabelIterator::takeWhile, breakableFilter, and toInMemoryList

## How was this patch tested?

- New UTs

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaLogActionUtils.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaLogActionUtils.java
@@ -35,11 +35,13 @@ import io.delta.kernel.internal.util.FileNames;
 import io.delta.kernel.internal.util.FileNames.DeltaLogFileType;
 import io.delta.kernel.types.*;
 import io.delta.kernel.utils.CloseableIterator;
+import io.delta.kernel.utils.CloseableIterator.BreakableFilterResult;
 import io.delta.kernel.utils.FileStatus;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.util.*;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -103,13 +105,14 @@ public class DeltaLogActionUtils {
 
     // Get any available commit files within the version range
     final List<FileStatus> commitFiles =
-        listDeltaLogFiles(
-            engine,
-            Collections.singleton(DeltaLogFileType.COMMIT),
-            tablePath,
-            startVersion,
-            Optional.of(endVersion),
-            false /* mustBeRecreatable */);
+        listDeltaLogFilesAsIter(
+                engine,
+                Collections.singleton(DeltaLogFileType.COMMIT),
+                tablePath,
+                startVersion,
+                Optional.of(endVersion),
+                false /* mustBeRecreatable */)
+            .toInMemoryList();
 
     // There are no available commit files within the version range.
     // This can be due to (1) an empty directory, (2) no valid delta files in the directory,
@@ -181,14 +184,14 @@ public class DeltaLogActionUtils {
   }
 
   /**
-   * Returns the list of files of type $fileTypes in the _delta_log directory of the given
-   * $tablePath, in increasing order from $startVersion to the optional $endVersion.
+   * Returns a {@link CloseableIterator} of files of type $fileTypes in the _delta_log directory of
+   * the given $tablePath, in increasing order from $startVersion to the optional $endVersion.
    *
    * @throws TableNotFoundException if the table or its _delta_log does not exist
    * @throws KernelException if mustBeRecreatable is true, endVersionOpt is present, and the
    *     _delta_log history has been truncated so that we cannot load the desired end version
    */
-  public static List<FileStatus> listDeltaLogFiles(
+  public static CloseableIterator<FileStatus> listDeltaLogFilesAsIter(
       Engine engine,
       Set<DeltaLogFileType> fileTypes,
       Path tablePath,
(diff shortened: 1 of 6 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaLogActionUtils.java
```diff
@@ -215,69 +218,59 @@ public static List<FileStatus> listDeltaLogFiles(
         startVersion,
         endVersionOpt);
 
-    final List<FileStatus> output = new ArrayList<>();
```
scottsand-db, line 218: verbatim translate this logic to use the breakableFilter API instead

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaLogActionUtils.java
```diff
@@ -215,69 +218,59 @@ public static List<FileStatus> listDeltaLogFiles(
...
+        .breakableFilter(
+            fs -> {
+              if (fileTypes.contains(DeltaLogFileType.COMMIT)
+                  && FileNames.isCommitFile(getName(fs.getPath()))) {
+                // Here, we do nothing (we will consume this file).
+              } else if (fileTypes.contains(DeltaLogFileType.CHECKPOINT)
+                  && FileNames.isCheckpointFile(getName(fs.getPath()))
+                  && fs.getSize() > 0) {
+                // Checkpoint files of 0 size are invalid but may be ignored silently when read,
+                // hence we
```
scottsand-db, line 234: fix this

huan233usc (approved):
LGTM with the under radar doc nit got fix(https://github.com/delta-io/delta/pull/4097/files#r1933026458)

kernel/kernel-api/src/main/java/io/delta/kernel/utils/CloseableIterator.java
```diff
@@ -36,6 +39,31 @@
 @Evolving
 public interface CloseableIterator<T> extends Iterator<T>, Closeable {
 
+  /**
+   * Represents the result of applying the filter condition in the {@link
+   * #breakableFilter(Function)} method of a {@link CloseableIterator}. This enum determines how
+   * each element in the iterator should be handled.
+   */
+  enum BreakableFilterResult {
```
allisonport-db, line 47: Is this a common paradigm? It seems like `breakableFilter` could also be a combo of filter + takeWhile. But I see those are implemented using this so just curious on your reasoning

kernel/kernel-api/src/test/scala/io/delta/kernel/CloseableIteratorSuite.scala
```diff
@@ -0,0 +1,78 @@
...
+  private def normalDataIter = toCloseableIter(Seq(1, 2, 3, 4, 5))
+
+  private def throwingDataIter = toCloseableIter(Seq(1, 2, 3, 4, 5)).map { x =>
+    if (x > 4) {
+      throw new RuntimeException("Underlying data evaluated at element > 4")
+    }
+    x
+  }
+
+  test("CloseableIterator::filter -- returns filtered result") {
```
allisonport-db, line 45: add a case where nothing is filtered and we reach the end of the iterator? I think none of these tests reach the end

allisonport-db: it can be for any of the methods

allisonport-db (commented):
just 1 tiny comment + a question

kernel/kernel-api/src/main/java/io/delta/kernel/utils/CloseableIterator.java
scottsand-db: > breakableFilter could also be a combo of filter + takeWhile

Are you saying that we can implement breakableFilter by using a combination of both filter + takeWhile?

One of my tests has an example of breakableFilter: we want to: INCLUDE 1, EXCLUDE 2, INCLUDE 3, and BREAK at 4. After thinking about this for > 30 seconds, maybe there is a way to implement this using filter + takeWhile?

But I think the semantics I lay out here are simpler than using `.filter(....).takeWhile(...)`

kernel/kernel-api/src/test/scala/io/delta/kernel/CloseableIteratorSuite.scala
scottsand-db: Doesn't this case here reach that? This is just a normal filter -- we will iterate over all of the elements in the list

allisonport-db: sorry they all reach the end; the last element is never returned though; just seems slightly different conceptually

kernel/kernel-api/src/main/java/io/delta/kernel/utils/CloseableIterator.java
allisonport-db:
```
    val result = throwingDataIter.breakableFilter { x =>
      if (x <= 1 || x == 3) {
        BreakableFilterResult.INCLUDE
      } else if (x == 2) {
        BreakableFilterResult.EXCLUDE
      } else if (x == 4) {
        BreakableFilterResult.BREAK
      } else {
        throw new RuntimeException("This should never be reached")
      }
    }
```
Wouldn't the test case just be? They seem equivalent to me

```
iter.filter( x => x != 2).takeWhile(x != 4)
```

allisonport-db: Equivalently the snapshot manager code would probably be
```
    // Must be final to be used in lambda
    final AtomicBoolean hasReturnedAnElement = new AtomicBoolean(false);

    return listLogDir(engine, tablePath, startVersion)
        .filter(
            fs -> {
              if (fileTypes.contains(DeltaLogFileType.COMMIT)
                  && FileNames.isCommitFile(getName(fs.getPath()))) {
                return true
              } else if (fileTypes.contains(DeltaLogFileType.CHECKPOINT)
                  && FileNames.isCheckpointFile(getName(fs.getPath()))
                  && fs.getSize() > 0) {
                return true
              } else {
                return false
              }
           }
      .takeWhile(
          fs -> {
              final long fileVersion = FileNames.getFileVersion(new Path(fs.getPath()));

              if (fileVersion < startVersion) {
                throw new RuntimeException(
                    String.format(
                        "Listing files in %s with startVersion %s yet found file %s.",
                        logPath, startVersion, fs.getPath()));
              }

              if (endVersionOpt.isPresent()) {
                final long endVersion = endVersionOpt.get();

                if (fileVersion > endVersion) {
                  if (mustBeRecreatable && !hasReturnedAnElement.get()) {
                    final long earliestVersion =
                        DeltaHistoryManager.getEarliestRecreatableCommit(engine, logPath);
                    throw DeltaErrors.versionBeforeFirstAvailableCommit(
                        tablePath.toString(), endVersion, earliestVersion);
                  } else {
                    logger.debug(
                        "Stopping listing; found file {} with version greater than endVersion {}",
                        fs.getPath(),
                        endVersion);
                    return false;
                  }
                }
              }

              hasReturnedAnElement.set(true);

              return true;
            });
```

allisonport-db: I actually kind of like it; since it nicely delineates the filter mechanism versus the terminate the loop (takeWhile) mechanism. But they honestly are super similar so I don't feel too strongly about it.

scottsand-db: `iter.filter( x => x != 2).takeWhile(x != 4)` --> ah, I understand now, thanks for showing that.

> I actually kind of like it; since it nicely delineates the filter mechanism versus the terminate the loop (takeWhile) mechanism. 

I actually still prefer the current version. When they are separate, I need to understand 4 return values: 2 for the filter, and 2 for the takeWhile. It also introduces some coupling -- the `takeWhile` logic is inherintly tied to the `filter` logic -- e.g. it would we weird to filter(x -> x < 10) and then takeWhile(x > 20). IMO when it's all together, I think it's easier to parse the logic.

kernel/kernel-api/src/test/scala/io/delta/kernel/CloseableIteratorSuite.scala
scottsand-db: Ah, now I understand. For sure, will add.

kernel/kernel-api/src/main/java/io/delta/kernel/utils/CloseableIterator.java
scottsand-db: The good news is that this CloseableIterator API doesn't preclude your suggestion. i.e. other areas in the code are free to do .filter and then .takeWhile if we decide later on that that is best there.

Overall -- I'm inclined to stick with what I have now. Happy to refactor it laterr -- but really I'd like to unblock @huan233usc 's PR.

allisonport-db: Sounds good

allisonport-db: I guess I just hesitate to add a new concept of this breakable filter when it can be distilled into two existing things. But I agree it's easy to understand so not a big deal

allisonport-db (approved):

PR #4105, opened 2025-01-30
[Kernel] Add better param validations to LogSegment constructor

scottsand-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Add better validations to LogSegment

## How was this patch tested?

New UTs.

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
@@ -21,6 +21,7 @@ import static java.util.Objects.requireNonNull;
 
 import io.delta.kernel.internal.fs.Path;
 import io.delta.kernel.internal.lang.Lazy;
+import io.delta.kernel.internal.lang.ListUtils;
 import io.delta.kernel.internal.util.FileNames;
 import io.delta.kernel.utils.FileStatus;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
```diff
@@ -62,6 +66,29 @@ public LogSegment(
       List<FileStatus> checkpoints,
       Optional<Long> checkpointVersionOpt,
       long lastCommitTimestamp) {
+    requireNonNull(logPath, "logPath is null");
+    requireNonNull(deltas, "deltas is null");
+    requireNonNull(checkpoints, "checkpoints is null");
+    requireNonNull(checkpointVersionOpt, "checkpointVersionOpt is null");
+    checkArgument(
+        checkpoints.isEmpty() == !checkpointVersionOpt.isPresent(),
+        "checkpoints and checkpointVersionOpt must either be both empty or both non-empty");
```
allisonport-db, line 75: it's not possible to hit this in the code (in theory) because we thrown an error during log segment building in this case right?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
```diff
@@ -62,6 +66,29 @@ public LogSegment(
       List<FileStatus> checkpoints,
       Optional<Long> checkpointVersionOpt,
```
allisonport-db, line 67: why do we take this as an argument? can't we find this from the checkpoint files provided?

I know it's already here, just got confused on what it even was, and now wondering why we even take it as an arg

allisonport-db (approved):
Changes LGTM just asked a question

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
scottsand-db: In hindsight, yeah, why don't I just remove it!

huan233usc (approved):

PR #4108, opened 2025-01-31
[Kernel] More LogSegment cleanup and refactor

scottsand-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

This PR does some misc improvements:

- turn public member values in LogSegment to be private; add proper getters
- add better validation
- add isComplete and toString methods
- add toString method in FileStatus

## How was this patch tested?

New UTs

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
@@ -32,19 +32,27 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 public class LogSegment {
-  public final Path logPath;
-  public final long version;
-  public final List<FileStatus> deltas;
-  public final List<FileStatus> checkpoints;
-  public final Optional<Long> checkpointVersionOpt;
-  public final long lastCommitTimestamp;
 
-  private final Lazy<List<FileStatus>> allFiles;
-  private final Lazy<List<FileStatus>> allFilesReversed;
+  //////////////////////////////////
+  // Static variables and methods //
+  //////////////////////////////////
 
   public static LogSegment empty(Path logPath) {
     return new LogSegment(logPath, -1, Collections.emptyList(), Collections.emptyList(), -1);
   }
+
+  //////////////////////////////////
+  // Member variables and methods //
+  //////////////////////////////////
+
+  private final Path logPath;
+  private final long version;
+  private final List<FileStatus> deltas;
+  private final List<FileStatus> checkpoints;
+  private final Optional<Long> checkpointVersionOpt;
+  private final long lastCommitTimestamp;
+  private final Lazy<List<FileStatus>> allFiles;
+  private final Lazy<List<FileStatus>> allFilesReversed;
 
   /**
    * Provides information around which files in the transaction log need to be read to create the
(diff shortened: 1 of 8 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
```diff
@@ -140,6 +178,49 @@ public LogSegment(
...
 
+  /////////////////
+  // Public APIs //
+  /////////////////
+
+  /**
+   * @return true if this LogSegment is complete and fully describes a Snapshot version. A partial
+   *     LogSegment is missing some information. We consider an empty LogSegment to be incomplete.
+   */
+  public boolean isComplete() {
```
scottsand-db, line 189: this will be helpful for CCv2

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
```diff
@@ -94,16 +115,31 @@ public LogSegment(
...
         checkArgument(
-            FileNames.deltaVersion(ListUtils.getLast(deltas).getPath()) == version,
+            ListUtils.getLast(deltaVersions) == version,
             "Last delta file version must equal the version of this LogSegment");
+
+        // Ensure the delta versions are contiguous
+        for (int i = 1; i < deltaVersions.size(); i++) {
+          if (deltaVersions.get(i) != deltaVersions.get(i - 1) + 1) {
+            throw new IllegalArgumentException(
+                String.format("Delta versions must be contiguous: %s", deltaVersions));
```
allisonport-db, line 140: Can we use the same check we use in `SnapshotManager`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
allisonport-db: Can you explain to me an example of an incomplete log segment? in the context of ccv2

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
scottsand-db: I considered that, yeah. My only intuition _not_ to use it was: this LogSegment will eventually be public API. So if the engine constructs and invalid LogSegment to pass in -- do we really want to throw an InvalidTableException, or would IllegalArgumentException be more appropriate? Open to your thoughts.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
scottsand-db: This will be the unbackfilled commits. e.g. Suppose the latest version is 17. Suppose 10.checkpoint and 11.json and 12.json have been backfilled. We will get back 13.json to 17.json as unbackfilled commits (inside of a LogSegment). This is an incomplete LogSegment (13.json to 17.json doesn't fully describe the table).

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
allisonport-db: yeah okay that makes sense to me. I think IllegalArgumentException is much better here!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
allisonport-db: Got it. Thanks

allisonport-db (approved):
LGTM

Issue #4045, opened 2025-01-14
[BUG][Spark] Auto Compaction trigger logic is not consistent with documentation


mwc360:
## Bug
The logic for when auto compaction is triggered does not work as documented: already compacted files (files that are >= minFileSize (or maxFileSize / 2) seem to be counted towards the minNumFiles for compaction to be triggered.

#### Which Delta project/connector is this regarding?
- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

### Describe the problem
Already compacted files (files that are >= minFileSize (or maxFileSize / 2) seem to be counted towards the minNumFiles for compaction to be triggered. This results in compactions running more frequently as the number of compacted files increases and approaches minNumFiles.

#### Steps to reproduce

```
# RUN ON CLUSTER w/ 2x8vCore Workers
spark.conf.set("spark.databricks.delta.autoCompact.minNumFiles", "50")
spark.conf.set("spark.databricks.delta.autoCompact.maxFileSize", "134217728b")

spark.sql(f"""
    CREATE TABLE dbo.ac_test
    TBLPROPERTIES ('delta.autoOptimize.autoCompact' = 'true')
""")

import pyspark.sql.functions as sf

for i in range(200):
    data = spark.range(1_000_000) \
            .withColumn("id", sf.monotonically_increasing_id()) \
            .withColumn("category", sf.concat(sf.lit("category_"), (sf.col("id") % 10))) \
            .withColumn("value1", sf.round(sf.rand() * (sf.rand() * 1000), 2)) \
            .withColumn("value2", sf.round(sf.rand() * (sf.rand() * 10000), 2)) \
            .withColumn("value3", sf.round(sf.rand() * (sf.rand() * 100000), 2)) \
            .withColumn("date1", sf.date_add(sf.lit("2022-01-01"), sf.round(sf.rand() * 1000, 0).cast("int"))) \
            .withColumn("date2", sf.date_add(sf.lit("2020-01-01"), sf.round(sf.rand() * 2000, 0).cast("int"))) \
            .withColumn("is_cancelled", (sf.col("id") % 3 != 0))

    data.write.mode('append').option("mergeSchema", "true").saveAsTable(f"dbo.ac_test")
```

#### Observed results
[...]

Issue #4043, opened 2025-01-13
[Feature Request][Spark] Auto Compaction shouldn't be trigged if compaction hasn't been run yet


mwc360:
## Feature request
The OSS implementation will run compaction when auto-compaction is enabled if compaction hasn't been run yet. I.e. running a CTAS w/ the table property enabled will perform compaction after the write even if the small file count doesn't meet minNumFiles.

https://github.com/delta-io/delta/blob/a9208853bdf8836af1e0897cb800964f5f732589/spark/src/main/scala/org/apache/spark/sql/delta/stats/AutoCompactPartitionStats.scala#L71-L78

Auto compaction in Databricks does not perform this unnecessary initial compaction operation. It should only be evaluated based on the presence of small files which meet or exceed the minNumFiles.

#### Which Delta project/connector is this regarding?
- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

### Motivation
Improve performance of tables that get created with auto compaction enabled.

### Willingness to contribute

- [ ] Yes. I can contribute this feature independently.
- [x] Yes. I would be willing to contribute this feature with guidance from the Delta Lake community.
- [ ] No. I cannot contribute this feature at this time.

mwc360:
@nicklan - what was the intent behind always triggering auto compact when the feature is enabled? Auto compact in DBX doesn't work this same way.

i.e. running a CTAS w/ the autoCompact property runs optimize right after the CTAS completes, even when the count of small files is well below minNumFiles.

mwc360:
closing as this is a result of [modifiedPartitionsOnly ](https://github.com/delta-io/delta/blob/92a8a2283d20a448b651ad199251946f3504e83f/spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala#L308) being enabled by default. I've proposed disabling this by default in this PR: https://github.com/delta-io/delta/pull/4109

PR #4109, opened 2025-01-31
[SPARK] Set autoCompact.modifiedPartitionsOnly.enabled to false as default

mwc360:
When True is the default, the behavior of AC doesn't align with the DBR implementation and results in too frequent compactions being run as minNumFiles isn't respected.

Example of odd result of AC in this issue: https://github.com/delta-io/delta/issues/4045
and resolves the odd behavior noted here: https://github.com/delta-io/delta/issues/4043

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Aligns behavior of AC with Databricks DBR implementation where it appears this is not enabled (also not exposed as a config option).

## How was this patch tested?
Ran 200 iterations of merging data into a table to confirm the expected behavior of AC in managing small files while respecting the minNumFiles threshold.

No as this session config isn't documented in the Delta.io docs


mwc360:
After more testing this config resolved the odd behavior in https://github.com/delta-io/delta/issues/4043 but did not fix the core issue described in https://github.com/delta-io/delta/issues/4045

PR #4112, opened 2025-02-01
[Kernel]Remove two file IO in CRC loading by reusing log listing

huan233usc:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

Incremental CRC loading for P&M was added in https://github.com/delta-io/delta/pull/4077.  

For snapshot N, we try to read N.crc, if N.crc missing, we list CRC files and find the one up to N-100. These two file listing operation could be actually avoided, because we have already listed files under _delta_log folder.

This PR collects latest CRC in the first file listing and remove two unnecessary IO.

## How was this patch tested?

LogReplay test

No


kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaLogActionUtils.java
```diff
@@ -219,7 +219,7 @@ public static CloseableIterator<FileStatus> listDeltaLogFilesAsIter(
         endVersionOpt);
 
     // Must be final to be used in lambda
-    final AtomicBoolean hasReturnedAnElement = new AtomicBoolean(false);
+    final AtomicBoolean hasReturnedlogOrCheckPoint = new AtomicBoolean(false);
```
scottsand-db, line 222: `hasReturnedCommitOrCheckpoint`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaLogActionUtils.java
```diff
@@ -266,7 +269,11 @@ public static CloseableIterator<FileStatus> listDeltaLogFilesAsIter(
                 }
               }
 
-              hasReturnedAnElement.set(true);
+              // Only log and checkpoint could use to construct table state.
+              if (FileNames.isCommitFile(getName(fs.getPath()))
```
scottsand-db, line 273: I just realized this. `isCommitFile` takes in a String path. so does `isCheckpointFile`. Want to update all of these occurences?

It can just be : `FileNames.isCommitFile(fs.getPath())`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaLogActionUtils.java
```diff
@@ -266,7 +269,11 @@ public static CloseableIterator<FileStatus> listDeltaLogFilesAsIter(
                 }
               }
 
-              hasReturnedAnElement.set(true);
+              // Only log and checkpoint could use to construct table state.
```
scottsand-db, line 272: You can delete this comment line

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaLogActionUtils.java
scottsand-db: You can add a comment here. Something like: `This variable is used to help determine if we should throw an error if the table history is not reconstructable. Only commit and checkpoint files are applicable.`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -319,10 +323,16 @@ public LogSegment getLogSegmentForVersion(Engine engine, Optional<Long> versionT
             listedFileStatuses,
```
scottsand-db, line 323: Errr ... we should just totally re-do this logic. For example, `listedCheckpointAndDeltaFileStatuses` is an incorrect variable name now since this list now also contains CRC files.

Something like this:

```java
Map<DeltaLogFileType, List<FileStatus>> partitionedFiles = files.stream()
            .collect(Collectors.groupingBy(
                file -> {
                    String name = file.getPath().getName();
                    if (FileNames.isCommitFile(name)) {
                        return DeltaLogFileType.COMMIT;
                    } else if (FileNames.isCheckpointFile(name)) {
                        return DeltaLogFileType.CHECKPOINT;
                    } else if (FileNames.isCrcFile(name)) {
                        return DeltaLogFileType.CRC;
                    } else {
                        return DeltaLogFileType.OTHER;
                    }
                },
                LinkedHashMap::new, // Ensure order is maintained
                Collectors.toList()
            ));

List<FileStatus> commitFiles = partitionedFiles.getOrDefault(DeltaLogFileType.COMMIT, Collections.emptyList());

List<FileStatus> checkpointFiles = partitionedFiles.getOrDefault(DeltaLogFileType.CHECKPOINT, Collections.emptyList());

List<FileStatus> crcFiles = partitionedFiles.getOrDefault(DeltaLogFileType.CRC, Collections.emptyList());
```

I just

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -493,6 +503,15 @@ public LogSegment getLogSegmentForVersion(Engine engine, Optional<Long> versionT
         newVersion,
         deltasAfterCheckpoint,
         latestCompleteCheckpointFileStatuses,
+        listedChecksumFileStatues.stream()
```
scottsand-db, line 506: wouldn't it just be the last one?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/LogReplayEngineMetricsSuite.scala
```diff
@@ -412,23 +412,23 @@ class LogReplayEngineMetricsSuite extends AnyFunSuite with TestUtils {
...
-          expJsonVersionsRead = Seq(11),
-          expParquetVersionsRead = Seq(10),
-          expParquetReadSetSizes = Seq(1),
-          expChecksumReadSet = Seq(11)
+        // 10.checkpoint found, so use it and combined with 11.crc
+        expJsonVersionsRead = Seq(11),
+        expParquetVersionsRead = Seq(10),
+        expParquetReadSetSizes = Seq(1),
+        expChecksumReadSet = Nil
         )
```
scottsand-db, line 432: I think the indentation here is all messaged up (not your fault). I think the bracket here needs to be 2 spaces to the left.

I also think the next `loadPandMCheckMetrics` is also wrong -- that entire block needs to go to the left.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
```diff
@@ -263,4 +269,8 @@ private String formatList(List<FileStatus> list) {
     return "\n    "
         + list.stream().map(FileStatus::toString).collect(Collectors.joining(",\n    "));
   }
+
+  public Optional<FileStatus> getLatestChecksum() {
```
scottsand-db, line 273: nit: this should be next to the other gettings, and it should be inserted to be in the same order as the parameter in the constructor

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -414,11 +415,15 @@ private Map<String, DomainMetadata> loadDomainMetadataMap(Engine engine) {
                 snapshotHint.map(SnapshotHint::getVersion).orElse(-1L) + 1,
                 logSegment.getCheckpointVersionOpt().orElse(0L),
                 // Only find the CRC within 100 versions.
-                snapshotVersion - 100,
-                0L));
```
scottsand-db, line 418: why remove the 0L? seems like a good safeguard?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
scottsand-db, line 273: and I think these needs method docs? This checksum might not be for the `version` that this LogSegment represents, right?

also ... suppose we have 10.checkpoint and 11.json and 12.json ...

what are the possible ranges of the version for which this CRC could be for? Could we have 8.crc? what are the guarantees of this version range?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -414,11 +415,15 @@ private Map<String, DomainMetadata> loadDomainMetadataMap(Engine engine) {
...
+                snapshotVersion - 100));
```
scottsand-db, line 418: do we still need this `crcSearchLowerBound`? we are not even doing a LIST call now?

scottsand-db (changes requested):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
huan233usc: For logSegment, we just have 8.crc, downstream(LogReplay) decides use it or not.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
huan233usc: Renamed to CRC read bound

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
huan233usc: yes, updated

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
```diff
@@ -217,6 +221,35 @@ public Optional<Long> getCheckpointVersionOpt() {
     return checkpointVersionOpt;
   }
 
+  /**
+   * Returns the most recent checksum file encountered during log file listing, if available.
```
scottsand-db, line 225: Is one of the guarantees that the checksum verison here is <= the `version` of this `LogSegment`?

Can we add a `checkArgument` for that?

scottsand-db: And is it ever useful to have a .crc file that is lower than the latest complete checkpoint of this logsegment?

i.e. are we able to define more strict bounds here?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -314,15 +317,37 @@ public LogSegment getLogSegmentForVersion(Engine engine, Optional<Long> versionT
...
+                    LinkedHashMap::new, // Ensure order is maintained
+                    Collectors.toList()));
+
+    List<FileStatus> listedDeltaFileStatuses =
+        partitionedFiles.getOrDefault(DeltaLogFileType.COMMIT, Collections.emptyList());
+
+    List<FileStatus> listedCheckpointFileStatuses =
+        partitionedFiles.getOrDefault(DeltaLogFileType.CHECKPOINT, Collections.emptyList());
+
+    List<FileStatus> listedChecksumFileStatues =
```
scottsand-db, line 345: typo: `listedChecksumFileStatuses`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -314,15 +317,37 @@ public LogSegment getLogSegmentForVersion(Engine engine, Optional<Long> versionT
...
-            listedFileStatuses,
-            fileStatus -> FileNames.isCheckpointFile(new Path(fileStatus.getPath()).getName()));
-    final List<FileStatus> listedCheckpointFileStatuses = listedCheckpointAndDeltaFileStatuses._1;
-    final List<FileStatus> listedDeltaFileStatuses = listedCheckpointAndDeltaFileStatuses._2;
+    Map<DeltaLogFileType, List<FileStatus>> partitionedFiles =
+        listedFileStatuses.stream()
+            .collect(
+                Collectors.groupingBy(
+                    file -> {
+                      String fileName = file.getPath();
```
scottsand-db, line 325: Could you move this to a public static helper method in FileNames?

```java
public static DeltaLogFileType determineFileType(FileStatus file) {
    final String fileName = file.getPath().getName();
    
    if (isCommitFile(fileName)) {
        return DeltaLogFileType.COMMIT;
    } else if (isCheckpointFile(fileName)) {
        return DeltaLogFileType.CHECKPOINT;
    } else if (isChecksumFile(fileName)) {
        return DeltaLogFileType.CHECKSUM;
    } else {
        throw new IllegalStateException("Unexpected file type: " + fileName);
    }
}
```

scottsand-db (changes requested):
Looks great! Left only some minor comments!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -408,25 +431,40 @@ private Map<String, DomainMetadata> loadDomainMetadataMap(Engine engine) {
...
-                // if hint is not present, list from version 0.
+                // if hint is not present, read from version 0.
                 snapshotHint.map(SnapshotHint::getVersion).orElse(-1L) + 1,
                 logSegment.getCheckpointVersionOpt().orElse(0L),
-                // Only find the CRC within 100 versions.
+                // Only read the CRC within 100 versions.
                 snapshotVersion - 100,
                 0L));
+
+    // Inside the method where checksum is processed from LogSegment
```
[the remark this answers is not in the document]
huan233usc: Removed

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
huan233usc: Done

>And is it ever useful to have a .crc file that is lower than the latest complete checkpoint of this logsegment?

One interesting thing is file size histogram, but I am not aware of the current use case needed to load it from previous CRC now or in the short term. We could make logSegment.getLastSeenChecksum() only return crc older than checkpoint for now as It is not a one way door.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/snapshot/LogSegmentSuite.scala
```diff
@@ -40,74 +41,86 @@ class LogSegmentSuite extends AnyFunSuite with MockFileSystemClientUtils {
   }
 
   test("constructor -- valid case (non-empty)") {
```
allisonport-db, line 43: test the constructor validations for checksum in test cases here? seems like we do this for the other validations?

allisonport-db (commented):
LGTM this is super clean! 2 minor comments to address before merge

allisonport-db (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -408,25 +432,38 @@ private Map<String, DomainMetadata> loadDomainMetadataMap(Engine engine) {
       snapshotHint = Optional.empty();
     }
 
-    long crcSearchLowerBound =
+    long crcReadLowerBound =
         max(
             asList(
                 // Prefer reading hint over CRC, so start listing from hint's version + 1,
```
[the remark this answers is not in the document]
huan233usc: It is still true, in terms of P&M loading true, as reading from crc is 1 extra io.

Loading dm is not always happen in the lifecycle of snapshot, so I would rather defer the attempt to load current crc once again for dm.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -390,6 +395,25 @@ private Map<String, DomainMetadata> loadDomainMetadataMap(Engine engine) {
...
+   *         <li>First check if there's a lastSeenChecksum in the LogSegment and if its version is
+   *             appropriate (>= lower bound, <= requested version)
+   *         <li>If a suitable checksum exists, generate version hint directly from it for
+   *             subsequence call use that
+   *         <li>If no suitable checksum exists, subsequence call fall back to reading from log
+   *             files
+   *       </ol>
+   * </ol>
    */
   private Tuple2<Optional<SnapshotHint>, Optional<CRCInfo>>
```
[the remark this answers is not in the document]
huan233usc: I make this method only return Optional<SnapshotHint> and cache crcinfo when it is read, so we could get rid of tuple.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -390,6 +395,25 @@ private Map<String, DomainMetadata> loadDomainMetadataMap(Engine engine) {
   /**
    * Calculates the latest snapshot hint before or at the current snapshot version, returns the
    * CRCInfo if checksum file at the current version is read
```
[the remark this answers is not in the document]
huan233usc: We call ChecksumReader.getCRCInfo(engine, checksum) in this method, so a crc file might be read.

Note: logSegment only knows the last seem crc file's path and this method decides whether to open the file and load crc info based on the versionHint's version.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -408,25 +432,38 @@ private Map<String, DomainMetadata> loadDomainMetadataMap(Engine engine) {
...
+
     Optional<CRCInfo> crcInfoOpt =
-        ChecksumReader.getCRCInfo(
-            engine, logSegment.getLogPath(), snapshotVersion, crcSearchLowerBound);
+        logSegment
+            .getLastSeenChecksum()
+            .filter(
+                checksum -> {
+                  long checksumVersion = FileNames.getFileVersion(new Path(checksum.getPath()));
+                  if (checksumVersion != snapshotVersion) {
```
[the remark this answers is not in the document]
huan233usc: >we should probably remove snapshotVersion as an input param to this class? Because that is already contains in logsegment.version? wdyt?

Good find, sent https://github.com/delta-io/delta/pull/4464 to avoid expanding this PR's change

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -388,50 +387,66 @@ private Map<String, DomainMetadata> loadDomainMetadataMap(Engine engine) {
...
-        max(
-            asList(
-                // Prefer reading hint over CRC, so start listing from hint's version + 1,
-                // if hint is not present, list from version 0.
-                snapshotHint.map(SnapshotHint::getVersion).orElse(-1L) + 1,
-                logSegment.getCheckpointVersionOpt().orElse(0L),
-                // Only find the CRC within 100 versions.
-                snapshotVersion - 100,
-                0L));
+    // Prefer reading hint over CRC to save 1 io, so start listing from hint's version + 1,
```
scottsand-db, line 429: `listing` --> out of date comment, no listing here

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
```diff
@@ -161,6 +175,19 @@ public LogSegment(
     this.version = version;
     this.deltas = deltas;
     this.checkpoints = checkpoints;
+    this.lastSeenChecksum =
+        lastSeenChecksum.filter(
+            file -> {
+              long checksumVersion = FileNames.checksumVersion(new Path(file.getPath()));
+              logger.debug(
+                  "Found checksum file for version {} in "
+                      + "LogSegment version {} and checkpoint version {}",
```
scottsand-db, line 184: IMO we could just do a logger.debug(this.toString) at the bottom of this constructor and print all the info?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/FileNames.java
```diff
@@ -77,6 +79,20 @@ public enum DeltaLogFileType {
 
   public static final String SIDECAR_DIRECTORY = "_sidecars";
 
+  public static DeltaLogFileType determineFileType(FileStatus file) {
```
scottsand-db, line 82: Simple test in FileNamesSuite?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -388,50 +387,66 @@ private Map<String, DomainMetadata> loadDomainMetadataMap(Engine engine) {
...
+   * Calculates the latest snapshot hint before or at the current snapshot version, cache the
    * CRCInfo if checksum file at the current version is read
+   *
+   * <p>Strategy for finding and using CRC (checksum) files:
+   *
+   * <ol>
+   *   <li>During initial log directory listing, the SnapshotManager captures all files including
+   *       commits, checkpoints, and checksums.
+   *   <li>The most recent checksum file discovered is stored in the LogSegment as
+   *       "lastSeenChecksum", regardless of its version, however, logSegment will make sure
```
scottsand-db, line 399: can you delete the `regardless of its version` -- since that's not true -- the 2nd half of this sentence is good and accurate

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -388,50 +387,66 @@ private Map<String, DomainMetadata> loadDomainMetadataMap(Engine engine) {
...
+   *       last_check_point_version <= crc_version <= logSegment_version
+   *   <li>When loading table state information like Protocol, Metadata, or Domain Metadata:
```
scottsand-db, line 401: is the snapshot hint used for the DM?

I think there's 2 parts:
- load the CRC Info as needed to get the newest possible snapshotHint that is used during Log Replay for P&M loading
- use the CRC Info if its version is equal to the snapshot version to load the domain metadatas; else load the DMs from the delta log

Is my understanding correct?

huan233usc: >is the snapshot hint used for the DM?
No, snapshot hint is not persisting DM and I am a bit hesitate to make it so. Unlike P&M which are mandatory in loading a snapshot, DM are not.(We only load DM when we try to get a dm of a table)

>load the CRC Info as needed to get the newest possible snapshotHint that is used during Log Replay('s constructor) for P&M loading
Correct

>use the CRC Info if its version is equal to the snapshot version to load the domain metadatas; else load the DMs from the delta log (lazily, e.g. only when we try to get dm of a table at a certain snapshot)
Correct and is under construction -> https://github.com/delta-io/delta/pull/4438

scottsand-db: So is this the best location for this comment? I also think this comment is incomplete in that

```java
   *         <li>First check if there's a lastSeenChecksum in the LogSegment and if its version is
   *             appropriate (>= lower bound, <= requested version)
   *         <li>If a suitable checksum exists, generate version hint directly from it for
   *             subsequence call use that
   *         <li>If no suitable checksum exists, subsequence call fall back to reading from log
   *             files
```

this does not mention that if there is no checksum, we will also try to use the hint. Currently it says if there is no crc then we just list from the log

Can we delete some of these comments and just ensure that the code is more self describing?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -388,50 +387,66 @@ private Map<String, DomainMetadata> loadDomainMetadataMap(Engine engine) {
...
+      Engine engine,
+      LogSegment logSegment,
+      Optional<SnapshotHint> snapshotHint,
+      long snapshotVersion) {
+
+    // Snapshot hint's version is current so we could use it in loading P&M.
+    // No need to read crc.
     if (snapshotHint.isPresent() && snapshotHint.get().getVersion() == snapshotVersion) {
-      return new Tuple2<>(snapshotHint, Optional.empty());
+      return snapshotHint;
```
scottsand-db, line 421: Can we move the `if (snapshotHint.isPresent() && snapshotHint.get().getVersion() > snapshotVersion) {` to the constructor? and just consider this part of the constructor input validation?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
huan233usc: I reworked a bit this method's documents. Reading twice I feel like the previous one is more verbose

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -182,7 +187,7 @@ public long getVersion() {
 
   /** Returns the crc info for the current snapshot if the checksum file is read */
   public Optional<CRCInfo> getCurrentCrcInfo() {
```
scottsand-db, line 189: So this is introducing a similar complexity to what we had in TransactionImpl a few weeks back

We now have a private member field that we should NOT use -- and instead we should use the getter method -- but both of these are available in the class!

So we either need to (1) wrap these behind a hidden scope/context, or (2) set it once and only once.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -182,7 +187,7 @@ public long getVersion() {
...
-    return currentCrcInfo;
+    return cachedCrcInfo.filter(crc -> crc.getVersion() == logSegment.getVersion());
```
scottsand-db, line 190: Also -- to clarify: the consumers of this API only ever want the CRCInfo if it was read during log replay? They never just want the CRCINfo for the latest version, even if it wasn't read during log replay? That semantic seems a bit odd to me -- can you please clarify / explain?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
huan233usc: Good catch, likewise I define a private class

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
huan233usc: the consumer just try to get crcInfo of the current version. We return if it is cached. Updating the cache is not necessary during log replay's construction.

scottsand-db: > the consumer just try to get crcInfo of the current version.

But this API will _only_ return it if it was read / cached?

My confusion is this:
- it seems like consumers want the CRC info for the current version. full stop. if it exists, they want it
- this API returns it if and only if it exists **and we had to read it to update the snapshot hint**

So, there seems to be a gap?

huan233usc: So the concern here is --- when version hint is used, then crcInfo will always empty. This was unfortunately the behavior before, although problematic.

I updated with a fix. (Previously it will cause CRC missing when snapshot hint is used, now it is getting worse as we load domain metadata, so agree to fix it along the way.)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -135,7 +135,7 @@ public static StructType getAddReadSchema(boolean shouldReadStats) {
   private final LogSegment logSegment;
   private final Tuple2<Protocol, Metadata> protocolAndMetadata;
   private final Lazy<Map<String, DomainMetadata>> domainMetadataMap;
-  private final Optional<CRCInfo> currentCrcInfo;
+  private final CrcInfoContext crcInfoContext;
 
   public LogReplay(
```
scottsand-db, line 140: Perhaps worth adding a class-level comment. Something like

```
*Implementation Note*: This class uses `maybeGetNewerSnapshotHintAndUpdateCache`
to calculate a `SnapshotHint` and also exposes a `getCurrentCrcInfo` method. Let's explain
their relationship:
- We want to find the latest `SnapshotHint` to use during log replay for Protocol and Metadata loading
- If we are not provided a SnapshotHint for this version, or are provided a stale hint, we will try to read the latest seen (by file listing) CRC file (if it exists). If so, we read it, cache it, and create a newer hint.
- Then, when `getCurrentCrcInfo` is called, we will either use the cached CRCInfo that we have already read, parsed, and cached; or, if it was never cached (because the hint was sufficiently new) we will read it, parse it, and cache it for the first time
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
```diff
@@ -109,6 +122,12 @@ public LogSegment(
         checkpoints.stream().allMatch(fs -> FileNames.isCheckpointFile(fs.getPath())),
         "checkpoints must all be actual checkpoint files");
 
+    lastSeenChecksum.ifPresent(
+        checksumFile ->
+            checkArgument(
+                FileNames.checksumVersion(new Path(checksumFile.getPath())) <= version,
+                "checksum file's version should be lower than logSegment's version"));
```
scottsand-db, line 129: `lower` --> `less than or equal to`

scottsand-db: and greater than the checkpoints version, right? add that check?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
huan233usc: Updated the class-level comment for CrcInfoContext below

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
```diff
@@ -171,6 +190,13 @@ public LogSegment(
     this.deltas = deltas;
     this.compactions = compactions;
     this.checkpoints = checkpoints;
+    this.lastSeenChecksum =
```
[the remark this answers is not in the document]
huan233usc: Done, made the SnapshotManager preparing the valid file and let LogSegment just do validation. Yeah it is cleaner.

scottsand-db (approved):

vkorukanti (approved):

PR #4113, opened 2025-02-01
[Kernel]Cache the crc info for a snapshot if its version's checksum file is read

huan233usc:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

Cache the crc info for a snapshot if its version's checksum file is read. This will be used for calculate the new CRC info for a new commit
## How was this patch tested?
Unit test

No

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
@@ -121,6 +121,7 @@ public class LogReplay {
   private final LogSegment logSegment;
   private final Tuple2<Protocol, Metadata> protocolAndMetadata;
   private final Lazy<Map<String, DomainMetadata>> domainMetadataMap;
+  private final Optional<CRCInfo> currentCrcInfo;
 
   public LogReplay(
       Path logPath,
@@ -131,12 +132,17 @@ public class LogReplay {
       Optional<SnapshotHint> snapshotHint,
       SnapshotMetrics snapshotMetrics) {
     assertLogFilesBelongToTable(logPath, logSegment.allLogFilesUnsorted());
-
+    Tuple2<Optional<SnapshotHint>, Optional<CRCInfo>> newerSnapshotHintAndCurrentCrcInfo =
+        maybeGetNewerSnapshotHintAndCurrentCrcInfo(
+            engine, logSegment, snapshotHint, snapshotVersion);
+    this.currentCrcInfo = newerSnapshotHintAndCurrentCrcInfo._2;
     this.dataPath = dataPath;
     this.logSegment = logSegment;
     this.protocolAndMetadata =
         snapshotMetrics.loadInitialDeltaActionsTimer.time(
-            () -> loadTableProtocolAndMetadata(engine, logSegment, snapshotHint, snapshotVersion));
+            () ->
+                loadTableProtocolAndMetadata(
+                    engine, logSegment, newerSnapshotHintAndCurrentCrcInfo._1, snapshotVersion));
     // Lazy loading of domain metadata only when needed
     this.domainMetadataMap = new Lazy<>(() -> loadDomainMetadataMap(engine));
   }
@@ -163,6 +169,10 @@ public class LogReplay {
 
   public long getVersion() {
     return logSegment.getVersion();
+  }
+
+  public Optional<CRCInfo> getCurrentCrcInfo() {
+    return currentCrcInfo;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -392,4 +372,57 @@ private Map<String, DomainMetadata> loadDomainMetadataMap(Engine engine) {
...
+                new SnapshotHint(
+                    crcInfo.getVersion(), crcInfo.getProtocol(), crcInfo.getMetadata())),
+            crcInfoOpt);
+      }
+      checkArgument(
+          crcInfo.getVersion() >= crcSearchLowerBound && crcInfo.getVersion() <= snapshotVersion);
+      // We found a CRCInfo of a version (a) older than the one we are looking for (snapshotVersion)
+      // but (b) newer than the current hint. Use this CRCInfo to create a new hint.
+      return new Tuple2<>(
+          Optional.of(
```
scottsand-db, line 422: You have duplicate creations of snapshot hints from the CRC Info. I think you could make a `SnapshotHint::fromCrcInfo` static helper in SnapshotHint.java

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -392,4 +372,57 @@ private Map<String, DomainMetadata> loadDomainMetadataMap(Engine engine) {
       throw new UncheckedIOException("Could not close iterator", ex);
     }
   }
+
+  // Calculates the latest snapshot hint before current snapshot version, returns the CRCInfo if
```
scottsand-db, line 376: Some thoughts:

1. Please use `/** */` for multi-line methoddocs, no `//`

2. `Calculates the latest snapshot hint before current snapshot version` --> below, you also return it if `snapshotHint.version == snapshotVersion`. So I don't think "before" is correct, right?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -392,4 +372,57 @@ private Map<String, DomainMetadata> loadDomainMetadataMap(Engine engine) {
...
+                snapshotHint.map(SnapshotHint::getVersion).orElse(0L) + 1,
+                logSegment.getCheckpointVersionOpt().orElse(0L),
+                // Only find the CRC within 100 versions.
+                snapshotVersion - 100,
+                0L));
+    Optional<CRCInfo> crcInfoOpt =
+        ChecksumReader.getCRCInfo(
+            engine, logSegment.getLogPath(), snapshotVersion, crcSearchLowerBound);
+    if (crcInfoOpt.isPresent()) {
+      CRCInfo crcInfo = crcInfoOpt.get();
```
scottsand-db, line 408: Couldn't you do:

```java
// < insert comments explaining the different cases >
checkArgument(crcInfo.getVersion() >= crcSearchLowerBound && crcInfo.getVersion() <= snapshotVersion);
val snapshotHInt = SnapshotHInt.fromCrcInfo(crcInfo)
return new Tuple2<>(
  snapshotHint,
  crcInfo.getVersion() == snapshotVersion ? crcInfoOpt : Optional.empty()
)
```

isn't this clearer + more succinct?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/LogReplaySuite.scala
```diff
@@ -298,4 +301,73 @@ class LogReplaySuite extends AnyFunSuite with TestUtils {
...
+      val tablePath = tempFile.getAbsolutePath
+      spark.sql(
+        s"CREATE TABLE delta.`$tablePath` USING DELTA AS " +
+          s"SELECT 0L as id"
+      )
+      spark.sql(
+        s"INSERT INTO delta.`$tablePath` SELECT 1L as id"
+      )
+      assert(
+        Files.deleteIfExists(
```
scottsand-db, line 336: worth making a helper function that deletes CRC at a given version in a given table path?

scottsand-db: you could probably add it to TestUtils

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/LogReplaySuite.scala
```diff
@@ -17,18 +17,21 @@ package io.delta.kernel.defaults
 
 import java.io.File
 import java.util.Optional
-
```
scottsand-db, line 20: nit: undo these deletes plz

scottsand-db (changes requested):
Looks good! Requested some changes

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -392,4 +372,52 @@ private Map<String, DomainMetadata> loadDomainMetadataMap(Engine engine) {
...
+    if (snapshotHint.isPresent() && snapshotHint.get().getVersion() > snapshotVersion) {
+      snapshotHint = Optional.empty();
+    }
+
+    long crcSearchLowerBound =
+        max(
+            asList(
+                // Prefer reading hint over CRC, so start listing from hint's version + 1,
+                // if hint is not present, list from version 0.
+                snapshotHint.map(SnapshotHint::getVersion).orElse(-1L) + 1,
```
huan233usc, line 402: After making change below, I observe a corner case of hit missing, read 0 will fail the check
checkArgument(crcInfo.getVersion() >= crcSearchLowerBound && crcInfo.getVersion() <= snapshotVersion);

Fix it.

(Old implementation is also correct because this check won't happen when version = 0 and find 0.crc)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
huan233usc: Yes it is cleaning and less error prone, updated

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/LogReplayEngineMetricsSuite.scala
```diff
@@ -384,7 +384,7 @@ class LogReplayEngineMetricsSuite extends AnyFunSuite with TestUtils {
   // Tests for loading P & M through checksums files                                            //
   /////////////////////////////////////////////////////////////////////////////////////////////////
 
-  Seq(-1L, 3L, 4L).foreach { version => // -1 means latest version
+  Seq(-1L, 0L, 3L, 4L).foreach { version => // -1 means latest version
```
huan233usc, line 387: 0L for the case https://github.com/delta-io/delta/pull/4113/files#r1940334805

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
huan233usc: Change it to "before or at the current snapshot version". Thanks for catching this boundary case.

scottsand-db (approved):
Please implement the one small change I've asked for, but then LGTM!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -165,6 +172,11 @@ public long getVersion() {
     return logSegment.getVersion();
   }
 
+  /** Returns the crc info for the current snapshot if the checksum file is read */
+  public Optional<CRCInfo> getCurrentCrcInfo() {
```
allisonport-db, line 176: is there any reason that we only save it if it's the current version? in the future won't we also want to still save the info for an earlier one? (to pass along as much info as possible)

vkorukanti (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
huan233usc: current snapshot is sufficient for unblocking simple crc write, so I only expose the currently version. We could incrementally expose more if needed in the future.

PR #4119, opened 2025-02-03
Add spark.testing=1 to build config

ctring:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Currently, the [`Utils.isTesting`](https://github.com/apache/spark/blob/105bed7d2cab807211540b166ce8b97549d4ede7/core/src/main/scala/org/apache/spark/util/Utils.scala#L1878) check in tests is not reliable because the `spark.testing` system property is not set before the suites start, so the tests are relying on [this line](https://github.com/apache/spark/blob/105bed7d2cab807211540b166ce8b97549d4ede7/core/src/test/scala/org/apache/spark/SparkFunSuite.scala#L94) in `SparkFunSuite` to set the test indicator. This is causing issues in CI because it means whether `Utils.isTesting` returns correct result depends on whether it is called before or after the function in `SparkFunSuite` is called.

For example, some deletion vector-related suites are relying on [the `TEST_DV_NAME_PREFIX`](https://github.com/delta-io/delta/blob/a6db4e05d943fec019522c9c08ff0b64125cbc5d/spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala#L2410-L2415) to be correctly initialized in tests. However, if a previous test in the same JVM group initialized this config before `spark.testing` is correctly set by `SparkFunSuite`, the deletion vector suites would see the incorrect value of `TEST_DV_NAME_PREFIX` and fail [this is currently not an issue in master by luck but if a PR added new tests that changed the JVM grouping in CI job, this would block the PR].

## How was this patch tested?

Check that the `TEST_DV_NAME_PREFIX` is correctly set for tests that run `Utils.testing` before `SparkFunSuite::beforeAll` (e.g., `org.apache.spark.sql.delta.DeltaWithNewTransactionWithCoordinatedCommitsBatch100Suite`)

No


felipepessoto:
@ctring could you help reviewing this related PR? https://github.com/delta-io/delta/pull/4074

PR #4120, opened 2025-02-03
[Spark] Honor codegen configs in DataSkippingStatsTracker

LukasRupprecht:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR uses MutableProjection.create instead of directly generating the MutableProjection via GeneratMutableProjection.generate. As the latter forces codegen, it can fail if the generated code is too large, which, e.g., can happen for very wide schemas. The new code uses the correct Spark API to automatically fall back to an InterpretedMutableProjection if codegen fails for any reason.

## How was this patch tested?

This PR just uses the correct, higher-level API call to create a MutableProjection so existing tests are sufficient.

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingStatsTracker.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingStatsTracker.scala
@@ -87,14 +87,11 @@ class DeltaTaskStatisticsTracker(
 
   // This projection combines the intermediate results stored by aggBuffer with the values of the
   // currently processed row and updates aggBuffer in place.
-  private val updateStats: MutableProjection = GenerateMutableProjection.generate(
-    expressions = JoinedProjection.bind(
-      aggBufferAttrs,
-      dataCols,
-      aggregates.flatMap(_.updateExpressions)),
-    inputSchema = Nil,
-    useSubexprElimination = true
-  )
+  private val updateStats: MutableProjection = {
+    val expressions = JoinedProjection.bind(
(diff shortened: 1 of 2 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingStatsTracker.scala
```diff
@@ -87,14 +87,11 @@ class DeltaTaskStatisticsTracker(
...
-    expressions = JoinedProjection.bind(
-      aggBufferAttrs,
-      dataCols,
-      aggregates.flatMap(_.updateExpressions)),
-    inputSchema = Nil,
-    useSubexprElimination = true
-  )
+  private val updateStats: MutableProjection = {
+    val expressions = JoinedProjection.bind(
+      aggBufferAttrs, dataCols, aggregates.flatMap(_.updateExpressions))
```
scovich, line 92: Tiny nit -- while we're cleaning up the nesting, it may be worth cleaning up all the way, for maximum readability?
```suggestion
    val aggs = aggregates.flatMap(_.updateExpressions)
    val expressions = JoinedProjection.bind(aggBufferAttrs, dataCols, aggs)
```

scovich (approved):
Nice stability fix!

spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingStatsTracker.scala
LukasRupprecht: Done

PR #4121, opened 2025-02-04
[Delta] When checks schema for writing, Delta enforces not null on a Nested Field only when its parent is not null

harperjiang:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
This PR corrects Delta's handling of nullable checks for nested fields when writing data.

For example, consider the schema:
```
STRUCT<foo: STRUCT<bar: INT NOT NULL, bar2: INT>>
```
Previously, Delta rejected NULL values for `foo`, even when it was marked as nullable. This was because a NULL `foo` would also make `foo.bar` NULL.

With this change, Delta now enforces `foo.bar` as non-null only when `foo` itself is non-null. As a result:
* foo = NULL is now accepted.
* foo = { bar = NULL, bar2 = 1 } remains rejected.

## How was this patch tested?
UT

No


Kimahriman:
Related issue and discussion: https://github.com/delta-io/delta/issues/860

felipepessoto:
@harperjiang @Kimahriman was this fixed?

Kimahriman:
No this is still an active problem, no fix has been accepted

PR #4128, opened 2025-02-06
[WIP] [Kernel] Catalog Commits (CCv2) Prototype

scottsand-db:
...


kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ccv2/Catalog.scala
```diff
@@ -0,0 +1,53 @@
...
+  override def resolveTable(tableName: String): ResolveTableResponse = ???
+  override def getCommits(): GetCommitsResponse = ???
+}
+
+////////////////////////
+// InMemoryCCv2Client //
+////////////////////////
+
+class InMemoryCCv2Client extends CCv2Client {
+  override def resolveTable(tableName: String): ResolvedMetadata = {
```
scottsand-db, line 50:
```
// UC_REST_CLIENT
uc_catalog.delta.protocol.$field1

// GLUE_REST_CLIENT
glue_catalog.delta.protocol.$field1

Should the UC_CCV2_CLIENT and the GLUE_CCV2_CLIENT then translate to KERNEL?

kernel.delta.protocol
```

Q: if we give them a bag of properties, will they add a prefix? are they responsible for stripping the prefix when they give the bag of properties back to us?

Q: is the goal that the CCv2 clients should not have to do a bunch of translation / deep understanding?

**Intuition: Kernel should be able to commit a bag of properties and then get back that exact same bag of properties. Maybe the CCv2 client needs to add/strip some prefixes when it goes back and forth to the catalog.**

scottsand-db: the interface `Kernel <> CCv2Client` should be the bag of properties

scottsand-db, line 50: `resolveTableAsBagOfProperties(tableName: String) -> Map<String, String>`

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ccv2/InMemoryCCv2Suite.scala
```diff
@@ -0,0 +1,109 @@
...
+
+    val txn = resolvedTable
+      .createTransactionBuilder(testEngineInfo, Operation.CREATE_TABLE)
+      .withSchema(defaultEngine, testSchema)
+      .build(defaultEngine)
+
+    val txnState = txn.getTransactionState(defaultEngine)
+    val stagedFiles = stageData(txnState, Map.empty, data.toList)
+    val stagedActionsIterable = CloseableIterable.inMemoryIterable(stagedFiles)
+    txn.commit(defaultEngine, stagedActionsIterable)
```
scottsand-db, line 56:
```
val fileActions = ...
val allActions = txn.finalizeActions(fileActions, ...)
ResolvedMetadata::commit(allActions, ...)
```

horizonzy:
Hi, this PR hasn't changed in a long time. Is it still being processed?

scottsand-db:
Hi @horizonzy ! Take a look at https://github.com/delta-io/delta/issues/4573! Kernel catalog-managed tables support is done!

horizonzy:
Is there some demo code to test the feature?

scottsand-db:
@horizonzy -- Kernel isn't trivial to "demo". Like, it's not a connector or compute engine like Delta-Spark or Delta-Flink. It's a building block to build those connectors.

e.g. I can point you to a past example where you point kernel/unitycatalog and kernel/api and kernel/defaultEngine to a real UC instance and you can read and write -- but that is a contrived examples. Developers probably aren't using just Raw Kernel. Instead they are building Kenel into connnectors

PR #4131, opened 2025-02-07
[SPARK] Add common sql test utility methods to `DeltaSQLTestUtils`

stefankandic:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Adding the most commonly used methods in testing; such as getting the delta log, snapshot, protocol, stats, columns etc in `DeltaSQLTestUtils` so that we don't have to keep reimplementing them over and over again and just use this trait instead.

## How was this patch tested?
No testing required.

No.


johanl-db (approved):
Nice! Thanks for adding these

PR #4134, opened 2025-02-08
[Kernel]Add simple crc post commit for incremental crc writing.

huan233usc:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

This PR introduces a new post commit hook - ChecksumSimple, for writing CRC file after txn commit.
CRC file will only be written only commit version - 1's snapshot reads CRC during state construction

Other case will be handled in a separate PR

## How was this patch tested?

E2e test

No


kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWritesSuite.scala
huan233usc, line 1: the change in this file mainly parameterized certain test case for verifying crc writing

kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/ChecksumWriter.java
```diff
@@ -0,0 +1,90 @@
...
+  private static final String PROTOCOL = "protocol";
+  private static final String TXN_ID = "txnId";
+
+  public ChecksumWriter(Path logPath) {
+    this.logPath = requireNonNull(logPath);
+  }
+
+  /** Writes a checksum file */
+  public void writeCheckSum(Engine engine, CRCInfo crcInfo) throws IOException {
+    // No sufficient information to write checksum file.
```
scottsand-db, line 55: what's this comment for?

scottsand-db (changes requested):
Looks good! Left some feedback on some of the code comments + organziation. Thanks!

Also -- please update the PR description.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/checksum/ChecksumWriterSuite.scala
```diff
@@ -0,0 +1,191 @@
...
+ * Mock implementation of JsonHandler for testing checksum file writing.
+ */
+class MockCheckSumFileJsonWriter extends BaseMockJsonHandler {
+  var capturedCrcRow: Row = new GenericRow(new StructType(), new util.HashMap[Integer, AnyRef])
+  var checksumFilePath: String = ""
+
+  override def writeJsonFileAtomically(
+    filePath: String,
+    data: CloseableIterator[Row],
+    overwrite: Boolean): Unit = {
```
scottsand-db, line 185: nit: should be 4 space indentation on continuation lines

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumSimpleComparisonSuite.scala
```diff
@@ -0,0 +1,376 @@
...
+      val kernelTablePath = tablePath + "kernel"
+      Table
+        .forPath(engine, kernelTablePath)
+        .createTransactionBuilder(engine, "test-engine", CREATE_TABLE)
+        .withSchema(engine, new StructType().add("id", INTEGER))
+        .build(engine)
+        .commit(engine, emptyIterable())
+        .getPostCommitHooks
+        .forEach(
+          hook => hook.threadSafeInvoke(engine)
```
scottsand-db, line 62: nit: this `hook => hook.` can be on the same line as the `forEach` call

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumSimpleComparisonSuite.scala
```diff
@@ -0,0 +1,376 @@
...
+      spark.sql(
+        s"CREATE OR REPLACE TABLE delta.`${tablePath + "spark"}` USING DELTA AS SELECT 1 as id"
+      )
+      val txn = Table
+        .forPath(engine, kernelTablePath)
+        .createTransactionBuilder(engine, "test-engine", CREATE_TABLE)
+        .withSchema(engine, new StructType().add("id", INTEGER))
+        .build(engine)
+
+      copyAndCommitTxnForUnpartitionedTable(txn, engine, sparkTablePath, versionAtCommit = 0)
```
scottsand-db, line 87: I couldn't quite understand what this method is doing -- can you explain to me (on this PR, not in the code)?

huan233usc: I was trying to make the test logic like

1.commit to spark table using ctas/dml, which produces a log with add file
2.read the log in step 1 and commit to the kernel table (make sore both table's log pointing to the same sets of parquet file to validate table size is correct in crc)
3.verify crc written stats are aligned

This method is mainly for step 2.

Does "copyDeltaLogFromSparkToKernel" a better name?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumSimpleComparisonSuite.scala
```diff
@@ -0,0 +1,376 @@
...
+      assertChecksumEquals(engine, sparkTablePath, kernelTablePath, 0)
+
+      (1 to 10).foreach(
+        version => insertIntoUnpartitionedTable(engine, sparkTablePath, kernelTablePath, version)
+      )
+
+    }
+  }
+
+  test("create table as select and verify checksum") {
```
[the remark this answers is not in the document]
huan233usc: The only difference is the 00000.crc, ctas has added file while create table doesn't, so I added this test for completeness.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -436,6 +442,53 @@ private void recordTransactionReport(
     engine.getMetricsReporters().forEach(reporter -> reporter.report(transactionReport));
   }
 
+  private Optional<CRCInfo> buildPostCommitCrcInfo(
```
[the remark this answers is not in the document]
huan233usc: This is an excellent suggestion, thanks!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/ChecksumWriter.java
huan233usc: Was accidental kept when copying some code from an old pr, removed

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumSimpleComparisonSuite.scala
```diff
@@ -0,0 +1,376 @@
...
+        .commit(engine, emptyIterable())
+        .getPostCommitHooks
+        .forEach(
+          hook => hook.threadSafeInvoke(engine)
+        )
+      spark.sql(s"CREATE OR REPLACE TABLE delta.`${sparkTablePath}` (id Integer) USING DELTA")
+      assertChecksumEquals(engine, sparkTablePath, kernelTablePath, 0)
+
+      (1 to 10).foreach(
+        version => insertIntoUnpartitionedTable(engine, sparkTablePath, kernelTablePath, version)
```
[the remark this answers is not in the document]
huan233usc: Good point, I renamed it to insertIntoUnpartitionedTableAndCheckCrc

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/checksum/ChecksumWriterSuite.scala
huan233usc: I manually fix it :-(

kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/CRCInfo.java
```diff
@@ -0,0 +1,122 @@
...
+          .add(NUM_PROTOCOL, LongType.LONG)
+          .add(METADATA, Metadata.FULL_SCHEMA)
+          .add(PROTOCOL, Protocol.FULL_SCHEMA)
+          .add(TXN_ID, StringType.STRING, /*nullable*/ true);
+
+  public static Optional<CRCInfo> fromColumnarBatch(
+      long version, ColumnarBatch batch, int rowId, String crcFilePath) {
+    Protocol protocol =
+        Protocol.fromColumnVector(
+            batch.getColumnVector(CRC_FILE_SCHEMA.indexOf("protocol")), rowId);
```
scottsand-db, line 56: nit: if you define `PROTOCOL` then you might as well use it here, eh?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/CRCInfo.java
```diff
@@ -0,0 +1,122 @@
...
+  // Constants for schema field names
+  public static final String TABLE_SIZE_BYTES = "tableSizeBytes";
+  public static final String NUM_FILES = "numFiles";
+  public static final String NUM_METADATA = "numMetadata";
+  public static final String NUM_PROTOCOL = "numProtocol";
+  public static final String METADATA = "metadata";
+  public static final String PROTOCOL = "protocol";
+  public static final String TXN_ID = "txnId";
+
+  public static StructType CRC_FILE_SCHEMA =
```
scottsand-db, line 42: make this `final`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/ChecksumWriter.java
```diff
@@ -0,0 +1,81 @@
...
+    logger.info("Writing checksum file to path: {}", newChecksumPath);
+    wrapEngineExceptionThrowsIO(
+        () -> {
+          engine
+              .getJsonHandler()
+              .writeJsonFileAtomically(
+                  newChecksumPath.toString(),
+                  singletonCloseableIterator(buildCheckSumRow(crcInfo)),
+                  false /* overwrite */);
+          return null;
```
scottsand-db, line 57: log that the write was a success

kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/ChecksumWriter.java
```diff
@@ -0,0 +1,81 @@
...
+        },
+        "Write checksum file `%s`",
+        newChecksumPath);
+  }
+
+  private Row buildCheckSumRow(CRCInfo crcInfo) {
```
scottsand-db, line 63: why not just have `toRow` on `CRCInfo`? I think that's a better spot for this than inside of `ChecksumWriter`

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/checksum/ChecksumWriterSuite.scala
```diff
@@ -0,0 +1,191 @@
...
+    assert(row.getLong(NUM_PROTOCOL) == 1L)
+
+    if (txn.isPresent) {
+      assert(row.getString(TXN_ID) == txn.get())
+    } else {
+      assert(row.isNullAt(TXN_ID))
+    }
+  }
+
+  private def verifyMetadataAndProtocol(row: Row, metadata: Metadata, protocol: Protocol): Unit = {
```
scottsand-db, line 97: `expectedMetadata` `expectedProtocol`

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/checksum/ChecksumWriterSuite.scala
```diff
@@ -0,0 +1,191 @@
...
+  private def createTestProtocol(): Protocol = {
+    new Protocol(
+      /* minReaderVersion= */ 1,
+      /* minWriterVersion= */ 2,
+      Collections.emptyList(),
+      Collections.emptyList()
+    )
+  }
+
+  private def checkMetadata(metadata: Metadata, metadataRow: Row): Unit = {
```
scottsand-db, line 127: `expectedMetadata` `actualMetadataRow`

scottsand-db (changes requested):

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumSimpleComparisonSuite.scala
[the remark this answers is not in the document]
huan233usc: Removed the CTAS case, we believe it is a bit redundant here.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumSimpleComparisonSuite.scala
huan233usc: Added some more documentation to the test class and method as follow up

kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/CRCInfo.java
huan233usc: Yes, my bad, I forgot to update them.

scottsand-db (changes requested):

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/checksum/ChecksumWriterSuite.scala
```diff
@@ -0,0 +1,198 @@
...
+        .getInt(Protocol.FULL_SCHEMA.indexOf("minWriterVersion"))
+    )
+  }
+}
+
+/**
+ * Mock implementation of JsonHandler for testing checksum file writing.
+ */
+class MockCheckSumFileJsonWriter extends BaseMockJsonHandler {
+  var capturedCrcRow: Row = new GenericRow(new StructType(), new util.HashMap[Integer, AnyRef])
```
vkorukanti, line 186: why do you need to initialize it?

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/checksum/ChecksumWriterSuite.scala
```diff
@@ -0,0 +1,198 @@
...
+  private def createTestProtocol(): Protocol = {
+    new Protocol(
+      /* minReaderVersion= */ 1,
+      /* minWriterVersion= */ 2,
+      Collections.emptyList(),
+      Collections.emptyList()
+    )
+  }
+
+  private def checkMetadata(expectedMetadata: Metadata, actualMetadataRow: Row): Unit = {
```
vkorukanti, line 130: can we use convert the Row back to Metadata and compare? same for the protocol

kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/ChecksumWriter.java
```diff
@@ -0,0 +1,82 @@
...
+
+  private final Path logPath;
+
+  public ChecksumWriter(Path logPath) {
+    this.logPath = requireNonNull(logPath);
+  }
+
+  /** Writes a checksum file */
+  public void writeCheckSum(Engine engine, CRCInfo crcInfo) throws IOException {
+    Path newChecksumPath = FileNames.checksumFile(logPath, crcInfo.getVersion());
```
vkorukanti, line 47: Should we add some validations to make sure we are not writing some incorrect values (e.g. numMetadata is not 1)

vkorukanti: If we add validations, make sure to add tests for it.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumSimpleComparisonSuite.scala
```diff
@@ -0,0 +1,338 @@
...
+    withTempDirAndEngine { (tablePath, engine) =>
+      val sparkTablePath = tablePath + "spark"
+      val kernelTablePath = tablePath + "kernel"
+      Table
+        .forPath(engine, kernelTablePath)
+        .createTransactionBuilder(engine, "test-engine", CREATE_TABLE)
+        .withSchema(engine, new StructType().add("id", INTEGER))
+        .build(engine)
+        .commit(engine, emptyIterable())
+        .getPostCommitHooks
```
vkorukanti, line 62: Extend the DeltaWriteSuiteBase. It has utility methods to create the txn and other helper methods

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumSimpleComparisonSuite.scala
```diff
@@ -0,0 +1,338 @@
...
+    new File(f"$basePath/_delta_log/$version%020d.crc").toPath
+  }
+
+  private def assertCrcInfoEquals(crc1: CRCInfo, crc2: CRCInfo): Unit = {
+    assert(crc1.getVersion === crc2.getVersion)
+    assert(crc1.getNumFiles === crc2.getNumFiles)
+    assert(crc1.getTableSizeBytes === crc2.getTableSizeBytes)
+    assert(crc1.getMetadata.getSchema === crc2.getMetadata.getSchema)
+    assert(crc1.getMetadata.getPartitionColNames === crc2.getMetadata.getPartitionColNames)
+  }
```
vkorukanti, line 136: can you just add equals/hashCode to CRCInfo?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWritesSuite.scala
```diff
@@ -329,103 +330,172 @@ class DeltaTableWritesSuite extends DeltaTableWriteSuiteBase with ParquetSuiteBa
...
+            assert(txn.getSchema(engine) === schema)
+            // Expect the partition column name is exactly same as the one in the schema
+            assert(txn.getPartitionColumns(engine) === Seq("Part1", "part2").asJava)
+            val txnResult = txn.commit(engine, emptyIterable())
+
+            assert(txnResult.getVersion === 0)
+            assertCheckpointReadiness(txnResult, isReadyForCheckpoint = false)
+            assertChecksumSimpleReadiness(txnResult)
+            if (executePostCommitHook) {
+              txnResult.getPostCommitHooks.forEach(hook => hook.threadSafeInvoke(engine))
```
vkorukanti, line 363: here we are exeucting the hooks, but not checking anything?

Wondering if there is a better way to do this.

vkorukanti (changes requested):

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumSimpleComparisonSuite.scala
huan233usc: Metadata and Protocol doesn't have equals/hashCode implemented so adding equals/hashCode to CRCInfo won't works without fixing Metadata and Protocol.

Added a todo and I will make it in a separate PR

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -437,6 +443,34 @@ private void recordTransactionReport(
...
```
allisonport-db, line 446: add simple to method name? or some other reference that this is only if previous crc is avail

kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/CRCInfo.java
```diff
@@ -0,0 +1,148 @@
...
+
+  public static Optional<CRCInfo> fromColumnarBatch(
+      long version, ColumnarBatch batch, int rowId, String crcFilePath) {
+    Protocol protocol =
+        Protocol.fromColumnVector(batch.getColumnVector(CRC_FILE_SCHEMA.indexOf(PROTOCOL)), rowId);
+    Metadata metadata =
+        Metadata.fromColumnVector(batch.getColumnVector(CRC_FILE_SCHEMA.indexOf(METADATA)), rowId);
+    long tableSizeBytes =
+        batch.getColumnVector(CRC_FILE_SCHEMA.indexOf(TABLE_SIZE_BYTES)).getLong(rowId);
+    long numFiles = batch.getColumnVector(CRC_FILE_SCHEMA.indexOf(NUM_FILES)).getLong(rowId);
```
allisonport-db, line 64: I think we should still use `requireNonNull` here, we should have a helper util for CV for this case

kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/CRCInfo.java
```diff
@@ -0,0 +1,148 @@
...
+    Optional<String> txnId =
+        Optional.ofNullable(
+            batch.getColumnVector(CRC_FILE_SCHEMA.indexOf(TXN_ID)).getString(rowId));
```
allisonport-db, line 67: needs to check `isNullAt`, column vector interface technically doesn't promise any specific return value when `isNullAt = true`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/CRCInfo.java
```diff
@@ -0,0 +1,148 @@
...
+      Protocol protocol,
+      long tableSizeBytes,
+      long numFiles,
+      Optional<String> txnId) {
+    this.version = version;
+    this.metadata = requireNonNull(metadata);
+    this.protocol = requireNonNull(protocol);
+    this.tableSizeBytes = tableSizeBytes;
+    this.numFiles = numFiles;
+    this.txnId = txnId;
```
allisonport-db, line 95: we still `requireNoneNull` here

kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/ChecksumWriter.java
allisonport-db: we could also instead add this validation as part of the `CRCInfo` constructor?

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/checksum/ChecksumWriterSuite.scala
allisonport-db: +1

Or maybe convert them both to `TestRow`s and compare

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/checksum/ChecksumWriterSuite.scala
```diff
@@ -0,0 +1,198 @@
...
+  private val FAKE_DELTA_LOG_PATH = new Path("/path/to/delta/log")
+
+  // Schema field indices in crc file
+  private val TABLE_SIZE_BYTES = CRC_FILE_SCHEMA.indexOf("tableSizeBytes")
+  private val NUM_FILES = CRC_FILE_SCHEMA.indexOf("numFiles")
+  private val NUM_METADATA = CRC_FILE_SCHEMA.indexOf("numMetadata")
+  private val NUM_PROTOCOL = CRC_FILE_SCHEMA.indexOf("numProtocol")
+  private val TXN_ID = CRC_FILE_SCHEMA.indexOf("txnId")
+  private val METADATA = CRC_FILE_SCHEMA.indexOf("metadata")
+  private val PROTOCOL = CRC_FILE_SCHEMA.indexOf("protocol")
```
allisonport-db, line 47: if we only use these variables once I'd prefer to just put them in-line it's more self-defining. if not please add something like `_IDX` to them, it's not intuitive that they are indices

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/checksum/ChecksumWriterSuite.scala
huan233usc: It seems that we need some refactoring here, either implement `compareTo()` on  Metadata or move TestRow from default to internal.

Let create a separate PR for add `compareTo()` to metadata.java but prefer not to block this PR on this

kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/CRCInfo.java
huan233usc: long is primitive so I added checkArgument to ensure it is not null

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/checksum/ChecksumWriterSuite.scala
```diff
@@ -0,0 +1,198 @@
...
+      new StructType(),
+      stringArrayValue(util.Arrays.asList("c3")),
+      Optional.of(123),
+      stringStringMapValue(new util.HashMap[String, String]() {
+        put("delta.appendOnly", "true")
+      })
+    )
+  }
+
+  private def createTestProtocol(): Protocol = {
```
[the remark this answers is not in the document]
huan233usc: I could not find it in generic utils in internal package, just some in 

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/tablefeatures/TableFeaturesSuite.scala

I create https://github.com/delta-io/delta/pull/4169 for refactoring but prefer not to let this PR blocked by the refactoring PR

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumSimpleComparisonSuite.scala
```diff
@@ -0,0 +1,338 @@
...
+      val row = rows.next()
+      val addIndex = row.getSchema.indexOf("add")
+
+      if (!row.isNullAt(addIndex)) {
+        processAddFile(row.getStruct(addIndex), partition, addFiles)
+      }
+    }
+  }
+
+  private def processAddFile(
```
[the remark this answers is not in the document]
huan233usc: I revisited the code and feel that both processAddFile and shouldIncludeFile are only called once so we could just not using these two helper lib.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -437,6 +444,34 @@ private void recordTransactionReport(
     engine.getMetricsReporters().forEach(reporter -> reporter.report(transactionReport));
   }
 
+  private Optional<CRCInfo> buildPostCommitCrcInfoIfCurrentCrcAvailable(
+      long commitAtVersion, TransactionMetricsResult metricsResult) {
+    // Create table
```
scottsand-db, line 449: nit: superfluous comment

kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/CRCInfo.java
```diff
@@ -0,0 +1,155 @@
...
+      new StructType()
+          .add(TABLE_SIZE_BYTES, LongType.LONG)
+          .add(NUM_FILES, LongType.LONG)
+          .add(NUM_METADATA, LongType.LONG)
+          .add(NUM_PROTOCOL, LongType.LONG)
+          .add(METADATA, Metadata.FULL_SCHEMA)
+          .add(PROTOCOL, Protocol.FULL_SCHEMA)
+          .add(TXN_ID, StringType.STRING, /*nullable*/ true);
+
+  public static Optional<CRCInfo> fromColumnarBatch(
```
scottsand-db, line 57: I think this method can be cleaned up a bit / presented a bit more structured:
- could you do all the gets together? (i.e. get from the vectors)
- could you move all the checks together?
- use getSchemaIndex? or delete it from the method below -- just be consistent
- save `batch.getColumnVector(CRC_FILE_SCHEMA.indexOf(TXN_ID))` to a local variable to not paste it twice
- make it a bit clearer which are the optional params and which are the required params of the CRC schema?

Basically, right now this is a bit of a wall of text and its a bit hard to read. Some grouping / new lines between groups would be nice

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -437,6 +444,34 @@ private void recordTransactionReport(
...
+    return readSnapshot
+        .getCurrentCrcInfo()
+        .filter(lastCrcInfo -> commitAtVersion == lastCrcInfo.getVersion() + 1)
+        .map(
+            lastCrcInfo ->
+                new CRCInfo(
+                    commitAtVersion,
+                    metadata,
+                    protocol,
+                    lastCrcInfo.getTableSizeBytes() + metricsResult.getTotalAddFilesSizeInBytes(),
```
scottsand-db, line 470: add a TODO that in the future we will need to handle RemoveFiles

kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/CRCInfo.java
huan233usc: Thanks for the suggestion!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -437,6 +445,34 @@ private void recordTransactionReport(
...
+              metadata,
+              protocol,
+              metricsResult.getTotalAddFilesSizeInBytes(),
+              metricsResult.getNumAddFiles(),
+              Optional.of(txnId.toString())));
+    }
+
+    return readSnapshot
+        .getCurrentCrcInfo()
+        .filter(lastCrcInfo -> commitAtVersion == lastCrcInfo.getVersion() + 1)
```
allisonport-db, line 463: nit: want to add a comment like `// in the case of a conflicting txn and successful retry the readSnapshot may not be commitVersion - 1`

When I first saw this it was intuitive to me immediately why we needed to filter

allisonport-db: Side note, in those cases, is it possible for us to see if there was a CRC written and use that for simple crc write?

This can be a P2+ but maybe we should track it somewhere (i.e. create a github issue?)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/CRCInfo.java
```diff
@@ -0,0 +1,159 @@
...
+
+    // Read optional fields
+    ColumnVector txnIdColumnVector = batch.getColumnVector(getSchemaIndex(TXN_ID));
+    Optional<String> txnId =
+        txnIdColumnVector.isNullAt(rowId)
+            ? Optional.empty()
+            : Optional.of(txnIdColumnVector.getString(rowId));
+
+    checkArgument(!batch.getColumnVector(getSchemaIndex(TABLE_SIZE_BYTES)).isNullAt(rowId));
+    checkArgument(!batch.getColumnVector(getSchemaIndex(NUM_FILES)).isNullAt(rowId));
```
allisonport-db, line 76: I think technically we want to do the isNullAt check before we access the value. Basically we should prefix any CV access with a null check

allisonport-db: Ah I found the util function, you should just use `InternalUtils.requireNonNull` for all the required fields

See Metadata.fromColumnVector for examples on how to use it

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
huan233usc: This could be a good optimization, filed https://github.com/delta-io/delta/issues/4177 for track

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWriteWithCrcSuite.scala
```diff
@@ -0,0 +1,125 @@
...
+        engine: Engine,
+        dataActions: CloseableIterable[Row]): TransactionCommitResult = {
+      val result = txn.commit(engine, dataActions)
+      result.getPostCommitHooks
+        .stream()
+        .filter(hook => hook.getType == PostCommitHookType.CHECKSUM_SIMPLE)
+        .forEach(hook => hook.threadSafeInvoke(engine))
+      result
+    }
+  }
```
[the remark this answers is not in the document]
allisonport-db: I think if we truly wanted to do it all w/out `commitTransaction` we could wrap our txn/txnBuilder with one that inserts custom commit logic. This is probably overkill. Wondering if there are any ideas on how to be sure that future developers use `commitTransaction` in their tests, not because it's easier (it's not) but because it inserts custom logic in child suites.

Please make sure to document this also! Test documentation is super helpful.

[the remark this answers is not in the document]
allisonport-db: Maybe there's some way to disallow a certain regex in a style check?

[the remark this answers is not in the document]
huan233usc: Ah, this is some code I was intended to remove, but some how that commit isn't pushed. Will move to commitTransaction

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWriteWithCrcSuite.scala
```diff
@@ -0,0 +1,125 @@
...
+      engine: Engine,
+      dataActions: CloseableIterable[Row]): TransactionCommitResult = {
+    txn.commitAndGenerateCrc(engine, dataActions)
+  }
+
+  override def verifyWrittenContent(
+      path: String,
+      expSchema: StructType,
+      expData: Seq[TestRow],
+      expPartitionColumns: Seq[String] = Seq(),
```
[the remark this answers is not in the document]
huan233usc: Building for snapshot will load the crc. So getting the partition columns from snapshot may not be a good idea.

However, verifying partitioning here is a bit overkill after a second thought, we could just ensure checksum file is readable.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWriteWithCrcSuite.scala
```diff
@@ -0,0 +1,125 @@
...
+    // need to set the timezone to UTC before reading the data using Spark to make the tests pass
+    withSparkTimeZone("UTC") {
+      val resultSpark = spark
+        .sql(s"SELECT * FROM delta.`$path`" + {
+          if (version.isDefined) s" VERSION AS OF ${version.get}" else ""
+        })
+        .collect()
+        .map(TestRow(_))
+      checkAnswer(resultSpark, expData)
+    }
```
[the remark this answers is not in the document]
huan233usc: super.verifyWrittenContent works, thanks

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumSimpleComparisonSuite.scala
```diff
@@ -0,0 +1,294 @@
...
+        s"CREATE OR REPLACE TABLE delta.`${sparkTablePath}` " +
+          s"(id Integer, part Integer) USING DELTA PARTITIONED BY (part)")
+      assertChecksumEquals(engine, sparkTablePath, kernelTablePath, 0)
+
+      (1 to 10).foreach(version =>
+        insertIntoPartitionedTableAndCheckCrc(engine, sparkTablePath, kernelTablePath, version))
+    }
+  }
+
+  /**
```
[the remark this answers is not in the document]
huan233usc: Spark does the a/b comparison, (not at file level though)

https://github.com/delta-io/delta/blob/3e3c01ba09fda2c73cc900f49966cb28bbed46a8/spark/src/test/scala/org/apache/spark/sql/delta/ChecksumSuite.scala#L127

I think the value of this test is to ensure our behavior(especially file size collection) is the same as spark


>Can you try using TableImpl.getRawChanges maybe?
This is an excellent suggestion! It simply the code a lot

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
scottsand-db: This should all just be a part of the rebase process. When we rebase, we need to list + read all the JSONs and do various conflict checkpoing. We might as well list + read the latest CRC file, too.

scottsand-db (approved):

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWriteWithCrcSuite.scala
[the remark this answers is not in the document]
allisonport-db: oop yeah that makes sense :)

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWriteSuiteBase.scala
```diff
@@ -460,4 +462,11 @@ trait DeltaTableWriteSuiteBase extends AnyFunSuite with TestUtils {
         .stream()
         .anyMatch(hook => hook.getType == PostCommitHookType.CHECKPOINT) === isReadyForCheckpoint)
   }
+
```
allisonport-db, line 465: please document that all child suites should use this instead of `.commit` directly, and explain why this is!! also maybe put it at the top of the file in all caps (in DeltaTableWriteSuite)

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWriteWithCrcSuite.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  override def verifyWrittenContent(
+      path: String,
+      expSchema: StructType,
+      expData: Seq[TestRow]): Unit = {
+    super.verifyWrittenContent(path, expSchema, expData)
+    checkChecksum(path, expSchema)
+  }
+
+  def checkChecksum(
```
allisonport-db, line 60: please document what this does and doesn't check

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWriteWithCrcSuite.scala
```diff
@@ -0,0 +1,71 @@
...
+import io.delta.kernel.hook.PostCommitHook.PostCommitHookType
+import io.delta.kernel.internal.{SnapshotImpl, TransactionImpl}
+import io.delta.kernel.internal.actions.Metadata
+import io.delta.kernel.internal.checksum.{ChecksumReader, CRCInfo}
+import io.delta.kernel.internal.fs.Path
+import io.delta.kernel.internal.util.Utils.singletonCloseableIterator
+import io.delta.kernel.types.StructType
+import io.delta.kernel.utils.{CloseableIterable, FileStatus}
+
+class DeltaTableWriteWithCrcSuite extends DeltaTableWritesSuite {
```
allisonport-db, line 38: class docs might also be useful here (i.e. we just check that crc is generated for all txn basically right? nothing more?)

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWritesSuite.scala
```diff
@@ -794,6 +797,8 @@ class DeltaTableWritesSuite extends DeltaTableWriteSuiteBase with ParquetSuiteBa
 
   test("insert into table - idempotent writes") {
     withTempDirAndEngine { (tblPath, engine) =>
+      // TODO: re-enable when CRC_FULL post commit hook is added, txn2 requires CRC_FULL
+      assume(this.suiteName != ("DeltaTableWriteWithCrcSuite"))
```
vkorukanti, line 801: can we add this to the ingore list in the `DeltaTableWriteWithCrcSuite`? This can be a separate PR. I think we may need support for ingoreList similar to delta-spark tests.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWriteSuiteBase.scala
```diff
@@ -460,4 +462,11 @@ trait DeltaTableWriteSuiteBase extends AnyFunSuite with TestUtils {
...
+  protected def commitTransaction(
```
vkorukanti, line 466: Please add an oneliner doc to this base class on the need to commit through this method.

vkorukanti (approved):
LGTM, some followup test work which can be in followup PRs.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWritesSuite.scala
allisonport-db: we can make `IgnoreTests` a mixin trait

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumSimpleComparisonSuite.scala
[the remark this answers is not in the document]
allisonport-db: https://github.com/delta-io/delta/commit/aae9d649fd6a9ca4b87c742beb048afffcfd571b is what I was trying to outline earlier but hard to explain

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWriteWithCrcSuite.scala
huan233usc: Update the method name to verifyChecksumValid and add docs

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWritesSuite.scala
huan233usc: Will address it seperately

PR #4142, opened 2025-02-11
[SPARK] Support CTAS from a table with a deleted default value

olaky:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

We are currently leaking values using the 'EXISTS_DEFAULT' config key that is used for default columns in Spark into table schemas. The 'EXISTS_DEFAULT' key holds values for rows that existed when a new column with a default value was added to the table. This use case is not supported in Delta.

In this PR, we remove the 'EXISTS_DEFAULT' entries every time we update the metadata of a table. This includes the case when a table is created. This ensures that new tables do not have 'EXISTS_DEFAULT' entries and operations that update metadata remove entries that existed before because previous versions added them.

This change also now allows us to CTAS from a table where a default value existed in the past but has been dropped. It does not enable CTAS from a table with an active default value.

The feature has a kill switch: REMOVE_EXISTS_DEFAULT_FROM_SCHEMA_ON_EVERY_METADATA_CHANGE

## How was this patch tested?

Added new tests

Previously it was not possible to CTAS from a table with a removed default without setting the delta.feature.allowColumnDefaults even though no default value was actually part of the new table. this is now possible.


spark/src/main/scala/org/apache/spark/sql/delta/commands/CreateDeltaTableCommand.scala
```diff
@@ -314,7 +315,29 @@ case class CreateDeltaTableCommand(
             txn.snapshot.domainMetadata, newDomainMetadata)
         taggedCommitData = taggedCommitData.copy(actions = newActions)
       }
-      val op = getOperation(txn.metadata, isManagedTable, Some(options),
+
+      def schemaContainsExistsColumnMetadata(schema: StructType): Boolean = {
+        schema.exists(
+          _.metadata.contains(ResolveDefaultColumnsUtils.EXISTS_DEFAULT_COLUMN_METADATA_KEY))
+      }
+      val metadataForOp =
```
c27kwan, line 323: This logic could use a comment to explain _why_ we're doing it this way. IIUC we need to preserve the metadata on the source because the defaults are applied on read retroactively in the "exists default" case? So we keep the metadata on the source and we remove it from the target table, which makes it ineligible for `DeltaTableUtils.removeInternalWriterMetadata` and `ColumnWithDefaultExprUtils.removeDefaultExpressions`.

spark/src/main/scala/org/apache/spark/sql/delta/commands/CreateDeltaTableCommand.scala
```diff
@@ -314,7 +315,29 @@ case class CreateDeltaTableCommand(
...
+        if (schemaContainsExistsColumnMetadata(txn.metadata.schema)) {
+          val newSchema = SchemaMergingUtils.transformColumns(schema) {
+            case (_, field: StructField, _) if field.metadata.contains(
```
c27kwan, line 326: Comment that default is only available as a top level field?

spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
```diff
@@ -683,6 +683,34 @@ trait OptimisticTransactionImpl extends DeltaTransaction
       }
     }
 
+    if (spark.sessionState.conf
```
cstavr, line 686: Can we move this code to a separate function? This method is already way too big.

scottsand-db (approved):

PR #4143, opened 2025-02-11
[Spark] Relax type check in `SchemaUtils.normalizeColumnNamesInDataType`

ala:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR modifies the nested filed name normalization code. This code is used when we write to a Delta table. It makes sure that the names of all the columns in the source data (meaning, the data that we are writing to the table) match the target table column names in terms of case. For example, if the source contains column `address.CITY`, while the table contains column `address.city`, the source column is renamed to match the table.

If the schema of the source data and the table data differs enough to interfere with the name normalization, we log `delta.assertions.schemaNormalization.nonNestedTypeMismatch`. However, this check is too restrictive, and as a result, we log this assertion too often. This PR relaxes the check. We now assume that any two atomic columns can be matched to each other.

## How was this patch tested?

New tests in `SchemaUtilsSuite`.

No.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/schema/SchemaUtils.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/schema/SchemaUtils.scala
@@ -286,8 +286,10 @@ def normalizeColumnNamesInDataType(
         // When schema evolution adds a new column during MERGE, it can be represented with
         // a NullType in the schema of the data written by the MERGE.
         sourceDataType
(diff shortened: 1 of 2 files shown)


c27kwan (approved):
lgtm! Don't forget to tag the PR title with [Spark] prefix. I would also remove the whitespace after `SchemaUtils.` in the PR title. :)

ala:
All right. Back to the original minimal change then.

PR #4144, opened 2025-02-11
Preserve enablement for converting Iceberg table with bucket partition to unpartitioned delta table in table property

ChengJiX:
## Description
Once bucket partition to nonpartition is enabled for a table, the clone from iceberg table will always convert bucket partition to non-partition for that table, regardless of if the flag is turned off or not in the future. This is to make those tables forward compatible with future Deltas where "truly" bucket conversion is supported so they will keep the old behavior and won't make regressions

## How was this patch tested?
Existing UTs


foss-contributor:
@ChengJiX, thank you for your contribution! This pull request has been inactive for over 18 months and will be closed in one week. To keep it open, please leave a comment on this pull request or contact us in the `#contributors` channel on [Delta Lake Slack](https://go.delta.io/slack). We appreciate your involvement in Delta Lake!

PR #4146, opened 2025-02-11
[Spark] Fix metadata cleanup by retaining files required for log reconstruction

### LINKED ISSUE #606: MetadataCleanUp may make the present versions not available
MetadataCleanup will delete the expired delta logs(json and checkpoint.parquet). But if the present version depends on these which will be cleaned up, it cannot replay to the whole commits. For example: we have delta logs from version 0 to version 10 as following: 000.json ~ 009.json, 010.json, 010.checkpoint.parquet.

When Commit 10 is operated, MetadataCleanup works. If we assume the logs before 9 (not contained) should be cleaned up, then the rest files are: 009.json, 010.json, 010.checkpoint.parquet.
In fact, Version 9 is not  available, and only Version 10 is shown by `desc history`.

felipepessoto:
- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Delete eligible delta log files only if there's a checkpoint newer than them before the cutoff window.

Resolves #606

Unit tests based on https://github.com/delta-io/delta/pull/2673

## How was this patch tested?
Unit Tests

Yes, tables with low rate of commit/checkpoints will have increased log retention beyond the cutoff window

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
@@ -865,10 +865,10 @@ object DeltaHistoryManager extends DeltaLogging {
             currentFile.getLen, currentFile.isDirectory, currentFile.getReplication,
             currentFile.getBlockSize, lastFile.getModificationTime + 1, currentFile.getPath)
         } else if (FileNames.isCheckpointFile(currentFile)
-          && shouldDeleteFile(currentFile)
+          && shouldDeleteFile(lastFile)
           && (versionGetter(currentFile.getPath) > versionGetter(lastFile.getPath))) {
-          // Need to keep at least one expired checkpoint
-          // Version check is to handle multi-part checkpoint
+          // Version check is to handle multi-part checkpoint.
+          // We don't want to flush part2 only because the part1 is expired
           flushBuffer()
           continueBuffering = false
         }
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
@@ -809,6 +809,9 @@ object DeltaHistoryManager extends DeltaLogging {
     private val maybeDeleteFiles = new mutable.ArrayBuffer[FileStatus]()
     private var lastFile: FileStatus = _
     private var hasNextCalled: Boolean = false
+    // A map to keep track of multi-part checkpoints.
+    val checkpointMap = new scala.collection.mutable.HashMap[(Long, Int),
+      collection.mutable.Buffer[FileStatus]]()
 
     private def init(): Unit = {
       if (underlying.hasNext) {
@@ -828,7 +831,7 @@ object DeltaHistoryManager extends DeltaLogging {
      * Files need a time adjustment if their timestamp isn't later than the lastFile.
      */
     private def needsTimeAdjustment(file: FileStatus): Boolean = {
-      isFileVersionGreaterThanLastFileVersion(file) &&
+      versionGetter(lastFile.getPath) < versionGetter(file.getPath) &&
         lastFile.getModificationTime >= file.getModificationTime
     }
 
@@ -863,24 +866,35 @@ object DeltaHistoryManager extends DeltaLogging {
           currentFile = new FileStatus(
             currentFile.getLen, currentFile.isDirectory, currentFile.getReplication,
             currentFile.getBlockSize, lastFile.getModificationTime + 1, currentFile.getPath)
-        } else if (FileNames.isCheckpointFile(currentFile)
-          && isFileVersionGreaterThanLastFileVersion(currentFile)) {
+          maybeDeleteFiles.append(currentFile)
+        } else if (FileNames.isCheckpointFile(currentFile)) {
           // Only flush the buffer when find a checkpoint. This is because we don't want to delete
           // the delta log files unless we have a checkpoint to ensure that non-expired subsequent
           // delta logs are valid.
+          val numParts = FileNames.numCheckpointParts(currentFile.getPath)
 
-          // Version check is to handle multi-part checkpoint.
-          // We don't want to flush part1 only because the part2 is expired
-          flushBuffer()
-          continueBuffering = false
+          if (numParts.isEmpty) { // Single-part or V2
+            flushBuffer()
+            maybeDeleteFiles.append(currentFile)
+            continueBuffering = false
+          } else {
+            // Multi-part checkpoint
+            val mpKey = versionGetter(currentFile.getPath) -> numParts.get
+            val partBuffer = checkpointMap.getOrElse(mpKey, mutable.ArrayBuffer())
+            partBuffer.append(currentFile)
+            checkpointMap.put(mpKey, partBuffer)
+            if (numParts.get == partBuffer.size) {
+              flushBuffer()
+              partBuffer.foreach(f => maybeDeleteFiles.append(f))
+              checkpointMap.remove(mpKey)
+              continueBuffering = false
+            }
+          }
+        } else {
+          maybeDeleteFiles.append(currentFile)
         }
-        maybeDeleteFiles.append(currentFile)
         lastFile = currentFile
       }
-    }
-
-    def isFileVersionGreaterThanLastFileVersion(file: FileStatus): Boolean = {
-      versionGetter(file.getPath) > versionGetter(lastFile.getPath)
     }
 
     override def hasNext: Boolean = {
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
@@ -859,16 +859,18 @@ object DeltaHistoryManager extends DeltaLogging {
       while (continueBuffering && underlying.hasNext) {
         var currentFile = underlying.next()
         require(currentFile != null, "FileStatus iterator returned null")
-
         if (needsTimeAdjustment(currentFile)) {
           currentFile = new FileStatus(
             currentFile.getLen, currentFile.isDirectory, currentFile.getReplication,
             currentFile.getBlockSize, lastFile.getModificationTime + 1, currentFile.getPath)
         } else if (FileNames.isCheckpointFile(currentFile)
-          && shouldDeleteFile(lastFile)
           && (versionGetter(currentFile.getPath) > versionGetter(lastFile.getPath))) {
+          // Only flush the buffer when find a checkpoint. This is because we don't want to delete
+          // the delta log files unless we have a checkpoint to ensure that non-expired subsequent
+          // delta logs are valid.
+
           // Version check is to handle multi-part checkpoint.
-          // We don't want to flush part2 only because the part1 is expired
+          // We don't want to flush part1 only because the part2 is expired
           flushBuffer()
           continueBuffering = false
         }
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
@@ -856,24 +856,23 @@ object DeltaHistoryManager extends DeltaLogging {
      */
     private def queueFilesInBuffer(): Unit = {
       var continueBuffering = true
-      while (continueBuffering) {
-        if (!underlying.hasNext) {
-          flushBuffer()
-          return
-        }
-
+      while (continueBuffering && underlying.hasNext) {
         var currentFile = underlying.next()
         require(currentFile != null, "FileStatus iterator returned null")
+
         if (needsTimeAdjustment(currentFile)) {
           currentFile = new FileStatus(
             currentFile.getLen, currentFile.isDirectory, currentFile.getReplication,
             currentFile.getBlockSize, lastFile.getModificationTime + 1, currentFile.getPath)
-          maybeDeleteFiles.append(currentFile)
-        } else {
+        } else if (FileNames.isCheckpointFile(currentFile)
+          && shouldDeleteFile(currentFile)
+          && (versionGetter(currentFile.getPath) > versionGetter(lastFile.getPath))) {
+          // Need to keep at least one expired checkpoint
+          // Version check is to handle multi-part checkpoint
           flushBuffer()
-          maybeDeleteFiles.append(currentFile)
           continueBuffering = false
         }
+        maybeDeleteFiles.append(currentFile)
         lastFile = currentFile
       }
     }
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
@@ -867,10 +867,10 @@ object DeltaHistoryManager extends DeltaLogging {
             currentFile.getLen, currentFile.isDirectory, currentFile.getReplication,
             currentFile.getBlockSize, lastFile.getModificationTime + 1, currentFile.getPath)
           maybeDeleteFiles.append(currentFile)
-        } else if (FileNames.isCheckpointFile(currentFile)) {
-          // Only flush the buffer when find a checkpoint. This is because we don't want to delete
-          // the delta log files unless we have a checkpoint to ensure that non-expired subsequent
-          // delta logs are valid.
+        } else if (FileNames.isCheckpointFile(currentFile) && currentFile.getLen > 0) {
+          // Only flush the buffer when we find a checkpoint. This is because we don't want to
+          // delete the delta log files unless we have a checkpoint to ensure that non-expired
+          // subsequent delta logs are valid.
           val numParts = FileNames.numCheckpointParts(currentFile.getPath)
 
           if (numParts.isEmpty) { // Single-part or V2
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
@@ -828,7 +828,7 @@ object DeltaHistoryManager extends DeltaLogging {
      * Files need a time adjustment if their timestamp isn't later than the lastFile.
      */
     private def needsTimeAdjustment(file: FileStatus): Boolean = {
-      versionGetter(lastFile.getPath) < versionGetter(file.getPath) &&
+      isFileVersionGreaterThanLastFileVersion(file) &&
         lastFile.getModificationTime >= file.getModificationTime
     }
 
@@ -864,7 +864,7 @@ object DeltaHistoryManager extends DeltaLogging {
             currentFile.getLen, currentFile.isDirectory, currentFile.getReplication,
             currentFile.getBlockSize, lastFile.getModificationTime + 1, currentFile.getPath)
         } else if (FileNames.isCheckpointFile(currentFile)
-          && (versionGetter(currentFile.getPath) > versionGetter(lastFile.getPath))) {
+          && isFileVersionGreaterThanLastFileVersion(currentFile)) {
           // Only flush the buffer when find a checkpoint. This is because we don't want to delete
           // the delta log files unless we have a checkpoint to ensure that non-expired subsequent
           // delta logs are valid.
@@ -877,6 +877,10 @@ object DeltaHistoryManager extends DeltaLogging {
         maybeDeleteFiles.append(currentFile)
         lastFile = currentFile
       }
+    }
+
+    def isFileVersionGreaterThanLastFileVersion(file: FileStatus): Boolean = {
+      versionGetter(file.getPath) > versionGetter(lastFile.getPath)
     }
 
     override def hasNext: Boolean = {
--- a/spark/src/test/scala/org/apache/spark/sql/delta/DeltaRetentionSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/DeltaRetentionSuite.scala
@@ -509,6 +509,295 @@ class DeltaRetentionSuite extends QueryTest
     }
   }
 
+  (Seq(("Default", Seq.empty[(String, String)])) ++ CheckpointPolicy.ALL.map {
+    case CheckpointPolicy.Classic =>
+      Seq(
+        ("Classic", Seq(
+          DeltaConfigs.CHECKPOINT_POLICY.defaultTablePropertyKey -> CheckpointPolicy.Classic.name)),
+        ("Multipart", Seq(
+          DeltaConfigs.CHECKPOINT_POLICY.defaultTablePropertyKey -> CheckpointPolicy.Classic.name,
+          DeltaSQLConf.DELTA_CHECKPOINT_PART_SIZE.key -> "1"))
+      )
+    case CheckpointPolicy.V2 =>
+      V2Checkpoint.Format.ALL_AS_STRINGS.map { v2CheckpointFormat =>
+        (s"V2 $v2CheckpointFormat",
+          Seq(DeltaConfigs.CHECKPOINT_POLICY.defaultTablePropertyKey -> CheckpointPolicy.V2.name,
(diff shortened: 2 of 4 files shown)


felipepessoto:
@andreaschat-db, could you help with the `Metadata cleanup respects requireCheckpointProtectionBeforeVersion` test failure, please?

I couldn't find any info in Delta protocol about checkpointProtection table feature

```
[info] - Metadata cleanup respects requireCheckpointProtectionBeforeVersion *** FAILED ***
[info]   HashSet(10, 14, 6, 9, 13, 12, 7, 11, 8, 15) did not equal HashSet(10, 14, 9, 13, 12, 11, 8, 15) (CheckpointProtectionTestUtilsMixin.scala:91)
[info]   Analysis:
[info]   HashSet(missingInRight: [6, 7])
[info]   org.scalatest.exceptions.TestFailedException:
[info]   at org.scalatest.Assertions.newAssertionFailedException(Assertions.scala:472)
[info]   at org.scalatest.Assertions.newAssertionFailedException$(Assertions.scala:471)
[info]   at org.scalatest.Assertions$.newAssertionFailedException(Assertions.scala:1231)
[info]   at org.scalatest.Assertions$AssertionsHelper.macroAssert(Assertions.scala:1295)
[info]   at org.apache.spark.sql.delta.CheckpointProtectionTestUtilsMixin.$anonfun$testRequireCheckpointProtectionBeforeVersion$1(CheckpointProtectionTestUtilsMixin.scala:91)
[info]   at org.apache.spark.sql.delta.CheckpointProtectionTestUtilsMixin.$anonfun$testRequireCheckpointProtectionBeforeVersion$1$adapted(CheckpointProtectionTestUtilsMixin.scala:38)
[info]   at org.apache.spark.sql.delta.test.DeltaSQLTestUtils.withTempDir(DeltaSQLTestUtils.scala:52)
[info]   at org.apache.spark.sql.delta.test.DeltaSQLTestUtils.withTempDir$(DeltaSQLTestUtils.scala:50)
[info]   at org.apache.spark.sql.delta.DeltaRetentionSuite.withTempDir(DeltaRetentionSuite.scala:38)
[info]   at org.apache.spark.sql.delta.test.DeltaSQLTestUtils.withTempDir(DeltaSQLTestUtils.scala:35)
[info]   at org.apache.spark.sql.delta.test.DeltaSQLTestUtils.withTempDir$(DeltaSQLTestUtils.scala:34)
[info]   at org.apache.spark.sql.delta.DeltaRetentionSuite.withTempDir(DeltaRetentionSuite.scala:38)
[info]   at org.apache.spark.sql.delta.CheckpointProtectionTestUtilsMixin.testRequireCheckpointProtectionBeforeVersion(CheckpointProtectionTestUtilsMixin.scala:38)
[info]   at org.apache.spark.sql.delta.CheckpointProtectionTestUtilsMixin.testRequireCheckpointProtectionBeforeVersion$(CheckpointProtectionTestUtilsMixin.scala:30)
[info]   at org.apache.spark.sql.delta.DeltaRetentionSuite.testRequireCheckpointProtectionBeforeVersion(DeltaRetentionSuite.scala:38)
[info]   at org.apache.spark.sql.delta.DeltaRetentionSuite.$anonfun$new$94(DeltaRetentionSuite.scala:819)
```

andreaschat-db:
> @andreaschat-db, could you help with the `Metadata cleanup respects requireCheckpointProtectionBeforeVersion` test failure, please?
> 
> I couldn't find any info in Delta protocol about checkpointProtection table feature
> 
> ```
> [info] - Metadata cleanup respects requireCheckpointProtectionBeforeVersion *** FAILED ***
> [info]   HashSet(10, 14, 6, 9, 13, 12, 7, 11, 8, 15) did not equal HashSet(10, 14, 9, 13, 12, 11, 8, 15) (CheckpointProtectionTestUtilsMixin.scala:91)
> [info]   Analysis:
> [info]   HashSet(missingInRight: [6, 7])
> [info]   org.scalatest.exceptions.TestFailedException:
> [info]   at org.scalatest.Assertions.newAssertionFailedException(Assertions.scala:472)
> [info]   at org.scalatest.Assertions.newAssertionFailedException$(Assertions.scala:471)
> [info]   at org.scalatest.Assertions$.newAssertionFailedException(Assertions.scala:1231)
> [info]   at org.scalatest.Assertions$AssertionsHelper.macroAssert(Assertions.scala:1295)
> [info]   at org.apache.spark.sql.delta.CheckpointProtectionTestUtilsMixin.$anonfun$testRequireCheckpointProtectionBeforeVersion$1(CheckpointProtectionTestUtilsMixin.scala:91)
> [info]   at org.apache.spark.sql.delta.CheckpointProtectionTestUtilsMixin.$anonfun$testRequireCheckpointProtectionBeforeVersion$1$adapted(CheckpointProtectionTestUtilsMixin.scala:38)
> [info]   at org.apache.spark.sql.delta.test.DeltaSQLTestUtils.withTempDir(DeltaSQLTestUtils.scala:52)
> [info]   at org.apache.spark.sql.delta.test.DeltaSQLTestUtils.withTempDir$(DeltaSQLTestUtils.scala:50)
> [info]   at org.apache.spark.sql.delta.DeltaRetentionSuite.withTempDir(DeltaRetentionSuite.scala:38)
> [info]   at org.apache.spark.sql.delta.test.DeltaSQLTestUtils.withTempDir(DeltaSQLTestUtils.scala:35)
> [info]   at org.apache.spark.sql.delta.test.DeltaSQLTestUtils.withTempDir$(DeltaSQLTestUtils.scala:34)
> [info]   at org.apache.spark.sql.delta.DeltaRetentionSuite.withTempDir(DeltaRetentionSuite.scala:38)
> [info]   at org.apache.spark.sql.delta.CheckpointProtectionTestUtilsMixin.testRequireCheckpointProtectionBeforeVersion(CheckpointProtectionTestUtilsMixin.scala:38)
> [info]   at org.apache.spark.sql.delta.CheckpointProtectionTestUtilsMixin.testRequireCheckpointProtectionBeforeVersion$(CheckpointProtectionTestUtilsMixin.scala:30)
> [info]   at org.apache.spark.sql.delta.DeltaRetentionSuite.testRequireCheckpointProtectionBeforeVersion(DeltaRetentionSuite.scala:38)
> [info]   at org.apache.spark.sql.delta.DeltaRetentionSuite.$anonfun$new$94(DeltaRetentionSuite.scala:819)
> ```

Hi @felipepessoto, please check the doc in TableFeature. AFAIU, your PR is changing which checkpoints are removed and this changes the expectations of the test.

felipepessoto:
@andreaschat-db, I fixed it, could you please review it?

andreaschat-db (commented):
Hi @felipepessoto, thank you for looking at this. I left a couple of comments. However, I am a bit confused about what is exactly the issue we are trying to fix. AFAIU, the `BufferingLogDeletionIterator` is called with a max version. Isn't it guaranteed that at maxVersion + 1 there is a checkpoint? If so, what is the scenario that this can go wrong? Is there a case where we may stop earlier than maxVersion?

felipepessoto:
> Hi @felipepessoto, thank you for looking at this. I left a couple of comments. However, I am a bit confused about what is exactly the issue we are trying to fix. AFAIU, the `BufferingLogDeletionIterator` is called with a max version. Isn't it guaranteed that at maxVersion + 1 there is a checkpoint? If so, what is the scenario that this can go wrong? Is there a case where we may stop earlier than maxVersion?

@andreaschat-db currently we remove the expired files as long they meet the maxVersion and maxTimestamp requirements. The problem is if we have the following:

- Expired: 0.json-5.json 
- Non-expired: 6.json-10.json + 10.checkpoint

The cleanup deletes 0-5, and the versions 6-9 are not expired, but can't be queried because they depend on 0-5 json files to build the Snapshot.
This is a known issue, the protocol has a suggestion to always keep the last expired checkpoint: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#metadata-cleanup. It was my initial implementation, but then I found it causes issues with the checkpointProtection, because it keeps logs files that are not strictly required, so I made more changes to keep only the files we really need.

Basically, I only flush the buffer when I find a checkpoint. Some example:

- Expired: 0.json-5.json 
- Non-expired: 6.json-10.json + 10.checkpoint
- Result: Don't delete anything

- Expired: 0.json-15.json + 10.checkpoint
- Non-expired: 16.json-20.json + 20.checkpoint
- Result: Deletes 0.json to 9.json

- Expired: 0.json-19.json + 10.checkpoint
- Non-expired: 20.json + 20.checkpoint
- Result: Deletes 0.json to 19.json + 10.checkpoint.  (in my initial implementation I was keeping versions starting from 10, the latest expired checkpoint)

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -856,24 +856,25 @@ object DeltaHistoryManager extends DeltaLogging {
      */
     private def queueFilesInBuffer(): Unit = {
       var continueBuffering = true
-      while (continueBuffering) {
-        if (!underlying.hasNext) {
-          flushBuffer()
-          return
-        }
-
+      while (continueBuffering && underlying.hasNext) {
```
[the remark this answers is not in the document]
felipepessoto: We only want to flush when we find a checkpoint, if the underlying ended we want to discard the items in the buffer.

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -856,24 +856,25 @@ object DeltaHistoryManager extends DeltaLogging {
...
         var currentFile = underlying.next()
         require(currentFile != null, "FileStatus iterator returned null")
         if (needsTimeAdjustment(currentFile)) {
           currentFile = new FileStatus(
             currentFile.getLen, currentFile.isDirectory, currentFile.getReplication,
             currentFile.getBlockSize, lastFile.getModificationTime + 1, currentFile.getPath)
-          maybeDeleteFiles.append(currentFile)
-        } else {
+        } else if (FileNames.isCheckpointFile(currentFile)
+          && (versionGetter(currentFile.getPath) > versionGetter(lastFile.getPath))) {
```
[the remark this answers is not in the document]
felipepessoto: I can change, but I'm not sure if the result will be clear, not sure if this is what you are suggesting?:

`isFileVersionBiggerThanLastFileVersion(currentFile)`

spark/src/test/scala/org/apache/spark/sql/delta/DeltaRetentionSuite.scala
```diff
@@ -713,20 +849,18 @@ class DeltaRetentionSuite extends QueryTest
       testRequireCheckpointProtectionBeforeVersion(
         createNumCommitsOutsideRetentionPeriod = 8,
         createNumCommitsWithinRetentionPeriod = 8,
-        createCheckpoints = Set(1),
+        createCheckpoints = Set(1, 8),
```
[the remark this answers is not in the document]
felipepessoto: I'm not sure what was the corner case tested here. What would be the other parameters?

spark/src/test/scala/org/apache/spark/sql/delta/DeltaRetentionSuite.scala
```diff
@@ -613,29 +749,29 @@ class DeltaRetentionSuite extends QueryTest
       testRequireCheckpointProtectionBeforeVersion(
         createNumCommitsOutsideRetentionPeriod = 2,
         createNumCommitsWithinRetentionPeriod = 14,
-        createCheckpoints = Set(1),
+        createCheckpoints = Set(2),
```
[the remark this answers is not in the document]
felipepessoto: Updated. For the 3rd case I also had to change expectedCommitsAfterCleanup to start from 0, since we don't have a checkpoint for version 2, the cleanup is required to keep checkpoint 1, and then because of protected checkpoint logic, we can't clean up 0 if 1 isn't deleted too

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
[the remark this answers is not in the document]
andreaschat-db: Yes `isFileVersionGreaterThanLastFileVersion`.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaRetentionSuite.scala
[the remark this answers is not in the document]
andreaschat-db: As I recall, the idea was to have a single checkpoint close to the boundary. But now this is shadowed by checkpoint 8. Perhaps we could reproduce the same behaviour with something like:

```
        createNumCommitsOutsideRetentionPeriod = 1,
        createNumCommitsWithinRetentionPeriod = 15,
        createCheckpoints = Set(1),
        requireCheckpointProtectionBeforeVersion = 10,
        expectedCommitsAfterCleanup = (1 to 15),
        expectedCheckpointsAfterCleanup = Set(10))
```
  
  If that does not work we can leave it as it is.

andreaschat-db (approved):

andreaschat-db:
@prakharjain09 could you also please have a look?

spark/src/test/scala/org/apache/spark/sql/delta/DeltaRetentionSuite.scala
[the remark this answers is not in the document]
felipepessoto: The `expectedCheckpointsAfterCleanup` is Set(1,10) in that case. I'm leaving as is for now. The same for the below

spark/src/test/scala/org/apache/spark/sql/delta/DeltaRetentionSuite.scala
```diff
@@ -713,20 +849,18 @@ class DeltaRetentionSuite extends QueryTest
...
         expectedCommitsAfterCleanup = (8 to 15),
-        // This is a bit weird. Cleanup should had created a checkpoint at 8.
-        expectedCheckpointsAfterCleanup = Set(10))
+        expectedCheckpointsAfterCleanup = Set(8, 10))
 
       testRequireCheckpointProtectionBeforeVersion(
         createNumCommitsOutsideRetentionPeriod = 8,
         createNumCommitsWithinRetentionPeriod = 8,
-        createCheckpoints = Set(0),
+        createCheckpoints = Set(0, 8),
```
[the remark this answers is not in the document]
felipepessoto: Same as https://github.com/delta-io/delta/pull/4146/files#r2097911433

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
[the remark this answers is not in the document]
felipepessoto: Done, thanks

felipepessoto:
+ @allisonport-db , could we merge it before 4.0 release?

raveeram-db:
@dhruvarya-db do you have some cycles to review this PR please?

felipepessoto:
Hi @dhruvarya-db, let me know if you have any questions about this change. Thanks

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -856,28 +856,33 @@ object DeltaHistoryManager extends DeltaLogging {
...
         require(currentFile != null, "FileStatus iterator returned null")
         if (needsTimeAdjustment(currentFile)) {
           currentFile = new FileStatus(
             currentFile.getLen, currentFile.isDirectory, currentFile.getReplication,
             currentFile.getBlockSize, lastFile.getModificationTime + 1, currentFile.getPath)
-          maybeDeleteFiles.append(currentFile)
-        } else {
+        } else if (FileNames.isCheckpointFile(currentFile)
+          && isFileVersionGreaterThanLastFileVersion(currentFile)) {
+          // Only flush the buffer when find a checkpoint. This is because we don't want to delete
```
dhruvarya-db: IMO we should also check whether it is a complete checkpoint in the case of a multi-part checkpoint.

felipepessoto: Since multi-part checkpoint are deprecated, instead, can I check if it is not a multi-part checkpoint?

felipepessoto: @dhruvarya-db, since multi-part is deprecated, can I check if the checkpoint is not a multi-part checkpoint?

dhruvarya-db: V2 Checkpoints (which replace multi-part checkpoints) is still an opt-in feature. My fear is that with this fix, users will start expecting log cleanup to always retain a recreatable history but any incomplete multi-part checkpoint will break that.

felipepessoto: I'm actually saying the opposite, changing this:

```
        } else if (FileNames.isCheckpointFile(currentFile)
          && isFileVersionGreaterThanLastFileVersion(currentFile)) {
```

to this:

```
        } else if (FileNames.isCheckpointFile(currentFile)
          && isFileVersionGreaterThanLastFileVersion(currentFile)
          && IsNotAMultiPartCheckpoint(currentFile)) {
```


That means multi-part checkpoint won't flush the buffer.

Checking the complete multi-part checkpoint would tricky because we would need to accumulate them in a different buffer (a map per version?), instead of the `maybeDeleteFiles.append(currentFile)` below, and add them to the buffer at once, after the flush.

dhruvarya-db: Could this potentially result in us never doing a cleanup if a table only has multi-part checkpoints? To handle multi-part checkpoints, we could potentially reuse the multi-part checkpoint logic from [DeltaHistoryManager](https://github.com/delta-io/delta/blob/2d05055e8f43716e4c30e1b2da681b73fa154b9f/spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala#L412)

felipepessoto: Right, that would be the trade-off.
Yes I'm using the exact same logic, I created this method:

```
    def isSingleOrMultiPartComplete(fileName: FileStatus): Boolean = {
      val parts = FileNames.numCheckpointParts(fileName.getPath)
      if (parts.isEmpty) {
        true
      } else {
        // if we have a multi-part checkpoint, we need to check that all parts exist
        val checkpointVersion = FileNames.checkpointVersion(fileName)
        val numParts = parts.get
        val preCount = checkpointMap.getOrElse(checkpointVersion -> numParts, 0)
        val newCount = preCount + 1
        checkpointMap.put(checkpointVersion -> numParts, newCount)
        numParts == newCount
      }
    }
```

And changed the if to:

```
} else if (FileNames.isCheckpointFile(currentFile)
          && isSingleOrMultiPartComplete(currentFile)) {
```

The problem is, when isSingleOrMultiPartComplete finally returns true, when currentFile is the last part, all the previous parts are already in the `maybeDeleteFiles`.

I can handle this but will be ugly. I need to keep a buffer of all parts of the last multi-part checkpoint, and only add it to `maybeDeleteFiles` when I find it is complete, or the next file is not part of that same multi-part checkpoint, meaning the checkpoint is incomplete. Let me try that

felipepessoto: @dhruvarya-db, I updated it to support multi-part checkpoint

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -856,23 +859,39 @@ object DeltaHistoryManager extends DeltaLogging {
...
+      while (continueBuffering && underlying.hasNext) {
         var currentFile = underlying.next()
         require(currentFile != null, "FileStatus iterator returned null")
         if (needsTimeAdjustment(currentFile)) {
           currentFile = new FileStatus(
             currentFile.getLen, currentFile.isDirectory, currentFile.getReplication,
             currentFile.getBlockSize, lastFile.getModificationTime + 1, currentFile.getPath)
           maybeDeleteFiles.append(currentFile)
+        } else if (FileNames.isCheckpointFile(currentFile)) {
+          // Only flush the buffer when find a checkpoint. This is because we don't want to delete
```
dhruvarya-db: [nit] 
```suggestion
          // Only flush the buffer when we find a checkpoint. This is because we don't want to delete
```

spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala
```diff
@@ -856,23 +859,39 @@ object DeltaHistoryManager extends DeltaLogging {
...
-
+      while (continueBuffering && underlying.hasNext) {
         var currentFile = underlying.next()
         require(currentFile != null, "FileStatus iterator returned null")
         if (needsTimeAdjustment(currentFile)) {
           currentFile = new FileStatus(
             currentFile.getLen, currentFile.isDirectory, currentFile.getReplication,
             currentFile.getBlockSize, lastFile.getModificationTime + 1, currentFile.getPath)
           maybeDeleteFiles.append(currentFile)
+        } else if (FileNames.isCheckpointFile(currentFile)) {
```
dhruvarya-db: For extra precaution, we should also ignore any checkpoint files with size 0

spark/src/test/scala/org/apache/spark/sql/delta/DeltaRetentionSuite.scala
```diff
@@ -509,6 +509,142 @@ class DeltaRetentionSuite extends QueryTest
...
+            }
+          }
+        } else {
+          spark.sql(sqlCommand).collect()
+        }
+      }
+    }
+    }
+  }
+  }
```
dhruvarya-db, line 646: Let's also add a few test cases for complete and incomplete multi-part checkpoints?

dhruvarya-db (commented):
Thanks for fixing this @felipepessoto ! The change makes sense to me, left a couple of suggestions.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaRetentionSuite.scala
felipepessoto: The `cleanup does not delete the checkpoint if it is required by non-expired versions.` covers the multi-part complete case. I added "cleanup does not delete the JSON logs if the multi-part checkpoint is incomplete."

felipepessoto: @dhruvarya-db, please let me know if it looks good now

Thanks.

felipepessoto:
Thanks @dhruvarya-db. I'm not sure what are the next steps to get it merged to master and if I should open a PR to 4.0 branch?

dhruvarya-db:
Looks like 4.0.0 has already been [released](https://github.com/delta-io/delta/releases). @allisonport-db does it make sense to backport this to the 4.0 branch? It might be useful in case there is another 4.0 maintenance release?

felipepessoto:
Hi @allisonport-db, could you help with the merge, and should I create a PR to cherry pick it to 4.0?

Thanks

felipepessoto:
Gentle ping @allisonport-db

felipepessoto:
@scottsand-db could you help with merge and help decide if it make sense to backport this to the 4.0 branch?

scottsand-db:
@dhruvarya-db and @felipepessoto -- is this considered a bug or correctness fix? Thats' the only reason we would want to backport it to the 4.0 branch

felipepessoto:
Hard to judge since protocol is not strict about it, it only provides recommendation. But the consequences can be severe, as data loss since user can't reconstruct old version.

felipepessoto:
@vkorukanti, @scottsand-db tests are all green after the rebase

felipepessoto:
Thanks @vkorukanti

@scottsand-db, since it is not following protocol recommendation and can cause data loss, I believe it fits on bug fix category. I created a PR for 4.0 branch: https://github.com/delta-io/delta/pull/4784

PR #4147, opened 2025-02-11
[Kernel] [Cleanup] Move `SnapshotManager::checkpoint` to `Checkpointer` class

scottsand-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

This PR moves `SnapshotManager::checkpoint` to `Checkpointer` class. This reduces coupling and increases cohesion.

I also tidy up some code and logger statements along the way.

## How was this patch tested?

Just a refactor. Existing UTs.

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/checkpoints/Checkpointer.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/checkpoints/Checkpointer.java
@@ -16,17 +16,27 @@
 package io.delta.kernel.internal.checkpoints;
 
 import static io.delta.kernel.internal.DeltaErrors.wrapEngineExceptionThrowsIO;
+import static io.delta.kernel.internal.TableConfig.EXPIRED_LOG_CLEANUP_ENABLED;
+import static io.delta.kernel.internal.TableConfig.LOG_RETENTION;
+import static io.delta.kernel.internal.TableFeatures.validateWriteSupportedTable;
+import static io.delta.kernel.internal.snapshot.MetadataCleanup.cleanupExpiredLogs;
 import static io.delta.kernel.internal.util.Utils.singletonCloseableIterator;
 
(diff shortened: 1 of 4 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/checkpoints/Checkpointer.java
```diff
@@ -16,29 +16,104 @@
...
+  ////////////////////////////////
+
   private static final Logger logger = LoggerFactory.getLogger(Checkpointer.class);
 
+  private static final int READ_LAST_CHECKPOINT_FILE_MAX_RETRIES = 3;
+
   /** The name of the last checkpoint file */
   public static final String LAST_CHECKPOINT_FILE_NAME = "_last_checkpoint";
 
+  public static void checkpoint(Engine engine, Clock clock, SnapshotImpl snapshot)
```
scottsand-db, line 59: nearly a verbatim copy. no logic change. i change the inputs (snapshot instead of version), define tablePath and logPath and version variables, and make some minor logger improvements.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -602,10 +533,7 @@ private Optional<Long> getStartCheckpointVersion(Engine engine, Optional<Long> v
                       });
             })
         .orElseGet(
-            () -> {
-              logger.info("Loading last checkpoint from the _last_checkpoint file");
```
scottsand-db, line 606: The invoker of a method should only log additional context that the invokee would not know about.

In this case, the invokee should be responsible for printing this.

huan233usc (approved):

vkorukanti (approved):
makes sense to separate it out.

PR #4150, opened 2025-02-12
[Spark] Pass catalog table through TahoeLogFileIndex

ctring:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
`TahoeLogFileIndex` contains `DeltaLog::update` calls but it does not pass in the catalog table yet. This causes the check for catalog table existence unreliable in `DeltaLog::update`. This PR fixes that by adding the catalog table argument to `TahoeLogFileIndex` and pass the catalog table from `DeltaLog::createRelation`

## How was this patch tested?

Run existing unit tests

No

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/delta/test/DeltaTestImplicits.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/test/DeltaTestImplicits.scala
@@ -24,6 +24,7 @@ import org.apache.spark.sql.delta.actions.{Action, AddFile, Metadata, Protocol}
 import org.apache.spark.sql.delta.catalog.DeltaTableV2
 import org.apache.spark.sql.delta.commands.optimize.OptimizeMetrics
 import org.apache.spark.sql.delta.coordinatedcommits.TableCommitCoordinatorClient
+import org.apache.spark.sql.delta.files.TahoeLogFileIndex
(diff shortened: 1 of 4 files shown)


spark/src/test/scala/org/apache/spark/sql/delta/test/DeltaTestImplicits.scala
```diff
@@ -181,6 +182,12 @@ object DeltaTestImplicits {
     def snapshot: Snapshot = deltaTable.initialSnapshot
   }
 
+  implicit class TahoeLogFileIndexObjectTestHelper(index: TahoeLogFileIndex.type) {
+    def apply(spark: SparkSession, deltaLog: DeltaLog): TahoeLogFileIndex = {
```
prakharjain09, line 186: Are there existing tests which uses the 2 arg constructor?

ctring: Yes. There are quite a few in this file: https://github.com/delta-io/delta/blob/5aee6354d167763922276f557ad45152691d1525/spark/src/test/scala/org/apache/spark/sql/delta/DeltaSuite.scala#L1446

Issue #4152, opened 2025-02-13
[PROTOCOL RFC] Checkpoint Protection


andreaschat-db:
### Overview

This RFC is about introducing a new Writer feature named `checkpointProtection`. When the feature is present in the protocol, no checkpoint removal/creation before `delta.requireCheckpointProtectionBeforeVersion` is allowed during metadata cleanup, unless everything is cleaned up in one go.

### Motivation

The motivation is to improve the drop feature functionality. Today, dropping a feature requires truncating the history of a Delta table at the version boundary where the feature is removed from the protocol. This is necessary because the Delta protocol only safely supports table protocols that are monotonically increasing with table versions. And because it is unsafe to truncate the history of a Delta table while transactions are running, dropping a feature requires a 24-hour wait time to avoid corrupting the table.

We can improve this process by setting up the table's history (including checkpoints) in such a way that older readers will be able to handle it correctly, i.e., to read correctly at versions for which they support the read protocol, and to reject reading of versions for which they do not support all features. The `checkpointProtection` feature is needed to ensure that this very specific setup of the history _stays in place_ until the feature removal is cleaned up from the retained version history.

A key component of this solution is a special set of protected checkpoints at the DROP FEATURE boundary that are guaranteed to persist until all history is truncated up to the checkpoints in one go. These checkpoints act as barriers that hide unsupported commit
records behind them. By "hiding", we mean that older readers will not need to replay those commits that they _don't_ support in order to reconstruct the table state at a later version that they _do_ support. With the `checkpointProtection`, we can guarantee these checkpoints will persist until history is truncated.

Furthermore, with the new drop feature method, it is no longer guaranteed that protocols are monotonically increasing. This means that clients that validate against the latest protocol can no longer assume that they can also operate on earlier versions correctly. In particular, writers are allowed to create checkpoints at earlier versions, but if they do this without checking the protocol at that specific version, and then they may write corrupted checkpoints for table versions for which they do not support the protocol. The `checkpointProtection` table feature also protects against these cases by requiring writers to check the protocol versions at historical table versions before creating a new checkpoint.

With these changes, we can drop table features without needing to truncate history. More importantly, they simplify the drop feature user journey by requiring a single execution of the DROP FEATURE command.

### Details

The `checkpointProtection` is a Writer feature that protects checkpoints before the version indicated by table property `delta.requireCheckpointProtectionBeforeVersion`, and that forbids writers from creating checkpoints before that version unless they confirm that they support the table protocol at that version.

[...]

PR #4153, opened 2025-02-13
[SPARK][PROTOCOL RFC] Checkpoint Protection

andreaschat-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This is an RFC for the `checkpointProtection` table feature. The related issue is https://github.com/delta-io/delta/issues/4152.

## How was this patch tested?

N/A.

No.

### DIFF
--- a/protocol_rfcs/checkpoint-protection.md
+++ b/protocol_rfcs/checkpoint-protection.md
@@ -16,25 +16,24 @@ With these changes, we can drop table features in a single command without needi
 
 --------
 
-
-> ***New Section***
+> ***Add a new section at the [Table Features](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#table-features) section***
 # Checkpoint Protection
 
-The `checkpointProtection` is a Writer feature that allows writers to clean up metadata iff metadata can be cleaned up to the `requireCheckpointProtectionBeforeVersion` table property in one go.
+The `checkpointProtection` is a Writer feature that allows writers to clean up metadata if and only if metadata can be cleaned up to the `requireCheckpointProtectionBeforeVersion` table property in one go.
 
 Enablement:
-- The table must be on Writer Version 7 and Reader Version 1.
+- The table must be at least on Writer Version 7 and Reader Version 1.
 - The feature `checkpointProtection` must exist in the table `protocol`'s `writerFeatures`.
 
 ## Writer Requirements for Checkpoint Protection
 
-For tables with `checkpointProtection` supported in the protocol, writers need to check `requireCheckpointProtectionBeforeVersion` before cleaning up metadata. Metadata clean up can proceed iff metadata can be cleaned up to the `requireCheckpointProtectionBeforeVersion` table property in one go. This means that a single cleanup operation should truncate up to `requireCheckpointProtectionBeforeVersion` as opposed to several cleanup operations truncating in chunks.
+For tables with `checkpointProtection` supported in the protocol, writers need to check `requireCheckpointProtectionBeforeVersion` before cleaning up metadata. Metadata clean up can proceed if and only if metadata can be cleaned up to the `requireCheckpointProtectionBeforeVersion` table property in one go. This means that a single cleanup operation should truncate up to `requireCheckpointProtectionBeforeVersion` as opposed to several cleanup operations truncating in chunks.
 
 There are two exceptions to this rule. If any of the two holds, the rule above can be ignored:
 
 a) The writer does not create any checkpoints during history cleanup and does not erase any checkpoints after the truncation version.
 
-b) The writer verifies it supports all protocols between `[start, min(requireCheckpointProtectionBeforeVersion, targetCleanupVersion)]`.
+b) The writer verifies it supports all protocols in the closed range `[start, min(requireCheckpointProtectionBeforeVersion, targetCleanupVersion)]`.
 
 The `checkpointProtection` feature can only be removed if history is truncated up to at least the `requireCheckpointProtectionBeforeVersion`.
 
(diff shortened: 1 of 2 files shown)


protocol_rfcs/checkpoint-protection.md
```diff
@@ -0,0 +1,43 @@
...
+
+The `checkpointProtection` is a Writer feature that allows writers to clean up metadata iff metadata can be cleaned up to the `requireCheckpointProtectionBeforeVersion` table property in one go.
+
+Enablement:
+- The table must be on Writer Version 7 and Reader Version 1.
+- The feature `checkpointProtection` must exist in the table `protocol`'s `writerFeatures`.
+
+## Writer Requirements for Checkpoint Protection
+
+For tables with `checkpointProtection` supported in the protocol, writers need to check `requireCheckpointProtectionBeforeVersion` before cleaning up metadata. Metadata clean up can proceed iff metadata can be cleaned up to the `requireCheckpointProtectionBeforeVersion` table property in one go. This means that a single cleanup operation should truncate up to `requireCheckpointProtectionBeforeVersion` as opposed to several cleanup operations truncating in chunks.
```
larsk-db, line 31: Write out "if and only if". Not everyone is necessarily familiar with the iff shorthand.

larsk-db: Can you clarify what "in one go" really means? What are the exact invariants that need to be upheld at any point here?

protocol_rfcs/checkpoint-protection.md
```diff
@@ -0,0 +1,43 @@
...
+
+There are two exceptions to this rule. If any of the two holds, the rule above can be ignored:
+
+a) The writer does not create any checkpoints during history cleanup and does not erase any checkpoints after the truncation version.
+
+b) The writer verifies it supports all protocols between `[start, min(requireCheckpointProtectionBeforeVersion, targetCleanupVersion)]`.
```
larsk-db, line 37: Either write "in the closed range" instead of "between" or actually make it two separate values, rather than a single closed range notation.

protocol_rfcs/checkpoint-protection.md
andreaschat-db: Added clarifications.

larsk-db (approved):
LGTM

protocol_rfcs/checkpoint-protection.md
larsk-db: Thanks

protocol_rfcs/checkpoint-protection.md
```diff
@@ -0,0 +1,43 @@
...
+
+The `checkpointProtection` is a Writer feature that allows writers to clean up metadata if and only if metadata can be cleaned up to the `requireCheckpointProtectionBeforeVersion` table property in one go.
+
+Enablement:
+- The table must be at least on Writer Version 7 and Reader Version 1.
+- The feature `checkpointProtection` must exist in the table `protocol`'s `writerFeatures`.
+
+## Writer Requirements for Checkpoint Protection
+
+For tables with `checkpointProtection` supported in the protocol, writers need to check `requireCheckpointProtectionBeforeVersion` before cleaning up metadata. Metadata clean up can proceed if and only if metadata can be cleaned up to the `requireCheckpointProtectionBeforeVersion` table property in one go. This means that a single cleanup operation should truncate up to `requireCheckpointProtectionBeforeVersion` as opposed to several cleanup operations truncating in chunks. Furthermore, before removing checkpoints, all associated commits need to be removed first. This operation should have the same atomicity guarantees (if any) as with the regular metadata cleanup operation.
```
bart-samwel, line 31: Trying to weaken the requirements so that writers can still do incremental cleanup if they can prove that it is safe.

```suggestion
For tables with `checkpointProtection` supported in the protocol:

a) Writers must not clean up any checkpoints for table versions before the version given by table property `delta.requireCheckpointProtectionBeforeVersion`. 

b) Writers must not create new checkpoints for table versions before the version given by table property `delta.requireCheckpointProtectionBeforeVersion` unless they support all of the features in the table protocol at that version.

c) Writers must not clean up version history for table versions for which they do not support the protocol. A writer is allowed to clean up a range of versions if it supports all table features for every version that is being cleaned up. If a writer does not support the protocol for some of the versions that are being cleaned up, then the cleanup is allowed if and only if the cleanup includes _all_ table versions before the version given by  `delta.requireCheckpointProtectionBeforeVersion`. In this case, a single cleanup operation should truncate the history up to that boundary version in one go as opposed to several cleanup operations truncating in chunks.

d) In version history cleanup, writers must remove commits _before_ removing the associated checkpoints, so that requirement (a) is satisfied even during the cleanup.
```

protocol_rfcs/checkpoint-protection.md
```diff
@@ -0,0 +1,43 @@
...
+
+We can allow history truncation at an earlier commit, as long as checkpoints are removed together with the associated commits, and if any of the two following exceptions hold:
+
+a) The writer does not create any checkpoints during history cleanup and does not erase any checkpoints after the truncation version.
+
+b) The writer verifies it supports all protocols in the closed range `[start, min(requireCheckpointProtectionBeforeVersion, targetCleanupVersion)]` (assuming a single checkpoint is created at `targetCleanupVersion`).
```
bart-samwel, line 37: I think this can be completely removed if we make the change above.

bart-samwel (commented):
I made a bunch of suggestions. In particular:

* Rewording to make it clear that the problem is that the Delta protocol only supports monotonically increasing protocol versions.

* Increasing leniency by allowing checkpoint creation for versions where the client understands the protocol.

bart-samwel (approved):

felipepessoto:
@andreaschat-db, since it is usually used temporarily, do we need to add the conditions when the feature can be removed?

andreaschat-db:
> @andreaschat-db, since it is usually used temporarily, do we need to add the conditions when the feature can be removed?

Hi @felipepessoto, when `checkpointProtection` is supported by the table's protocol, all invariants need to be respected. If `checkpointProtection` is dropped, no invariants need to be respected.

PR #4155, opened 2025-02-13
[INFRA][Spark][Iceberg] Separate out the iceberg tests to a separate github action

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [X] Other (INFRA)

## Description

Separates out the iceberg tests to a separate github action. Also makes the unidoc settings dependent on an environmental variable `DISABLE_UNIDOC`. This is useful because due to unidoc, running tests compiles all the projects in the entire repo. If you disable unidoc, you can, for example, run `spark / test` without compiling the iceberg projects.

## How was this patch tested?

CI jobs look good.

No

### DIFF
--- a/.github/workflows/iceberg_test.yaml
+++ b/.github/workflows/iceberg_test.yaml
@@ -8,12 +8,9 @@ jobs:
       matrix:
         # These Scala versions must match those in the build.sbt
         scala: [2.12.18, 2.13.13]
-        # Important: This list of shards must be [0..NUM_SHARDS - 1]
-        shard: [0, 1, 2, 3]
     env:
       SCALA_VERSION: ${{ matrix.scala }}
       # Important: This must be the same as the length of shards in matrix
-      NUM_SHARDS: 4
     steps:
       - uses: actions/checkout@v3
         # TODO we can make this more selective
@@ -90,5 +87,5 @@ jobs:
       - name: Run Scala/Java and Python tests
         # when changing TEST_PARALLELISM_COUNT make sure to also change it in spark_master_test.yaml
         run: |
-          TEST_PARALLELISM_COUNT=4 pipenv run python run-tests.py --group iceberg --shard ${{ matrix.shard }}
+          TEST_PARALLELISM_COUNT=4 pipenv run python run-tests.py --group iceberg
         if: steps.git-diff.outputs.diff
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -178,10 +178,6 @@ def crossSparkSettings(): Seq[Setting[_]] = getSparkVersion() match {
     Antlr4 / antlr4Version := "4.9.3",
     Test / javaOptions ++= Seq("-Dlog4j.configurationFile=log4j2.properties"),
 
-    // Java-/Scala-/Uni-Doc Settings
```
allisonport-db, line 181: This has to be moved to `Unidoc` settings because when unidoc is disabled this fails compilation since it's an unrecognized option.

.github/workflows/iceberg_test.yaml
```diff
@@ -0,0 +1,91 @@
+name: "Delta Iceberg Latest"
+on: [push, pull_request]
+jobs:
+  test:
+    name: "DSL: Scala ${{ matrix.scala }}, Shard ${{ matrix.shard }}"
```
scottsand-db, line 5: nit: `DIL`

.github/workflows/iceberg_test.yaml
```diff
@@ -0,0 +1,91 @@
...
+          eval "$(pyenv virtualenv-init -)"
+          pyenv install 3.8.18
+          pyenv global system 3.8.18
+          pipenv --python 3.8 install
+          # Update the pip version to 24.0. By default `pyenv.run` installs the latest pip version
+          # available. From version 24.1, `pip` doesn't allow installing python packages
+          # with version string containing `-`. In Delta-Spark case, the pypi package generated has
+          # `-SNAPSHOT` in version (e.g. `3.3.0-SNAPSHOT`) as the version is picked up from
+          # the`version.sbt` file.
+          pipenv run pip install pip==24.0 setuptools==69.5.1 wheel==0.43.0
```
scottsand-db, line 65: does this iceberg stuff need python?

allisonport-db: we do use python for packaging iceberg but these packages I think are all for building delta-spark. Tried removing them and we can see

scottsand-db (approved):

PR #4164, opened 2025-02-17
[Spark] Add order-agnostic type widening modes

johanl-db:
## Description
All code paths implementing type widening so far had a clear 'before' and 'after' schema and checked that 'before' could be widened to 'after'.
An additional use case is, given two schemas from two separate flows, to compute wider schema from the two.
To support this, 'bidirectional' widening modes are added to `SchemaMergingUtils.mergeSchemas()`.
These allow picking the wider of the two input types, or, for decimals, picking a decimal type that is wider than each input.

## How was this patch tested?
Added unit tests for `SchemaMergingUtils.mergeSchemas()`.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMode.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMode.scala
@@ -18,26 +18,27 @@ package org.apache.spark.sql.delta
 
 import org.apache.spark.sql.util.ScalaExtensions._
 
-import org.apache.spark.sql.types.AtomicType
+import org.apache.spark.sql.catalyst.analysis.DecimalPrecisionTypeCoercion
+import org.apache.spark.sql.types.{AtomicType, DecimalType}
 
 /**
  * A type widening mode captures a specific set of type changes that are allowed to be applied.
  * Currently:
  *  - NoTypeWidening: No type change is allowed.
- *  - AllTypeWidening: All supported type widening changes are allowed.
- *  - TypeEvolution(uniformIcebergCompatibleOnly = true): Type changes that are eligible to be
- *    applied automatically during schema evolution and that are supported by Iceberg are allowed.
- *  - TypeEvolution(uniformIcebergCompatibleOnly = false): Type changes that are eligible to be
- *    applied automatically during schema evolution are allowed, even if they are not supported by
- *    Iceberg.
- *
- * AllTypeWidening & TypeEvolution also have a "bidirectional" variant: instead of only allowing
- * 'from' to be widened to 'to', these also allow 'to' to be widened to 'from'. Useful when there's
- * no actual relation between 'from' and 'to' and we just want to use the wider type of the two,
- * e.g. when merging two unrelated schemas.
+ *  - AllTypeWidening: Allows widening to the target type using any supported type change.
+ *  - TypeEvolution: Only allows widening to the target type if the type change is eligible to be
+ *      applied automatically during schema evolution.
+ *  - AllTypeWideningToCommonWiderType: Allows widening to a common (possibly different) wider type
+ *      using any supported type change.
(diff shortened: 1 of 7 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMode.scala
```diff
@@ -16,44 +16,96 @@
...
+
+import org.apache.spark.sql.catalyst.analysis.DecimalPrecisionTypeCoercion
+import org.apache.spark.sql.types.{AtomicType, DecimalType}
 
 /**
  * A type widening mode captures a specific set of type changes that are allowed to be applied.
  * Currently:
  *  - NoTypeWidening: No type change is allowed.
- *  - AllTypeWidening: All supported type widening changes are allowed.
- *  - TypeEvolution(uniformIcebergCompatibleOnly = true): Type changes that are eligible to be
```
olaky, line 26: I think we should still mention in here that some of the have the Iceberg compat flag

spark/src/test/scala/org/apache/spark/sql/delta/schema/SchemaUtilsSuite.scala
```diff
@@ -2569,6 +2573,244 @@ class SchemaUtilsSuite extends QueryTest
...
+        TypeEvolution(uniformIcebergCompatibleOnly = false),
+        TypeEvolution(uniformIcebergCompatibleOnly = true),
+        AllTypeWideningToCommonWiderType,
+        TypeEvolutionToCommonWiderType(uniformIcebergCompatibleOnly = false),
+        TypeEvolutionToCommonWiderType(uniformIcebergCompatibleOnly = true))) {
+      // Widening is allowed, unless mode is NoTypeWidening.
+      val merged = mergeSchemas(narrow, wide, typeWideningMode = typeWideningMode)
+      assert(merged === wide)
+    }
+    expectAnalysisErrorClass("DELTA_MERGE_INCOMPATIBLE_DATATYPE",
```
olaky, line 2657: is this not duplicate as it is also handled via line 2625 and checked in 2630?

spark/src/test/scala/org/apache/spark/sql/delta/schema/SchemaUtilsSuite.scala
```diff
@@ -2569,6 +2573,244 @@ class SchemaUtilsSuite extends QueryTest
...
+    val wide = new StructType()
+      .add("a", ShortType)
+      .add("b", IntegerType)
+      .add("c", IntegerType)
+      .add("s", new StructType().add("x", IntegerType))
+      .add("m", MapType(ShortType, IntegerType))
+      .add("ar", ArrayType(IntegerType))
+
+    for (typeWideningMode <- Seq(
+        NoTypeWidening,
```
olaky, line 2595: Does it make sense to create a list of all modes? I think we would use it more than once

olaky (approved):
Very nice, I like how flexible the feature is now. I only have some style suggestions, the code and tests look all correct

spark/src/test/scala/org/apache/spark/sql/delta/schema/SchemaUtilsSuite.scala
johanl-db: :+1:, I added it

spark/src/test/scala/org/apache/spark/sql/delta/schema/SchemaUtilsSuite.scala
johanl-db: No, previous one covers `narrow` -> `wide`, this one covers `wide` -> `narrow`

spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMode.scala
johanl-db: I added back a small description of the iceberg compat flag

PR #4165, opened 2025-02-18
[Docs] Add line continuation characters for the explicitly update schema command

chong0929:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [x] Other (docs)

## Description

In the documentation regarding, the command to modify the table schema using Spark: "Change a column type" is missing a line continuation character.

## How was this patch tested?

n/a

Yes, this is a fix about documentation. URL：https://docs.delta.io/latest/delta-batch.html#change-a-column-type


scottsand-db (approved):
Thanks for this!

PR #4167, opened 2025-02-18
Disallow enabling column mapping if invalid column mapping metadata is already present

juliuszsompolski:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

As effect of earlier bugs (e.g. fixed in https://github.com/delta-io/delta/pull/3487) there can exists tables where column mapping is disabled, but there is column mapping metadata on the table. Enabling column mapping metadata on such a table could lead to unexpected corruption. Simply stripping such metadata could also lead to curruptions, as the invalid metadata can be already used in other places (e.g. column statistics) via DeltaColumnMapping.getPhysicalName, which returns the name from the metadata even when column mapping is disabled.

After https://github.com/delta-io/delta/pull/3688 it should no longer be possible to end up with tables having such invalid metadata, so the issue only concerns existing tables created before that fix.

To avoid corruption, we want to disallow enabling column mapping on such tables.

## How was this patch tested?

Added tests to DeltaColumnMappingSuite.

No.
We are disallowing an operation on tables that would lead to Delta table corruption on tables that are already in an invalid state entering which is fixed already, so it can only concern old tables in the wild.

### DIFF
--- a/spark/src/main/resources/error/delta-error-classes.json
+++ b/spark/src/main/resources/error/delta-error-classes.json
@@ -869,6 +869,12 @@
     ],
     "sqlState" : "42K03"
   },
+  "DELTA_ENABLING_COLUMN_MAPPING_DISALLOWED_WHEN_COLUMN_MAPPING_METADATA_ALREADY_EXISTS" : {
+    "message" : [
+      "Enabling column mapping when column mapping metadata is already present in schema is not supported."
+    ],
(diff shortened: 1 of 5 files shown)


juliuszsompolski:
@cstavr

cstavr (approved):
Good catch! LGTM

PR #4169, opened 2025-02-18
[Kenrnel][Test only] Refactor and move method to generate protocol, metadata, schema to a separate class.

huan233usc:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

No prod code change, just refactor and move method to generate protocol, metadata, schema to a separate class. These utils could be shared in other case such as https://github.com/delta-io/delta/pull/4134

## How was this patch tested?

Existing

No


allisonport-db (approved):

kernel/kernel-api/src/test/scala/io/delta/kernel/test/TableMetadataTestUtils.scala
```diff
@@ -0,0 +1,126 @@
...
+      // minReaderVersion - it doesn't matter as the read fails anyway before the writer check
+      0,
+      minWriterVersion,
+      // reader features - it doesn't matter as the read fails anyway before the writer check
+      Collections.emptyList(),
+      writerFeatures.toSeq.asJava
+    )
+  }
+
+  def testMetadata(
```
scottsand-db, line 42: createTestMetadata ?

PR #4179, opened 2025-02-20
Update delta protocol to specify remove file path must byte match add file path

lzlfred:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [x] Other (fill in here)

## Description

Update delta protocol to specify remove file path must string match add file path. The current spec doesn't actually say how to compare the paths from two string actions. An implementation could choose any of the following, with very different outcomes for ambiguous cases:
* Do a raw string comparison. Not reliable because different URI encoding libraries could choose to encode a different subset of "illegal" characters. e.g. abc vs %65bc.
* URI-decode the strings before comparing them. Moderately expensive, but still not reliable, e.g. /path/to/table and /path//to//table refer to the same URI.
* Construct a full URI from the string is the most expensive, and still not reliable, e.g. /path/to/table and /path/to/other/../table are not considered the same URI but some systems still treat them as the same path

In practice, all Delta clients we know of just copy whatever was in add.path as remove.path, so they should anyway compare byte-equal. Many Delta clients already perform raw string comparisons.

Since none of the approaches is reliable, and raw string comparisons are already used in practice, hereby we just update the spec to make that the required behavior.

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -486,7 +486,11 @@ The `remove` action includes a timestamp that indicates when the removal occurre
 Physical deletion of physical files can happen lazily after some user-specified expiration time threshold.
 This delay allows concurrent readers to continue to execute against a stale snapshot of the data.
 A `remove` action should remain in the state of the table as a _tombstone_ until it has expired.
-A tombstone expires when *current time* (according to the node performing the cleanup) exceeds the expiration threshold added to the `remove` action timestamp.
+A tombstone expires when *current time* (according to the node performing the cleanup) exceeds the expiration
+threshold added to the `remove` action timestamp. Because it is possible to represent the same physical file name with
+many different URI layouts and URI encodings, all add and remove actions that refer to the same physical file must use
+the same (byte-identical) path string. For example, [Action Reconciliation](#action-reconciliation) will treat
+a relative path as distinct from an absolute path, and `a` as distinct from `%65`, because they are not byte-identical.
 
 In the following statements, `dvId` can refer to either the unique id of a specific Deletion Vector (`deletionVector.uniqueId`) or to `NULL`, indicating that no rows are invalidated. Since actions within a given Delta commit are not guaranteed to be applied in order, a **valid** version is restricted to contain at most one file action *of the same type* (i.e. `add`/`remove`) for any one combination of `path` and `dvId`. Moreover, for simplicity it is required that there is at most one file action of the same type for any `path` (regardless of `dvId`).
 That means specifically that for any commit…
@@ -553,7 +557,7 @@ The schema of the `remove` action is as follows:
 
 Field Name | Data Type | Description | optional/required
 -|-|-|-
-path| String | A relative path to a file from the root of the table or an absolute path to a file that should be removed from the table. The path is a URI as specified by [RFC 2396 URI Generic Syntax](https://www.ietf.org/rfc/rfc2396.txt), which needs to be decoded to get the data file path. The path must string match the corresponding `add` action. | required
+path| String | A relative path to a file from the root of the table or an absolute path to a file that should be removed from the table. The path is a URI as specified by [RFC 2396 URI Generic Syntax](https://www.ietf.org/rfc/rfc2396.txt), which needs to be decoded to get the data file path. The path must be a byte-identical copy of the corresponding `add.path`. | required
 deletionTimestamp | Option[Long] | The time the deletion occurred, represented as milliseconds since the epoch | optional
 dataChange | Boolean | When `false` the records in the removed file must be contained in one or more `add` file actions in the same version | required
 extendedFileMetadata | Boolean | When `true` the fields `partitionValues`, `size`, and `tags` are present | optional
@@ -842,7 +846,7 @@ To achieve the requirements above, related actions from different delta files ne
  - The latest `metaData` action seen wins
  - For `txn` actions, the latest `version` seen for a given `appId` wins
  - For `domainMetadata`, the latest `domainMetadata` seen for a given `domain` wins. The actions with `removed=true` act as tombstones to suppress earlier versions. Snapshot reads do _not_ return removed `domainMetadata` actions.
- - Logical files in a table are identified by their `(path, deletionVector.uniqueId)` primary key. File actions (`add` or `remove`) reference logical files, and a log can contain any number of references to a single file. `add` and `remove` are considered the same only if the paths are string equal.
+ - Logical files in a table are identified by their `(path, deletionVector.uniqueId)` primary key. File actions (`add` or `remove`) reference logical files, and a log can contain any number of references to a single file. Two file actions refer to the same file only if their path strings are byte-identical (before URI decoding).
  - To replay the log, scan all file actions and keep only the newest reference for each logical file.
  - `add` actions in the result identify logical files currently present in the table (for queries). `remove` actions in the result identify tombstones of logical files no longer present in the table (for VACUUM).
  - [v2 checkpoint spec](#v2-spec) actions are not allowed in normal commit files, and do not participate in log replay.
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -553,7 +553,7 @@ The schema of the `remove` action is as follows:
 
 Field Name | Data Type | Description | optional/required
 -|-|-|-
-path| String | A relative path to a file from the root of the table or an absolute path to a file that should be removed from the table. The path is a URI as specified by [RFC 2396 URI Generic Syntax](https://www.ietf.org/rfc/rfc2396.txt), which needs to be decoded to get the data file path. The path must string match the corresponding add action. | required
+path| String | A relative path to a file from the root of the table or an absolute path to a file that should be removed from the table. The path is a URI as specified by [RFC 2396 URI Generic Syntax](https://www.ietf.org/rfc/rfc2396.txt), which needs to be decoded to get the data file path. The path must string match the corresponding `add` action. | required
 deletionTimestamp | Option[Long] | The time the deletion occurred, represented as milliseconds since the epoch | optional
 dataChange | Boolean | When `false` the records in the removed file must be contained in one or more `add` file actions in the same version | required
 extendedFileMetadata | Boolean | When `true` the fields `partitionValues`, `size`, and `tags` are present | optional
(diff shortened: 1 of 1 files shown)


PROTOCOL.md
```diff
@@ -842,7 +842,7 @@ To achieve the requirements above, related actions from different delta files ne
  - The latest `metaData` action seen wins
  - For `txn` actions, the latest `version` seen for a given `appId` wins
  - For `domainMetadata`, the latest `domainMetadata` seen for a given `domain` wins. The actions with `removed=true` act as tombstones to suppress earlier versions. Snapshot reads do _not_ return removed `domainMetadata` actions.
- - Logical files in a table are identified by their `(path, deletionVector.uniqueId)` primary key. File actions (`add` or `remove`) reference logical files, and a log can contain any number of references to a single file.
+ - Logical files in a table are identified by their `(path, deletionVector.uniqueId)` primary key. File actions (`add` or `remove`) reference logical files, and a log can contain any number of references to a single file. `add` and `remove` are considered the same only if the paths are string equal.
```
scovich, line 845: > Two file actions refer to the same file only if their path strings are byte-identical (before URI decoding).

We need that extra requirement because there are ~infinite ways to URI-encode the same string, e.g.
`abc` could _technically_ URI-encode as `%65bc` even tho it's completely unnecessary.

PROTOCOL.md
scovich, line 1: In https://github.com/delta-io/delta/blob/master/PROTOCOL.md#add-file-and-remove-file there is a top-level description of how adds and removes work, what is legal/illegal, etc. We should probably update it as well. Somewhere near this paragraph:

> The `remove` action includes a timestamp that indicates when the removal occurred. Physical deletion of physical files can happen lazily after some user-specified expiration time threshold. This delay allows concurrent readers to continue to execute against a stale snapshot of the data. A `remove` action should remain in the state of the table as a tombstone until it has expired. A tombstone expires when current time (according to the node performing the cleanup) exceeds the expiration threshold added to the `remove` action timestamp.

We should probably add a sentence explaining:
> Because it is possible to represent the same physical file name with many different URI layouts and URI encodings, all `add` and `remove` actions that refer to the same physical file must use the same (byte-identical) `path` string. For example, [action reconciliation] will treat a relative path as distinct from an absolute path, and `a` as distinct from `%65`, because they are not byte-identical.

scovich (commented):
The description needs to explain why the spec change is necessary. From what I understand:
1. The current spec doesn't actually _say_ how to compare the paths from two string actions. An implementation could choose any of the following, with very different outcomes for ambiguous cases:
   * Do a raw string comparison. Not reliable because different URI encoding libraries could choose to encode a different subset of "illegal" characters. e.g. `abc` vs `%65bc`.
   * URI-decode the strings before comparing them. Moderately expensive, but still not reliable, e.g. `/path/to/table` and `/path//to//table` refer to the same URI.
   * Construct a full URI from the string. Most expensive, and _still_ not reliable, e.g. `/path/to/table` and `/path/to/other/../table` are not considered the same URI but some systems still treat them as the same path
2. In practice, all Delta clients we know of just copy whatever was in `add.path` as `remove.path`, so they should anyway compare byte-equal.
3. Many Delta clients already perform raw string comparisons.

Since none of the approaches is reliable, and raw string comparisons are already used in practice, we should just update the spec to make that the required behavior.

felipepessoto:
> The description needs to explain why the spec change is necessary. From what I understand:
> 
> 1. The current spec doesn't actually _say_ how to compare the paths from two string actions. An implementation could choose any of the following, with very different outcomes for ambiguous cases:
>    
>    * Do a raw string comparison. Not reliable because different URI encoding libraries could choose to encode a different subset of "illegal" characters. e.g. `abc` vs `%65bc`.
>    * URI-decode the strings before comparing them. Moderately expensive, but still not reliable, e.g. `/path/to/table` and `/path//to//table` refer to the same URI.
>    * Construct a full URI from the string. Most expensive, and _still_ not reliable, e.g. `/path/to/table` and `/path/to/other/../table` are not considered the same URI but some systems still treat them as the same path
> 2. In practice, all Delta clients we know of just copy whatever was in `add.path` as `remove.path`, so they should anyway compare byte-equal.
> 3. Many Delta clients already perform raw string comparisons.
> 
> Since none of the approaches is reliable, and raw string comparisons are already used in practice, we should just update the spec to make that the required behavior.

@lzlfred can paths be considered equal if only query string is different?

lzlfred:
@felipepessoto may i ask to clarify "query string" ?

PROTOCOL.md
```diff
@@ -486,7 +486,11 @@ The `remove` action includes a timestamp that indicates when the removal occurre
 Physical deletion of physical files can happen lazily after some user-specified expiration time threshold.
 This delay allows concurrent readers to continue to execute against a stale snapshot of the data.
 A `remove` action should remain in the state of the table as a _tombstone_ until it has expired.
-A tombstone expires when *current time* (according to the node performing the cleanup) exceeds the expiration threshold added to the `remove` action timestamp.
+A tombstone expires when *current time* (according to the node performing the cleanup) exceeds the expiration
+threshold added to the `remove` action timestamp. Because it is possible to represent the same physical file name with
```
scovich, line 490: Seems like this addition should be a new paragraph?

scovich:
> can paths be considered equal if only query string is different?

Normally two URI would _not_ compare equal if they have different query strings? 

What use case do you have in mind, that would require it?

felipepessoto:
> @felipepessoto may i ask to clarify "query string" ?

for example:

abfss://..../file.parquet?copy=1
abfss://..../file.parquet?copy=2

felipepessoto:
> > can paths be considered equal if only query string is different?
> 
> Normally two URI would _not_ compare equal if they have different query strings, e.g. [java.net.URI](https://docs.oracle.com/javase/8/docs/api/java/net/URI.html#equals-java.lang.Object-) docs state:
> 
> > For two hierarchical URIs to be considered equal, their paths must be equal and their queries must either both be undefined or else be equal.
> 
> What use case do you have in mind, that would require comparisons to ignore the query fragment?

Not sure if there is a valid reason to do it, I was reviewing the list of examples you gave, like `e.g. /path/to/table and /path/to/other/../table`

And wonder if having the same file referenced twice in the Delta Log, but with different query strings are valid.

scovich:
> > > can paths be considered equal if only query string is different?
> > 
> > 
> > Normally two URI would _not_ compare equal if they have different query strings, e.g. [java.net.URI](https://docs.oracle.com/javase/8/docs/api/java/net/URI.html#equals-java.lang.Object-) docs state:
> > > For two hierarchical URIs to be considered equal, their paths must be equal and their queries must either both be undefined or else be equal.
> > 
> > 
> > What use case do you have in mind, that would require comparisons to ignore the query fragment?
> 
> Not sure if there is a valid reason to do it, I was reviewing the list of examples you gave, like `e.g. /path/to/table and /path/to/other/../table`
> 
> And wonder if having the same file referenced twice in the Delta Log, but with different query strings are valid.

Gotcha. I'm not aware of any valid use case for that, and a big part of this change is to accept the fact that URI are too generic and too domain-specific to _really_ be able to say whether two of them are "equivalent" or not. So we bake in the simplest (and already prevailing) way of comparing them to eliminate that whole source of confusion.

felipepessoto:
We may need to change Kernel to avoid the relative path -> absolute path conversion here:

https://github.com/delta-io/delta/blob/cad2672376dcc13172526f0b91e553f845bb715a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/InternalScanFileUtils.java#L110-L112

and

https://github.com/delta-io/delta/blob/cad2672376dcc13172526f0b91e553f845bb715a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/InternalScanFileUtils.java#L63-L65

scovich:
> We may need to change Kernel to avoid the relative path -> absolute path conversion here

Not sure I follow? As far as I know those code sites are converting to absolute paths for consumption by the engine, e.g. for benefit of the object store (which always requires an absolute path). I don't think it influences the deduplication logic of log replay?

felipepessoto:
I was planning to use this API to make modifications to Delta Log, and adding a remove action from the data returned by it cause problems because path doesn’t match.

Do we have a different API in Kernel that could help?

felipepessoto:
Hi @lzlfred, @scovich I guess we forgot about this PR, it is a nice clarification to the spec.

scovich:
@felipepessoto -- good catch, this indeed did fall through the cracks. It's always a bit hair raising to tweak the spec like this (there's a risk of breaking some client/workload/customer), which makes it a bit too easy to procrastinate.

PROTOCOL.md
xupefei: +1.

PROTOCOL.md
```diff
@@ -842,7 +846,7 @@ To achieve the requirements above, related actions from different delta files ne
  - The latest `metaData` action seen wins
  - For `txn` actions, the latest `version` seen for a given `appId` wins
  - For `domainMetadata`, the latest `domainMetadata` seen for a given `domain` wins. The actions with `removed=true` act as tombstones to suppress earlier versions. Snapshot reads do _not_ return removed `domainMetadata` actions.
- - Logical files in a table are identified by their `(path, deletionVector.uniqueId)` primary key. File actions (`add` or `remove`) reference logical files, and a log can contain any number of references to a single file.
+ - Logical files in a table are identified by their `(path, deletionVector.uniqueId)` primary key. File actions (`add` or `remove`) reference logical files, and a log can contain any number of references to a single file. Two file actions refer to the same file only if their path strings are byte-identical (before URI decoding).
```
xupefei, line 849:
```suggestion
 - Logical files in a table are identified by their `(path, deletionVector.uniqueId)` primary key. File actions (`add` or `remove`) reference logical files, and a log can contain any number of references to a single file. Two file actions refer to the same file only if their path strings are byte-identical.
```


I don't think we should allow any processing. They should be exactly the same.

xupefei (commented):
Lgtm with a small change.

felipepessoto:
@xupefei, @lzlfred @scovich, the branch has conflicts. Let me know if there is anything I can do to help merge it.

felipepessoto:
@xupefei, @lzlfred @scovich. Can I help get this merged?
Thanks

prakharjain09 (approved):

prakharjain09 (approved):

felipepessoto:
@lzlfred could you resolve the conflicts?

felipepessoto:
@prakharjain09 PR is approved but has merge conflicts. Is @lzlfred planning to resolve the conflicts? Or else I can create a  new PR

felipepessoto:
@lzlfred, @prakharjain09, @xupefei. ping to merge it.

foss-contributor:
@lzlfred, thank you for your contribution! This pull request has been inactive for over 18 months and will be closed in one week. To keep it open, please leave a comment on this pull request or contact us in the `#contributors` channel on [Delta Lake Slack](https://go.delta.io/slack). We appreciate your involvement in Delta Lake!

felipepessoto:
Not stale

PR #4195, opened 2025-02-27
Uniform: Fix microsecond conversion for timestamp/ntz partition value

amogh-jahagirdar:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [x] Other (fill in here)

## Description

When converting timestamp partition values from Delta to Iceberg, currently we're doing the microsecond conversion incorectly. This leads to every conversion just returning the Unix epoch timestamp.

 In the current approach we're using some utilities from Iceberg, such as DateTimeUtil.microsFromInstant which already have implemented this logic correctly.

 There were unit tests that should've caught this earlier by testing the Iceberg spark session as a reader, but they were passing just because the wrong read method was used. This PR fixes that as well.

## How was this patch tested?

I fixed existing unit tests since they should've caught this.

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/delta/uniform/UniFormE2ESuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/uniform/UniFormE2ESuite.scala
@@ -64,7 +64,7 @@ abstract class UniFormE2EIcebergSuiteBase extends UniFormE2ETest {
       "to_date('2016-12-31', 'yyyy-MM-dd')",
       "'asdf'",
       true,
-      "TIMESTAMP_NTZ'2021-12-06 00:00:00'",
+      "TIMESTAMP_NTZ'2021-12-06 05:12:34'",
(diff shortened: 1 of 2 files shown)


spark/src/test/scala/org/apache/spark/sql/delta/uniform/UniFormE2ESuite.scala
```diff
@@ -83,7 +83,7 @@ abstract class UniFormE2EIcebergSuiteBase extends UniFormE2ETest {
             s"where ${partitionColumnName}=${partitionColumnsAndValues._2}"
           // Verify against Delta read and Iceberg read
           checkAnswer(spark.sql(verificationQuery), Seq(Row(123)))
-          checkAnswer(createReaderSparkSession.sql(verificationQuery), Seq(Row(123)))
```
amogh-jahagirdar, line 91: it's important to go through the `read` function, for some reason when I go through `createReaderSparkSession` directly that doesn't actually use the Iceberg spark session even though we are overriding it correctly.

PR #4203, opened 2025-02-28
[Spark] Throw an exception on write for DV size mismatch

c27kwan:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Right now, we are only logging for delta.deletionVector.write.offsetMismatch. In this PR, we trigger an exception at the time of writing the DV, so before we make a commit to the table. It'll fail the specific query, but won't put the table in a broken and unreadable state. We also change the log from delta.deletionVector.write.offsetMismatch into a delta assertion. 

## How was this patch tested?

Existing tests pass.

Yes. If we detect an issue related to the file size while writing Deletion Vector files, we will now throw an exception at the time of writing the DV, before we make a commit to the table. This will fail the specific query, but will prevent the table from being in an unreadable state.


larsk-db (approved):
LGTM

PR #4208, opened 2025-03-02
[Kernel] New tests for Timestamp_NTZ writes

### LINKED ISSUE #4103: [Feature Request] [Kernel] Add Timestamp-NTZ write support
Today, Kernel has Timestamp-NTZ read support. See the protocol spec [here](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#timestamp-without-timezone-timestampntz). See the read support PR [here](https://github.com/delta-io/delta/pull/2855).

This issue requests that we add Timestamp-NTZ _write_ support. Please refer to the delta-spark implementation as needed.

A rough sketch would include:
- Figureout from Delta-Spark on what triggers enabling the table features. Is it having a column with type timestampNtx or delta property
- [ALREADY COMPLETE] Kernel: Add support for writing timestampNTZ in the ParquetFileWriter (it may already have it, can't recall now)
- Update the Kernel TxnBuilderImpl and TxnImpl to upgrade the protocol appropriately and generate a new Protocol action before commit

anoopj:
- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Adds new unit tests for timestampNtz writes. Note that most of the write functionality (creation, inserts, insert with partition columns) are already exercised by  DeltaTableWritesSuite. This PR adds a few more tests.

Fixes #4103

## How was this patch tested?

New unit tests

No

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/Protocol.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/Protocol.java
@@ -45,6 +45,25 @@ public class Protocol {
   /// Public static variables and methods                                                       ///
   /////////////////////////////////////////////////////////////////////////////////////////////////
 
+  /**
+   * Helper method to get the Protocol from the row representation.
+   *
+   * @param row Row representation of the Protocol.
+   * @return the Protocol object
+   */
+  public static Protocol fromRow(Row row) {
+    requireNonNull(row);
+    Set<String> readerFeatures =
+        row.isNullAt(2)
+            ? Collections.emptySet()
+            : Collections.unmodifiableSet(new HashSet<>(VectorUtils.toJavaList(row.getArray(2))));
+    Set<String> writerFeatures =
+        row.isNullAt(3)
+            ? Collections.emptySet()
+            : Collections.unmodifiableSet(new HashSet<>(VectorUtils.toJavaList(row.getArray(3))));
+    return new Protocol(row.getInt(0), row.getInt(1), readerFeatures, writerFeatures);
+  }
+
   public static Protocol fromColumnVector(ColumnVector vector, int rowId) {
     if (vector.isNullAt(rowId)) {
       return null;
--- a/kernel/kernel-api/src/test/scala/io/delta/kernel/internal/actions/ProtocolSuite.scala
+++ b/kernel/kernel-api/src/test/scala/io/delta/kernel/internal/actions/ProtocolSuite.scala
@@ -17,7 +17,10 @@ package io.delta.kernel.internal.actions
 
 import scala.collection.JavaConverters._
 
+import io.delta.kernel.internal.data.GenericRow
 import io.delta.kernel.internal.tablefeatures.TableFeatures
+import io.delta.kernel.internal.util.VectorUtils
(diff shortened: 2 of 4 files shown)


kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWriteSuiteBase.scala
```diff
@@ -226,6 +226,30 @@ trait DeltaTableWriteSuiteBase extends AnyFunSuite with TestUtils {
       }).map { case protocol: Row => Some(protocol) }.getOrElse(Option.empty)
   }
 
+  /**
+   * Helper method to get the Protocol from the row representation.
+   *
+   * @param row Row representation of the Protocol.
+   * @return the Protocol object
+   */
+  def getProtocolFromRow(row: Row): Protocol = {
```
scottsand-db, line 235: I'd be fine if you just moved this to a static utility method in Protocol.java right above `fromColumnVector`

anoopj: Done.

vkorukanti (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/Protocol.java
```diff
@@ -45,6 +45,24 @@ public class Protocol {
   /// Public static variables and methods                                                       ///
   /////////////////////////////////////////////////////////////////////////////////////////////////
 
+  /**
+   * Helper method to get the Protocol from the row representation.
+   *
+   * @param row Row representation of the Protocol.
+   * @return the Protocol object
+   */
+  public static Protocol fromRow(Row row) {
```
scottsand-db, line 54: nit: use `requireNonNull` to assert that row isn't null before we access `row.issNullAt` etc

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/actions/ProtocolSuite.scala
```diff
@@ -369,4 +374,24 @@ class ProtocolSuite extends AnyFunSuite {
         assert(merged.getWriterFeatures.asScala === expWriterFeatures)
       }
   })
+
+  test("extract protocol from the row representation") {
+    val values: util.Map[Integer, Object] = {
+      val map = new util.HashMap[Integer, Object]()
```
scottsand-db, line 380: nit: this map = { create a map, then return a map } seems a bit odd? could you just (a) do an inline map in scala and then use `.asJava` or (b) just create and mutate the `values` map itself (i.e. no inner `map`?)

scottsand-db (changes requested):
Just left two nits! Will approve and merge once revised! Thanks!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/Protocol.java
anoopj: Done.

vkorukanti (approved):

PR #4212, opened 2025-03-03
[Spark] Enable Fast Drop feature by default

andreaschat-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Enable Fast Drop feature by default.

## How was this patch tested?

Existing tests.

Yes. The current drop feature command requires the execution of the command twice with a 24 hour waiting time in between. In addition, it also results in the truncation of the history of the Delta table to the last 24 hours.

We introduce a new DROP FEATURE implementation that allows to drop features instantly without truncating history. Dropping a feature introduces a new writer feature to the table, the `CheckpointProtectionTableFeature`.

Dropping a feature with the new behaviour can be achieved as follows:
`ALTER TABLE x DROP FEATURE y`

We can still drop a feature with the old behaviour as follows:
`ALTER TABLE x DROP FEATURE y TRUNCATE HISTORY`.

Finally, the CheckpointProtectionTableFeature can be dropped similarly to any other feature:
`ALTER TABLE x DROP FEATURE CheckpointProtectionTableFeatureTRUNCATE HISTORY`.


larsk-db (commented):
Post-merge review: LGTM

PR #4238, opened 2025-03-07
[Spark] Remove internal metadata from output of time travel queries

johanl-db:
## Description
We usually strip internal delta metadata (column mapping, generated columns) in schemas returned from e.g. `df.schema`.

One missing case is for partition columns during time-travel, which removes some metadata (column mapping, type widening), but not others (generated columns, identity columns).

Exposing generated column metadata can cause issues as it may then be propagated to other tables during CTAS.

## How was this patch tested?
- Adding test case covering stripping generated column metadata for partition columns in time-travel query.
The test fails without this fix.


foss-contributor:
@johanl-db, thank you for your contribution! This pull request has been inactive for over 18 months and will be closed in one week. To keep it open, please leave a comment on this pull request or contact us in the `#contributors` channel on [Delta Lake Slack](https://go.delta.io/slack). We appreciate your involvement in Delta Lake!

PR #4239, opened 2025-03-07
Use iceberg jar instead of icebeg code in icebergShaded module

lzlfred:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

The existing Delta Uniform code relies on a custom build iceberg shaded jar, which is made of:

* pulling iceberg source code from branch 1.1.x
* apply a series of patches
* build a shaded jar

This process is lengthy and expensive. It has caused flaky build and we had to often disable the module.

The new approach made the code directly rely on the iceberg maven jar, with HiveCatalog.java and HiveTableOperations.java which contains customized code to support Delta Uniform: They take a list of "MetadataUpdates" from caller
HiveTableOperations applies those updates to iceberg TableMetadata before writing to the disk, so the updated iceberg TableMedata has the schema & partitionSpec with fieldIds assigned by Delta.

## How was this patch tested?

existing unit tests.


build.sbt
```diff
@@ -861,37 +864,72 @@ lazy val iceberg = (project in file("iceberg"))
...
+      cp.filter { jar =>
+        val doExclude = jar.data.getName.contains("jackson-annotations") ||
+          jar.data.getName.contains("RoaringBitmap")
+        doExclude
+      }
+    },
+    // all following clases have Delta customized implementation under icebergShaded/src and thus
+    // require them to be 'first' to replace the class from iceberg jar
+    assembly / assemblyMergeStrategy := {
+       case PathList("shadedForDelta", "org", "apache", "iceberg", "PartitionSpec$Builder.class") =>
```
tdas, line 912: can you put this in a functions in the project/ directory to hide all the complexity from this main build?

project/IcebergBuild.scala
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import sbt._
+object IcebergBuild {
```
tdas, line 18: ShadedIcebergBuild

build.sbt
```diff
@@ -861,37 +864,72 @@ lazy val iceberg = (project in file("iceberg"))
...
+        icebergExclusionRules: _*
+      ),
+      "org.apache.iceberg" % "iceberg-hive-metastore" % "1.8.0" excludeAll (
+        icebergExclusionRules: _*
+      ),
+      "org.apache.hadoop" % "hadoop-client" % "2.7.3" excludeAll (
+        hadoopClientExclusionRules: _*
+      ),
+      "org.apache.hive" % "hive-metastore" % "2.3.8" excludeAll (
+        hiveMetastoreExclusionRules: _*
```
tdas, line 889: lets just IcebergBuild and use IcebergBuild.hiveMetastoreExclusionRules

build.sbt
```diff
@@ -861,37 +864,72 @@ lazy val iceberg = (project in file("iceberg"))
...
+         MergeStrategy.first
+       case PathList("shadedForDelta", "org", "apache", "iceberg", "hive", "HiveTableOperations.class") =>
+         MergeStrategy.first
+       case PathList("shadedForDelta", "org", "apache", "iceberg", "hive", "HiveTableOperations$1.class") =>
+         MergeStrategy.first
+       case x => (assemblyMergeStrategy in assembly).value(x)
+    },
     assemblyPackageScala / assembleArtifact := false,
     // Make the 'compile' invoke the 'assembly' task to generate the uber jar.
+    Compile / packageBin := assembly.value
```
tdas, line 932: isnt this duplicated 
https://github.com/delta-io/delta/pull/4239/files#diff-5634c415cd8c8504fdb973a3ed092300b43c4b8fc1e184f7249eb29a55511f91R894

build.sbt
```diff
@@ -861,37 +864,72 @@ lazy val iceberg = (project in file("iceberg"))
...
+      "org.apache.iceberg" % "iceberg-core" % "1.8.0" excludeAll (
+        icebergExclusionRules: _*
+      ),
+      "org.apache.iceberg" % "iceberg-hive-metastore" % "1.8.0" excludeAll (
+        icebergExclusionRules: _*
+      ),
+      "org.apache.hadoop" % "hadoop-client" % "2.7.3" excludeAll (
+        hadoopClientExclusionRules: _*
+      ),
+      "org.apache.hive" % "hive-metastore" % "2.3.8" excludeAll (
```
tdas, line 888: isnt there a hive-metastore client just like hadoop-client? do you really the entirely of hive-metastore classes as an unshaded dependencies in delta-spark? 

cant we have minimize the dependencies further .. and move some of the stuff as test dependencies for running tests with real hive metastore which needs all the classes?

build.sbt
```diff
@@ -861,37 +864,72 @@ lazy val iceberg = (project in file("iceberg"))
...
+      // Fix Iceberg's legacy java.lang.NoClassDefFoundError: scala/jdk/CollectionConverters$ error
+      // due to legacy scala.
+      "org.scala-lang.modules" %% "scala-collection-compat" % "2.1.1" % "provided",
+      "org.apache.iceberg" % "iceberg-core" % "1.8.0" excludeAll (
+        icebergExclusionRules: _*
+      ),
+      "org.apache.iceberg" % "iceberg-hive-metastore" % "1.8.0" excludeAll (
+        icebergExclusionRules: _*
+      ),
+      "org.apache.hadoop" % "hadoop-client" % "2.7.3" excludeAll (
```
tdas, line 885: what in this icebergShaded project requires hadoop for compilation ... that we are not getting from transitive dependencies?

lzlfred: without this, running example gives 

Caused by: java.lang.NoSuchMethodError: 'org.apache.hadoop.fs.FutureDataInputStreamBuilder org.apache.hadoop.fs.FileSystem.openFile(org.apache.hadoop.fs.Path)'

build.sbt
lzlfred: fixed. thx for catching.

build.sbt
lzlfred: see https://github.com/apache/iceberg/blob/main/build.gradle#L699

iceberg-hive-metastore uses hadoop client and hive-metastore as compile time dependency and thus the iceberg-hive-metastore jar does not have those classes. 

These added deps are not on delta-spark; they are on delta-iceberg. Either we go for the current way, or we make those as compile time dependency but user needs to include extra jars when they use it. I prefer the current way as uniform always need delta-iceberg jar, and uniform always needs those HMS classes to talk to HMS. 

I did not find "hadoop-client" equivalent... I choosed hive-metastore because iceberg-hive-metastore depends on it.

build.sbt
```diff
@@ -860,38 +863,55 @@ lazy val iceberg = (project in file("iceberg"))
     assemblyPackageScala / assembleArtifact := false
   )
 // scalastyle:on println
-
-lazy val generateIcebergJarsTask = TaskKey[Unit]("generateIcebergJars", "Generate Iceberg JARs")
-
+val icebergShadedVersion = "1.8.0"
```
tdas, line 866: nit: icebergVersion

tdas (approved):
this looks great! I am so glad we dont have to clone + compile all of iceberg any more! thank you for doing this.

felipepessoto:
I think we can close it, replaced by #5325

felipepessoto:
@newfront this can be closed

foss-contributor:
@lzlfred, thank you for your contribution! This pull request has been inactive for over 18 months and will be closed in one week. To keep it open, please leave a comment on this pull request or contact us in the `#contributors` channel on [Delta Lake Slack](https://go.delta.io/slack). We appreciate your involvement in Delta Lake!

PR #4240, opened 2025-03-07
Upgrade DescribeDeltaHistorySuite to be compatible with scala 2.13 - ignore order in column properties

ChengJi-db:
## Description

Upgrade `DescribeDeltaHistorySuite` to be compatible with scala `2.13`: we will validate properties for describing delta history separately to ignore order difference

This is because pairs inside map have different orders between scala `2.12` and `2.13`

## How was this patch tested?
Existing UTs


LukasRupprecht (approved):
LGTM!

PR #4244, opened 2025-03-08
[Spark] Abstract out a generic SkippingEligibleExpression for general use

andylam-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Pulls out a generic abstract class of `SkippingEligibleExpression` so that it can be used outside of DataFiltersBuilder.

## How was this patch tested?

NA, refactor only.

No.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingReader.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingReader.scala
@@ -170,6 +170,26 @@ object SkippingEligibleDataType {
   def unapply(f: StructField): Option[DataType] = unapply(f.dataType)
 }
 
+/**
+ * An extractor that matches expressions that are eligible for data skipping predicates.
+ *
+ * @return A tuple of 1) column name referenced in the expression, 2) date type for the
+ *         expression, 3) [[DataSkippingPredicateBuilder]] that builds the data skipping
+ *         predicate for the expression, if the given expression is eligible.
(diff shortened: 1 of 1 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingReader.scala
```diff
@@ -170,6 +171,26 @@ object SkippingEligibleDataType {
...
+/**
+ * An extractor that matches expressions that are eligible for data skipping predicates.
+ *
+ * @return A tuple of 1) column name referenced in the expression, 2) date type for the
+ *         expression, 3) [[DataSkippingPredicateBuilder]] that builds the data skipping
+ *         predicate for the expression, if the given expression is eligible.
+ *         Otherwise, return None.
+ */
+abstract class GenericSkippingEligibleExpression(
+  dataSkippingType: DeltaDataSkippingType, conf: SQLConf) {
```
chirag-s-db, line 183: Why is this conf needed?

chirag-s-db (commented):
seems reasonable overall to me, just one question

spark/src/main/scala/org/apache/spark/sql/delta/stats/DataSkippingReader.scala
andylam-db: Removed unnecessary params

chirag-s-db (approved):

PR #4260, opened 2025-03-12
[Spark] Remove listFromOrNone in SnapshotManagement

ctring:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
The method `listFromOrNone` only has one call site so removing it and inline its implementation at that call site.

## How was this patch tested?

Existing UTs

No


LukasRupprecht (approved):
LGTM!

PR #4265, opened 2025-03-13
[Kernel] Support withClusteringColumn API in kernel 

KaiqiJinWow:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This PR implement the `withClusteringColumn` API in kernel to support table creation as a clustered table.

It contains steps below,

1. withClusteringColumn takes logicalColumns as input
2. validation (column exist and cannot present together with partitionColumns)
3. convert the logical column name to physical column names to create a metadataDomain
4. Add the domainMetadata to domainMetadatasAdded.
5. Update the protocol to include `clustering` writer feature

## How was this patch tested?

Unit tests


kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -239,6 +254,8 @@ private void validate(Engine engine, SnapshotImpl snapshot, boolean isNewTable)
       SchemaUtils.validateSchema(schema.get(), isColumnMappingModeEnabled(mappingMode));
       SchemaUtils.validatePartitionColumns(
           schema.get(), partitionColumns.orElse(Collections.emptyList()));
+      SchemaUtils.validateClusteringColumns(
```
vkorukanti, line 257: do we need to enable `clustering` table feature?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -163,6 +163,22 @@ public static Map<String, Literal> casePreservingPartitionColNames(
                 Map.Entry::getValue));
   }
 
+  /**
+   * Validate the clustering columns exists in the table schema
+   */
+  public static void validateClusteringColumns(StructType schema, List<String> clusteringCols) {
+    List<String> flattenColNames = flattenNestedFieldNames(schema);
```
vkorukanti, line 170: can a nested column be selected as clustering column?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -231,6 +240,12 @@ private void validate(Engine engine, SnapshotImpl snapshot, boolean isNewTable)
             "Table already exists, but provided new partition columns. "
                 + "Partition columns can only be set on a new table.");
       }
+      if (clusteringColumns.isPresent()) {
```
vkorukanti, line 243: also should we check if partition columns are present and throw error?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
KaiqiJinWow: yes we need, will add it in this PR.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
KaiqiJinWow: Yes, it supported

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
KaiqiJinWow: added the check that partition column and clustering column cannot present together.

kernel/kernel-api/src/main/java/io/delta/kernel/TransactionBuilder.java
```diff
@@ -52,6 +52,16 @@ public interface TransactionBuilder {
    */
   TransactionBuilder withPartitionColumns(Engine engine, List<String> partitionColumns);
 
+  /**
+   * Set the list of clustering columns when create a new clustered table
+   *
+   * @param engine {@link Engine} instance to use.
+   * @param clusteringColumns The partition columns of the table. These should be a subset of the
```
imback82, line 59:
```suggestion
   * @param clusteringColumns The clustering columns of the table. These should be a subset of the
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -387,6 +409,15 @@ private Protocol getInitialProtocol() {
     return new Protocol(DEFAULT_READ_VERSION, DEFAULT_WRITE_VERSION);
   }
 
+  private void addSystemDomainMetadata() {
```
imback82, line 412: @allisonport-db IIRC, you mentioned that we need a separate API for setting system-controlled domain?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/clustering/ClusteringMetadataDomain.java
```diff
@@ -0,0 +1,104 @@
...
+                          .toLowerCase(Locale.ROOT)
+                          .equals(namePart.toLowerCase(Locale.ROOT)))
+              .findFirst();
+
+      if (!fieldOpt.isPresent()) {
+        throw new IllegalArgumentException("Column not found in schema: " + logicalName);
+      }
+
+      StructField field = fieldOpt.get();
+      currentType = field.getDataType(); // Move deeper into the nested structure
```
imback82, line 97: I didn't get what this comment is saying.

imback82 (commented):
the overall approach looks reasonable

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
allisonport-db: This is going to conflict with https://github.com/delta-io/delta/pull/4245 by the way.

I don't think we need a separate API. We probably should pass the `clusteringColumns` through to the txn, and then in the txn we add any internals system DM (using `txnImpl.addDomainMetadataInternal`)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/clustering/ClusteringMetadataDomain.java
```diff
@@ -0,0 +1,108 @@
...
+  public ClusteringMetadataDomain(List<String> logicalClusteringColumns, StructType schema) {
+    List<List<String>> physicalClusteringColumns = new ArrayList<>();
+    for (String logicalName : logicalClusteringColumns) {
+      physicalClusteringColumns.add(getPhysicalColumnNames(schema, logicalName));
+    }
+    this.clusteringColumns = physicalClusteringColumns;
+  }
+
+  /** Returns the physical column names for logical columns. */
+  private List<String> getPhysicalColumnNames(StructType schema, String logicalName) {
```
allisonport-db, line 71: I think we should have a util for this somewhere? If not, let's add this to a utils somewhere instead of here. Maybe in `ColumnMapping`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/tablefeatures/TableFeatures.java
```diff
@@ -434,6 +452,11 @@ public static Optional<Tuple2<Protocol, Set<TableFeature>>> autoUpgradeProtocolB
           Stream.concat(allNeededTableFeatures.stream(), Stream.of(DOMAIN_METADATA_W_FEATURE))
               .collect(toSet());
     }
+    if (needClusteringSupport) {
+      allNeededTableFeatures =
```
allisonport-db, line 456: We need to enable dependent features somewhere

allisonport-db (commented):
Do we need to enforce somewhere that stats are present for clustering columns? Or am I misunderstanding that as a hard requirement in the protocol

kernel/kernel-api/src/main/java/io/delta/kernel/internal/clustering/ClusteringMetadataDomain.java
vkorukanti, line 71: I think ColumnMapping has a utility for similar method.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/clustering/ClusteringMetadataDomainSuite.scala
```diff
@@ -0,0 +1,80 @@
...
+        true)
+
+    val clusteringMetadata =
+      new ClusteringMetadataDomain(List("name").asJava, metadata.get.getSchema)
+
+    assert(clusteringMetadata.toDomainMetadata.getDomain == "delta.clustering")
+    assert(clusteringMetadata.getClusteringColumns.asScala == List("name"))
+  }
+
+  test("Correctly maps nested logical column names to physical column names") {
```
vkorukanti, line 60: where are the physical column names set in the test input?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWritesSuite.scala
```diff
@@ -376,6 +378,73 @@ class DeltaTableWritesSuite extends DeltaTableWriteSuiteBase with ParquetSuiteBa
     }
   }
 
+  test("create clustered table - clustering column is not part of the schema") {
```
vkorukanti, line 381: Move these to a new suite: DeltaClusteringSuite

Also there are utilities in DeltaTableWritesBaseSuite for createEmptyTable, updateableMetadata - simplifies the tests.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWritesSuite.scala
```diff
@@ -376,6 +378,73 @@ class DeltaTableWritesSuite extends DeltaTableWriteSuiteBase with ParquetSuiteBa
...
+          .withPartitionColumns(engine, Seq("part1").asJava)
+          .build(engine)
+      }
+      assert(
+        ex.getMessage
+          .contains("Partition Columns and Clustering Columns cannot be set at the same time"))
+    }
+  }
+
+  test("create a clustered table") {
```
vkorukanti, line 415: can we add a test where we can write to the table using Kernel?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWritesSuite.scala
```diff
@@ -376,6 +378,73 @@ class DeltaTableWritesSuite extends DeltaTableWriteSuiteBase with ParquetSuiteBa
...
+    }
+  }
+
+  test(
+    "create clustered table - clustering column and partition column cannot be set at same time") {
+    withTempDirAndEngine { (tablePath, engine) =>
+      val table = Table.forPath(engine, tablePath)
+      val txnBuilder = table.createTransactionBuilder(engine, testEngineInfo, CREATE_TABLE)
+
+      val ex = intercept[IllegalArgumentException] {
```
vkorukanti, line 402: should it be a KernelException?

vkorukanti (changes requested):
LGTM overall. One qn is: should clustering require additional checks when writing AddFile?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -90,6 +92,15 @@ public TransactionBuilder withPartitionColumns(Engine engine, List<String> parti
     return this;
   }
 
+  @Override
+  public TransactionBuilder withClusteringColumns(Engine engine, List<String> clusteringColumns) {
```
raveeram-db, line 96: Should we also require clusteringColumns is non null?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWritesSuite.scala
KaiqiJinWow: Hi @vkorukanti can you clarify why this is a kernel exception? My understanding is that this is because customers incorrectly set up two incompatible parameters.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWritesSuite.scala
KaiqiJinWow: done

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWritesSuite.scala
KaiqiJinWow: done

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -92,6 +94,14 @@ public TransactionBuilder withPartitionColumns(Engine engine, List<String> parti
...
+    if (!clusteringColumns.isEmpty()) {
```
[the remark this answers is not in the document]
KaiqiJinWow: this check would be done later in `validateTransactionInputs`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/clustering/ClusteringUtils.java
```diff
@@ -0,0 +1,51 @@
...
+    return clusteringMetadataDomain.toDomainMetadata();
+  }
+
+  /**
+   * Validate the given {@link DataFileStatus} that is being added as a {@code add} action to Delta
+   * Log. Currently, it checks that the statistics are not empty.
+   *
+   * @param dataFileStatus The {@link DataFileStatus} to validate.
+   */
+  public static void validateDataFileStatus(DataFileStatus dataFileStatus) {
```
allisonport-db, line 46: Can you clarify the validation needed here for me? I was under the impression that we need to check that the stats are present for the clustering columns - not just that there are some stats present (it could possibly just be `numRows` even).

kernel/kernel-api/src/main/java/io/delta/kernel/internal/tablefeatures/TableFeatures.java
allisonport-db: If you address this by the change you made in `TransactionBuilder` then can we enforce the condition that if `needsClusteringSupport = true` -> we require `needsDomainMetadataSupport = true`?

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/clustering/ClusteringMetadataDomainSuite.scala
allisonport-db: +1

allisonport-db (commented):
@amogh-jahagirdar @KaiqiJinWow Can you two confer on how this affects schema evolution? Do we need to validate anything specific for clustering columns?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -255,11 +269,20 @@ private void validateTransactionInputs(Engine engine, SnapshotImpl snapshot, boo
...
     } else {
       // New table verify the given schema and partition columns
       ColumnMappingMode mappingMode =
           ColumnMapping.getColumnMappingMode(tableProperties.orElse(Collections.emptyMap()));
 
+      checkArgument(
+          !(partitionColumns.isPresent() && clusteringColumns.isPresent()),
+          "Partition Columns and Clustering Columns cannot be set at the same time");
       SchemaUtils.validateSchema(schema.get(), isColumnMappingModeEnabled(mappingMode));
       SchemaUtils.validatePartitionColumns(
```
vkorukanti, line 287: we should validate clustering columns to make sure they are part of the table schema?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -192,6 +196,14 @@ public void removeDomainMetadata(String domain) {
     removeDomainMetadataInternal(domain);
   }
 
+  private void generateSystemControlledDomainMetadata() {
+    if (!clusteringColumns.isEmpty()) {
+      DomainMetadata clusteringDomainMetadata =
```
vkorukanti, line 201: do we need to make sure the case sensitive of these columns matches the one in schema? I know for partition columns this is a requirement. We basically the preserve the case given in the schema.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -192,6 +196,14 @@ public void removeDomainMetadata(String domain) {
     removeDomainMetadataInternal(domain);
   }
 
+  private void generateSystemControlledDomainMetadata() {
```
vkorukanti, line 199: nit: move this private API to below out of the public method sections.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/data/TransactionStateRow.java
```diff
@@ -39,21 +43,25 @@ public class TransactionStateRow extends GenericRow {
...
+      Metadata metadata, Protocol protocol, String tablePath, StructType physicalSchema) {
     HashMap<Integer, Object> valueMap = new HashMap<>();
     valueMap.put(COL_NAME_TO_ORDINAL.get("logicalSchemaString"), metadata.getSchemaString());
     valueMap.put(COL_NAME_TO_ORDINAL.get("physicalSchemaString"), physicalSchema.toJson());
     valueMap.put(COL_NAME_TO_ORDINAL.get("partitionColumns"), metadata.getPartitionColumns());
     valueMap.put(COL_NAME_TO_ORDINAL.get("configuration"), metadata.getConfigurationMapValue());
     valueMap.put(COL_NAME_TO_ORDINAL.get("tablePath"), tablePath);
+    valueMap.put(
+        COL_NAME_TO_ORDINAL.get("writerFeatures"),
+        buildArrayValue(new ArrayList<>(protocol.getWriterFeatures()), StringType.STRING));
```
vkorukanti, line 64: given we just need to know whether cluster is supported or not (boolean), can we just pass that in the state than just the writerFeatures? It is odd to just pass writerFeatures from the protocol and nothing else.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/tablefeatures/TableFeatures.java
vkorukanti: IMO, we don't need to set the needsDomainMetadataSupport to true. The clsutering feature already has the domainMetadata as the required features and it will be automatically added. By enforcing the `needsDomainMetadataSupport = true` here, we are duplicating the feature dependency logic here.

Also remove the `needsDomainMetadataSupport = true` in TxnBuilder.withClusteringColumns.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaClusteringSuite.scala
```diff
@@ -0,0 +1,136 @@
...
+    withTempDirAndEngine { (tablePath, engine) =>
+      val table = Table.forPath(engine, tablePath)
+      val txnBuilder = table.createTransactionBuilder(engine, testEngineInfo, CREATE_TABLE)
+
+      val ex = intercept[IllegalArgumentException] {
+        val txn = txnBuilder
+          .withSchema(engine, testPartitionSchema)
+          .withClusteringColumns(engine, List(new Column("PART1"), new Column("part3")).asJava)
+          .build(engine)
+        commitTransaction(txn, engine, emptyIterable())
```
vkorukanti, line 45: can you update the `createEmptyTable` to take clustering as one of the optional args and use that here?

vkorukanti (changes requested):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -192,6 +196,14 @@ public void removeDomainMetadata(String domain) {
...
+    if (!clusteringColumns.isEmpty()) {
+      DomainMetadata clusteringDomainMetadata =
+          getClusteringDomainMetadata(clusteringColumns, metadata.getSchema());
```
scottsand-db, line 202: Please do `ClusteringUtils.getClusteringDomainMetadata`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/ColumnMapping.java
```diff
@@ -107,6 +108,31 @@ public static StructType convertToPhysicalSchema(
     }
   }
 
+  public static List<String> convertToPhysicalColumnNames(StructType schema, Column logicalColumn) {
```
scottsand-db, line 111: I think there should be individual unit tests for this in ColumnMappingSuite

scottsand-db (changes requested):
Looks great! Left some comments. Will review the tests during the work week.

kernel/kernel-api/src/main/java/io/delta/kernel/Transaction.java
```diff
@@ -241,12 +242,16 @@ static CloseableIterator<Row> generateAppendActions(
         "DataWriteContext is not created by the `Transaction.getWriteContext()`");
 
     boolean isIcebergCompatV2Enabled = isIcebergCompatV2Enabled(transactionState);
```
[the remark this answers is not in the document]
scottsand-db: Actually -- please make them `TransactionStateRow.isIcebergCompatV2Enabled` etc -- in which case the line will probably get line and we should just assign to a final variable

kernel/kernel-api/src/main/java/io/delta/kernel/TransactionBuilder.java
```diff
@@ -47,11 +48,23 @@ public interface TransactionBuilder {
...
+  /**
+   * Set the list of clustering columns when create a new clustered table
+   *
+   * @param engine {@link Engine} instance to use.
+   * @param clusteringColumns The clustering columns of the table. These should be a subset of the
+   *     columns in the schema. Both top-level and nested columns are allowed to be clustered. Note:
+   *     Clustering columns and partition columns cannot coexist in a table.
+   * @return updated {@link TransactionBuilder} instance.
+   */
+  TransactionBuilder withClusteringColumns(Engine engine, List<Column> clusteringColumns);
```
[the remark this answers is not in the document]
KaiqiJinWow: To support multiple nested columns as clustering columns, I think here we have two options, one is use `List<Column>` and the other one is using `List<List<String>>`. Does the latter one sounds better to you?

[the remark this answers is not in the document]
scottsand-db: How will engines create the `List<Column>`? Each `Column` would require passing in a `String[]` anyways, right?

Are they just going to take some string arrays and map them to create a bunch of `Column`s?

scottsand-db:
@KaiqiJinWow -- does this PR also have to _do_ anything with the set clustering columns? Is this as simple as writing the correct Domain Metadata or do we actually need to go and "cluster" our addfiles? (Sorry if this is a naive question!)

kernel/kernel-api/src/main/java/io/delta/kernel/TransactionBuilder.java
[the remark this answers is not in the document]
KaiqiJinWow: > Each Column would require passing in a String[] anyways, right?

yes this is the case. Agreed, will update using `List<List<String>>`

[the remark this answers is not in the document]
scottsand-db: Oh -- sorry, i'm just asking questions :P I don't yet have a firm opinion! I'd like to hear from @allisonport-db and @vkorukanti first :)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
KaiqiJinWow: good catch. I would add this part

kernel/kernel-api/src/main/java/io/delta/kernel/internal/tablefeatures/TableFeatures.java
KaiqiJinWow: > The clsutering feature already has the domainMetadata as the required features and it will be automatically added.


From the func `autoUpgradeProtocolBasedOnMetadata`, only the feature who implements `FeatureAutoEnabledByMetadata` could automatically add the dependency feature. However, clusteringTableFeature is not part of it as we only store it in domain metadata so we need to manually enable it.

KaiqiJinWow:
> @KaiqiJinWow -- does this PR also have to _do_ anything with the set clustering columns? Is this as simple as writing the correct Domain Metadata or do we actually need to go and "cluster" our addfiles? (Sorry if this is a naive question!)

Hi @scottsand-db , on kernel side we just enable this table as a clustered table and store its clustering column in domain metadata, the actual clustering operations would be done on spark-delta side

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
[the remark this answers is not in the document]
KaiqiJinWow: I would go with `generateClusteringDomainMetadataIfNeeded`, as it provides better clarity and visibility. We could revisit it later once we have more system controlled domain metadata

kernel/kernel-api/src/main/java/io/delta/kernel/TransactionBuilder.java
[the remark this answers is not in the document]
raveeram-db: I think `List<Column>` makes sense? We do use it in other public APIs like DataFileStatistics and a couple of other places as well I think. And does convey semantic meaning correctly (including that of nested columns).

PR #4267, opened 2025-03-13
[Kernel] Reorganize metadata updates in TransactionBuilder to clearly delineate what happens when for updates + validation

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Reorganizes the order of metadata updates and validation in `TransactionBuilder`. When building the transaction the metadata may be updated many different ways including new user provided properties, new user provided schema, transitive property changes based on enabled confs (e.g. enabling `IcebergCompatV2` enables column mapping). We should be sure to perform any validation AFTER any metadata changes have been made, less we miss something. This PR reorganizes the transformations to do transformations first and then validation.

Examples of how this could go wrong
- Right now we only block changing cmMode none -> id (which not allowed for existing tables) when newProperties are set by the user. However, cmMode confs may change after during the `IcebergCompat` checks. We should perform this validation only AFTER any config changes have been made.
   - Note - currently this invalid change is not possible since we don't automatically enable table properties for iceberg compats, but this could change in the future.
- Example in this PR where we should be using the latest updated metadata to perform validation https://github.com/delta-io/delta/pull/4196/files/70cdbc9067079705e3f68e2b9cbbc16156d19015#r1994099944 but this is not easy given the current set up

## How was this patch tested?

Existing tests suffice.

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
@@ -106,8 +106,8 @@ public class TransactionBuilderImpl implements TransactionBuilder {
 
   @Override
   public TransactionBuilder withTableProperties(Engine engine, Map<String, String> properties) {
-    Map<String, String> validatedProperties = TableConfig.validateDeltaProperties(properties);
-    this.tableProperties = Optional.of(Collections.unmodifiableMap(validatedProperties));
+    this.tableProperties =
+        Optional.of(Collections.unmodifiableMap(TableConfig.validateDeltaProperties(properties)));
     return this;
   }
 
@@ -245,7 +245,7 @@ public class TransactionBuilderImpl implements TransactionBuilder {
    * <ul>
    *   <li>We can write to the current version of the table
    *   <li>Partition columns are not specified for an existing table
-   *   <li>The schema provides is valid (e.g. no duplicate columns, valid names)
+   *   <li>The provided schema is valid (e.g. no duplicate columns, valid names)
    *   <li>Partition columns provided are valid (e.g. they exist, valid data types)
    *   <li>Concurrent txn has not already committed to the table with same txnId
    * </ul>
@@ -300,7 +300,6 @@ public class TransactionBuilderImpl implements TransactionBuilder {
    */
   private void validateMetadataChange(
       Metadata oldMetadata, Metadata newMetadata, boolean isNewTable) {
-    // Validate configuration changes
     ColumnMapping.verifyColumnMappingChange(
         oldMetadata.getConfiguration(), newMetadata.getConfiguration(), isNewTable);
     // TODO In the future block enabling IcebergWriterCompatV1 for existing tables
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
@@ -106,7 +106,8 @@ public class TransactionBuilderImpl implements TransactionBuilder {
 
   @Override
   public TransactionBuilder withTableProperties(Engine engine, Map<String, String> properties) {
-    this.tableProperties = Optional.of(new HashMap<>(properties));
+    Map<String, String> validatedProperties = TableConfig.validateDeltaProperties(properties);
+    this.tableProperties = Optional.of(Collections.unmodifiableMap(validatedProperties));
     return this;
   }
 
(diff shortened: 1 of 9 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -268,6 +273,17 @@ private void validate(Engine engine, SnapshotImpl snapshot, boolean isNewTable)
         });
   }
 
+  /** Validate that the change from oldMetadata to newMetadata is a valid change. */
```
allisonport-db, line 276: invalid changes examples

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -217,18 +229,28 @@ public Transaction build(Engine engine) {
...
         getDomainMetadatasToCommit(snapshot));
   }
 
-  /** Validate the given parameters for the transaction. */
-  private void validate(Engine engine, SnapshotImpl snapshot, boolean isNewTable) {
+  /**
+   * Validates the transaction as built given the parameters input by the user. This includes
+   *
+   * <ul>
+   *   <li>We can write to the current version of the table
```
scottsand-db, line 246: `We can write to the current version of the table` --> if the latest version of the table is `N`, we are not writing to version `N`. Could you clarify this a bit? i.e. I think you mean to express: `Ensures that the table, as defined by its latest protocol and metadata, is writable by Kernel.`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/icebergcompat/IcebergCompatMetadataValidatorAndUpdater.java
```diff
@@ -105,6 +105,22 @@ static class IcebergCompatRequiredTablePropertyEnforcer<T> {
...
+     * Constructor for RequiredDeltaTableProperty
+     *
+     * @param property DeltaConfig we are checking
+     * @param validator A generic method to validate the given value
+     * @param autoSetValue The value to set if we can auto-set this value (e.g. during table
+     *     creation)
+     */
+    IcebergCompatRequiredTablePropertyEnforcer(
+        TableConfig<T> property, Predicate<T> validator, String autoSetValue) {
+      this.property = property;
```
scottsand-db, line 118: can you call `this(...)` instead

scottsand-db (changes requested):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/icebergcompat/IcebergCompatMetadataValidatorAndUpdater.java
allisonport-db: oops this is what I meant to do

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -162,53 +163,64 @@ public Transaction build(Engine engine) {
...
 
-    Optional<Metadata> metadataWithCMInfo =
-        ColumnMapping.updateColumnMappingMetadataIfNeeded(metadata, isNewTable);
-    if (metadataWithCMInfo.isPresent()) {
-      shouldUpdateMetadata = true;
-      metadata = metadataWithCMInfo.get();
-    }
-
-    Optional<Metadata> metadataWithIcebergCompatInfo =
+    /* ----- 3: Update the METADATA with new table properties based on set properties ----- */
```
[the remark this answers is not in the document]
allisonport-db: It's not really based on the protocol changes in step 2 but it assumes that they have happened (since it validates that the protocol is upgraded). I'll clarify it a bit more.

This is some of the ordering that is confusing and would be nice to improve / change the dependencies in the future (but out of scope for this)

scottsand-db (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -162,53 +163,67 @@ public Transaction build(Engine engine) {
...
+    Optional<Protocol> newProtocol = Optional.empty();
 
-    if (!newProperties.isEmpty()) {
-      shouldUpdateMetadata = true;
-      Map<String, String> oldConfiguration = metadata.getConfiguration();
-
-      metadata = metadata.withNewConfiguration(newProperties);
+    // The metadata + protocol transformations get complex with the addition of IcebergCompat which
+    // can mutate the configuration. We walk through an example of this for clarity.
+    /* ----- 1: Update the METADATA with new table properties or schema set in the builder ----- */
```
raveeram-db, line 175: This makes the flow very clear, great change @allisonport-db !

raveeram-db (approved):

PR #4299, opened 2025-03-20
[Spark] Add user facing error for null predicate

dusantism-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

In the case when `DeltaCommand.parsePredicates` throws a NullPointerException due to null predicate, throw a more user friendly error.

## How was this patch tested?

Not necessary.

Yes, user will get an `DELTA_FAILED_RECOGNIZE_PREDICATE` error instead of `NullPointerException`.


allisonport-db (approved):

felipepessoto:
@dusantism-db would be better if this validation was done in sqlParser.parseExpression?

PR #4302, opened 2025-03-20
Convert stats for Decimal data type in converting iceberg metadata to delta

ChengJi-db:
## Description

While converting iceberg metadata to delta metadata, convert stats for decimal type in the fast path

## How was this patch tested?

UTs


lzlfred (approved):
LGTM

PR #4303, opened 2025-03-21
[Kernel][icebergWriterCompatV1 - PR#3] E2E integration of IcebergWriterCompatV1 metadata validator + updater into transaction

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Parent issue https://github.com/delta-io/delta/issues/4289

Integrates the `IcebergWriterCompatV1MetadataValidatorAndUpdater` added in https://github.com/delta-io/delta/pull/4291 to the transaction write path. This means after this PR when writing to a table with `IcebergWriterCompatV1` enabled, or when enabling `IcebergWriterCompatV1` we will perform the necessary metadata validations and updates.

We currently limit this feature to only be enabled on new tables.

## How was this patch tested?

`IcebergWriterCompatV1MetadataValidatorAndUpdater` is already thoroughly tested in `IcebergWriterCompatV1MetadataValidatorAndUpdaterSuite` but we also add an E2E test suite that tests all the validation  & use cases (& covers the protocol upgrades for this feature).

Yes, after all tasks in https://github.com/delta-io/delta/issues/4289 are complete a new writer feature will be supported.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
@@ -33,6 +33,7 @@ import io.delta.kernel.exceptions.TableNotFoundException;
 import io.delta.kernel.internal.actions.*;
 import io.delta.kernel.internal.fs.Path;
 import io.delta.kernel.internal.icebergcompat.IcebergCompatV2MetadataValidatorAndUpdater;
+import io.delta.kernel.internal.icebergcompat.IcebergWriterCompatV1MetadataValidatorAndUpdater;
 import io.delta.kernel.internal.metrics.SnapshotMetrics;
 import io.delta.kernel.internal.metrics.SnapshotQueryContext;
 import io.delta.kernel.internal.replay.LogReplay;
@@ -187,6 +188,19 @@ public class TransactionBuilderImpl implements TransactionBuilder {
     // tables if needed (e.g. enables column mapping)
     // Ex: We enable column mapping mode in the configuration such that our properties now include
     // Map(delta.enableIcebergCompatV2 -> true, delta.columnMapping.mode -> name)
+
+    // We must do our icebergWriterCompatV1 checks/updates FIRST since it has stricter column
+    // mapping requirements (id mode) than icebergCompatV2. It also may enable icebergCompatV2.
+    Optional<Metadata> icebergWriterCompatV1 =
+        IcebergWriterCompatV1MetadataValidatorAndUpdater
+            .validateAndUpdateIcebergWriterCompatV1Metadata(
+                isNewTable,
+                newMetadata.orElse(snapshotMetadata),
+                newProtocol.orElse(snapshotProtocol));
+    if (icebergWriterCompatV1.isPresent()) {
+      newMetadata = icebergWriterCompatV1;
+    }
+
     Optional<Metadata> icebergCompatV2Metadata =
         IcebergCompatV2MetadataValidatorAndUpdater.validateAndUpdateIcebergCompatV2Metadata(
             isNewTable, newMetadata.orElse(snapshotMetadata), newProtocol.orElse(snapshotProtocol));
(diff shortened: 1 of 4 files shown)


kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/IcebergWriterCompatV1Suite.scala
```diff
@@ -0,0 +1,489 @@
...
+      // Disable icebergCompatV2 (disallowed by the checks for icebergWriterCompatV1)
+      updateTableMetadata(
+        engine,
+        tablePath,
+        tableProperties = Map(TableConfig.ICEBERG_COMPAT_V2_ENABLED.getKey -> "false"))
+      assert(!TableConfig.ICEBERG_COMPAT_V2_ENABLED.fromMetadata(getMetadata(engine, tablePath)))
+    }
+  }
+
+  // TODO once we support schema evolution test adding columns of these types
```
allisonport-db, line 173: This is tracked in https://github.com/delta-io/delta/issues/4289

kernel/kernel-api/src/main/java/io/delta/kernel/internal/icebergcompat/IcebergWriterCompatV1MetadataValidatorAndUpdater.java
```diff
@@ -66,6 +67,30 @@
 public class IcebergWriterCompatV1MetadataValidatorAndUpdater
     extends IcebergCompatMetadataValidatorAndUpdater {
 
+  /**
+   * Validates that any change to property {@link TableConfig#ICEBERG_WRITER_COMPAT_V1_ENABLED} is
+   * valid. Currently, the changes we support are
+   *
+   * <ul>
+   *   <li>No change in enablement (true -> true or false -> false)
```
scottsand-db, line 75: Might be worth commenting exactly what you do _not_ support, which is enablement on an existing table?

Q: what's the value of oldConfig on an new table? Just empty map? Just curious.

vkorukanti (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/icebergcompat/IcebergWriterCompatV1MetadataValidatorAndUpdater.java
allisonport-db: Clarified a bit in the last line -- what we don't support.

> Q: what's the value of oldConfig on an new table? Just empty map? Just curious.

Correct

scottsand-db (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -187,6 +188,19 @@ public Transaction build(Engine engine) {
     // tables if needed (e.g. enables column mapping)
     // Ex: We enable column mapping mode in the configuration such that our properties now include
     // Map(delta.enableIcebergCompatV2 -> true, delta.columnMapping.mode -> name)
+
+    // We must do our icebergWriterCompatV1 checks/updates FIRST since it has stricter column
+    // mapping requirements (id mode) than icebergCompatV2. It also may enable icebergCompatV2.
```
nicklan, line 193: Doesn't it have to enable icebergCompatV2? (or do you mean that may already be enabled?)

nicklan (approved):
lgtm! just one small comment

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
allisonport-db: Yeah I mean that it may additional enable it (if not already enabled). I will clarify the comment

"It will enable icebergCompatV2 if it is not already enabled" is more clear

PR #4308, opened 2025-03-22
[SPARK] Do not remove EXISTS_DEFAULT on every metadata update

olaky:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This is a follow up to 6e9498cfc4559c5e12d8d1594edfdccf86afaca3

The previous commit assumed that that the 'EXISTS_DEFAULT' metadata key is never required in Delta. While that was probably the original intention, it is possible to create a table that does actually require the key using `CONVERT TO DELTA`.
With this PR, we do not remove the 'EXISTS_DEFAULT' from a table on every metadata operation anymore, but only when a new table is created using either `CREATE TABLE`, `REPLACE TABLE`, or `CREATE TABLE AS SELECT`.

## How was this patch tested?

Validate that key is removed in CREATE, CTAS, REPLACE, but not in CLONE and other write operations.

No


olaky:
cc @cstavr as you reviewed the previous PR

spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
```diff
@@ -817,6 +808,23 @@ trait OptimisticTransactionImpl extends DeltaTransaction
     updateMetadata(proposedNewMetadata)
   }
 
+  /**
+   * Remove the 'EXISTS_DEFAULT' metadata key from the schema. This is used for new tables, but
+   * not on existing tables because we cannot assure that the 'EXISTS_DEFAULT' values is actually
```
cstavr, line 813:
```suggestion
   * not on existing tables because we cannot assure that the 'EXISTS_DEFAULT' values is actually not
```

spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
```diff
@@ -817,6 +808,23 @@ trait OptimisticTransactionImpl extends DeltaTransaction
...
+   * required without reading the data.
+   */
+  def updateSchemaRemoveExistsDefault(): Unit = {
+    if (spark.sessionState.conf.getConf(DeltaSQLConf.REMOVE_EXISTS_DEFAULT_AT_TABLE_CREATION)) {
```
cstavr, line 817: It is a little weird to have a method called "updateSchemaRemoveExistsDefault()" that checks a "at table creation" config.

spark/src/main/scala/org/apache/spark/sql/delta/commands/CreateDeltaTableCommand.scala
```diff
@@ -301,6 +301,9 @@ case class CreateDeltaTableCommand(
         // overwrite.
         isTableReplace = isReplace && !isV1Writer
       )
+      // The 'deltaWriter' initialized the schema. Remove 'EXISTS_DEFAULT' metadata keys because
+      // they are not required on tables created by CTAS.
+      txn.updateSchemaRemoveExistsDefault()
```
cstavr, line 306: I am not sure I get this. Doesn't append with DFV1 use the same code here?

spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala
```diff
@@ -645,15 +645,13 @@ trait DeltaSQLConfBase {
...
-    buildConf("allowColumnDefaults.removeExistsDefaultFromSchemaOnMetadataChange")
-      .internal()
-      .doc("When enabled, remove all field metadata entries using the 'EXISTS_DEFAULT' key " +
-        "from the schema whenever the table metadata is updated. 'EXISTS_DEFAULT' holds values " +
-        "that are used in Spark for existing rows when a new column with a default value is " +
-        "added to a table. Since we do not support adding columns with a default value in " +
-        "Delta, this configuration should always be removed, also when it was written by an " +
-        "older version that still put it into the schema.")
+  val REMOVE_EXISTS_DEFAULT_AT_TABLE_CREATION =
+    buildConf("allowColumnDefaults.storeExistsDefaultAtTableCreation")
```
cstavr, line 649: There is a confusing mismatch here between the val and the config name.

spark/src/main/scala/org/apache/spark/sql/delta/commands/CreateDeltaTableCommand.scala
cstavr: If the intention is to do this only at table creation, why we don't we detect this during commit?

olaky: You mean deltaWriter is also used by datasource V1? That is why I do not make the change (at least unconditionall( in there.

So why here and not in commit() or updateMetdata()
* updateMetadata is designed to and enforces to be only called once
* updateMetadata does not have a parameter or in general the info to allow us to distinguish between the different commands. So we can for example detect that we have a new table in updateMetadata, but not if it is a CLONE or CREATE
* Also in commit, I did not find a way to distinguish the different commands at the level we need to
* And finally, we have a lot of code that depends on the schema that is set in the transaction. In this file we for example have

```
      val op = getOperation(txn.metadata, isManagedTable, Some(options),
        clusterBy = ClusteredTableUtils.getLogicalClusteringColumnNames(
          txn, taggedCommitData.actions)
      )
```
So the operation (that is built before we call commit) is taking the schema from the metadata of the transaction.

spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
olaky: There is indeed the problem that tables created using 'CONVERT TO DELTA' might not conform to the Delta spec and return wrong results for non-Spark based readers. My suggestion to solve this would be to not allow 'CONVERT TO DELTA' on tables with a default, or with an actually used 'EXISTS_DEFAULT'. After doing this we still have a lot of tables in existence which do require 'EXISTS_DEFAULT'

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTableCreationTests.scala
```diff
@@ -2417,66 +2418,123 @@ class DeltaTableCreationSuite
...
-    metadataOperations.foreach { metadataUpdatingQuery =>
-      withTable(testTableName) {
-        // Create the table and ensure that EXISTS_DEFAULT is part of the schema.
-        withSQLConf(DeltaSQLConf
-            .REMOVE_EXISTS_DEFAULT_FROM_SCHEMA_ON_EVERY_METADATA_CHANGE.key -> "false") {
-          sql(s"""CREATE TABLE $testTableName(int_col INT DEFAULT 2)
-                 |USING delta
-                 |TBLPROPERTIES ('delta.feature.allowColumnDefaults' = 'supported')""".stripMargin)
-          assert(schemaContainsExistsDefaultKey(testTableName))
+    writeOperations.foreach { metadataUpdatingQuery =>
```
cstavr, line 2454:
```suggestion
    writeOperations.foreach { writeOperation =>
```

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTableCreationTests.scala
```diff
@@ -2417,66 +2418,123 @@ class DeltaTableCreationSuite
...
+             |TBLPROPERTIES('delta.feature.allowColumnDefaults' = 'supported')""".stripMargin)
+    }),
+    "REPLACE TABLE" -> (() => {
+      sql("CREATE TABLE test_table(id INT) USING DELTA")
+      sql(s"""REPLACE TABLE test_table(int_with_default INT DEFAULT 2)
+             |USING DELTA
+             |TBLPROPERTIES ('delta.feature.allowColumnDefaults'= 'supported')""".stripMargin)
+    }),
+    "CREATE OR REPLACE TABLE that REPLACES" -> (() => {
+      sql("CREATE TABLE test_table(id INT) USING DELTA")
```
cstavr, line 2484: "that REPLACES" -> The create will already run with the config set to `true` or `false`, so this doesn't actually test that it replaces.

cstavr (approved):

PR #4310, opened 2025-03-23
Validate nested ids

amogh-jahagirdar:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Breaks away icebergCompatV2 nested ID validation changes from https://github.com/delta-io/delta/pull/4196/files
This PR adds more logic to the SchemaUtils#validateUpdatedSchema function to validate that under iceberg compat v2, map/array types have consistent nested field IDs across updates.

This PR builds on top of https://github.com/delta-io/delta/pull/4309

## How was this patch tested?

Added unit tests

No


kernel/kernel-api/src/test/scala/io/delta/kernel/internal/util/SchemaUtilsSuite.scala
```diff
@@ -896,6 +806,105 @@ class SchemaUtilsSuite extends AnyFunSuite {
...
+    .add(
+      "top_level_struct",
+      new StructType().add(
+        "array",
+        structWithArrayOfStructs
+          .get("top_level_struct").getDataType.asInstanceOf[StructType].get("array").getDataType,
+        false,
+        FieldMetadata.builder().fromMetadata(fieldMetadata(2L, "array_field")).putFieldMetadata(
+          COLUMN_MAPPING_NESTED_IDS_KEY,
+          FieldMetadata.builder().putLong("array.element", 3L).build()).build()),
```
amogh-jahagirdar, line 831: Open to concrete suggestions to reduce duplication here since there was concern around this @vkorukanti. We can either extract the fields from the existing schema and use that (which I'm doing) or just be explicit. Either way is more or less the same lines of code. The fundamental bottleneck in reducing duplication is really just the immutability of StructType (beyond just adding fields). 

 I feel like the right thing to do long term is introduce `SchemaVisitor` and have an implementation for applying updates to a structtype which produces a new structtype. We'll need that for intent based APIs anyways.

In the tests, we'd just run through the visitor with just a single updated field with the "failing attribute" we desire (missing IDs etc).

For now, I think a bit of duplication is probably OK (again unless there's a concrete way to avoid it).

amogh-jahagirdar: Since we're manipulating the same inner fields across the tests, one thing we could do is just introduce a helper function which accepts the metadata for that inner field.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/util/SchemaUtilsSuite.scala
```diff
@@ -895,6 +744,123 @@ class SchemaUtilsSuite extends AnyFunSuite {
...
+          .get("array").getMetadata) = {
+    new StructType()
+      .add(
+        "top_level_struct",
+        new StructType().add(
+          "array",
+          arrayType,
+          false,
+          arrayMetadata),
+        fieldMetadata(1L, "top_level_struct"))
```
amogh-jahagirdar, line 862: Added some helper functions to help reduce duplication, there's very specific inner fields we're modifying for each of these cases so we can just pass in the inner types/inner metadata specific for each case

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -436,6 +440,124 @@ private static Map<Long, StructField> fieldsById(StructType schema) {
...
+        }
+
+        for (NestedIdEntry existingNestedIdEntry : existingNestedIdEntries) {
+          Long updatedNestedColumnId = updatedNestedKeyToColumnId.get(existingNestedIdEntry.key);
+          if (updatedNestedColumnId == null
+              || !updatedNestedColumnId.equals(existingNestedIdEntry.id)) {
+            throw new IllegalArgumentException(
+                String.format(
+                    "Expected field with id %s to have nested key %s with value %s",
+                    fieldId, existingNestedIdEntry.key, existingNestedIdEntry.id));
```
amogh-jahagirdar, line 531: checkArgument

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -436,6 +440,124 @@ private static Map<Long, StructField> fieldsById(StructType schema) {
...
+        checkArgument(
+            getNestedFieldIds(field).size() >= 2,
+            "Map field %s needs at least 2 nested IDs, one for the key and one for the value",
+            field.getName());
+      }
+    }
+  }
+
+  /* Mapping of field ID to nested IDs for that field */
+  private static Map<Long, List<NestedIdEntry>> columnIdToNestedIds(
```
scottsand-db, line 474: nit: probably just a noise comment? if anything -- you could tweak the function name to be clearer -- but I don't think this comment adds anything?

Aside: I've recently been reading Clean Code by Robert C Martin -- I'm trying to apply some of those learnings :p

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -436,6 +440,124 @@ private static Map<Long, StructField> fieldsById(StructType schema) {
...
+      if (hasNestedColumnIds(entry.getValue())) {
+        nestedIdEntries.put(entry.getKey(), getNestedFieldIds(entry.getValue()));
+      }
+    }
+
+    return nestedIdEntries;
+  }
+
+  /**
+   * Validate that any nested IDs that were in the current field are consistent with the ones in the
```
scottsand-db, line 487: what does "consistent" mean? equal to? unchanged? maybe tweak that verbiage?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -436,6 +440,124 @@ private static Map<Long, StructField> fieldsById(StructType schema) {
...
+    return getNestedColumnIds(field).getEntries().entrySet().stream()
+        .map(
+            nestedIdEntry ->
+                new NestedIdEntry(nestedIdEntry.getKey(), (Long) nestedIdEntry.getValue()))
+        .collect(Collectors.toList());
+  }
+
+  /**
+   * Wrapper object for icebergCompatV2 nested ID entry to simplify validation e.g. a schema with
+   * col1 as map[int, array[int]] could look like <key: "col1.key", id: 1> <key: "col1.value", id:
```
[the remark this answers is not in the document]
scottsand-db: also -- put the static class at the bottom/top of the file maybe? or -- wdyt? any preferred structure of static classes mixed in with functions?

scottsand-db (approved):
LGTM with some minor comments!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
amogh-jahagirdar: I think that's a great suggestion, I feel the same way after reading it again tbh, the comment isn't adding anything. I'll remove it

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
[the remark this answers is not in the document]
amogh-jahagirdar: I'm good either way, I was mostly aiming for proximity to where it's actually used since this section is all private anyways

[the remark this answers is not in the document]
amogh-jahagirdar: >Does something somewhere need to handle escaping column names? i.e. topLevelColA.withDot (top level column) is different from colB.withNestedCol (nested column)

I don't think so because at this level of abstraction what we care about is just the raw values for `delta.columnMapping.nested.ids` which may have dots intended as nested columns or not. And the consistency logic isn't doing any inference on that, just taking the raw values. 
 I think it's also something we don't really care about at this level of validation since we're essentially taking what's just set in `delta.columnMapping.nested.ids`. https://github.com/delta-io/delta/blob/master/PROTOCOL.md#writer-requirements-for-icebergcompatv2

 implementations though should make sure that this property is being *set* correctly in both the case where the dot is part of the column name or part of a nested column.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
amogh-jahagirdar: Unchanged is a better indicator of what I mean here

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -436,6 +440,123 @@ private static Map<Long, StructField> fieldsById(StructType schema) {
     return columnIdToField;
   }
 
+  private static void validateNestedIds(
+      SchemaChanges schemaChanges,
+      Map<Long, StructField> currentFieldsById,
+      Map<Long, StructField> updatedFieldsById) {
+    for (Tuple2<StructField, StructField> updatedField : schemaChanges.updatedFields()) {
+      validateNestedIdsExist(updatedField._2);
```
allisonport-db, line 448: why not validate consistency here? just curious

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -436,6 +440,123 @@ private static Map<Long, StructField> fieldsById(StructType schema) {
     return columnIdToField;
   }
 
+  private static void validateNestedIds(
```
allisonport-db, line 443: Can you add docs for this fx?

Checks that all fields that should have nested ids have them? And that they are consistent?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -436,6 +440,123 @@ private static Map<Long, StructField> fieldsById(StructType schema) {
...
+    }
+
+    for (StructField addedField : schemaChanges.addedFields()) {
+      validateNestedIdsExist(addedField);
+    }
+
+    validateNestedIdConsistency(currentFieldsById, updatedFieldsById);
+  }
+
+  private static void validateNestedIdsExist(StructField field) {
```
allisonport-db, line 458: Can you also add docs here? It seems like we validate that there are at least nested ids present for the data types but not necessarily that they are correct for nested complex types?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -436,6 +440,123 @@ private static Map<Long, StructField> fieldsById(StructType schema) {
...
+   */
+  private static void validateNestedIdConsistency(
+      Map<Long, StructField> currentFieldsById, Map<Long, StructField> updatedFieldsById) {
+    Map<Long, List<NestedIdEntry>> currentFieldIdToNestedIds =
+        columnIdToNestedIds(currentFieldsById);
+    Map<Long, List<NestedIdEntry>> updatedFieldIdToNestedIds =
+        columnIdToNestedIds(updatedFieldsById);
+    /* Validate that nested field IDs exist in the schema
+    and are consistent with the previous schemas */
+    for (Map.Entry<Long, List<NestedIdEntry>> entry : updatedFieldIdToNestedIds.entrySet()) {
```
allisonport-db, line 497: Wondering if we can/should use the `SchemaChanges` here to check updated fields since it seems applicable and a useful breakdown?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -436,6 +440,123 @@ private static Map<Long, StructField> fieldsById(StructType schema) {
...
+      Long fieldId = entry.getKey();
+      List<NestedIdEntry> updatedNestedIdEntries = entry.getValue();
+      Set<Long> seenNestedIds = new HashSet<>();
+      for (NestedIdEntry nestedIdEntry : updatedNestedIdEntries) {
+        // Duplicate nested ID with top level field
```
allisonport-db, line 502: Is this the only place we do this check? This seems more like (1) validating a single input schema rather than (2) validating a oldSchema -> newSchema update

allisonport-db: Same question I guess with `validateNestedIdsExist`

allisonport-db: Or are we merging these two checks and we should always call `validateUpdatedSchema` even for new table?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -436,6 +440,123 @@ private static Map<Long, StructField> fieldsById(StructType schema) {
...
+
+        for (NestedIdEntry existingNestedIdEntry : existingNestedIdEntries) {
+          Long updatedNestedColumnId = updatedNestedKeyToColumnId.get(existingNestedIdEntry.key);
+          checkArgument(
+              updatedNestedColumnId != null
+                  && updatedNestedColumnId.equals(existingNestedIdEntry.id),
+              "Expected field with id %s to have nested key %s with value %s",
+              fieldId,
+              existingNestedIdEntry.key,
+              existingNestedIdEntry.id);
```
allisonport-db, line 531: wondering if this can be a simple subset operation?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
amogh-jahagirdar: We could do it in the same method, I just found it easier to read if I broke it up between existence and consistency checks.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
amogh-jahagirdar: Sure, I can add docs. Yes, this is just an existence check

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
amogh-jahagirdar: Yeah I originally punted on using the SchemaChanges here because we needed the grouping by field ID for the nested ID consistency check and SchemaChanges currently is agnostic to that (doesn't care about how the changes are produced, that's up to the person building the changes). 

But I just realized we probably could build that grouping ourselves from the changes, and still just go through the updates and added columns ourselves for now and filter out the entries we want to do the nested ID check. That'll save the extra unnecessary validations against unchanged fields.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
amogh-jahagirdar: This check only applies when updating a schema on an existing table. When we're creating a new table, since the schema building is all internal to Kernel, we should never be in a state where Kernel can somehow produce duplicate IDs for the nested fields, which means we should never really need to validate for a new table.

For an existing table though since we're allowing outside callers to pass in a schema we're just being defensive here since it's a simple enough check.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -436,6 +441,138 @@ private static Map<Long, StructField> fieldsById(StructType schema) {
     return columnIdToField;
   }
 
+  /* Validates that array and map types have valid nested IDs
```
scottsand-db, line 444: Sorry -- please follow the method-doc convetion used throughout kernel. Nit comment but it is important for readability that we follow a consistent code style throughout the code base.

Mehtod docs use

```java
  /**
   *
   */
```

amogh-jahagirdar: No problem, just noticed this and fixed it and in another location

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
allisonport-db: what do you mean by "grouping by field ID"?

Isn't the schemChanges updates determined via fieldId -- so we have oldField and newField (same fieldId) and can just verify the metadata?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
allisonport-db: I thought DBI is creating tables with schemas with the fieldIds populated?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
amogh-jahagirdar: That's what's happening now. I just meant that since SchemaChanges itself wasn't aware of grouping by field ID, we have to extract the ID from the changes again to be able to figure out which fields we want do the nested ID validation on

amogh-jahagirdar:
I'm going to table this PR for now for a couple reasons. Let's hold off on merging please cc @vkorukanti @allisonport-db @scottsand-db .

1. We're going to need validation that fields cannot be moved outside of their containing struct. We're also going to need to validation that in icebergCompatV2 map keys are not modified (in Delta this is allowed). This validation is made simpler by building an idToParent field ID mapping. Building this mapping when icebergCompatV2 is enabled will internally require enforcement of the existence of these nested fields.
2. I think it's more important to get https://github.com/delta-io/delta/pull/4335 in first.
3. Then I'll have a separate PR up for the idToParent and move logic
4. Then I can update the compatibility check to verify the map key isn't modified based on idToParent and a separate parentToAdded fields

foss-contributor:
@amogh-jahagirdar, thank you for your contribution! This pull request has been inactive for over 18 months and will be closed in one week. To keep it open, please leave a comment on this pull request or contact us in the `#contributors` channel on [Delta Lake Slack](https://go.delta.io/slack). We appreciate your involvement in Delta Lake!

PR #4314, opened 2025-03-25
[Spark] Improve the API of the run method of PostCommitHook

ctring:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Currently, the `PostCommitHook::run` method is taking in a `Seq[Action]`. This means the actions must be materialized in memory before passing to this method, which may cause OOM for commit with a large number of actions. This PR changes the type to be an `Iterator[Action]` instead to allow future optimizations where the Actions are read from file on demand.

This change also prompts a change in `GenerateSymlinkManifest` to make sure that the iterator is only used once. Some test refactoring for `DeltaGenerateSymlinkManifestSuite` is also included

## How was this patch tested?

Existing unit tests

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
@@ -1454,7 +1454,7 @@ trait OptimisticTransactionImpl extends DeltaTransaction
         throw e
     }
 
-    runPostCommitHooks(version, postCommitSnapshot, actualCommittedActions)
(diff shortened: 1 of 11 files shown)


LukasRupprecht (approved):
LGTM!

prakharjain09 (approved):

felipepessoto:
@ctring if `actualCommittedActions` is a Seq, is there any benefit of using Iterator?

ctring:
@felipepessoto There is no benefit at the moment. This is a preparation for a future optimization in the case when the list of committed actions is too large relative to available memory.

PR #4317, opened 2025-03-25
[Kernel][Clean up]Add fromRow() method for Metadata and directly compare P&M in test

huan233usc:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This PR address some leftover todos in the test by removing custom helper method for checking equality of P&M.
As a part of this effort, added fromRow() method for Metadata which may address several different use cases.

No functional changes

## How was this patch tested?

Existing tests

No

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/Metadata.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/Metadata.java
@@ -31,9 +31,10 @@ import java.util.stream.Collectors;
 public class Metadata {
 
   public static Metadata fromRow(Row row) {
+    requireNonNull(row);
     checkArgument(FULL_SCHEMA.equals(row.getSchema()));
     return fromColumnVector(
-        VectorUtils.buildColumnVector(Collections.singletonList(row), FULL_SCHEMA), 0);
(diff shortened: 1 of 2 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/Metadata.java
```diff
@@ -30,6 +30,12 @@
 
 public class Metadata {
 
+  public static Metadata fromRow(Row row) {
```
scottsand-db, line 33: check argument that row is not null

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/Metadata.java
```diff
@@ -30,6 +30,12 @@
...
+    checkArgument(FULL_SCHEMA.equals(row.getSchema()));
+    return fromColumnVector(
+        VectorUtils.buildColumnVector(Collections.singletonList(row), FULL_SCHEMA), 0);
```
scottsand-db, line 36: 0 is a magic number -- can you include the param name at least `, 0 /* paramName */);`

scottsand-db (approved):
LGTM after 2 minor nits

PR #4328, opened 2025-03-26
[Kernel]Keep track of all file size related stats correctly in CRC

huan233usc:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

This PR keep track of file size histogram in CRC[1], and support collecting file size/count for the remove file.

[1] https://github.com/delta-io/delta/blob/master/PROTOCOL.md#file-size-histogram-schema

## How was this patch tested?

Added unit and functional tests

No


kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -444,6 +445,10 @@ private TransactionCommitResult doCommit(
       // Action counters may be partially incremented from previous tries, reset the counters to 0
       transactionMetrics.resetActionCounters();
 
+      Optional<FileSizeHistogram> fileSizeHistogram =
+          isNewTable
+              ? Optional.of(FileSizeHistogram.createDefaultHistogram())
+              : readSnapshot.getCurrentCrcInfo().flatMap(CRCInfo::getFileSizeHistogram);
```
vkorukanti, line 451: Add a comment in what conditions we update the file size historing: new table, or table that already has filesize histogram available through CRC

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -457,8 +462,10 @@ private TransactionCommitResult doCommit(
                           // TODO: handle RemoveFiles.
                           if (!action.isNullAt(ADD_FILE_ORDINAL)) {
                             transactionMetrics.addFilesCounter.increment();
-                            transactionMetrics.addFilesSizeInBytesCounter.increment(
-                                new AddFile(action.getStruct(ADD_FILE_ORDINAL)).getSize());
+                            long addFileSize =
```
vkorukanti, line 465: fyi: @allisonport-db I think we need to handle this when we commit removes.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/stats/FileSizeHistogram.java
```diff
@@ -54,6 +54,15 @@ public static FileSizeHistogram createDefaultHistogram() {
     return new FileSizeHistogram(defaultBoundaries, zeroCounts, zeroBytes);
   }
 
+  public static Optional<FileSizeHistogram> fromRow(Row row) {
+    requireNonNull(row);
+    if (!FULL_SCHEMA.equals(row.getSchema())) {
+      return Optional.empty();
+    }
```
vkorukanti, line 61: why do we need this check? if the schemas don't match, we should throw error right?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumSuite.scala
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * Functional e2e test suite for verifying checksum correctness.
+ */
+class ChecksumSuite extends DeltaTableWriteSuiteBase {
+
+  test("Check checksum fields are correct") {
+    withTempDirAndEngine { (tablePath, engine) =>
+      // Create transaction for a new table with test schema
+      val txn = createTxn(engine, tablePath, isNewTable = true, testSchema, Seq.empty)
+      val txnState = txn.getTransactionState(engine)
```
vkorukanti, line 42: we don't need to write each step to create the table. You can just call appendData. Get the list of parquet file status using fsclient.listFrom and compute the expected file count and size.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumSuite.scala
```diff
@@ -0,0 +1,88 @@
...
+      commitResult.getPostCommitHooks.forEach(hook => hook.threadSafeInvoke(engine))
+
+      // Verify checksum exists and content are correct.
+      val crcInfo =
+        Option(ChecksumReader.getCRCInfo(engine, new Path(tablePath + "/_delta_log"), 0L, 0L))
+          .filter(_.isPresent)
+          .map(_.get()).getOrElse {
+            fail("CRC information should be present")
+          }
+      assert(crcInfo.getNumFiles === 2)
```
vkorukanti, line 81: does this work (instead of filter/map). `Optional.orElseGet`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -457,8 +462,10 @@ private TransactionCommitResult doCommit(
                           // TODO: handle RemoveFiles.
                           if (!action.isNullAt(ADD_FILE_ORDINAL)) {
```
vkorukanti, line 463: can you add if this action is a remove, we throw exception, that way we get to update the file size histogram when remove support is added.

vkorukanti (changes requested):
Looks awesome!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -457,8 +462,10 @@ private TransactionCommitResult doCommit(
...
                             transactionMetrics.addFilesCounter.increment();
-                            transactionMetrics.addFilesSizeInBytesCounter.increment(
-                                new AddFile(action.getStruct(ADD_FILE_ORDINAL)).getSize());
+                            long addFileSize =
+                                new AddFile(action.getStruct(ADD_FILE_ORDINAL)).getSize();
+                            transactionMetrics.addFilesSizeInBytesCounter.increment(addFileSize);
+                            fileSizeHistogram.ifPresent(histogram -> histogram.insert(addFileSize));
```
scottsand-db, line 468: can we group this a little better? this is a bit easier to read; first we get the size; then we go and increment various things

```java
long addFileSize = ...

transactionMetrics.addFilesCounter.increment();
transactionMetrics.addFilesSizeInBytesCounter.increment(addFileSize);
fileSizeHistogram.ifPresent(histogram -> histogram.insert(addFileSize));
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/stats/FileSizeHistogram.java
scottsand-db: +1

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumSuite.scala
scottsand-db: might be that or `.orElseThrow`?

scottsand-db (commented):
Looks great! Minor comments.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/stats/FileSizeHistogram.java
```diff
@@ -54,6 +54,13 @@ public static FileSizeHistogram createDefaultHistogram() {
     return new FileSizeHistogram(defaultBoundaries, zeroCounts, zeroBytes);
   }
 
+  public static Optional<FileSizeHistogram> fromRow(Row row) {
```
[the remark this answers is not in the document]
allisonport-db: I agree we need to clarify this, but I think it's different right because for `fromColumnVector` we are checking if the value in the vector is null and if so, returning empty (seems correct to me). Here, we are just checking if a null pointer is provided right?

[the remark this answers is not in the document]
scottsand-db: > I think it's different right because for fromColumnVector we are checking if the value in the vector is null and if so, returning empty (seems correct to me)

I agree that `fromColumnVector` is correct 👍 

My comment was about this method, `fromRow`. Let's look at `fromRow`, which is returning an `Optional`.

If I give you `null` --> you throw an exception.

If I give you `non-null` --> you create a new column vector with that single row, you pass it into `fromColumnVector`, the row is not null, and we get back `Some(FileSizeHistogram)`.

So, can `fromRow` return `Empty` ? (If my analysis above is correct, I don't think it can). If that's the case -- why is it returning an `Optional` at all?

Please let me know though if I'm missing a nuance here, thanks!

P.S. My suggestion here is to (a) either do not return Optional, or (b) do not throw on `null` and instead return `Optional.empty`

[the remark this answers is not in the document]
huan233usc: (a) is makes more sense to me.  

I will check the result from fromColumnVector(
        VectorUtils.buildColumnVector(Collections.singletonList(row), FULL_SCHEMA), 0) non empty.

WDYT cc @allisonport-db @vkorukanti @scottsand-db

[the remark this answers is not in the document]
scottsand-db: (a) requires that this is coupled with a `parentRow.isNullAt(idx)` call before we call `parent.getStruct(child_idx)` ... which is fine ... please ensure you add the javadocs to document this.

this isn't a one way door :) so this SGTM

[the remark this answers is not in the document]
allisonport-db: yep this works for me

kernel/kernel-api/src/main/java/io/delta/kernel/internal/stats/FileSizeHistogram.java
```diff
@@ -54,6 +54,25 @@ public static FileSizeHistogram createDefaultHistogram() {
...
+   * Creates a FileSizeHistogram from a Row object.
+   * This method converts a Row object into a FileSizeHistogram instance. The Row must conform
+   * to the FULL_SCHEMA defined for FileSizeHistogram.
+   *
+   * @param row The Row object containing histogram data. Must not be null and must match FULL_SCHEMA.
+   * @return A new FileSizeHistogram instance created from the provided Row.
+   * @throws NullPointerException if the row is null
+   * @throws IllegalArgumentException if the row's schema doesn't match FULL_SCHEMA
+   **/
+  public static FileSizeHistogram fromRow(Row row) {
```
allisonport-db, line 67: @huan233usc actually is this only used once in tests or is it used more in future PRs? we could just put this hack to use `fromColumnVector` in the test code in that case

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumSuite.scala
```diff
@@ -0,0 +1,84 @@
...
+      commitResult.getPostCommitHooks.forEach(hook => hook.threadSafeInvoke(engine))
+
+      // Verify checksum exists and content are correct.
+      val crcInfo = ChecksumReader.getCRCInfo(engine, new Path(tablePath + "/_delta_log"), 0L, 0L)
+        .orElseThrow(() => new AssertionError("CRC information should be present"))
+      assert(crcInfo.getNumFiles === 2)
+      assert(crcInfo.getTableSizeBytes === 1026)
+      assert(crcInfo.getFileSizeHistogram === Optional.of(expectedFileSizeHistogram))
+      assert(crcInfo.getMetadata === expectedMetadata)
+      assert(crcInfo.getProtocol === expectedProtocol)
```
allisonport-db, line 81: nit - don't we load these using the crc anyways? probably no need to verify this

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumSuite.scala
```diff
@@ -0,0 +1,84 @@
...
+import io.delta.kernel.internal.util.Utils.toCloseableIterator
+import io.delta.kernel.utils.CloseableIterable.inMemoryIterable
+import io.delta.kernel.utils.DataFileStatus
+
+/**
+ * Functional e2e test suite for verifying checksum correctness.
+ */
+class ChecksumSuite extends DeltaTableWriteSuiteBase {
+
+  test("Check checksum fields are correct") {
```
allisonport-db, line 38: Should we also validate it's updated for txn on existing table?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/metrics/TransactionReportSuite.scala
```diff
@@ -267,6 +267,8 @@ class TransactionReportSuite extends AnyFunSuite with MetricsReportTestUtils {
   }
 
   test("TransactionReport: manually commit a remove file") {
+    // TODO: re-enable this test once remove file is supported.
```
allisonport-db, line 270: When we add fileSizeHistogram to transactionMetrics let's be sure to update these tests in this suite

allisonport-db (commented):
Just a few minor comments looks good

kernel/kernel-api/src/main/java/io/delta/kernel/internal/stats/FileSizeHistogram.java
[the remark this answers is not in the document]
huan233usc: removed this method as we could defer it for the future follow up PR. This PR doesn't need it. follow suggestion form https://github.com/delta-io/delta/pull/4328#discussion_r2019601649

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumSuite.scala
huan233usc: we also test remove file in the PR, which at this moment only works with icebergwritercompatv1, however appendData is not working with column name mapping iiuc.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumStatsSuite.scala
```diff
@@ -0,0 +1,170 @@
...
+import io.delta.kernel.internal.fs.Path
+import io.delta.kernel.internal.stats.FileSizeHistogram
+import io.delta.kernel.internal.util.Utils.toCloseableIterator
+import io.delta.kernel.utils.{CloseableIterable, DataFileStatus}
+import io.delta.kernel.utils.CloseableIterable.inMemoryIterable
+
+/**
+ * Functional e2e test suite for verifying file stats collection in CRC are correct.
+ */
+class ChecksumStatsSuite extends DeltaTableWriteSuiteBase {
```
huan233usc, line 37: I feel like in the future we could make this suite replace ComparisionSuite in the end. This suite is more self-contained in kernel

vkorukanti (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/stats/FileSizeHistogram.java
```diff
@@ -160,7 +160,15 @@ private static long[] createDefaultBinBoundaries() {
     return boundaries;
   }
 
-  // TODO factory from FileSizeHistogramResult
+  /** Creates a FileSizeHistogram instance from a FileSizeHistogramResult. */
```
scottsand-db, line 163: nit: so `public static FileSizeHistogram fromFileSizeHistogramResult(` already tells me that this method `Creates a FileSizeHistogram instance from a FileSizeHistogramResult.`. IMO we should delete this comment.

However, what would be a helpful comment, is reminding the reader / giving some context on what is a `FileSizeHistogramResult` and how does it relate to this `FileSizeHistogram`? i.e. which code path is this factory involved in?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -606,12 +608,16 @@ private Optional<CRCInfo> buildPostCommitCrcInfoIfCurrentCrcAvailable(
                     commitAtVersion,
                     metadata,
                     protocol,
-                    // TODO: handle RemoveFiles for calculating table size and num of files.
-                    lastCrcInfo.getTableSizeBytes() + metricsResult.getTotalAddFilesSizeInBytes(),
-                    lastCrcInfo.getNumFiles() + metricsResult.getNumAddFiles(),
+                    lastCrcInfo.getTableSizeBytes()
```
raveeram-db, line 611: should we add some basic validation here that it's never < 0 and Math.Min(0,..)?

raveeram-db (approved):
Just one nit, LGTM!

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumStatsSuite.scala
```diff
@@ -0,0 +1,179 @@
...
+        engine,
+        tablePath,
+        Map("file1.parquet" -> 100),
+        expectedFileSizeHistogram)
+      checkCrcCorrect(
+        engine,
+        tablePath,
+        version = 2,
+        expectedFileCount = 1,
+        expectedTableSize = 100802,
```
scottsand-db, line 74: idea / question: would it be clearer to just write `100902 - 100`? I trust whatever you think 👍

scottsand-db (approved):
Minor comments -- LGTM! Test code is great -- simple to read and understand. Thanks!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
huan233usc: We had check already when constructing crc object here -> https://github.com/huan233usc/delta/blob/9292dfc2c2b01fd70e326c7603f905bbccf0908c/kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/CRCInfo.java#L115-L116

huan233usc:
Oops, missed https://github.com/huan233usc/delta/blob/master/kernel/kernel-api/src/test/scala/io/delta/kernel/internal/checksum/ChecksumWriterSuite.scala#L112, will do it in a different pr

felipepessoto:
@huan233usc, @allisonport-db is it the plan to also add support to file size histogram to delta-spark?
Thanks

huan233usc:
As far as I know, delta-spark's effort is here -> https://github.com/delta-io/delta/pull/3907 cc:@dhruvarya-db

PR #4346, opened 2025-03-31
[SPARK] Check all not null constraints

### LINKED ISSUE #4345: [Feature Request][Spark] Support enforcing all not null constraints

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

### Overview

Currently Delta only supports not-null columns and constraints on top level fields and nested struct fields, but not fields inside arrays and maps, and has a config to allow unenforceable not nulls columns by making any fields inside arrays or maps nullable. This should be improved to allow non-nullable fields inside arrays and maps.

### Motivation

Allow non-nullable columns in arrays and maps for better performance and data quality.

### Willingness to contribute

Kimahriman:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Resolves #4345 

Update the invariant checker to support checking for null values in non-nullable fields inside arrays and maps, and updates all nested field handling to consider the nullability of it's parents.

The config `constraints.allowUnenforcedNotNull.enabled` is removed, as it is no longer needed.

## How was this patch tested?

New and updated UTs

Yes, allows non-nullable fields in arrays and maps and removes the unenforced not null config.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
@@ -581,19 +581,6 @@ trait OptimisticTransactionImpl extends DeltaTransaction
       newMetadataTmp = newMetadataTmp.copy(schemaString = schema.json)
     }
 
-    newMetadataTmp = if (snapshot.metadata.schemaString == newMetadataTmp.schemaString) {
-      // Shortcut when the schema hasn't changed to avoid generating spurious schema change logs.
-      // It's fine if two different but semantically equivalent schema strings skip this special
-      // case - that indicates that something upstream attempted to do a no-op schema change, and
-      // we'll just end up doing a bit of redundant work in the else block.
-      newMetadataTmp
-    } else {
-      val fixedSchema = SchemaUtils.removeUnenforceableNotNullConstraints(
-        newMetadataTmp.schema, spark.sessionState.conf).json
-      newMetadataTmp.copy(schemaString = fixedSchema)
-    }
-
-
     if (canAssignAnyNewProtocol) {
       // Check for the new protocol version after the removal of the unenforceable not null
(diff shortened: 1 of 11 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
```diff
@@ -581,19 +581,6 @@ trait OptimisticTransactionImpl extends DeltaTransaction
...
-      // Shortcut when the schema hasn't changed to avoid generating spurious schema change logs.
-      // It's fine if two different but semantically equivalent schema strings skip this special
-      // case - that indicates that something upstream attempted to do a no-op schema change, and
-      // we'll just end up doing a bit of redundant work in the else block.
-      newMetadataTmp
-    } else {
-      val fixedSchema = SchemaUtils.removeUnenforceableNotNullConstraints(
-        newMetadataTmp.schema, spark.sessionState.conf).json
-      newMetadataTmp.copy(schemaString = fixedSchema)
-    }
```
Kimahriman, line 594: This is no longer needed so it only involved removing the unenforceable not null constraints

spark/src/test/scala/org/apache/spark/sql/delta/schema/InvariantEnforcementSuite.scala
```diff
@@ -167,54 +167,163 @@ class InvariantEnforcementSuite extends QueryTest
     }
   }
 
-  test("write empty DataFrame - zero columns") {
```
Kimahriman, line 170: I removed this test since all it was really testing was that a not null exception was thrown before a can't write an empty dataframe exception. This is a weird edge case that would make the code even more complicated, and an exception is thrown either way so it doesn't seem very useful

PR #4348, opened 2025-03-31
[Kernel] Minor refactor to column mapping related code for clarity

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Rename some fields, improve docs, and move some code for better clarity / readability.

## How was this patch tested?

Existing tests suffice.

No

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/ColumnMapping.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/ColumnMapping.java
@@ -89,20 +89,24 @@ public class ColumnMapping {
    * of the data stored in data files based on the table's column mapping mode. Field-id column
    * metadata is preserved when cmMode = ID, all column metadata is otherwise removed.
    *
+   * <p>We require {@code fullSchema} in addition to the pruned schema we want to convert since
+   * we need the complete field metadata as it is stored in the schema in the _delta_log. We
+   * cannot be sure (and do not enforce) that this metadata is preserved by the connector.
+   *
    * @param prunedSchema the logical read schema requested by the connector
-   * @param fullDeltaSchema the full delta schema (with complete metadata) as read from the
+   * @param fullSchema the full delta schema (with complete metadata) as read from the
    *     _delta_log
    * @param columnMappingMode Column mapping mode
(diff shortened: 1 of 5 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/ColumnMapping.java
```diff
@@ -200,20 +202,20 @@ private static int findMaxColumnId(StructField field, int maxColumnId) {
   /**
    * Utility method to convert the given logical schema to physical schema, recursively converting
    * sub-types in case of complex types. When {@code includeFieldId} is true, converted physical
-   * schema will have field ids in the metadata.
+   * schema will have field ids in the metadata. Column metadata is otherwise removed.
    */
   private static StructType convertToPhysicalSchema(
```
scottsand-db, line 207: I think it is note-worthy and worth adding as a method doc _why_ this method required taking in the `fullSchema`.

For example -- a naive reader would wonder why we do not just take in the `prunedSchema` and apply the column mapping on top of that.

Is it because `prunedSchema` comes from the user input? And if so, we cannot trust that it has all of the FieldMetadata that we require to calculate the physical schema? And only the fullSchema from the Delta Log will have that? I think it is worth adding some of these details

raveeram-db: +1 - this can be perhaps done later but this terminology (I see `dataSchema`, `prunedSchema`, `fullSchema`, `physicalDataSchema` etc) would be best defined in one place and used everywhere uniformly too? (like in the README or an eventual dev guide) too perhaps?

raveeram-db (approved):
Overall LGTM

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/ColumnMapping.java
allisonport-db: Updated the docs for the public method

scottsand-db (approved):
Thanks for making this!

PR #4352, opened 2025-03-31
Fix Redirect so that redirect-rw take precedence over redirect-w

ami7o:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

When both redirect-rw and redirect-w configuration exist, we want redirect-rw take precedence over redirect-w. 

## How was this patch tested?
`TableRedirectSuite.scala`

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/redirect/TableRedirect.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/redirect/TableRedirect.scala
@@ -490,6 +490,10 @@ object RedirectFeature {
     mapper.readValue(configString, classOf[TableRedirectConfiguration])
   }
 
+  /**
+   * Get the current `TableRedirectConfiguration` object from the table properties.
+   * Note that the redirect-rw takes precedence over redirect-w.
+   */
   def getRedirectConfiguration(
       properties: Map[String, String]): Option[TableRedirectConfiguration] = {
     properties.get(DeltaConfigs.REDIRECT_READER_WRITER.key)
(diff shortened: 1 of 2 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/redirect/TableRedirect.scala
```diff
@@ -490,6 +490,10 @@ object RedirectFeature {
     mapper.readValue(configString, classOf[TableRedirectConfiguration])
   }
 
+  /**
+   * Get the current `TableRedirectConfiguration` object from the table properties.
+   * Note that the redirect-rw takes precedence over redirect-w.
```
weiluo-db, line 495: s/redirect-rw/REDIRECT_READER_WRITER

weiluo-db, line 495: s/redirect-w/REDIRECT_WRITER_ONLY

spark/src/test/scala/org/apache/spark/sql/delta/TableRedirectSuite.scala
```diff
@@ -396,4 +397,23 @@ class TableRedirectSuite extends QueryTest
...
+      noRedirectRules = Set.empty)
+    val properties2 = RedirectWriterOnly.generateRedirectMetadata(
+      PathBasedRedirectSpec.REDIRECT_TYPE,
+      RedirectReady,
+      redirectSpec,
+      noRedirectRules = Set.empty)
+
+    val configuration = RedirectFeature.getRedirectConfiguration(properties1 ++ properties2)
+    assert(configuration.isDefined)
+    // RedirectReaderWriter should be preferred over RedirectWriterOnly.
```
weiluo-db, line 416: This should also check the type?

spark/src/main/scala/org/apache/spark/sql/delta/redirect/TableRedirect.scala
```diff
@@ -514,13 +518,12 @@ object RedirectFeature {
     }
   }
 
-  /** Get the current `TableRedirectConfiguration` object from the snapshot. */
+  /**
+   * Get the current `TableRedirectConfiguration` object from the snapshot.
+   * Note that the redirect-rw takes precedence over redirect-w.
```
weiluo-db, line 523: Same changes here.

spark/src/main/scala/org/apache/spark/sql/delta/redirect/TableRedirect.scala
ami7o: updated to redirect-reader-writer to match the format in this file

spark/src/test/scala/org/apache/spark/sql/delta/TableRedirectSuite.scala
ami7o: updated to compare the whole configuration

kamcheungting-db (approved):

PR #4358, opened 2025-04-02
[Spark] Improve `DeltaVersionsNotContiguousException` error message & add extra logging

xzhseh:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
The current error message of `DeltaVersionsNotContiguousException` needs to be improved in order to provide detailed context regarding the table state when throwing the exception.

Specifically, we add `startVersion`, `endVersion`, and `versionToLoad` to the error message, which could be useful to identify the table state and potentially help investigate the root cause of `DeltaVersionsNotContiguousException`.

In addition, we add detailed logging (`usage_log`) to provide more specific context when throwing `DeltaVersionsNotContiguousException`, which includes the entire calling stack and could help improve future visibility.

## How was this patch tested?
Through UTs in `SnapshotManagementSuite`.

Improve the error message of the user-facing `DeltaVersionsNotContiguousException`.


junlee-db (approved):
LGTM, thanks for adding more logs and error message

PR #4360, opened 2025-04-02
ID to parent

amogh-jahagirdar:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -484,28 +528,164 @@ private static void validateFieldCompatibility(StructField existingField, Struct
...
+    public int id() {
+      return id;
     }
 
-    return columnIdToField;
+    public Integer parentId() {
+      return parentId;
+    }
+
+    public String columnPath() {
```
amogh-jahagirdar, line 592: Plan on using this in all the error messages, just didn't want to change all of those right now since it'll widen the changes since quite a few test assertions will need to change. This was always planned in a follow on (and I think it should be tackled along with making sure the actual exceptions we want are being thrown, not just IllegalArgumentExceptions) , but just including in the structure for now.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -484,28 +528,164 @@ private static void validateFieldCompatibility(StructField existingField, Struct
...
+      return columnPath;
+    }
+
+    @Override
+    public StructField get() {
+      return field;
+    }
+  }
+
+  private static void idToFields(
```
amogh-jahagirdar, line 602: Add some comments

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -401,35 +408,74 @@ private static void validateSchemaEvolution(
...
     validatePhysicalNameConsistency(schemaChanges.updatedFields());
+    validateNoInvalidFieldMoves(currentFieldsById, updatedFieldsById);
     // Validates that the updated schema does not contain breaking changes in terms of types and
     // nullability
-    validateUpdatedSchemaCompatibility(schemaChanges);
+    validateUpdatedSchemaCompatibility(schemaChanges, currentFieldsById, icebergCompatV2);
     // ToDo Potentially validate IcebergCompatV2 nested IDs
   }
 
+  private static void validateNoInvalidFieldMoves(
```
amogh-jahagirdar, line 426: add a comment

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaChanges.java
```diff
@@ -35,61 +36,75 @@
...
  * level schema
  */
-class SchemaChanges {
-  private List<StructField> addedFields;
-  private List<StructField> removedFields;
-  private List<Tuple2<StructField, StructField>> updatedFields;
+class SchemaChanges<T extends Supplier<StructField>> {
+  private final List<T> addedFields;
+  private final List<T> removedFields;
+  private final List<T> renamedFields;
```
amogh-jahagirdar, line 42: I separated out fields which *only* went through renames into renamed fields as opposed to including those in updatedFields. The reason I did it was because the validation for if a map key was modified needs to allow for renames as well. To distinguish if two fields are only different by name I added DataType#equalsIgnoringNames. See those changes below.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -484,28 +531,164 @@ private static void validateFieldCompatibility(StructField existingField, Struct
...
+    } else if (field.getDataType() instanceof MapType) {
+      MapType mapType = (MapType) field.getDataType();
+      int mapKeyId;
+      int mapValueId;
+      String keyPath = fieldPath + ".key";
+      String valuePath = fieldPath + ".value";
+      if (icebergCompatV2) {
+        checkArgument(
+            hasNestedColumnIds(field),
+            "Map data types in icebergCompatV2 must have nested ID metadata defined for key and value");
```
amogh-jahagirdar, line 647: This is one of the keys in this PR, most of the validation that was done in https://github.com/delta-io/delta/pull/4310 is implicitly done here. The realization I had was that there's no reason to really distinguish the map.key/map.value and array.element field IDs in icebergCompatV2. We can include those in the mapping along with parent ID metadata for later checks.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -401,35 +408,77 @@ private static void validateSchemaEvolution(
...
+          if (updatedField.field().getDataType() instanceof StructType
+              && parentDataType instanceof MapType
+              && updatedField.columnPath().contains("key")) {
+            throw new KernelException(
+                "Cannot modify map field %s with struct key in icebergCompatV2");
+          }
+        }
+      }
+
       // ToDo: See if recursion can be avoided by incorporating map key/value and array element
```
amogh-jahagirdar, line 479: After this change in the case of IcebergCompatV2 we can actually avoid recursion entirely because now the fields include map.key/value array element. In the non IcebergCompatV2 case we don't have those fields registered as changes since they don't have IDs, but we still need to do validations on the map key/v or array element fields don't go through invalid changes. 

To make an optimization generalize for non icebergCompatV2 we may need to somehow have a sentinel ID for map key/value or array element that indicates that it's not a real ID but more so just a placeholder to make the optimized diff computation generalize across both cases. I'll table this for now

PR #4371, opened 2025-04-04
[Kernel] Cleanup E2E tests for schema evolution

amogh-jahagirdar:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This change addresses some low hanging fruit cleanup in the recently added E2E tests for schema evolution

## How was this patch tested?

I updated the tests, so it's the same tests

no


kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableSchemaEvolutionSuite.scala
```diff
@@ -729,22 +631,11 @@ class DeltaTableSchemaEvolutionSuite extends DeltaTableWriteSuiteBase with Colum
...
 
       // Create a table initially without column mapping
       createEmptyTable(engine, tablePath, initialSchema)
 
-      val currentSchema = table.getLatestSnapshot(engine).getSchema()
-      val newSchema = new StructType()
-        .add("a", StringType.STRING, true, currentSchema.get("a").getMetadata)
-        .add(
-          "b",
-          new StructType()
```
amogh-jahagirdar, line 742: This test doesn't need a nested field so I simplified the structure

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableSchemaEvolutionSuite.scala
```diff
@@ -59,17 +59,10 @@ class DeltaTableSchemaEvolutionSuite extends DeltaTableWriteSuiteBase with Colum
...
-      val structType = table.getLatestSnapshot(engine).getSchema
-      assertColumnMapping(structType.get("a"), 1)
+      updateTableMetadata(engine, tablePath, newSchema)
 
-      val innerStruct = structType.get("b").getDataType.asInstanceOf[StructType]
-      assertColumnMapping(innerStruct.get("d"), 4, "d")
-      assertColumnMapping(innerStruct.get("e"), 5, "e")
-      assertColumnMapping(structType.get("c"), 2)
+      val actualSchema = getMetadata(engine, tablePath).getSchema
+      assert(actualSchema == newSchema)
```
amogh-jahagirdar, line 65: I changed all the asserts to just do equality checks against the schema, `equals` verifies metadata like field IDs as well. It also makes sure we're not missing any assertions on fields that should not be there etc.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableSchemaEvolutionSuite.scala
```diff
@@ -88,16 +81,14 @@ class DeltaTableSchemaEvolutionSuite extends DeltaTableWriteSuiteBase with Colum
...
 
-      table.createTransactionBuilder(engine, testEngineInfo, Operation.MANUAL_UPDATE)
-        .withSchema(engine, newSchema)
-        .build(engine).commit(engine, emptyIterable())
+      updateTableMetadata(engine, tablePath, newSchema)
 
-      val structType = table.getLatestSnapshot(engine).getSchema
-      assertColumnMapping(structType.get("a"), 1)
+      val actualSchema = getMetadata(engine, tablePath).getSchema
+      assert(actualSchema == newSchema)
```
allisonport-db, line 91: These tests (that you updated) are mostly just checking that we don't (1) fail when the operation should succeed (2) validate the newSchema is correctly committed to the table right?

allisonport-db: Like these test cases updated don't include ones where Kernel is updating any column metadata. We are pre-populating the field ids and everything

PR #4377, opened 2025-04-07
[Spark ]Add config to disable schema tracking for type widening

johanl-db:
## Description
Add an internal config to disable tracking schema used to block widening type changes when streaming from a Delta table.
The default will remain to have schema tracking enabled, this only provides flexibility in case we need to disable this feature.

## How was this patch tested?
Added a test to cover the new config.

No


fanyue-xia (approved):
LGTM

Issue #4381, opened 2025-04-07
[PROTOCOL RFC] Catalog-managed Tables


scottsand-db:
See the actual proposed protocol change here: https://github.com/delta-io/delta/pull/4382

### Description of the protocol change

The corresponding RFC proposes a new reader-writer table feature `catalogManaged` which changes the way Delta Lake accesses tables.

Today’s Delta protocol relies entirely on the filesystem for read-time discovery as well as write-time commit atomicity. This feature request is to allow catalog-managed Delta tables whose discovery and commits go through the table's owning catalog instead of going directly to the filesystem (s3, abfs, etc). In particular, the catalog becomes the source of truth about whether a given commit attempt succeeded or not, instead of relying exclusively on filesystem PUT-if-absent primitives.

Making the catalog the source of truth for commits to a table brings several important advantages:

1. Allows the catalog to broker all commits to the tables it owns, and to reject filesystem-based commits that would bypass the catalog. Otherwise, the catalog cannot reliably stay in sync with the table state, nor can it reject invalid commits, because it doesn’t even know about writes until they are already durable and visible to readers. For example, a catalog would want to block attempts to drop the NOT NULL constraint on a column which is referenced by a FOREIGN KEY constraint in a different table.

2. Opens a clear path to transactions that could span multiple tables and/or involve non-table catalog updates. Otherwise, the catalog cannot participate in commit at all, because filesystem-based commits (i.e. using PUT-if-absent) do not admit any way to coordinate with other entities.

3. Allows the catalog to facilitate efficient writes of the table, e.g. by directly hosting the content of small commits instead of forcing clients to write them to cloud storage first. Otherwise, the catalog is not a source of truth, and at best it can only mirror stale copies of table state.

4. Allows the catalog to facilitate efficient reads of the table. Examples include vending storage credentials, as well as serving up the content of small commits and/or table state such as version checksum file so that clients do not have to read those files from cloud storage. Otherwise, the catalog is not even involved with reads, let alone a source of truth about the table, and so it cannot help readers in any way.

5. Allows the catalog to trigger followup actions based on a commit, such as VACUUMing, data layout optimizations, automatic UniForm conversions, or triggering arbitrary listeners such as downstream ETL or streaming pipelines.

[...]

PR #4382, opened 2025-04-07
[PROTOCOL RFC] Catalog-managed Tables

scottsand-db:
This PR

1. Adds the new `catalogManaged` table feature protocol RFC for issue https://github.com/delta-io/delta/issues/4381
2. Formally rejects the previous Coordinated-Commits RFC (https://github.com/delta-io/delta/issues/2598)

For details about why we are proposing this RFC and why we are rejecting the Coordinated Commits RFC, please see the tracking issue #4381

Design doc: [Design Doc: Catalog-Managed Delta Table Feature](https://docs.google.com/document/d/1u7SpfY7OLUhAYolqDQ-5VT6OU95hKSlHhtn9PFGAHks)


scovich (commented):
Clear and easy to read, thanks! 

The PR description and associated issue don't seem to indicate why this new RFC replaces the original one, tho?

xzhseh (commented):
Great RFC, thanks for the excellent work!

+1 on @scovich's comment to have detailed reasons why this RFC replaces the [previous one](https://github.com/delta-io/delta/blob/master/protocol_rfcs/managed-commits.md).

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,481 @@
...
+  the contents to the server-side component) and atomically record the content of the commit as
+  version `v` of the table.
+- Propose, ratify, and publish all-in-one by atomically writing a [published commit file](#published-commit)
+  in the `_delta_log` directory. Note that this commit will be considered to have succeeded as soon
+  as the file becomes visible in the filesystem, regardless of when or whether the catalog is made
+  aware of the successful publish. The catalog does not need to track these files.
+
+Note that when a catalog receives a request to ratify version `v`, it must first verify that the
+previous version `v-1` already exists (ratified or published), and that version `v` does not yet
+exist.
```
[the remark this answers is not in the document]
scottsand-db: I like this 👍

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,481 @@
...
+catalog.
+
+This component then has several options for proposing, ratifying, and publishing the commit,
+detailed below.
+
+- Write the actions (likely client-side) to an [staged commit file](#staged-commit) in the
+  `_delta_log/_staged_commits` directory and then ratify the staged commit by atomically recording
+  (likely server-side) (in persistent storage of some kind) that the file corresponds to version
+  `v`.
+- Treat this as an [inline commit](#inline-commit) (i.e. likely that the client-side component sends
```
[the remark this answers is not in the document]
scottsand-db: Yup!

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,481 @@
...
+
+## Commit Protocol
+
+To start, Delta Clients send the desired actions to be commited to the client-side component of the
+catalog.
+
+This component then has several options for proposing, ratifying, and publishing the commit,
+detailed below.
+
+- Write the actions (likely client-side) to an [staged commit file](#staged-commit) in the
```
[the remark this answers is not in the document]
scottsand-db: This makes it much clearer, thanks!

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,481 @@
...
+primitives.
+
+Making the catalog the source of truth for commits to a table brings several important advantages:
+
+1. Allows the catalog to broker all commits to the tables it owns, and to reject filesystem-based
+   commits that would bypass the catalog. Otherwise, the catalog cannot reliably stay in sync with the
+   table state, nor can it reject invalid commits, because it doesn’t even know about writes until they
+   are already durable and visible to readers. For example, a catalog would want to block attempts to
+   drop the NOT NULL constraint on a column which is referenced by a FOREIGN KEY constraint in a
+   different table.
```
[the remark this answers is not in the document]
scottsand-db: You mean in our Delta Catalog? Hm. I'm not sure. I'd think this would instead be the responsibility of, say, a Glue catalog or Unity catalog implementation.

cc @scovich, @LukasRupprecht

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,481 @@
...
+3. **Inline commit** <a name="inline-commit"></a>: A proposed commit that is not written to disk but
+   rather has its content sent to the catalog for the catalog to track directly.
+
+4. **Ratified commit**: A proposed commit that a catalog has determined has won the commit at the
+   desired version of the table.
+    - It may or may not yet be "published".
+    - It may or may not even be _tracked_ by the catalog at all - the catalog may have just
+      atomically published it to the filesystem directly, relying on PUT-if-absent primitives to
+      facilitate the ratification and publication all in one step.
+    - The catalog must track ratified commits until they are published to the `_delta_log` directory.
```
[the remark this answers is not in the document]
scottsand-db: Great feedback, thanks!

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,481 @@
...
+     *
+     * If neither start nor end version is specified, the catalog will return all available ratified
+     * commits (possibly empty, if all commits have been published).
+     *
+     * In all cases, the response also includes the table's latest ratified commit version.
+     *
+     * @return GetCommitsResponse which contains an ordered list of ratified commits
+     *         tracked by the catalog, as well as table's latest commit version.
+     */
+    def getRatifiedCommits(
```
scottsand-db, line 472: Should this be `getTable` instead?

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,489 @@
...
+      listing operations to directly discover them.
+    - Published commits do not need to be tracked by the catalog.
+
+## Terminology: Delta Client
+
+This is the component that implements support for reading and writing Delta tables, and implements
+the logic required by the `catalogOwned` table feature. Among other things, it
+- triggers the filesystem listing, if needed, to discover published commits
+- generates the commit content (the set of [actions](#actions))
+- works together with the query engine to trigger the commit process and invoke the client-side
```
xzhseh, line 198: `Delta Client` is also responsible for **defining** the interface/APIs/contract with the client-side catalog component (i.e., `Catalog Client`), right?

scottsand-db: Yes. Added!

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,489 @@
...
+2. Opens a clear path to transactions that could span multiple tables and/or involve non-table
+   catalog updates. Otherwise, the catalog cannot participate in commit at all, because
+   filesystem-based commits (i.e. using PUT-if-absent) do not admit any way to coordinate with other
+   entities.
+
+3. Allows the catalog to facilitate efficient writes of the table, e.g. by directly hosting the
+   content of small commits instead of forcing clients to write them to cloud storage first. Otherwise,
+   the catalog is not a source of truth, and at best it can only mirror stale copies of table state.
+
+4. Allows the catalog to facilitate efficient reads of the table. Examples include vending storage
```
bart-samwel, line 32: In theory the caching of such files doesn't need to be done by the catalog because they are immutable.

What is a great advantage of a catalog is that you can avoid having to do a LIST to get the latest committed version. That can save time, and it can also allow implementations of Delta on file systems where LIST is not ordered, such as S3 Express One Zone.

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,489 @@
...
+the `_delta_log/_staged_commits` directory correspond to ratified versions, and Delta clients should
+not attempt to directly interpret the contents of that directory. Refer to
+[catalog-owned tables](#catalog-owned-tables) for more details.</ins>
+
+~~Delta files use new-line delimited JSON format, where every action is stored as a single line JSON
+document. A delta file, `n.json`, contains an atomic set of [_actions_](#actions) that should be
+applied to the previous table state, `n-1.json`, in order to the construct `n`th snapshot of the
+table. An action changes one aspect of the table's state, for example, adding or removing a file.~~
+
+<ins>Delta files use newline-delimited JSON format, where every action is stored as a single-line
```
bart-samwel, line 99: This uses "Delta files", but the preceding text uses "commits". And the section title says "Delta Log Entries". Nothing actually says that a Delta Log Entry == a Delta file == a commit.

It might make sense to move this paragraph above the whole catalogOwned sequence. The first part is really an introduction into what's in these Delta files, and then the catalogOwned is a modification to that. The intermezzo about catalogOwned is really making it hard to pop back up the stack to this higher level intro description.

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,489 @@
...
+> ***Change to [existing section](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#metadata-cleanup)***
+
+2. Identify the newest checkpoint that is not newer than the `cutOffCommit`. A checkpoint at the
+   `cutOffCommit` is ideal, but an older one will do. Lets call it `cutOffCheckpoint`. We need to
+   preserve the `cutOffCheckpoint` and all <ins>published</ins> commits after it, because we need
+   them to enable time travel for commits between `cutOffCheckpoint` and the next available
+   checkpoint.
+    - <ins>If no `cutOffCheckpoint` can be found, do not proceed with metadata cleanup as there is
+      nothing to cleanup.</ins>
+3. Delete all [delta log entries](#delta-log-entries) and [checkpoint files](#checkpoints) before
```
bart-samwel, line 123: Existing: should we have the CRC files listed here as well for deletion?

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,489 @@
...
+might attempt to access a catalog-owned table's storage location without going through the catalog
+first, and to block filesystem-based writers who could otherwise corrupt both the table and the
+catalog by failing to commit through the catalog.
+
+Before we can go into details of this protocol feature, we must first align our terminology.
+
+## Terminology: Commits
+
+A commit is a set of [actions](#actions) that transform a Delta table from version `v - 1` to `v`.
+A commit contains the same content as a [Delta file](#delta-log-entries).
```
bart-samwel, line 156: We could say "A commit may be stored in the file system as a Delta file." A commit doesn't always contain the content of a (particular) Delta file, i.e., there may not actually be a Delta file for a commit.

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,489 @@
...
+
+There are several types of commits:
+
+1. **Proposed commit**:  A commit that a Delta client has proposed for the next version of the
```
bart-samwel, line 160: "Has proposed" suggests that the client must have already contacted the catalog. A staged commit is a proposed commit, but what if I write a file in _delta_log/_staged_commits but wait with actually proposing it to the catalog? Is it still a proposed commit if I haven't proposed it? Can it be a staged commit if it's not a proposed commit?

We could work around this issue by calling this a "prepared commit". Preparing doesn't require that you already contact the catalog.

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,489 @@
...
+   table. It could be "staged" or "inline". It will either become "ratified" or be rejected.
+
+2. **Staged commit** <a name="staged-commit"></a>: A proposed commit that is written to disk at
```
bart-samwel, line 163: Do we need to say that this has the format of a normal Delta file?

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,489 @@
...
+   `_delta_log/_staged_commits/<v>.<uuid>.json`.
+    - Here, the `uuid` is a random UUID that is generated for each commit and `v` is the version
+      which is being committed, zero-padded to 20 digits.
```
bart-samwel, line 166:
```suggestion
      which is proposed to be committed, zero-padded to 20 digits.
```

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,489 @@
...
+    - The catalog tracks only the location, not the content, of a staged commit.
+    - The mere existence of one of these files does not mean that the file is a _ratified_ commit.
+      It might correspond to a failed or in-progress commit attempt.
+    - The catalog is the source of truth around which staged commits are "ratified".
+
+3. **Inline commit** <a name="inline-commit"></a>: A proposed commit that is not written to disk but
```
bart-samwel, line 172: Inline commits can also be ratified, right?

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,489 @@
...
+   rather has its content sent to the catalog for the catalog to track directly.
+
+4. **Ratified commit**: A proposed commit that a catalog has determined has won the commit at the
+   desired version of the table.
+    - The catalog must track ratified commits until they are published to the `_delta_log` directory.
+    - A ratified commit may or may not yet be published.
+    - A ratified commit may or may not even be _tracked_ by the catalog at all - the catalog may
```
bart-samwel, line 179: The concept of "tracked" is new. We could say that the state for unpublished ratified commits is always stored by the catalog, while published commits may or may not be stored by the catalog. ("Stored" makes it explicit what "tracked" means.)

Changing this phrasing would require changes across the RFC. I could also live with phrasing where the catalog "keeps track of" commits. The use of "to track" doesn't work for me. To me (and to the dictionary definition), tracking means following the direction or location of something that moves or changes, but we use it just to mean "store".

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,489 @@
...
+- generates the commit content (the set of [actions](#actions))
+- works together with the query engine to trigger the commit process and invoke the client-side
+  catalog component with the commit content
+
+## Terminology: Catalogs
+
+1. **Catalog**: A catalog is a entity which manages a Delta table, including its creation, writes,
+   reads, and eventual deletion.
+    - It could be backed by a database, a filesystem, or any other persistence mechanism.
+    - Each catalog has its own spec around how Delta clients should contact them, and how they
```
bart-samwel, line 206: "contact them" implies the presence of a server. How about "interact with them"?

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,489 @@
...
+    - This client-side component has two primary responsibilities:
+        - implement any client-side catalog-specific logic (such as [publishing](#publishing-commits))
+        - communicate with the Catalog Server, if any
+
+3. **Catalog Server**: The catalog may also involve a server-side component which the client-side
+   component would be responsible to communicate with.
+    - This server is responsible for coordinating commits and potentially persisting table metadata
+      and enforcing authorization policies.
+    - Not all catalogs require a server; some may be entirely client-side (e.g. filesystem-backed
+      catalogs).
```
bart-samwel, line 221:
```suggestion
    - Not all catalogs require a server; some may be entirely client-side, e.g. filesystem-backed
      catalogs, or they may make use of a generic database server and implement all of the logic
      client-side.
```

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,489 @@
...
+- Retrieve information about recent ratified commits and the latest ratified version on the table.
+  This is explained in detail in the [Getting Ratified Commits from the Catalog](#getting-ratified-commits-from-the-catalog) section.
+- Though not required, it is encouraged that catalogs also return the latest table-level metadata,
+  such as the latest Protocol and Metadata actions, for the table. This can provide significant
+  performance advantages to conforming Delta clients, who may forgo log replay and instead trust
+  the information provided by the catalog during query planning.
+
+## Reading Catalog-owned Tables
+
+A catalog-owned table can have a mix of (a) published and (b) ratified and non-published commits.
```
bart-samwel, line 248:
```suggestion
A catalog-owned table can have a mix of (a) published and (b) ratified but non-published commits.
```

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,489 @@
...
+
+To start, Delta Clients send the desired actions to be commited to the client-side component of the
+catalog.
+
+This component then has several options for proposing, ratifying, and publishing the commit,
+detailed below.
+
+- Option 1: Write the actions (likely client-side) to an [staged commit file](#staged-commit) in the
+  `_delta_log/_staged_commits` directory and then ratify the staged commit by atomically recording
+  (likely server-side) (in persistent storage of some kind) that the file corresponds to version
```
bart-samwel, line 318: Double parenthesized asides looks weird :)

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,489 @@
...
+  `v`.
+- Option 2: Treat this as an [inline commit](#inline-commit) (i.e. likely that the client-side
+  component sends the contents to the server-side component) and atomically record the content of
+  the commit as version `v` of the table.
+- Option 3: Propose, ratify, and publish all-in-one by atomically writing a [published commit file](#published-commit)
+  in the `_delta_log` directory. Note that this commit will be considered to have succeeded as soon
+  as the file becomes visible in the filesystem, regardless of when or whether the catalog is made
+  aware of the successful publish. The catalog does not need to track these files.
```
bart-samwel, line 326: This paragraph makes it sound like the catalog client can always publish a commit directly without contacting the catalog server. The last sentence doesn't help, it just reinforces that "the catalog (server) is not involved at all". I think this particular case needs to be strongly qualified that this approach only works for catalog clients that use PUT-if-absent. We can also mention the option that the server does the PUT-if-absent.

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,489 @@
...
+
+A catalog must not ratify version `v` until it has ratified version `v - 1`, and it must ratify
+version `v` at most once.
+
+The catalog must track both flavors of ratified commits (staged or inline) and make them available
+to readers until they are [published](#publishing-commits).
+
+Delta clients are encouraged to establish an API contract where the catalog provides the latest
```
bart-samwel, line 334: Mention "for performance reasons"? Because it's really just a performance optimization.

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,489 @@
...
+
+## Publishing Commits
+
+Publishing is the process of copying the ratified commit with version `<v>` to
+`_delta_log/<v>.json`. The ratified commit may be a staged commit located in
+`_delta_log/_staged_commits/<v>.<uuid>.json`, or it may be an inline commit whose content the
+catalog tracks itself. Because the content of a ratified commit is immutable, it does not matter
+whether the client-side, server-side, or both catalog components initiate publishing.
+
+Implementations are strongly encouraged to publish commits promptly. This reduces the number of
```
bart-samwel, line 359: This is not necessary. In theory a catalog could be fine tracking many commits, and clients could benefit from getting those commits from the catalog instead of from the file system. E.g. if checkpoint interval is 2500 commits, then it might be beneficial if the catalog serves up 2500 commits.

I would suggest simply making the choice of when to publish the joint responsibility of the client and the catalog.

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,489 @@
...
+section for more details.
+
+## Maintenance Operations on Catalog-owned Tables
+
+Delta clients must follow the owning catalog's rules, if any, when performing maintenance
+operations like [Checkpoints](#checkpoints-1), [Log Compaction Files](#log-compaction-files),
+[Metadata Cleanup](#metadata-cleanup), and VACUUM. These rules can be unique to each catalog and may
+evolve over time.
+
+In particular, maintenance operations like Checkpoints, Log Compaction Files, and Metadata Cleanup
```
bart-samwel, line 385: The previous paragraph says that all of these things are based on the catalog's rules, but then this says "in particular", as if it's an example of a catalog-based rule, but it actually states a universal rule, not a rule that a catalog may choose to require or not.

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,489 @@
...
+can only be performed on published versions, unless the catalog specifies otherwise.
+
+To highlight: Because the catalog controls the lifetime and potentially the storage of the table,
```
bart-samwel, line 388: This is also not a catalog rule, it's a universal rule. If the catalog is OK with the clients vacuuming, then that's OK. Clients should not try to VACUUM the table without having the catalog's permission (according to whatever the catalog's rules are).

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,489 @@
...
+
+## Writer Requirements for Catalog-owned tables
+
+When supported and active:
+
+- Writers must discover and access the table using catalog calls, which happens before the table's
+  protocol is known.
+- The [in-commit-timestamps](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#in-commit-timestamps)
+  table feature must be supported and active.
+- The `commitInfo` action must also a field `txnId` that stores a unique transaction identifier
```
bart-samwel, line 427:
```suggestion
- The `commitInfo` action must also contain a field `txnId` that stores a unique transaction identifier
```

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,489 @@
...
+
+## Reader Requirements for Catalog-owned tables
+
+When supported and active:
+
+- Readers must discover the table using catalog calls, which happens before the table's protocol
+  is known.
+- Readers must contact the catalog for information about unpublished ratified commits.
+
+## Sample client API
```
bart-samwel, line 441:
```suggestion
## Sample Catalog Client API
```

protocol_rfcs/catalog-owned.md
```diff
@@ -0,0 +1,489 @@
...
+     * @return GetCommitsResponse which contains an ordered list of ratified commits
+     *         tracked by the catalog, as well as table's latest commit version.
+     */
+    def getRatifiedCommits(
+        startVersion: Option[Long],
+        endVersion: Option[Long]): GetCommitsResponse
+}
+```
+
+Note that the above is only one example of a possible Delta Client API. It is also _NOT_ a catalog
```
bart-samwel, line 486:
```suggestion
Note that the above is only one example of a possible Catalog Client API. It is also _NOT_ a catalog
```

protocol_rfcs/catalog-owned.md
scottsand-db: > A commit doesn't always contain the content of a (particular) Delta file, i.e., there may not actually be a Delta file for a commit.

Hi Bart. The goal here was to communicate the general rules of the content that commits contain. The goal isn't to communicate where or how a commit can be stored (e.g. as you said, "in the file system as a delta file".

So --> I don't quite understand your emphasis of it always containing the same content as a _particular_ Delta file? Could you please elaborate?

I am certainly traying to convey that: commits represent the same information, thus "contain the same content" as Delta Log Entries. They contain a list of actions.

I realize now, by the way, that inline commits, for example, do *not* have to store actions as newline-delimited JSON? inline commits can probably be stored in whatever format the catalog wants?

protocol_rfcs/catalog-owned.md
scottsand-db: > [From the Commit Protocol section] To start, Delta Clients send the desired actions to be commited to the client-side component of the
catalog.
This component then has several options for proposing, ratifying, and publishing the commit,
detailed below.

I am totally open to that suggestion Bart. And thanks for calling out those great questions. My worry with "prepared" is that it seems you jump from "prepared" to "ratified" without any "proposing"? And I am inclined to avoid including _both_ prepared and proposed?

I think this section should actually say `A commit that the _client-side catalog component_ has proposed for ...`

- Step 1: Delta Clients send actions to the client-side catalog component
- Step 2: Client-side catalog component must **propose** them. It can write them to a staged commit file or send them inline to the server-side catalog component
- Suppose the client-side component wrote to a staged commit file but then never forwarded that location to the server-side catalog component. I'm okay still calling that a proposed commit? It's a proposed commit that the server-side catalog component never learned about, but the Delta Client sent the actions and the client-side catalog component did _something_ with them. It never got ratified -- this all seems fine?

I'm not trying to push back _too_ hard against your suggestion btw. Let me know what you think of the above. Thanks!

protocol_rfcs/catalog-owned.md
scottsand-db: Yes. I think I need to make it clearer that both staged commits and inline commits are (1) proposed and then (2) ratified.

protocol_rfcs/catalog-owned.md
scottsand-db: I'd be happy to change this to "store" -- why don't we do that?

protocol_rfcs/catalog-owned.md
scottsand-db: Fixed. @bart-samwel -- does the catalog _have_ to store the location of the staged ratified commit and the content of the inline ratified commit in persistent storage?

Can't it keep track of them purely in memory?

bart-samwel: If the catalog process dies, it shouldn't lose the information on ratified but unpublished commits. In theory it could store things in memory, if that memory is replicated N ways across multiple regions and has a lot of 9s of reliability. But at that point it's actually "persistent" in spirit. :)

protocol_rfcs/catalog-owned.md
bart-samwel: > So --> I don't quite understand your emphasis of it always containing the same content as a particular Delta file? Could you please elaborate?
It's the way it is phrased. "A commit contains the same content as a Delta file" can be interpreted as "it contains the same content as a particular Delta file, and a commit only exists by the grace of the existence of that Delta file". It could also be interpreted as "It contains the same *kind of* content as a Delta file", which is the interpretation that you mean. I was proposing to clarify this ambiguity.

> I realize now, by the way, that inline commits, for example, do not have to store actions as newline-delimited JSON? inline commits can probably be stored in whatever format the catalog wants?
Agreed!

protocol_rfcs/catalog-owned.md
bart-samwel: "Prepared" really is "staged", right? I.e., it lives in the `staged_commits` directory. I think the key problem that I have with the phrasing isn't here, but it is in the definition of "staged commit", where it says that a staged commit is always a proposed commit. What I'm saying is that a staged commit doesn't have to be proposed! The text below actually says that it could be a failed commit attempt. But I think it could also be a commit attempt that hasn't even reached the "proposed" state yet.

michelleon:
Thanks for the great work here! One minor naming suggestion: would the term `catalogManaged` be a clearer fit than `catalogOwned`?

“Managed” may better reflect the catalog’s role in governing both metadata and data operations like commits and reads. It also aligns with common terminology in systems like Snowflake-managed Iceberg tables[ [1] ](https://docs.snowflake.com/en/user-guide/tables-iceberg#use-snowflake-as-the-catalog)and Unity Catalog managed tables [[2],](https://docs.databricks.com/aws/en/tables/) where “managed” indicates that the catalog is responsible for the lifecycle of the table.

protocol_rfcs/catalog-owned.md
scottsand-db: I'll just delete this sentence then?

protocol_rfcs/catalog-owned.md
scottsand-db: Yes!

protocol_rfcs/catalog-owned.md
scottsand-db: > It's the way it is phrased. "A commit contains the same content as a Delta file" can be interpreted as "it contains the same content as a particular Delta file, and a commit only exists by the grace of the existence of that Delta file". It could also be interpreted as "It contains the same kind of content as a Delta file", which is the interpretation that you mean. I was proposing to clarify this ambiguity.

Ah, thanks for explaining! Makes sense.

protocol_rfcs/catalog-owned.md
scottsand-db: > What I'm saying is that a staged commit doesn't have to be proposed!

Sounds good 👍

scottsand-db:
@michelleon -- That's a great suggest. Will implement that. Thanks!

protocol_rfcs/catalog-owned.md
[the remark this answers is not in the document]
scovich: > Is this functionality already implemented in the current Delta codebase? For Catalog-Owned enabled tables, should Catalog Client takes the responsibility to check/block the attempt - or it depends on the actual Catalog implementation?

This sort of functionality belongs to the catalog's own client (Glue, UC, etc) that interacts with Delta. The RFC only cares to ensure that catalogs are _able_ to perform such checks, not whether/how they do it.

protocol_rfcs/catalog-managed.md
```diff
@@ -0,0 +1,492 @@
...
+## Terminology: Delta Client
+
+This is the component that implements support for reading and writing Delta tables, and implements
+the logic required by the `catalogManaged` table feature. Among other things, it
+- triggers the filesystem listing, if needed, to discover published commits
+- generates the commit content (the set of [actions](#actions))
+- works together with the query engine to trigger the commit process and invoke the client-side
+  catalog component with the commit content
+
+The Delta client is also responsible for defining the client-side API that catalogs should target.
```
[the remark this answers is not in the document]
scovich: Exactly. A given catalog client (UC, Glue, etc) will already have some API it uses to communicate with the catalog backend. But we also need an API the catalog client can use to communicate to Delta the subset of catalog-managed information that Delta cares about. This RFC is concerned with _what_ information Delta cares about, but leaves to Delta clients to design the API they use to obtain that information.

protocol_rfcs/catalog-managed.md
```diff
@@ -0,0 +1,492 @@
...
+The `catalogManaged` table feature is supported and active when:
+- The table is on Reader Version 3 and Writer Version 7.
+- The table has a `protocol` action with `readerFeatures` and `writerFeatures` both containing the
+  feature `catalogManaged`.
+
+## Writer Requirements for Catalog-managed tables
+
+When supported and active:
+
+- Writers must discover and access the table using catalog calls, which happens before the table's
```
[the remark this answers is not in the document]
scovich: Yes. Like v2 checkpoints, this table feature changes the way a Delta client discovers the table. In a sense, the table feature itself mostly serves as a warning to older/unaware Delta clients that this isn't an ordinary table.

That said, this requirement really just means that clients should be using name-based table access, which is already true in Delta-spark and several other clients. Those clients rely on the catalog to resolve a name into a table (which crucially includes the table's physical storage location). This RFC proposes that the catalog resolution step should also indicate whether this is a catalog-managed Delta table. 

Whether the catalog backend chooses to return ratified commits as part of the original resolution, or the catalog client acts on the first response to trigger a second call to the backend, is up to the catalog implementation to decide. The RFC just relieves the Delta itself from having to make "side" calls to the catalog in order to learn this kind of information, which was a major flaw in the original Coordinated Commits RFC.

protocol_rfcs/catalog-managed.md
```diff
@@ -0,0 +1,492 @@
...
+  catalog component with the commit content
+
+The Delta client is also responsible for defining the client-side API that catalogs should target.
+
+## Terminology: Catalogs
+
+1. **Catalog**: A catalog is a entity which manages a Delta table, including its creation, writes,
+   reads, and eventual deletion.
+    - It could be backed by a database, a filesystem, or any other persistence mechanism.
+    - Each catalog has its own spec around how Delta clients should interact with them, and how they
```
scovich, line 216:
```suggestion
    - Each catalog has its own spec around how catalog clients should interact with them, and how they
```
(the Delta client has an API for how catalog clients interact with _it_, not the other way around)

protocol_rfcs/catalog-managed.md
```diff
@@ -0,0 +1,492 @@
...
+    - This client-side component has two primary responsibilities:
+        - implement any client-side catalog-specific logic (such as [publishing](#publishing-commits))
+        - communicate with the Catalog Server, if any
+
+3. **Catalog Server**: The catalog may also involve a server-side component which the client-side
+   component would be responsible to communicate with.
+    - This server is responsible for coordinating commits and potentially persisting table metadata
+      and enforcing authorization policies.
+    - Not all catalogs require a server; some may be entirely client-side, e.g. filesystem-backed
+      catalogs, or they may make use of a generic database server and implement all of the logic
```
scovich, line 230:
```suggestion
      catalogs, or they may make use of a generic database server and implement all of the catalog's business logic
```

scovich (approved):
LGTM, but please address grammar nits before merging.

prakharjain09 (approved):

protocol_rfcs/catalog-managed.md
```diff
@@ -0,0 +1,492 @@
...
+   `_delta_log/_staged_commits/<v>.<uuid>.json`. It has the same content and format as a published
+   Delta file.
+    - Here, the `uuid` is a random UUID that is generated for each commit and `v` is the version
+      which is proposed to be committed, zero-padded to 20 digits.
+    - The catalog stores only the location, not the content, of a staged commit.
+    - The mere existence of one of these files does not mean that the file has been ratified or even
+      proposed. It might correspond to a failed or in-progress commit attempt.
+    - The catalog is the source of truth around which staged commits are ratified.
+
+3. **Inline commit** <a name="inline-commit"></a>: A proposed commit that is not written to disk but
```
xzhseh: Is catalog the only entity that can initiate the publishing of inline commits since the other components do not have enough information regarding them? Or this depends on the actual catalog implementation/interface/spec?

protocol_rfcs/catalog-managed.md
```diff
@@ -0,0 +1,492 @@
...
+  filesystem was dropped
+- commit 9 is an inline commit
+- neither of the two staged commits for version 10 have been ratified
+
+To read such tables, Delta clients must first contact the catalog to get the ratified commits. This
+informs the Delta client of commits [7, 9] as well as the latest ratified version, 9.
+
+If this information is insufficient to construct a complete snapshot of the table, Delta clients
+must LIST the `_delta_log` directory to get information about the published commits. For commits
+that are both returned by the catalog and already published, Delta clients can choose to read
```
xzhseh: What does the "_already published_" refer to? The published commits returned by the additional file-system LIST?

protocol_rfcs/README.md
```diff
@@ -16,12 +16,12 @@ Here is the history of all the RFCs propose/accepted/rejected since Feb 6, 2024,
...
+|:--------------|:---------------------------------------------------------------------------------------------------------------------------------|:----------------------------------------------|:------------------------------|
+| 2023-02-26    | [column-mapping-usage.tracking.md](https://github.com/delta-io/delta/blob/master/protocol_rfcs/column-mapping-usage-tracking.md) | https://github.com/delta-io/delta/issues/2682 | Column Mapping Usage Tracking |
+| 2023-04-24    | [variant-type.md](https://github.com/delta-io/delta/blob/master/protocol_rfcs/variant-type.md)                                   | https://github.com/delta-io/delta/issues/2864 | Variant Data Type             |
+| 2024-04-30    | [collated-string-type.md](https://github.com/delta-io/delta/blob/master/protocol_rfcs/collated-string-type.md)                   | https://github.com/delta-io/delta/issues/2894 | Collated String Type          |
+| 2025-04-07    | [catalog-owned.md](https://github.com/delta-io/delta/blob/master/protocol_rfcs/catalog-owned.md)                                 | https://github.com/delta-io/delta/issues/4381 | Catalog-owned Tables          |
```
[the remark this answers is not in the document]
scottsand-db: yup! Thanks!

protocol_rfcs/catalog-managed.md
```diff
@@ -0,0 +1,492 @@
...
+--------
+
+# Changes to existing sections
+
+### Delta Log Entries
+
+> ***Change to [existing section](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#delta-log-entries)***
+
+<ins>Delta Log Entries, also known as Delta files</ins>, are stored as JSON in a directory at the
+root of the table named `_delta_log`, and together with checkpoints make up the log of all changes
```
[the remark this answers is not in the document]
scottsand-db: Thanks!

xzhseh (approved):
LGTM w/ two clarification questions!

protocol_rfcs/catalog-managed.md
```diff
@@ -0,0 +1,492 @@
...
+```
+
+<ins>Delta files use newline-delimited JSON format, where every action is stored as a single-line
+JSON document. A Delta file, corresponding to version `v`, contains an atomic set of
+[_actions_](#Actions) that should be applied to the previous table state corresponding to version
+`v-1`, in order to construct the `v`th snapshot of the table. An action changes one aspect of the
+table's state, for example, adding or removing a file.</ins>
+
+<ins>**Note:** If the [`catalogManaged` table feature](#catalog-managed-tables) is enabled on the table,
+recently ratified commits may not yet be published to the `_delta_log` directory as normal Delta
```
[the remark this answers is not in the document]
scottsand-db: Done!

protocol_rfcs/catalog-managed.md
```diff
@@ -0,0 +1,492 @@
...
+./_delta_log/_staged_commits/00000000000000000001.7d17ac10-5cc3-401b-bd1a-9c82dd2ea032.json
+./_delta_log/_staged_commits/00000000000000000001.016ae953-37a9-438e-8683-9a9a4a79a395.json
+./_delta_log/_staged_commits/00000000000000000002.3ae45b72-24e1-865a-a211-34987ae02f2a.json
+```
+
+<ins>NOTE: The (proposed) version number of a staged commit is authoritative - file
+`00000000000000000100.<uuid>.json` always corresponds to a commit attempt for version 100. Besides
+simplifying implementations, it also acknowledges the fact that commit files cannot safely be reused
+for multiple commit attempts. For example, resolving conflicts in a table with [row
+tracking](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#row-tracking) enabled requires
```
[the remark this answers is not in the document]
scottsand-db: Yup! I guess we can deal with that when we accept and merge the RFC into the protocol. Great callout!

protocol_rfcs/catalog-managed.md
```diff
@@ -0,0 +1,492 @@
...
+
+A commit may be stored in the file system as a Delta file - either _published_ or _staged_ - or
+stored _inline_ in the managing catalog, using whatever format the catalog prefers.
+
+There are several types of commits:
+
+1. **Proposed commit**:  A commit that a Delta client has proposed for the next version of the
+   table. It could be _staged_ or _inline_. It will either become _ratified_ or be rejected.
+
+2. **Staged commit** <a name="staged-commit"></a>: A commit that is written to disk at
```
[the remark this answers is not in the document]
scottsand-db: I did intend it -- yeah. We want to be able to link to this section ... but I didn't feel the need to make the text `Staged Commit` itself a link (that points to itself). Happy to change it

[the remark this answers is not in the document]
scottsand-db: Actually -- scratch that -- I can just make it clickable / linkable 👍

protocol_rfcs/catalog-managed.md
```diff
@@ -0,0 +1,492 @@
...
+   `_delta_log/_staged_commits/<v>.<uuid>.json`. It has the same content and format as a published
+   Delta file.
+    - Here, the `uuid` is a random UUID that is generated for each commit and `v` is the version
+      which is proposed to be committed, zero-padded to 20 digits.
+    - The catalog stores only the location, not the content, of a staged commit.
+    - The mere existence of one of these files does not mean that the file has been ratified or even
```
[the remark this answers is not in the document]
scottsand-db: thx!

protocol_rfcs/catalog-managed.md
```diff
@@ -0,0 +1,492 @@
...
+
+1. **Proposed commit**:  A commit that a Delta client has proposed for the next version of the
+   table. It could be _staged_ or _inline_. It will either become _ratified_ or be rejected.
+
+2. **Staged commit** <a name="staged-commit"></a>: A commit that is written to disk at
+   `_delta_log/_staged_commits/<v>.<uuid>.json`. It has the same content and format as a published
+   Delta file.
+    - Here, the `uuid` is a random UUID that is generated for each commit and `v` is the version
+      which is proposed to be committed, zero-padded to 20 digits.
+    - The catalog stores only the location, not the content, of a staged commit.
```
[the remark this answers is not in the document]
scottsand-db: clarified!

protocol_rfcs/catalog-managed.md
```diff
@@ -0,0 +1,492 @@
...
+    - A ratified commit may or may not yet be published.
+    - A ratified commit may or may not even be stored by the catalog at all - the catalog may
+      have just atomically published it to the filesystem directly, relying on PUT-if-absent
+      primitives to facilitate the ratification and publication all in one step.
+
+5. **Published commit** <a name="published-commit"></a>: A ratified commit that has been copied into
+   the `_delta_log` as a normal Delta file, i.e. `_delta_log/<v>.json`.
+    - Here, the `v` is the version which is being committed, zero-padded to 20 digits.
+    - The existence of a `<v>.json` file proves that the corresponding version `v` is ratified,
+      regardless of whether the table is catalog-managed for filesystem-based. The catalog is allowed
```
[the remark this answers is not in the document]
scottsand-db: nice catch!

protocol_rfcs/catalog-managed.md
scovich: Essentially, the Delta client always passes an inline commit to the catalog client; the catalog client decides whether to forward the bytes directly to catalog, or stage them to a uuid commit file and pass the pointer to catalog, or locally initiate an atomic put-if-absent (i.e. for HMS or the fake catalog implementation used by path-based tables).

protocol_rfcs/catalog-managed.md
scovich: Exactly.

protocol_rfcs/catalog-managed.md
scottsand-db: "initiate" is a bit vague here. Let me try to answer this as best I can:

- engines want to write.
- engines will work with the delta connector to write the right parquet data and generate the right delta metadata information for the commit
- engines and the delta connector will work together to invoke the client-side catalog component APIs to send it all the necessary delta commit metadata information. that client-side catalog component API will then "commit" the "the catalog" using whatever API it wants (RPC, REST, etc.)

protocol_rfcs/catalog-managed.md
scottsand-db: yup!

protocol_rfcs/catalog-managed.md
```diff
@@ -0,0 +1,492 @@
...
+
+For performance reasons, Delta clients are encouraged to establish an API contract where the catalog
+provides the latest ratified commit information whenever a commit fails due to version conflict.
+
+## Getting Ratified Commits from the Catalog
+
+Even after a commit is ratified, it is not discoverable through filesystem operations until it is
+[published](#publishing-commits).
+
+The catalog is responsible to implement an API (defined by the Delta client) that Delta clients can
```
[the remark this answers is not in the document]
scottsand-db: Yup! Done!

protocol_rfcs/catalog-managed.md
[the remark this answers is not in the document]
scottsand-db: Im proposing I add this seciton below to address your feedback @LukasRupprecht ?

```
## Table Discovery

The requirements above state that readers and writers must discover and access the table using
catalog calls, which occurs _before_ the table's protocol is known. This raises an important
question: how can a client discover a `catalogManaged` Delta table without first knowing that it
_is_, in fact, `catalogManaged` (according to the protocol)?

To solve this, this protocol encourages that the catalog resolution step - which occurs, for example,
during name-based access to Delta tables - should also indicate whether the table is catalog-managed.
```

PR #4387, opened 2025-04-08
[Spark] Add option to DeltaParquetFileFormat to allow missing Row Tracking metadata fields

tomvanbussel:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR:

- Renames `nullableRowTrackingFields` in `DeltaParquetFileFormat` to `nullableRowTrackingGeneratedFields`.
- Adds another field called `nullableRowTrackingConstantFields` in `DeltaParquetFileFormat`, which allows reading row ids and row commit versions when the `baseRowId` and `defaultRowCommitVersion`s are missing (returning a NULL in this case).
- It adds a helper function `transformFileFormat` to `DeltaTableUtils` to make it easier to set the new field.

## How was this patch tested?

Added a test to `RowIdSuite`.

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DefaultRowCommitVersion.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DefaultRowCommitVersion.scala
@@ -41,9 +41,9 @@ object DefaultRowCommitVersion {
   }
 
   def createDefaultRowCommitVersionField(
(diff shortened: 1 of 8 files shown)


larsk-db (approved):
LGTM

PR #4390, opened 2025-04-09
[Spark] Make MERGE source checkpoint materialization a SQL operation

juliuszsompolski:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Because forcing source checkpoint in MergeIntoMaterializeSource.prepareMergeSource is triggered with RDD operation, it doesn't show in SQL UI and in QPL. Now in Spark 4.0, Dataset.localCheckpoint supports passing storageLevel natively, so we can use that directly without using eager = false and triggering evaluation ourselves.

We also remove the MERGE_MATERIALIZE_SOURCE_EAGER flag, as it was shown that it's incorrect to use lazy materialization.

## How was this patch tested?

Existing tests.
* MergeIntoMaterializedSourceSuite: no changes needed. There are some tests checking plans but they check for the existance of RDD named "mergeMaterializedSource" in the plan. Before: it didn't appear during materialization, because it was not a SQL operation, so there was no plan recorded. After: it doesn't appear, because the RDD is only named after materialization finishes.

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/commands/merge/MergeIntoMaterializeSource.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/commands/merge/MergeIntoMaterializeSource.scala
@@ -83,6 +83,18 @@ trait MergeIntoMaterializeSource extends DeltaLogging with DeltaSparkPlanUtils {
   protected var materializedSourceRDD: Option[RDD[InternalRow]] = None
 
   /**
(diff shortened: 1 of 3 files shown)


juliuszsompolski:
@larsk-db Could you review?
This can finally be merged because Delta dropped spark-3.5

PR #4392, opened 2025-04-09
[SPARK] Refactor error test suites to use `checkError()`

ala:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This change refactor Delta errors testing to use `checkError()` instead of `checkErrorMessage()`. This has multiple benefits:
- we can edit the error messages for style without having to alter the test code every time
- we can make sure that the error message template parameters are correctly filled, the number of parameters matches what is expected by the template, and the parameters with the same name are filled with the same value
- we are more consistently testing if the SQL states are set and propagated correctly

To that end, it was necessary to edit some of the exception classes to make sure they expose `getMessageParamters()` method correctly.

Some of the error message were edited to avoid unnamed parameters, and to correct obvious problems with formatting.

I also discovered and address an issue where using `AnalysisException.withPosition()` method would transform a `DeltaAnalysisException` into a malformed `AnalysisException` without a SQL state.

## How was this patch tested?

The changes are primarily impacting the tests themselves.

- Minor changes to a handful of error message templates with obvious formatting defects.
- A `DeltaAnalysisException` might be returned in pace of malformed `AnalysisException` in certain scenarios.

### DIFF
--- a/spark/src/main/resources/error/delta-error-classes.json
+++ b/spark/src/main/resources/error/delta-error-classes.json
@@ -299,13 +299,13 @@
   },
   "DELTA_CANNOT_UPDATE_ARRAY_FIELD" : {
     "message" : [
-      "Cannot update %1$s field %2$s type: update the element by updating %2$s.element"
+      "Cannot update <tableName> field <fieldName> type: update the element by updating `<fieldName>.element`."
     ],
     "sqlState" : "429BQ"
   },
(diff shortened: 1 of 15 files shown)


c27kwan (approved):
lgtm! Nice cleanup and thanks for fixing some of the error messages! This will make our error testing framework more flexible and robust moving forward :D

bart-samwel (approved):

PR #4398, opened 2025-04-10
[Kernel] Add `removeFileSizeInBytes` to transaction metrics and increment when removes are committed

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Follow-up to https://github.com/delta-io/delta/commit/bdd3c520ca20c03733d07bfff4261c2ebfc8db90.

## How was this patch tested?

Adds and updates unit tests.

No

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaErrorsInternal.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaErrorsInternal.java
@@ -22,6 +22,8 @@ package io.delta.kernel.internal;
  */
 public class DeltaErrorsInternal {
 
+  private DeltaErrorsInternal() {}
+
   public static IllegalStateException missingRemoveFileSizeDuringCommit() {
     return new IllegalStateException(
         "Kernel APIs for creating remove file rows require that "
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaErrorsInternal.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaErrorsInternal.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright (2025) The Delta Lake Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.delta.kernel.internal;
+
+/**
+ * Contains methods to create developer-facing exceptions. See <a
+ * href="https://github.com/delta-io/delta/blob/master/kernel/EXCEPTION_PRINCIPLES.md">Exception
+ * Principles</a> for more information on what these are and how to use them.
+ */
+public class DeltaErrorsInternal {
+
+  public static IllegalStateException missingRemoveFileSizeDuringCommit() {
+    return new IllegalStateException(
+        "Kernel APIs for creating remove file rows require that "
+            + "file size be provided but found null file size");
+  }
+}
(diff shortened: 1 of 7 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -486,6 +479,30 @@ private long doCommit(
...
+          removeFile
+              .getSize()
+              .orElseThrow(
+                  () ->
+                      new IllegalStateException(
+                          "Kernel APIs for creating remove file rows "
+                              + "require that file size be provided but "
+                              + "found null file size"));
+      txnMetrics.removeFilesSizeInBytesCounter.increment(fileSize);
+      // TODO increment fileSizeHistogram
```
huan233usc, line 502: nit: decrease fileSizeHistogram

huan233usc (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -486,6 +479,30 @@ private long doCommit(
     }
   }
 
+  private void incrementMetricsForFileActionRow(Row fileActionRow, TransactionMetrics txnMetrics) {
```
scottsand-db, line 482: nit: In general, the object being acted upon should come before the object doing the acting.

i.e. we are applying the fileActionRow (i.e. extracting metrics from it) and then applying that information to the txnMetrics, then the param order should be `txnMetrics, fileActionRow`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/TransactionMetrics.java
```diff
@@ -52,6 +52,7 @@ public void resetActionCounters() {
     addFilesSizeInBytesCounter.reset();
     removeFilesCounter.reset();
     totalActionsCounter.reset();
+    removeFilesSizeInBytesCounter.reset();
```
scottsand-db, line 55: Should we rename this method to `resetCounters` ? since `removeFilesSizeInBytesCounter` is not an action counter?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -486,6 +479,30 @@ private long doCommit(
...
+      // TODO increment fileSizeHistogram
+    } else if (!fileActionRow.isNullAt(REMOVE_FILE_ORDINAL)) {
+      txnMetrics.removeFilesCounter.increment();
+      RemoveFile removeFile = new RemoveFile(fileActionRow.getStruct(REMOVE_FILE_ORDINAL));
+      long fileSize =
+          removeFile
+              .getSize()
+              .orElseThrow(
+                  () ->
+                      new IllegalStateException(
```
scottsand-db, line 497: Question: is it worth creating our internal delta errors class? would that clean up the code?

scottsand-db (changes requested):
Looks great! Requested tiny changes and left 1 question, then will stamp!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
scottsand-db: And we should have a kernel java style guide that lists these preferences :D

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
allisonport-db: I'm not opposed to this but in what way does this clean up the code?

scottsand-db: It would be 1 line `DeltaErrorsXXX.yyy()` instead of 4

allisonport-db: Ah I see :) I was thinking error class like `KernelIllegalStateException` haha

Sure that makes sense to me

raveeram-db (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
allisonport-db: Done

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaErrorsInternal.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+package io.delta.kernel.internal;
+
+/**
+ * Contains methods to create developer-facing exceptions. See <a
+ * href="https://github.com/delta-io/delta/blob/master/kernel/EXCEPTION_PRINCIPLES.md">Exception
+ * Principles</a> for more information on what these are and how to use them.
+ */
+public class DeltaErrorsInternal {
```
scottsand-db, line 23: private constructor and then ship it! :)

scottsand-db (approved):
LGTM after private constructor in DeltaErrorsInternal

vkorukanti (approved):

PR #4402, opened 2025-04-11
[Kernel] Schema validation for clustering columns

amogh-jahagirdar:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This PR includes validation for clustering columns. Namely, that clustering columns cannot be dropped.
It allows for schema evolutions where clustering columns are renamed.

## How was this patch tested?

Updated/added new unit tests

The `withSchema` API will allow certain operations like renames to be performed on clustering columns

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
@@ -373,24 +373,26 @@ public class TransactionBuilderImpl implements TransactionBuilder {
         throw new KernelException("Cannot update mapping mode and perform schema evolution");
       }
 
+      // If the column mapping restriction is removed, clustering columns
+      // will need special handling during schema evolution since they won't have physical names
+      // ToDo: Support adding clustering columns
+
       if (!isColumnMappingModeEnabled(updatedMappingMode)) {
         throw new KernelException("Cannot update schema for table when column mapping is disabled");
       }
 
-      // TODO: revisit this once we want to support schema evolution with clustering columns
-      Optional<List<Column>> clusteringColumns =
-          ClusteringUtils.getClusteringColumnsOptional(snapshot);
-      if (clusteringColumns.isPresent() && !clusteringColumns.get().isEmpty()) {
-        throw new KernelException(
-            format(
-                "Update schema for table with clustering columns %s is not yet supported",
-                clusteringColumns.get()));
-      }
+      // Clustering columns will be guaranteed to have physical names at this point
+      Set<String> clusteringColumnPhysicalNames =
+          ClusteringUtils.getClusteringColumnsOptional(snapshot).orElse(Collections.emptyList())
+              .stream()
+              .map(col -> col.getNames()[0])
+              .collect(toSet());
 
       SchemaUtils.validateUpdatedSchema(
           oldMetadata.getSchema(),
           newMetadata.getSchema(),
           oldMetadata.getPartitionColNames(),
+          clusteringColumnPhysicalNames,
           newMetadata);
     }
   }
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
@@ -382,10 +382,14 @@ public class TransactionBuilderImpl implements TransactionBuilder {
       }
 
       // Clustering columns will be guaranteed to have physical names at this point
+      // Only the leaf part of the overall column needs to be taken since
+      // validation is performed on the leaf struct fields
+      // E.g. getClusteringColumns returns <physical_name_of_struct>.<physical_name_inner>,
+      // Only physical_name_inner is required for validation
       Set<String> clusteringColumnPhysicalNames =
           ClusteringUtils.getClusteringColumnsOptional(snapshot).orElse(Collections.emptyList())
               .stream()
-              .map(col -> col.getNames()[0])
+              .map(col -> col.getNames()[col.getNames().length - 1])
               .collect(toSet());
 
       SchemaUtils.validateUpdatedSchema(
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
@@ -104,6 +104,7 @@ public class SchemaUtils {
       StructType currentSchema,
       StructType newSchema,
       Set<String> currentPartitionColumns,
+      Set<String> clusteringColumnPhysicalNames,
       Metadata newMetadata) {
     checkArgument(
         isColumnMappingModeEnabled(
@@ -111,7 +112,7 @@ public class SchemaUtils {
         "Cannot validate updated schema when column mapping is disabled");
     validateSchema(newSchema, true /*columnMappingEnabled*/);
     validatePartitionColumns(newSchema, new ArrayList<>(currentPartitionColumns));
-    validateSchemaEvolution(currentSchema, newSchema, newMetadata);
+    validateSchemaEvolution(currentSchema, newSchema, newMetadata, clusteringColumnPhysicalNames);
   }
 
   /**
@@ -423,13 +424,16 @@ public class SchemaUtils {
 
   /* Validate if a given schema evolution is safe for a given column mapping mode*/
   private static void validateSchemaEvolution(
-      StructType currentSchema, StructType newSchema, Metadata metadata) {
+      StructType currentSchema,
+      StructType newSchema,
+      Metadata metadata,
+      Set<String> clusteringColumnPhysicalNames) {
     ColumnMappingMode columnMappingMode =
         ColumnMapping.getColumnMappingMode(metadata.getConfiguration());
     switch (columnMappingMode) {
       case ID:
       case NAME:
-        validateSchemaEvolutionById(currentSchema, newSchema);
+        validateSchemaEvolutionById(currentSchema, newSchema, clusteringColumnPhysicalNames);
         return;
       case NONE:
         throw new UnsupportedOperationException(
@@ -444,7 +448,8 @@ public class SchemaUtils {
    * Validates a given schema evolution by using field ID as the source of truth for identifying
    * fields
    */
-  private static void validateSchemaEvolutionById(StructType currentSchema, StructType newSchema) {
+  private static void validateSchemaEvolutionById(
+      StructType currentSchema, StructType newSchema, Set<String> clusteringColumnPhysicalNames) {
     Map<Integer, StructField> currentFieldsById = fieldsById(currentSchema);
     Map<Integer, StructField> updatedFieldsById = fieldsById(newSchema);
     SchemaChanges schemaChanges = computeSchemaChangesById(currentFieldsById, updatedFieldsById);
@@ -452,7 +457,19 @@ public class SchemaUtils {
     // Validates that the updated schema does not contain breaking changes in terms of types and
     // nullability
     validateUpdatedSchemaCompatibility(schemaChanges);
+    validateClusteringColumnsNotDropped(
+        schemaChanges.removedFields(), clusteringColumnPhysicalNames);
     // ToDo Potentially validate IcebergCompatV2 nested IDs
+  }
+
+  private static void validateClusteringColumnsNotDropped(
+      List<StructField> droppedFields, Set<String> clusteringColumnPhysicalNames) {
+    for (StructField droppedField : droppedFields) {
+      if (clusteringColumnPhysicalNames.contains(getPhysicalName(droppedField))) {
+        throw new KernelException(
+            String.format("Cannot drop clustering column %s", droppedField.getName()));
+      }
+    }
   }
 
   /**
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
@@ -465,6 +465,8 @@ public class SchemaUtils {
   private static void validateClusteringColumnsNotDropped(
       List<StructField> droppedFields, Set<String> clusteringColumnPhysicalNames) {
     for (StructField droppedField : droppedFields) {
+      // ToDo: At some point plumb through mapping of ID to full name, so we get better error
+      // messages
       if (clusteringColumnPhysicalNames.contains(getPhysicalName(droppedField))) {
         throw new KernelException(
             String.format("Cannot drop clustering column %s", droppedField.getName()));
--- a/kernel/kernel-api/src/test/scala/io/delta/kernel/internal/util/SchemaUtilsSuite.scala
+++ b/kernel/kernel-api/src/test/scala/io/delta/kernel/internal/util/SchemaUtilsSuite.scala
@@ -17,6 +17,7 @@ package io.delta.kernel.internal.util
 
 import java.util
 import java.util.{Locale, Optional}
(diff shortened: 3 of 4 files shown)


amogh-jahagirdar:
We still currently block adding new clustering columns, which I think we'll need. I'm discussing with @KaiqiJinWow

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -423,13 +424,16 @@ private static void validatePhysicalNameConsistency(
...
+      StructType newSchema,
+      Metadata metadata,
+      Set<String> clusteringColumnPhysicalNames) {
     ColumnMappingMode columnMappingMode =
         ColumnMapping.getColumnMappingMode(metadata.getConfiguration());
     switch (columnMappingMode) {
       case ID:
       case NAME:
-        validateSchemaEvolutionById(currentSchema, newSchema);
+        validateSchemaEvolutionById(currentSchema, newSchema, clusteringColumnPhysicalNames);
```
vkorukanti, line 436: Add a comment at the place where you block schema change if column mapping is disabled. Just incase if we remove that constraint in future to handle the clustering columns in schema evolution (e.g. adding a new clustering column).

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/util/SchemaUtilsSuite.scala
```diff
@@ -732,7 +738,12 @@ class SchemaUtilsSuite extends AnyFunSuite {
 
   test("validateUpdatedSchema succeeds with valid ID and physical name") {
     forAll(validUpdatedSchemas) { (schemaBefore, schemaAfter) =>
-      validateUpdatedSchema(schemaBefore, schemaAfter, Set.empty.asJava, metadata(schemaBefore))
+      validateUpdatedSchema(
+        schemaBefore,
+        schemaAfter,
+        Set.empty.asJava,
+        Set.empty.asJava,
+        metadata(schemaBefore))
```
vkorukanti, line 746:
```suggestion
      validateUpdatedSchema(schemaBefore, schemaAfter, emptySet, emptySet, metadata(schemaBefore))
```

vkorukanti (approved):
Looks great! One qn: what is delta-spark behavior for adding new clustering columns?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -444,17 +448,30 @@ private static void validateSchemaEvolution(
...
+  private static void validateSchemaEvolutionById(
+      StructType currentSchema, StructType newSchema, Set<String> clusteringColumnPhysicalNames) {
     Map<Integer, StructField> currentFieldsById = fieldsById(currentSchema);
     Map<Integer, StructField> updatedFieldsById = fieldsById(newSchema);
     SchemaChanges schemaChanges = computeSchemaChangesById(currentFieldsById, updatedFieldsById);
     validatePhysicalNameConsistency(schemaChanges.updatedFields());
     // Validates that the updated schema does not contain breaking changes in terms of types and
     // nullability
     validateUpdatedSchemaCompatibility(schemaChanges);
+    validateClusteringColumnsNotDropped(
```
vkorukanti, line 460: qn: should we check if the columns are relocated or moved?

vkorukanti (changes requested):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
amogh-jahagirdar: I think whether a clustering column is reordered is inconsequential insofar as it is not an invalid reordering (like moving it out of its containing struct). That invalid reordering will happen in a separate PR, but clustering columns are just columns for that validation's purpose.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
amogh-jahagirdar: I added a comment, though I think the main thing to consider with regards to clustering columns in case the column mapping restraint is removed is that we won't have the ability to properly validate changes since we won't have physical names. I added a separate ToDo comment for supporting adding a new clustering columns.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -373,24 +373,26 @@ private void validateMetadataChange(
...
-        throw new KernelException(
-            format(
-                "Update schema for table with clustering columns %s is not yet supported",
-                clusteringColumns.get()));
-      }
+      // Clustering columns will be guaranteed to have physical names at this point
+      Set<String> clusteringColumnPhysicalNames =
+          ClusteringUtils.getClusteringColumnsOptional(snapshot).orElse(Collections.emptyList())
+              .stream()
+              .map(col -> col.getNames()[0])
```
allisonport-db, line 388: why do we only want the top-level name?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableSchemaEvolutionSuite.scala
```diff
@@ -1160,33 +1196,31 @@ class DeltaTableSchemaEvolutionSuite extends DeltaTableWriteSuiteBase with Colum
     }
   }
 
-  // TODO: need to remove this once we support schema evolution with clustering columns
-  test("Cannot update schema with clustering columns") {
+  test("Cannot drop clustering columns") {
```
allisonport-db, line 1199: Test dropping a nested clustering column?

allisonport-db (changes requested):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
amogh-jahagirdar: I realized it should actually be only the leaf field when testing this a bit more. The getClusteringColumn API is a bit odd in that it returns a `Column` where every part is just the physical name so like `<physical_coluumn_for_struct>.<physical_column_for_inner_field>. `

I feel like the right thing to do is just return the leaf field for that API. There's no use as far as I can tell to the other parts when column mapping is enabled.


Anyways, since validation is performed on every leaf field, we just need to the leaf field physical name, that uniquely identifies the inner field anyways.

amogh-jahagirdar: I'll add a comment for this since I realize it's not obvious

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableSchemaEvolutionSuite.scala
```diff
@@ -1196,31 +1200,63 @@ class DeltaTableSchemaEvolutionSuite extends DeltaTableWriteSuiteBase with Colum
...
+            schemaBefore,
+            clusteringCols = List(clusteringColumn),
+            tableProperties = Map(
+              TableConfig.COLUMN_MAPPING_MODE.getKey -> "id",
+              TableConfig.ICEBERG_COMPAT_V2_ENABLED.getKey -> "true"))
+          assertSchemaEvolutionFails[KernelException](
+            table,
+            engine,
+            schemaAfter,
+            "Cannot drop clustering column clustering_col")
```
amogh-jahagirdar, line 1258: I added a ToDo to make sure our failure messages have the full path name, it's something we should really fix now :(. For now I've left it as is.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableSchemaEvolutionSuite.scala
amogh-jahagirdar: Done!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
amogh-jahagirdar: @allisonport-db let me know if the comment makes sense, I'll leave this thread unresolved for now

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableSchemaEvolutionSuite.scala
```diff
@@ -1160,33 +1200,63 @@ class DeltaTableSchemaEvolutionSuite extends DeltaTableWriteSuiteBase with Colum
...
+          val table = Table.forPath(engine, tablePath)
+          createEmptyTable(
+            engine,
+            tablePath,
+            schemaBefore,
+            clusteringCols = List(clusteringColumn),
+            tableProperties = Map(
+              TableConfig.COLUMN_MAPPING_MODE.getKey -> "id",
+              TableConfig.ICEBERG_COMPAT_V2_ENABLED.getKey -> "true"))
+          assertSchemaEvolutionFails[KernelException](
```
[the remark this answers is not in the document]
allisonport-db: A simple test for this might also be not dropping the clustering column from the schema and making sure it passes

allisonport-db (changes requested):
Please make sure the tests are correctly testing dropped columns

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableSchemaEvolutionSuite.scala
```diff
@@ -1160,33 +1200,63 @@ class DeltaTableSchemaEvolutionSuite extends DeltaTableWriteSuiteBase with Colum
...
-      val newSchema = new StructType()
-        .add("c", StringType.STRING, true, currentSchema.get("c").getMetadata)
-        .add("a", StringType.STRING, true, currentSchema.get("a").getMetadata)
-
-      assertSchemaEvolutionFails[KernelException](
-        table,
-        engine,
-        newSchema,
-        "Update schema for table with clustering columns")
+  val primitiveSchemaWithClusteringColumn = new StructType()
```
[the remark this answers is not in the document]
amogh-jahagirdar: Oh I see you mean dropping the whole struct. Sure we can do that

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableSchemaEvolutionSuite.scala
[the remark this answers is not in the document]
amogh-jahagirdar: > I'm concerned that we're seeing all columns as "dropped" since we don't have fieldIds or physical names in the schemaAfter

I don't think that's true, if you look at schemaAfter we're carrying over the metadata attributes which includes the field Ids and physical names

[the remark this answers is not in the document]
allisonport-db: Yeah but I don't think that metadata is populated yet when you build the schemaAfter (we don't populate until in the txn or if you manually call the method to add the CM metadata)

```
new StructType()
    .add("clustering_col", IntegerType.INTEGER)
    .add("data", IntegerType.INTEGER)
```
does NOT have any CM metadata!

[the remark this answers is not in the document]
amogh-jahagirdar: Oh...I previousy was carrying over the metadata attributes but now with the parametrized I'm not. Ok let me fix this

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableSchemaEvolutionSuite.scala
```diff
@@ -1160,33 +1200,68 @@ class DeltaTableSchemaEvolutionSuite extends DeltaTableWriteSuiteBase with Colum
...
+    .add("clustering_col", IntegerType.INTEGER, fieldMetadataForColumn(1, "clustering_col"))
```
allisonport-db, line 1204: nit - but I feel like the physicalName != the logical name for this test case (to make sure we use the right one when they are not the same)

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableSchemaEvolutionSuite.scala
```diff
@@ -1160,33 +1200,68 @@ class DeltaTableSchemaEvolutionSuite extends DeltaTableWriteSuiteBase with Colum
...
+      new StructType()
+        .add(
+          "data",
+          IntegerType.INTEGER,
+          true,
+          primitiveSchemaWithClusteringColumn.get("data").getMetadata),
+      new Column("clustering_col")),
+    (
+      nestedSchemaWithClusteringColumn,
+      new StructType()
```
allisonport-db, line 1229: not necessary for this PR but it might simplify the tests a lot (not just here, in all the schema code) if we added a test utils that supported something like `schema.withColumnDropped(x)` -- then we don't need to redefine the schema in each place and can just drop columns!

allisonport-db: just an idea to improve the tests -- maybe note this down since I think you're already tracking test improvements

amogh-jahagirdar: Yeah I was planning on doing this on the test cleanup PR, I think the util in test to just drop/rename by field name is a fine way of handling things for now.

 I think in the long run, with a visitor we'd be able to more generically apply a set of whatever changes to the schema and produce a new schema (and this would also be on the general schema update path as well) so we don't have to redefine things.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableSchemaEvolutionSuite.scala
amogh-jahagirdar: Sure I think that's a good call, I'll fix this

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -444,17 +448,32 @@ private static void validateSchemaEvolution(
    * Validates a given schema evolution by using field ID as the source of truth for identifying
    * fields
    */
-  private static void validateSchemaEvolutionById(StructType currentSchema, StructType newSchema) {
+  private static void validateSchemaEvolutionById(
+      StructType currentSchema, StructType newSchema, Set<String> clusteringColumnPhysicalNames) {
```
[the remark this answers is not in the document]
amogh-jahagirdar: For now, we're not allowing widening of _any_ columns or any type change. Clustering columns are undifferentiated in that regard (after all it's still just a field going through the existing validation). 

 I'd say when we start opening up type promotion, then we'd do whatever clustering specific validation we need to do then.

There's an existing test for making sure no types can be changed https://github.com/delta-io/delta/blob/master/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableSchemaEvolutionSuite.scala#L1033 so I think we're covered since clustering columns still just go through this same validation.

PR #4407, opened 2025-04-11
[Kernel]Add domain metadata to CRC Info and implements the toRow() 

huan233usc:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This is 1/n for domain metadata in crc support, it adds domain metadata to crcInfo.java and implements the toRow() method. 

(see https://github.com/delta-io/delta/pull/4330/files for e2e poc)

Follow up will be: (1)e2e integration to populate domain metadata in crc in a txn (2)load domain metadata from crc and use it for building snapshot.

## How was this patch tested?

Unit test

No

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/CRCInfo.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/CRCInfo.java
@@ -25,7 +25,6 @@ import io.delta.kernel.internal.actions.DomainMetadata;
 import io.delta.kernel.internal.actions.Metadata;
 import io.delta.kernel.internal.actions.Protocol;
 import io.delta.kernel.internal.data.GenericRow;
-import io.delta.kernel.internal.data.StructRow;
 import io.delta.kernel.internal.stats.FileSizeHistogram;
 import io.delta.kernel.internal.util.InternalUtils;
 import io.delta.kernel.internal.util.VectorUtils;
@@ -86,14 +85,6 @@ public class CRCInfo {
         txnIdColumnVector.isNullAt(rowId)
             ? Optional.empty()
             : Optional.of(txnIdColumnVector.getString(rowId));
-    ColumnVector domainMetadataVector = batch.getColumnVector(getSchemaIndex(DOMAIN_METADATA));
-    Optional<Set<DomainMetadata>> domainMetadata =
-        domainMetadataVector.isNullAt(rowId)
-            ? Optional.empty()
-            : Optional.of(
-                VectorUtils.toJavaList(domainMetadataVector.getArray(rowId)).stream()
-                    .map(row -> DomainMetadata.fromRow((StructRow) row))
-                    .collect(Collectors.toSet()));
     Optional<FileSizeHistogram> fileSizeHistogram =
         FileSizeHistogram.fromColumnVector(
             batch.getColumnVector(getSchemaIndex(FILE_SIZE_HISTOGRAM)), rowId);
@@ -111,7 +102,8 @@ public class CRCInfo {
             tableSizeBytes,
             numFiles,
             txnId,
-            domainMetadata,
+            // TODO: load domain metadata from CRC.
+            Optional.empty(),
             fileSizeHistogram));
   }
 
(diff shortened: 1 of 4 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/CRCInfo.java
```diff
@@ -196,6 +233,7 @@ public boolean equals(Object o) {
         && metadata.equals(other.metadata)
         && protocol.equals(other.protocol)
         && txnId.equals(other.txnId)
+        && domainMetadata.equals(other.domainMetadata)
```
raveeram-db, line 236: Perhaps not that important but should we be mindful of ordering the List here? IIRC, `List.equals()` is order-sensitive? I would assume it's not guaranteed that they'll be written in the same order always

huan233usc: Good point, I think we could just make domainMetadata a Optional<Set>

kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/CRCInfo.java
```diff
@@ -82,6 +86,14 @@ public static Optional<CRCInfo> fromColumnarBatch(
         txnIdColumnVector.isNullAt(rowId)
             ? Optional.empty()
             : Optional.of(txnIdColumnVector.getString(rowId));
+    ColumnVector domainMetadataVector = batch.getColumnVector(getSchemaIndex(DOMAIN_METADATA));
```
allisonport-db, line 89: If load is gonna be in follow up PR shouldn't we not read it in `fromColumnarBatch`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/CRCInfo.java
```diff
@@ -111,15 +131,19 @@ public CRCInfo(
...
     checkArgument(numFiles >= 0);
+    // Live Domain Metadata actions at this version, excluding tombstones.
+    domainMetadata.ifPresent(dms -> dms.forEach(dm -> checkArgument(!dm.isRemoved())));
     this.version = version;
     this.metadata = requireNonNull(metadata);
     this.protocol = requireNonNull(protocol);
     this.tableSizeBytes = tableSizeBytes;
     this.numFiles = numFiles;
     this.txnId = requireNonNull(txnId);
+    this.domainMetadata = requireNonNull(domainMetadata);
```
allisonport-db, line 146: nit - probably requireNonNull before first reference above

kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/CRCInfo.java
allisonport-db: Nvm, I see that we are updating the tests for CheckSumWriter. But doesn't this PR contain all the changes to read DM from CRC? We just don't write it with Kernel yet so it is always empty?

allisonport-db (approved):
Minor comment LGTM overall

But can you clarify the description a bit I was confused until I read through all the code. We are adding read/write to CrcInfo including writing with CheckSumWriter, we just never populate the domain metadatas yet in the txn so it isn't E2E integrated.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/checksum/CRCInfo.java
allisonport-db: I actually also think we don't need this in the PR? But if we do can you update description that we are also adding read support?

huan233usc: Done, Replace this with a TODO.

vkorukanti (approved):
LGTM, please address the comments before merge!

raveeram-db (commented):
LGTM!

PR #4410, opened 2025-04-14
[Kernel]Persisit Domain Metadata in CRC

huan233usc:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## How was this patch tested?
Updates 3 integration tests about CRC/Domain metadata

No


kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -200,17 +201,17 @@ public void removeDomainMetadata(String domain) {
    * @return A list of {@link DomainMetadata} containing domain metadata to be committed in this
    *     transaction.
    */
-  public List<DomainMetadata> getDomainMetadatas() {
+  public List<DomainMetadata> getDomainMetadatasToCommit() {
```
scottsand-db, line 204: I am confused by the existence of both a member variable `Optional<List<DomainMetadata>> domainMetadatasToCommit` and a getter method `List<DomainMetadata> getDomainMetadatasToCommit()`

when should code in this class use one or another? I think this is indicative that our semantics are not clear here

scottsand-db: Fundamentally, the isssue is that (1) `List<DomainMetadata> getDomainMetadatasToCommit()` should really be called `computeAndGetDomainMetadatasToCommit` and (2) we have no way to enforce that users in the class call `computeAndGet` before they access the member variable.

There's two options I see.

## Option 1: Better naming and documentation
- rename `domainMetadatasToCommit` to `computedDomainMetadatasToCommit` and add javadocs to the member variable declaration that None means it has NOT been computed yet
- rename `getDomainMetadatasToCommit` to `computeAndGetDomainMetadatasToCommit`


## Option 2: Encapsulate this information
- Create an inner/static/nester class `DomainMetadataState` that stores the caching and exposes the computed state --> that way, there is no way to peek into the internal state.

I prefer Option 2

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -245,10 +246,40 @@ public List<DomainMetadata> getDomainMetadatas() {
...
     }
-    domainMetadatas = Optional.of(finalDomainMetadatas);
+    domainMetadatasToCommit = Optional.of(finalDomainMetadatas);
     return finalDomainMetadatas;
   }
 
+  /**
+   * Returns the set of active domain metadata of the table, removed domain metadata are excluded.
+   */
+  private Optional<Set<DomainMetadata>> getActiveDomainMetadatas() {
```
scottsand-db, line 256: IMO this Transaction class is getting too big and it is violating the single responsibility principle.

For example, when this says "active" --> does that mean active at the time before the transaction has started? active after the transaction has completed? I know it means the latter -- but it is fair to interpret `Transaction:: getActiveDomainMetadatas` as the former.

Can you think of a way to separate this out to another class? Perhaps most of the post-transaction stuff?

scottsand-db: We can punt that to another PR -- but at the very least here
(1) rename to `getPostCommitDomainMetadatas`
(2) move this private helper method to the bottom of the class where the private members are

huan233usc: Renamed and moved the helper

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
scottsand-db: I do advocate for doing Option 2 now as a quick and small prefactor PR -- We should make the world (code) a better place as we develop it. IMO, followups often don't get done or don't get prioritized as we want them to be.

Just my two cents

scottsand-db (approved):
LGTM once you rebase on the prefactor PR

raveeram-db (approved):
LGTM

PR #4413, opened 2025-04-15
[SPARK] Block vacuum on catalog owned managed tables

Omar-Saleh:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR blocks running vacuum command on catalog owned managed tables. This is a safe guard since with catalog owned tables the clients running vacuum doesn't have complete ownership of the table and instead should go through their catalog to run the vacuum command.

## How was this patch tested?
UTs. Added a test to ensure that vacuum is block on catalog owned tables while all the other vacuum test suite still ensures existing behavior is unaffected.

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/commands/VacuumCommand.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/commands/VacuumCommand.scala
@@ -41,6 +41,7 @@ import org.apache.spark.broadcast.Broadcast
 import org.apache.spark.internal.MDC
 import org.apache.spark.paths.SparkPath
 import org.apache.spark.sql.{Column, DataFrame, Dataset, Encoder, SparkSession}
+import org.apache.spark.sql.catalyst.catalog.CatalogTableType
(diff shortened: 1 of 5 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/commands/VacuumCommand.scala
```diff
@@ -244,6 +245,11 @@ object VacuumCommand extends VacuumCommandImpl with Serializable {
       val snapshot = table.update()
       deltaLog.protocolWrite(snapshot.protocol)
 
+      if (table.catalogTable.exists(_.tableType == CatalogTableType.MANAGED)
```
johanl-db, line 248: Add a comment to explain this: Vacuum can break clones by removing files that the clone still references. For managed tables, we eventually want the catalog to track this dependency to avoid breaking clones, but for now we forbid running vacuum on managed tables.

johanl-db (approved):

johanl-db:
@Omar-Saleh Can you also update the PR description?

PR #4415, opened 2025-04-15
[Spark] Issue with MERGE/UPDATE on target with CHAR column, generated column and readSideCharPadding

juliuszsompolski:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Currently, GeneratedColumn is very strict in MERGE and UPDATE, not accepting anything else than a pure LogicalRelation as the target plan.
This is because, as generated column is defined as a piece of SQL text, it would have issues resolving against a view that does anything more tricky, which we (unfortunately and originally by accident) support, like e.g. if the view was select a as b, b as a from deltaTable, it would confuse the generated column which input columns to use.

Unfortunately, when CHAR columns are present on the table, ApplyCharTypePadding adds a projection with padding, which then Generated Columns does not accept. Refine the check, so that the char padding is allowed. Similar handling has been added in https://github.com/delta-io/delta/pull/1852/files#diff-e68090b34015f52fa61e949ad9d0b3fa527c7781a606ca8260796b441c996de8R374 in the past.

## How was this patch tested?

Test added.

No.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/GeneratedColumn.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/GeneratedColumn.scala
@@ -603,7 +603,7 @@ object GeneratedColumn extends DeltaLogging with AnalysisHelper {
    * not handle the aliasing).
    *
    * This function checks if the target plan is a bare reference to the Delta table, or if the
-   * transofrmations are the result of internal processing introduced not by the user, but
+   * transformations are the result of internal processing introduced not by the user, but
    * internally during analysis, which need to be taken into account and allowed.
    *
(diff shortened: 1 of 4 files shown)


juliuszsompolski:
@johanl-db

johanl-db (approved):
Discussed with @juliuszsompolski: there's no particularly good alternative here, we could check for generated columns when stripping the view in `DeltaViewHelper`, that comes with its own set of issues - e.p. it's very hard to ensure we don't break behavior that works today.

This fix is safe at least since it only impacts cases that would already fail today

PR #4425, opened 2025-04-16
[Kernel] Support enabling row tracking in Kernel

qiyuandong-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This PR addresses the missing [writer requirements](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#writer-requirements-for-row-tracking) to enable row tracking in Kernel. Specifically, on top of the existing implementation, it adds the assignment of materialized row tracking column names at table creation.

With this change, users will be able to create a new table with row tracking enabled. Note that enabling/disabling row tracking on existing tables is still blocked, since enabling requires backfilling row IDs/commit versions, which is not supported, and disabling is irreversible.

## How was this patch tested?

Existing and new tests in `RowTrackingSuite`.

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
@@ -15,8 +15,7 @@
  */
 package io.delta.kernel.internal;
 
-import static io.delta.kernel.internal.DeltaErrors.requiresSchemaForNewTable;
-import static io.delta.kernel.internal.DeltaErrors.tableAlreadyExists;
+import static io.delta.kernel.internal.DeltaErrors.*;
 import static io.delta.kernel.internal.TransactionImpl.DEFAULT_READ_VERSION;
 import static io.delta.kernel.internal.TransactionImpl.DEFAULT_WRITE_VERSION;
 import static io.delta.kernel.internal.util.ColumnMapping.isColumnMappingModeEnabled;
@@ -44,6 +43,8 @@ import io.delta.kernel.internal.icebergcompat.IcebergWriterCompatV1MetadataValid
 import io.delta.kernel.internal.metrics.SnapshotMetrics;
 import io.delta.kernel.internal.metrics.SnapshotQueryContext;
 import io.delta.kernel.internal.replay.LogReplay;
+import io.delta.kernel.internal.rowtracking.MaterializedRowTrackingColumn;
+import io.delta.kernel.internal.rowtracking.RowTracking;
 import io.delta.kernel.internal.snapshot.LogSegment;
 import io.delta.kernel.internal.snapshot.SnapshotHint;
 import io.delta.kernel.internal.tablefeatures.TableFeature;
@@ -485,7 +486,15 @@ public class TransactionBuilderImpl implements TransactionBuilder {
         initialClusteringColumns.map(
             cols -> SchemaUtils.casePreservingEligibleClusterColumns(updatedSchema, cols));
 
-    /* ----- 6: Validate the metadata change ----- */
+    /* ----- 6: Update the METADATA with materialized row tracking column name if applicable----- */
+    Optional<Metadata> rowTrackingMetadata =
+        MaterializedRowTrackingColumn.assignMaterializedColumnNamesIfNeeded(
+            newMetadata.orElse(baseMetadata));
+    if (rowTrackingMetadata.isPresent()) {
+      newMetadata = rowTrackingMetadata;
+    }
+
+    /* ----- 7: Validate the metadata change ----- */
     // Now that all the config and schema changes have been made validate the old vs new metadata
     if (newMetadata.isPresent()) {
       // Use physicalClusteringColumns if clustering column is set in this txn,
@@ -597,6 +606,8 @@ public class TransactionBuilderImpl implements TransactionBuilder {
    *         <li>the schema change is a valid schema change given the tables partition and
    *             clustering columns
    *       </ul>
+   *   <li>Enabling/disabling row tracking on existing tables is blocked
+   *   <li>Materialized row tracking column names do not conflict with schema
    * </ul>
    */
   private void validateMetadataChange(
@@ -674,6 +685,15 @@ public class TransactionBuilderImpl implements TransactionBuilder {
             true /* allowNewRequiredFields */);
       }
     }
+
+    // Block enabling/disabling row tracking on existing tables because:
+    // 1. Enabling requires backfilling row IDs/commit versions, which is not supported in Kernel
+    // 2. Disabling is irreversible in Kernel (re-enabling not supported)
+    if (!isCreateOrReplace) {
+      RowTracking.throwIfRowTrackingToggled(oldMetadata, newMetadata);
+    }
+
+    MaterializedRowTrackingColumn.throwIfColumnNamesConflictWithSchema(newMetadata);
   }
 
   private SnapshotImpl getInitialEmptySnapshot(
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
@@ -1,0 +1,121 @@
+/*
+ * Copyright (2025) The Delta Lake Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.delta.kernel.internal.rowtracking;
+
+import io.delta.kernel.internal.DeltaErrors;
+import io.delta.kernel.internal.TableConfig;
+import io.delta.kernel.internal.actions.Metadata;
+import io.delta.kernel.internal.util.ColumnMapping;
+import io.delta.kernel.types.StructField;
+import io.delta.kernel.types.StructType;
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/** A collection of helper methods for working with materialized row tracking columns. */
+public final class MaterializedRowTrackingColumn {
+
+  /** Static instance for the materialized row ID column. */
+  public static final MaterializedRowTrackingColumn ROW_ID =
+      new MaterializedRowTrackingColumn(
+          "delta.rowTracking.materializedRowIdColumnName", "_row-id-col-");
+
+  /** Static instance for the materialized row commit version column. */
+  public static final MaterializedRowTrackingColumn ROW_COMMIT_VERSION =
+      new MaterializedRowTrackingColumn(
+          "delta.rowTracking.materializedRowCommitVersionColumnName", "_row-commit-version-col-");
+
+  private final String materializedColumnNameProperty;
+  private final String materializedColumnNamePrefix;
+
+  private MaterializedRowTrackingColumn(String property, String prefix) {
+    this.materializedColumnNameProperty = property;
+    this.materializedColumnNamePrefix = prefix;
+  }
+
+  /** Returns the configuration property name associated with this materialized column. */
+  public String getMaterializedColumnNameProperty() {
+    return materializedColumnNameProperty;
+  }
+
+  /** Returns the prefix to use for generating the materialized column name. */
+  public String getMaterializedColumnNamePrefix() {
+    return materializedColumnNamePrefix;
+  }
+
+  /**
+   * Validates that the materialized column names for ROW_ID and ROW_COMMIT_VERSION do not conflict
+   * with any existing logical or physical column names in the table's schema.
+   *
+   * @param metadata The current {@link Metadata} of the table.
+   */
+  public static void throwIfColumnNamesConflictWithSchema(Metadata metadata) {
+    StructType schema = metadata.getSchema();
+    Set<String> logicalColNames =
+        schema.fields().stream().map(StructField::getName).collect(Collectors.toSet());
+    Set<String> physicalColNames =
+        schema.fields().stream().map(ColumnMapping::getPhysicalName).collect(Collectors.toSet());
+
+    Stream.of(ROW_ID, ROW_COMMIT_VERSION)
+        .map(column -> metadata.getConfiguration().get(column.getMaterializedColumnNameProperty()))
+        .filter(Objects::nonNull)
+        .forEach(
+            columnName -> {
+              if (logicalColNames.contains(columnName) || physicalColNames.contains(columnName)) {
+                throw DeltaErrors.conflictWithReservedInternalColumnName(columnName);
+              }
+            });
+  }
+
+  /**
+   * Assigns materialized column names for ROW_ID and ROW_COMMIT_VERSION if row tracking is enabled
+   * and the column names have not been assigned yet.
+   *
+   * @param metadata The current Metadata of the table.
+   * @return An Optional containing updated metadata if any assignments occurred; Optional.empty()
+   *     otherwise.
+   */
+  public static Optional<Metadata> assignMaterializedColumnNamesIfNeeded(Metadata metadata) {
+    // No assignment if row tracking is disabled
+    if (!TableConfig.ROW_TRACKING_ENABLED.fromMetadata(metadata)) {
+      return Optional.empty();
+    }
+
+    Map<String, String> configsToAdd = new HashMap<>();
+
+    Stream.of(ROW_ID, ROW_COMMIT_VERSION)
+        .filter(
+            column ->
+                !metadata
+                    .getConfiguration()
+                    .containsKey(column.getMaterializedColumnNameProperty()))
+        .forEach(
+            column -> {
+              configsToAdd.put(
+                  column.getMaterializedColumnNameProperty(),
+                  column.generateMaterializedColumnName());
+            });
+
+    return configsToAdd.isEmpty()
+        ? Optional.empty()
+        : Optional.of(metadata.withMergedConfiguration(configsToAdd));
+  }
+
+  /** Generates a random name by concatenating the prefix with a random UUID. */
+  private String generateMaterializedColumnName() {
+    return materializedColumnNamePrefix + UUID.randomUUID().toString();
+  }
+}
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
@@ -31,24 +31,25 @@ public final class MaterializedRowTrackingColumn {
   /** Static instance for the materialized row ID column. */
   public static final MaterializedRowTrackingColumn ROW_ID =
       new MaterializedRowTrackingColumn(
-          "delta.rowTracking.materializedRowIdColumnName", "_row-id-col-");
+          TableConfig.MATERIALIZED_ROW_ID_COLUMN_NAME, "_row-id-col-");
 
   /** Static instance for the materialized row commit version column. */
   public static final MaterializedRowTrackingColumn ROW_COMMIT_VERSION =
       new MaterializedRowTrackingColumn(
-          "delta.rowTracking.materializedRowCommitVersionColumnName", "_row-commit-version-col-");
+          TableConfig.MATERIALIZED_ROW_COMMIT_VERSION_COLUMN_NAME, "_row-commit-version-col-");
 
-  private final String materializedColumnNameProperty;
+  private final TableConfig<String> tableConfig;
   private final String materializedColumnNamePrefix;
 
-  private MaterializedRowTrackingColumn(String property, String prefix) {
-    this.materializedColumnNameProperty = property;
+  /** Private constructor to enforce the use of static instances. */
+  private MaterializedRowTrackingColumn(TableConfig<String> tableConfig, String prefix) {
+    this.tableConfig = tableConfig;
     this.materializedColumnNamePrefix = prefix;
   }
 
   /** Returns the configuration property name associated with this materialized column. */
   public String getMaterializedColumnNameProperty() {
-    return materializedColumnNameProperty;
+    return tableConfig.getKey();
   }
 
   /** Returns the prefix to use for generating the materialized column name. */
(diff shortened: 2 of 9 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -0,0 +1,117 @@
...
+   *
+   * @param metadata The current Metadata of the table.
+   * @param isNewTable Flag indicating if the table is new.
+   * @return An Optional containing updated metadata if the column name was assigned;
+   *     Optional.empty() otherwise.
+   */
+  public Optional<Metadata> assignMaterializedColumnNameIfNeeded(
+      Metadata metadata, boolean isNewTable) {
+    Boolean isRowTrackingEnabled = TableConfig.ROW_TRACKING_ENABLED.fromMetadata(metadata);
+    if (!isRowTrackingEnabled || !isNewTable) {
```
johanl-db, line 80: If the table isn't new and has row tracking enabled, then either:
- row tracking was already enabled and there should already be column names assigned, making this method a no-op
- row tracking wasn't enabled and is being enabled as part of this commit: Kernel will fail.

I would remove the `isNewTable` check and instead check whether column names are already assigned, which is what we'll want eventually once we support enabling row tracking on existing tables

qiyuandong-db: I’ve updated this to be more future-proof. `assignMaterializedColumnNameIfNeeded` now assigns the column name if row tracking is enabled and no name has been set yet—regardless of whether the table is new.  The check for blocking enabling row tracking on existing tables is handled in `TransactionBuilderImpl`.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -442,7 +442,22 @@ protected Tuple2<Optional<Protocol>, Optional<Metadata>> validateAndUpdateProtoc
...
+    if (rowTrackingMetadata.isPresent()) {
+      newMetadata = rowTrackingMetadata;
+    }
+
+    rowTrackingMetadata =
+        MaterializedRowTrackingColumn.ROW_COMMIT_VERSION.assignMaterializedColumnNameIfNeeded(
+            newMetadata.orElse(baseMetadata));
+    if (rowTrackingMetadata.isPresent()) {
+      newMetadata = rowTrackingMetadata;
+    }
```
vkorukanti, line 458: can we combine these into one method? assignMaterializedColumnNamesIfNeeded?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -583,6 +600,23 @@ private void validateMetadataChange(
...
+    // 1. Enabling requires backfilling row IDs/commit versions, which is not supported in Kernel
+    // 2. Disabling is irreversible in Kernel (re-enabling not supported)
+    if (!isCreateOrReplace) {
+      boolean oldRowTrackingEnabledValue =
+          TableConfig.ROW_TRACKING_ENABLED.fromMetadata(oldMetadata);
+      boolean newRowTrackingEnabledValue =
+          TableConfig.ROW_TRACKING_ENABLED.fromMetadata(newMetadata);
+      if (oldRowTrackingEnabledValue != newRowTrackingEnabledValue) {
+        throw DeltaErrors.cannotToggleRowTrackingOnExistingTable();
+      }
```
vkorukanti, line 614: can combine into one method, may be in RowTrackingUtils (similarly other methods).

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -583,6 +600,23 @@ private void validateMetadataChange(
...
+    }
+
+    MaterializedRowTrackingColumn.ROW_ID.throwIfColumnNameConflictsWithSchema(newMetadata);
+    MaterializedRowTrackingColumn.ROW_COMMIT_VERSION.throwIfColumnNameConflictsWithSchema(
+        newMetadata);
```
vkorukanti, line 619: are these checks done when the table is created for the first time?

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/tablefeatures/TableFeaturesSuite.scala
```diff
@@ -130,8 +130,7 @@ class TableFeaturesSuite extends AnyFunSuite {
     ("typeWidening", testMetadata(tblProps = Map("delta.enableTypeWidening" -> "true")), true),
     ("typeWidening", testMetadata(tblProps = Map("delta.enableTypeWidening" -> "false")), false),
     // Disable this until we have support to enable row tracking through metadata
-    // ("rowTracking", testMetadata(tblProps = Map("delta.enableRowTracking" -> "true")), true),
-    ("rowTracking", testMetadata(tblProps = Map("delta.enableRowTracking" -> "false")), false),
```
vkorukanti, line 134: why is this test line deleted?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/IcebergWriterCompatV1Suite.scala
```diff
@@ -474,44 +474,6 @@ class IcebergWriterCompatV1Suite extends DeltaTableWriteSuiteBase with ColumnMap
     // We throw an error earlier for variant for some reason
     expectedErrorMessage = "Kernel doesn't support writing data of type: variant")
 
-  // For some reason rowTracking throws an UnsupportedOperationException (due to partial support?)
```
vkorukanti, line 477: cc @allisonport-db

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/RowTrackingSuite.scala
```diff
@@ -46,29 +46,14 @@ class RowTrackingSuite extends DeltaTableWriteSuiteBase with ParquetSuiteBase {
...
-
-  private def createTableWithRowTrackingSupported(
-      engine: Engine,
-      tablePath: String,
-      schema: StructType = testSchema): Unit = {
-    createTxn(engine, tablePath, isNewTable = true, schema, Seq.empty)
+      schema: StructType = testSchema,
+      extraProps: Map[String, String] = Map.empty): Unit = {
+    val tableProps = Map(TableConfig.ROW_TRACKING_ENABLED.getKey -> "true") ++ extraProps
+    createTxn(engine, tablePath, isNewTable = true, schema, tableProperties = tableProps)
```
vkorukanti, line 55: use createEmptyTable

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/RowTrackingSuite.scala
```diff
@@ -468,47 +453,133 @@ class RowTrackingSuite extends DeltaTableWriteSuiteBase with ParquetSuiteBase {
...
-      // Create a table without row tracking
-      createTxn(engine, tablePath, isNewTable = true, testSchema, Seq.empty)
-        .commit(engine, emptyIterable())
+  private val ROW_TRACKING_ENABLED_PROP = Map(TableConfig.ROW_TRACKING_ENABLED.getKey -> "true")
+  private val ROW_TRACKING_DISABLED_PROP = Map(TableConfig.ROW_TRACKING_ENABLED.getKey -> "false")
 
-      // Create a txn but don't commit it yet
-      val dataBatch1 = generateData(testSchema, Seq.empty, Map.empty, 100, 1)
-      val txn1 = createTxn(engine, tablePath, isNewTable = false, testSchema, Seq.empty)
+  private def assertRowTrackingEnabledProp(
```
vkorukanti, line 459: nit: move these private methods together with other private methods.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/RowTrackingSuite.scala
```diff
@@ -468,47 +453,133 @@ class RowTrackingSuite extends DeltaTableWriteSuiteBase with ParquetSuiteBase {
...
+    val table = TableImpl.forPath(engine, tablePath)
+    val snapshot = table.getLatestSnapshot(engine).asInstanceOf[SnapshotImpl]
+    assertMetadataProp(snapshot, TableConfig.ROW_TRACKING_ENABLED, expectedValue)
+  }
 
-      // A concurrent txn makes row tracking supported
-      setWriterFeatureSupported(engine, tablePath, testSchema, Seq("rowTracking", "domainMetadata"))
+  test("row tracking can be enabled/disabled on new table") {
+    withTempDirAndEngine { (tablePath, engine) =>
+      createTxn(
```
vkorukanti, line 470: use createEmptyTable

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/RowTrackingSuite.scala
vkorukanti, line 459: there may be already a utility for this in the DeltaTableWriteSuiteBase.

vkorukanti (changes requested):
Looks great! thanks for adding this feature. One qn: do we want to add the support for reading the materialized columns or is that only necessary when Kernel has the DML support?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/IcebergWriterCompatV1Suite.scala
```diff
@@ -538,7 +500,7 @@ class IcebergWriterCompatV1Suite extends DeltaTableWriteSuiteBase with ColumnMap
 
   testIncompatibleTableFeature(
     "rowTracking inactive",
-    tablePropertiesToEnable = Map("delta.feature.rowTracking" -> "supported"),
```
allisonport-db, line 541: Why did you make this change? The point of this test is to keep it as is, we want to check when the feature is supported (in the protocol) but not active (not enabled via the config key you added here)

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/IcebergWriterCompatV1Suite.scala
allisonport-db: All of these tests should not be deleted -- they just need to be changed to catch different exceptions.

I think this should do it
```
  testIncompatibleTableFeature(
    "rowTracking",
    tablePropertiesToEnable = Map("delta.enableRowTracking" -> "true"))
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
scottsand-db: +1

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -0,0 +1,110 @@
...
+    this.materializedColumnNamePrefix = prefix;
+  }
+
+  /** Static instance for the materialized row ID column. */
+  public static final MaterializedRowTrackingColumn ROW_ID =
+      new MaterializedRowTrackingColumn(
+          "delta.rowTracking.materializedRowIdColumnName", "_row-id-col-");
+
+  /** Static instance for the materialized row commit version column. */
+  public static final MaterializedRowTrackingColumn ROW_COMMIT_VERSION =
```
scottsand-db, line 48: Final Static variables come first at the top of the class

scottsand-db (changes requested):
Wow, looks great! Only left one minor comment! And +1 to the other feedback left on the PR

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/RowTrackingSuite.scala
qiyuandong-db: I removed this and just use the `assertMetadataProp` from `DeltaTableWriteSuiteBase` when we want to assert `TableConfig.ROW_TRACKING_ENABLED`. It is already pretty handy.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
qiyuandong-db: Combined them into `MaterializedRowTrackingColumn.assignMaterializedColumnNamesIfNeeded()`. Thanks!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
qiyuandong-db: Moved this to `RowTracking.throwIfRowTrackingToggled()`.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/tablefeatures/TableFeaturesSuite.scala
qiyuandong-db: That was a mistake. I’ve added the test line back. Thanks for catching it!

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/IcebergWriterCompatV1Suite.scala
qiyuandong-db: I misunderstood the purpose of this test. Thanks for clarifying! I’ve reverted the change.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/IcebergWriterCompatV1Suite.scala
qiyuandong-db: Makes sense! I've added back a `testIncompatibleTableFeature` for `rowTracking`. Thanks for the clarification!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
qiyuandong-db: Not necessarily. If there are no updates to the initial metadata from `getInitialMetadata()`, these checks will be skipped.

That shouldn't be a problem since the row tracking column names are generated during the same commit that creates the table. They include UUIDs, so users wouldn’t know or be able to add these names to the schema at table creation.

qiyuandong-db:
Thanks for the reviews!

Reply to @vkorukanti 's question
> One qn: do we want to add the support for reading the materialized columns or is that only necessary when Kernel has the DML support?

We don’t need to support reading materialized row tracking columns in Kernel at this time. That will only be necessary once Kernel adds DML support. The [readers requirements](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#reader-requirements-for-row-tracking) to access row ids / row commit versions are implicitly met, because we simply can't access these fields in Kernel.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/tablefeatures/TableFeaturesSuite.scala
```diff
@@ -130,7 +130,7 @@ class TableFeaturesSuite extends AnyFunSuite {
     ("typeWidening", testMetadata(tblProps = Map("delta.enableTypeWidening" -> "true")), true),
     ("typeWidening", testMetadata(tblProps = Map("delta.enableTypeWidening" -> "false")), false),
     // Disable this until we have support to enable row tracking through metadata
```
vkorukanti, line 132:
```suggestion
```

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/IcebergWriterCompatV1Suite.scala
```diff
@@ -474,43 +474,11 @@ class IcebergWriterCompatV1Suite extends DeltaTableWriteSuiteBase with ColumnMap
...
-        updateTableMetadata(
-          engine,
-          tablePath,
-          tableProperties = Map("delta.enableRowTracking" -> "true"))
-      }
-      assert(e.getMessage.contains("Feature `rowTracking` is not yet supported in Kernel"))
-    }
-  }
-
-  test(s"Cannot enable feature rowTracking and icebergWriterCompatV1 on a new table") {
```
vkorukanti, line 500: we may want to keep this test: where tblProp enabled both row tracking and icebergWriterCompatV1

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -0,0 +1,121 @@
...
+    return materializedColumnNameProperty;
+  }
+
+  /** Returns the prefix to use for generating the materialized column name. */
+  public String getMaterializedColumnNamePrefix() {
+    return materializedColumnNamePrefix;
+  }
+
+  /** Generates a random name by concatenating the prefix with a random UUID. */
+  private String generateMaterializedColumnName() {
```
vkorukanti, line 60: nit: move these private methods below the public ones.

vkorukanti (approved):
lgtm, minor comments, once addressed we can merge this.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -442,7 +443,15 @@ protected Tuple2<Optional<Protocol>, Optional<Metadata>> validateAndUpdateProtoc
       newMetadata = columnMappingMetadata;
     }
 
-    /* ----- 5: Validate the metadata change ----- */
+    /* ----- 5: Update the METADATA with materialized row tracking column name if applicable----- */
```
allisonport-db, line 446: Does this only happen on table creation?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -0,0 +1,121 @@
...
+
+  /**
+   * Assigns materialized column names for ROW_ID and ROW_COMMIT_VERSION if row tracking is enabled
+   * and the column names have not been assigned yet.
+   *
+   * @param metadata The current Metadata of the table.
+   * @return An Optional containing updated metadata if any assignments occurred; Optional.empty()
+   *     otherwise.
+   */
+  public static Optional<Metadata> assignMaterializedColumnNamesIfNeeded(Metadata metadata) {
```
allisonport-db, line 96: Should we validate somewhere that these are not changed? (is this allowed?)

allisonport-db: Also wondering if the keys should be in `TableConfig`? Then we are also able to make them "non-editable" by the user (our other checks will throw an exception if the user tries to set this or if it's set to an invalid value such as an int or something)

allisonport-db: Then instead of `String property` in the constructor it would just be `TableConfig property`

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/RowTrackingSuite.scala
allisonport-db, line 1: Tests look good. Along with my other comments, I have questions about user table properties involving the setting for the col names and whether this should be blocked (I think so?). If so we should test that

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/IcebergWriterCompatV1Suite.scala
qiyuandong-db: I used the helper `testIncompatibleTableFeature()` to replace the tests on both a new table and an existing table. It should test both cases by default. ([code](https://github.com/delta-io/delta/blob/0b8511ad213e8e1b9cb41732a80a33e326dc4490/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/IcebergWriterCompatV1Suite.scala#L350))

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
qiyuandong-db: Adding them to `TableConfig` as non-editable makes sense. I’ve updated the code accordingly and added tests to validate they can’t be modified. Thanks for flagging this!

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/RowTrackingSuite.scala
qiyuandong-db: Changing row tracking column names should be blocked. I've added tests for that. Thanks!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -485,7 +486,15 @@ protected Tuple2<Optional<Protocol>, Optional<Metadata>> validateAndUpdateProtoc
         initialClusteringColumns.map(
             cols -> SchemaUtils.casePreservingEligibleClusterColumns(updatedSchema, cols));
 
-    /* ----- 6: Validate the metadata change ----- */
+    /* ----- 6: Update the METADATA with materialized row tracking column name if applicable----- */
+    Optional<Metadata> rowTrackingMetadata =
+        MaterializedRowTrackingColumn.assignMaterializedColumnNamesIfNeeded(
```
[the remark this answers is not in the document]
allisonport-db: > Just wondering if we should basically check: rowTracking configs are not missing in the metadata when rowTracking already enabled (& then Kernel goes and adds them silently)
> 
> > This is actually what assignMaterializedColumnNamesIfNeeded() does exactly:
> >
> > rowTracking not enabled -> no-op
> > rowTracking enabled -> if missing rowTracking configs, Kernel will add them to the metadata

Yeah I know this is the behavior that would happen, I'm wondering if we should check this so that this behavior doesn't happen silently. It seems undesirable to silently add the missing configs.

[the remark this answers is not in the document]
allisonport-db: Basically -- rowTracking configs should only be missing (& thus added by us) when rowTracking is being enabled in the transaction. For now, this is just during create table (but in the future possibly other times).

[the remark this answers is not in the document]
qiyuandong-db: Thanks for clarifying! That makes sense. We should explicitly check and avoid the silent addition of missing configs. I’ll open a follow-up PR to address this.

PR #4427, opened 2025-04-16
[Kernel][Refactor]Introduce a internal class for grouping all domain metadata members of TransactionImpl class

huan233usc:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This PR implements the refactor idea in https://github.com/delta-io/delta/pull/4410#discussion_r2047439689
No functional change but make dm to commit only accessible from getComputedDomainMetadatasToCommit().

## How was this patch tested?
Existing tests

No

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
@@ -258,12 +258,12 @@ public class TransactionImpl implements Transaction {
 
   @Override
   public void removeDomainMetadata(String domain) {
+    checkState(
+            TableFeatures.isDomainMetadataSupported(protocol),
+            "Unable to add domain metadata when the domain metadata table feature is disabled");
     checkArgument(
         DomainMetadata.isUserControlledDomain(domain),
         "Removing a system-controlled domain is not allowed: " + domain);
-    checkState(
-        TableFeatures.isDomainMetadataSupported(protocol),
-        "Unable to add domain metadata when the domain metadata table feature is disabled");
     domainMetadataState.removeDomain(domain);
   }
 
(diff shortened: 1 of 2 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -84,16 +84,99 @@ public class TransactionImpl implements Transaction {
...
+  private final DomainMetadataState domainMetadataState = new DomainMetadataState();
   private Metadata metadata;
   private boolean shouldUpdateMetadata;
   private int maxRetries;
   private Optional<CRCInfo> currentCrcInfo;
 
   private boolean closed; // To avoid trying to commit the same transaction again.
 
+  /** Encapsulates the state of domain metadata within a transaction. */
+  private class DomainMetadataState {
```
scottsand-db, line 96: Looks great! Please move this to the end of the file (and, yes, we should develop a dev guide where order and best practices like this are stated)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -188,65 +262,20 @@ public void removeDomainMetadata(String domain) {
...
-        // 4. txnA needs to perform conflict resolution against the V1 commit from txnB
-        // Conflict resolution should fail but since the domain does not exist we cannot create
-        // a tombstone to mark it as removed and correctly perform conflict resolution.
-        throw new DomainDoesNotExistException(
-            dataPath.toString(), domainName, readSnapshot.getVersion());
-      }
-    }
-    domainMetadatas = Optional.of(finalDomainMetadatas);
-    return finalDomainMetadatas;
+  public List<DomainMetadata> getComputedDomainMetadatasToCommit() {
```
scottsand-db, line 277: nit: delete this method, and just use `domainMetadataState.getComputedDomainMetadatasToCommit()` directly -- and move this comment to that method? what do you think?

scottsand-db (approved):
Looks great! Left 2 nits (please implement them before merge)

This code is much cleaner and safer now -- thank you!

PR #4432, opened 2025-04-17
[DELTA] Pass catalogTable while constructing deltaLog for DeltaTableV2

Omar-Saleh:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
This PR passes catalogTable while constructing deltaLog in DeltaTableV2, this allows spark to interact with catalog owned tables as an external client.

## How was this patch tested?
E2E integration testing using Delta, Spark and Unity-Catalog

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala
@@ -848,6 +848,13 @@ object DeltaLog extends DeltaLogging {
     apply(spark, rawPath, options = Map.empty, initialCatalogTable, clock)
 
 
+  /** Helper for creating a log for the table. */
+  private[delta] def forTable(spark: SparkSession,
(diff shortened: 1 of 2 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala
```diff
@@ -848,6 +848,13 @@ object DeltaLog extends DeltaLogging {
     apply(spark, rawPath, options = Map.empty, initialCatalogTable, clock)
 
 
+  /** Helper for creating a log for the table. */
+  private[delta] def forTable(spark: SparkSession,
+      dataPath: Path,
```
johanl-db, line 853: nit:
```suggestion
  private[delta] def forTable(
      spark: SparkSession,
      dataPath: Path,
```

PR #4443, opened 2025-04-21
[Delta Kernel] Allow reading of compacted log files

nicklan:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description
**First Draft** of reading compressed log files. This basically just plumbs everything through that allows reading these files and optionally using them in place of the covered json files.

A few things of note:
- Since this feature is backwards compatible (i.e. no table feature) we can _always_ assume that all the delta commit files are present and can verify things in the same way as previously
- This associates all the actions in the compressed file with the starting version of the compression
- We still need a way to configure this. Currently just hard coded to "on"
- (more to come)

## How was this patch tested?

Unit tests and by adding a compaction file to one test table and ensuring the tests still work.

It will


kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
```diff
@@ -321,29 +321,12 @@ private CloseableIterator<ActionWrapper> getNextActionsIter() {
         case COMMIT:
           {
             final long fileVersion = FileNames.deltaVersion(nextFilePath);
-            // We can not read multiple JSON files in parallel (like the checkpoint files),
-            // because each one has a different version, and we need to associate the
-            // version with actions read from the JSON file for further optimizations later
-            // on (faster metadata & protocol loading in subsequent runs by remembering
-            // the version of the last version where the metadata and protocol are found).
-            final CloseableIterator<ColumnarBatch> dataIter =
```
nicklan, line 329: code moved into `readCommitOrCompactionFile` below

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -131,6 +132,8 @@ public static StructType getAddReadSchema(boolean shouldReadStats) {
   private final Lazy<Map<String, DomainMetadata>> domainMetadataMap;
   private final Optional<CRCInfo> currentCrcInfo;
 
+  private boolean read_log_compaction_files = true;
```
nicklan, line 135: this needs to become a config _somewhere_

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
```diff
@@ -176,6 +193,69 @@ public LogSegment(
...
+    this.compactionsReversed =
+        new Lazy<>(
+            () ->
+                compactions.stream()
+                    .sorted(
+                        Comparator.comparing((FileStatus a) -> new Path(a.getPath()).getName())
+                            .reversed())
+                    .collect(Collectors.toList()));
+
+    this.allFilesWithCompactionsReversed =
```
scottsand-db, line 206: A few thoughts

1. Let's move this logic into a new private helper method `loadAllFilesWithCompactionsReversed`.

2. I think we should split this PR into two.
  a. One for gathering the list of compaction files (from the LIST) and
  b. Another for actually selecting the right compaction files for LogReplay. For example, this method alone (part b) could have a ton of unit tests for the various different edge cases

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -240,6 +240,8 @@ private SnapshotImpl createSnapshot(
    */
   @VisibleForTesting
   public LogSegment getLogSegmentForVersion(Engine engine, Optional<Long> versionToLoadOpt) {
+    final boolean USE_COMPACTED_FILES = true; // TODO: Where to config this
```
scottsand-db, line 243: Does delta-spark have a knob for this? I _think_ that we could have some heuristic for when we should or should not use these compaction files? (Not sure).

nicklan:
Closing, was split into multiple PRs

PR #4444, opened 2025-04-21
[Spark] Log if change feed is enabled for Delta Source

fanyue-xia:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

As the title, right now there is no intuitive way of checking if the option is set.

## How was this patch tested?

Unit test

N/A

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSource.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSource.scala
@@ -792,6 +792,11 @@ case class DeltaSource(
 
   logInfo(log"Filters being pushed down: ${MDC(DeltaLogKeys.FILTER, filters)}")
 
+  if (options.readChangeFeed) {
+    logInfo(log"Change feed enabled for delta_log_path" +
+      log"= ${MDC(DeltaLogKeys.PATH, deltaLog.logPath)}")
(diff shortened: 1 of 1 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSource.scala
```diff
@@ -792,6 +792,11 @@ case class DeltaSource(
 
   logInfo(log"Filters being pushed down: ${MDC(DeltaLogKeys.FILTER, filters)}")
 
+  if (options.readChangeFeed) {
+    logInfo(log"Change feed enabled for delta_log_path" +
```
anishshri-db, line 796: Can we move this to the existing logs for `Getting DataFrame` ?

fanyue-xia:
Closed the PR since `Getting CDC DataFrame for` might be enough to infer the config.

spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSource.scala
fanyue-xia: Discussed offline, we log `Getting CDC DataFrame for` if change data feed enabled.

PR #4451, opened 2025-04-23
[Kernel] Initial refactor to TransactionBuilder to support replace table in the future

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Initial refactoring of `TransactionBuilderImpl` to enable future support for replace table. A full prototype can be seen here https://github.com/delta-io/delta/commit/168f295d5fa5d059554c2640a5cf596f8bf6cb54

## How was this patch tested?

Just a refactor, existing tests should suffice.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
@@ -24,6 +24,7 @@ import static io.delta.kernel.internal.util.Preconditions.checkArgument;
 import static io.delta.kernel.internal.util.SchemaUtils.casePreservingPartitionColNames;
 import static io.delta.kernel.internal.util.VectorUtils.buildArrayValue;
 import static io.delta.kernel.internal.util.VectorUtils.stringStringMapValue;
+import static java.util.Collections.emptyList;
 import static java.util.Objects.requireNonNull;
 import static java.util.stream.Collectors.toSet;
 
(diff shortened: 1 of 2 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -188,25 +189,145 @@ public Transaction build(Engine engine) {
...
+      return buildTransactionInternal(engine, true /* isNewTableDef */, Optional.empty());
     }
+  }
+
+  /**
+   * Returns a built {@link Transaction} for this transaction builder (with the input provided by
+   * the user) given the provided parameters. This includes validation and updates as defined in the
+   * builder.
+   *
+   * @param isNewTableDef whether we are defining a new table definition or not. This determines
```
scottsand-db, line 206: `whether we are defining a new table definition or not` -> would a simpler variable name be `isCreateOrReplace`?

Basically, what is missing is the exact list of operations that "create a new table definition".  I _think_ it is only create or replace but I am not sure

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -188,25 +189,145 @@ public Transaction build(Engine engine) {
...
+   *     what metadata to commit in the returned transaction, and what operations to allow or block.
+   * @param existingSnapshot the existing latest snapshot of the table (or empty for a new table).
```
scottsand-db, line 208: would `latestSnapshot` be a better name? the `Optional` data type tells me either it will exist or not, but the `latest` is the context that I'm missing

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -188,25 +189,145 @@ public Transaction build(Engine engine) {
...
+      // Otherwise, use the existing table metadata
+      baseMetadata = existingSnapshot.get().getMetadata();
+    }
+    if (existingSnapshot.isPresent()) {
+      // If the table exists, start with the prior table protocol - this ensures we never downgrade
+      baseProtocol = existingSnapshot.get().getProtocol();
+    } else {
+      // Otherwise, start with initial protocol
+      baseProtocol = getInitialProtocol();
+    }
```
scottsand-db, line 232: These two if statements do basically the same thing but they are checking different values for the `if` -- can we make these check the same value? i.e. `if (existingSnapshot.isPresent()) {` seems good to me for both

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -188,25 +189,145 @@ public Transaction build(Engine engine) {
...
+          newProtocol.orElse(baseProtocol),
+          newMetadata.orElse(baseMetadata),
+          setTxnOpt,
+          casePreservingClusteringColumnsOpt,
+          newMetadata.isPresent() || isNewTableDef /* shouldUpdateMetadata */,
+          newProtocol.isPresent() || isNewTableDef /* shouldUpdateProtocol */,
+          maxRetries,
+          logCompactionInterval,
+          table.getClock());
+
```
scottsand-db, line 291: nit: unecessary empty line

scottsand-db (changes requested):
Looks great! Left some comments and questions

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
huan233usc, line 208: "empty for a new table" what will it be for replace table case? `isNewTableDef` is true and existingSnapshot is present?

vkorukanti (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
allisonport-db: Agree this is cleaner, updating it to be `isCreateOrReplace`. I also cannot think of a different scenario currently and if we come up with one in the future we can always update this internal naming

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
allisonport-db: yeah this makes sense to me updating it

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -411,18 +512,31 @@ private void validateTransactionInputs(Engine engine, SnapshotImpl snapshot, boo
    *   <li>Column mapping mode can only go from none->name for existing table
    *   <li>icebergWriterCompatV1 cannot be enabled on existing tables (only supported upon table
    *       creation)
+   *   <li>Validates the universal format configs are valid.
+   *   <li>If there is schema evolution validates
+   *       <ul>
+   *         <li>column mapping is enabled
```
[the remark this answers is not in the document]
allisonport-db: Honestly not inclined to do this in this PR; I didn't even change the functionality of this method at all the docs were just already out of date so did a best effort to at least improve them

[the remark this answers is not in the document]
allisonport-db: I can add this to the backlog if you'd like

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
allisonport-db: updating the docs

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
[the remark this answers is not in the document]
allisonport-db: Hm actually I don't think this improves much since we also want to use the latestProtocol for regular writes. I'll just try to clarify the comments better

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -188,25 +189,145 @@ public Transaction build(Engine engine) {
...
+    existingSnapshot.ifPresent(snapshot -> validateWriteToExistingTable(engine, snapshot));
+    validateTransactionInputs(engine, isNewTableDef);
+
+    boolean enablesDomainMetadataSupport =
+        needDomainMetadataSupport
+            && !baseProtocol.supportsFeature(TableFeatures.DOMAIN_METADATA_W_FEATURE);
+
+    // If this a new table definition or there is a metadata or protocol update, we need to execute
+    // our protocol & metadata validation/update logic
+    if (isNewTableDef
```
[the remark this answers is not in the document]
allisonport-db: Good idea; Updated it. I think this also makes the flow of logic clearer

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
allisonport-db: We actually have different behavior here; for replace we want to start with initial empty metadata but reuse the protocol. For create we want initial empty for both

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -188,25 +189,145 @@ public Transaction build(Engine engine) {
...
+
+    } else {
+      return new TransactionImpl(
+          false, // isNewTableDef
+          table.getDataPath(),
+          table.getLogPath(),
+          existingSnapshot.get(),
+          engineInfo,
+          operation,
+          existingSnapshot.get().getProtocol(),
```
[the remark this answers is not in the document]
allisonport-db: Hard-coded to use the existing protocol for when there are no updates -- I think this makes it easy to read

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -188,25 +189,145 @@ public Transaction build(Engine engine) {
...
+          existingSnapshot.get().getMetadata(),
+          setTxnOpt,
+          Optional.empty(), /* clustering cols */
+          false /* shouldUpdateMetadata */,
+          false /* shouldUpdateProtocol */,
+          maxRetries,
+          logCompactionInterval,
+          table.getClock());
```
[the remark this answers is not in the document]
allisonport-db: Refactored it to exit early

huan233usc (approved):

allisonport-db:
@scottsand-db gonna merge so I can rebase; if you have any more comments I can address them separately!

PR #4455, opened 2025-04-23
Add V2 table exists

amaliujia:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Existing Delta Catalog when creating table, only check HMS whether the table exists or no. This is not compatible with Spark V2 API.
## How was this patch tested?

N/A

No


spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala
```diff
@@ -520,6 +520,20 @@ class DeltaCatalog extends DelegatingCatalogExtension
 
   /** Checks if a table already exists for the provided identifier. */
   def getExistingTableIfExists(table: TableIdentifier): Option[CatalogTable] = {
+    try {
+      val loadedTable =
+        super.loadTable(Identifier.of(Array(table.database.getOrElse("")), table.table))
```
cloud-fan, line 525: shall we check table existence before loading the table? then we can save the try-catch and it's consistent with `getExistingTableIfExistsFromHMS`

PR #4457, opened 2025-04-24
[Delta Kernel] List compacted log files in SnapshotManager

nicklan:
- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Splitting up #4443 as requested. This part just lists relevant compaction files and stores them in the `LogSegment`. A follow-on PR will do the work to actually use them during log reply.

This PR includes a boolean to turn on/off listing, but it's just hard-coded to true. I'm happy to just remove that all together, but I suspect we'll want to configure it in the future so figured I'd leave that in.

## How was this patch tested?

Unit tests. 

No

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaLogActionUtils.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaLogActionUtils.java
@@ -224,11 +224,14 @@ public class DeltaLogActionUtils {
     return listLogDir(engine, tablePath, startVersion)
         .breakableFilter(
             fs -> {
-              if (fileTypes.contains(DeltaLogFileType.COMMIT)
-                  && FileNames.isCommitFile(getName(fs.getPath()))) {
+              String fileName = getName(fs.getPath());
+              if (fileTypes.contains(DeltaLogFileType.COMMIT) && FileNames.isCommitFile(fileName)) {
+                // Here, we do nothing (we will consume this file).
+              } else if (fileTypes.contains(DeltaLogFileType.LOG_COMPACTION)
+                  && FileNames.isLogCompactionFile(fileName)) {
                 // Here, we do nothing (we will consume this file).
               } else if (fileTypes.contains(DeltaLogFileType.CHECKPOINT)
-                  && FileNames.isCheckpointFile(getName(fs.getPath()))
+                  && FileNames.isCheckpointFile(fileName)
                   && fs.getSize() > 0) {
                 // Checkpoint files of 0 size are invalid but may be ignored silently when read,
                 // hence we ignore them so that we never pick up such checkpoints.
@@ -238,7 +241,12 @@ public class DeltaLogActionUtils {
                 return BreakableFilterResult.EXCLUDE; // Here, we exclude and filter out this file.
               }
 
-              final long fileVersion = FileNames.getFileVersion(new Path(fs.getPath()));
+              final long fileVersion;
+              if (FileNames.isLogCompactionFile(fileName)) {
+                fileVersion = FileNames.logCompactionVersions(new Path(fs.getPath()))._1;
+              } else {
+                fileVersion = FileNames.getFileVersion(new Path(fs.getPath()));
+              }
 
               if (fileVersion < startVersion) {
                 throw new RuntimeException(
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaLogActionUtils.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaLogActionUtils.java
@@ -243,6 +243,10 @@ public class DeltaLogActionUtils {
 
               final long fileVersion;
               if (FileNames.isLogCompactionFile(fileName)) {
+                // We use start version here. Below this is used to determine if we should stop
+                // listing because we've listed past the required version. But with a log compaction
+                // file, if the end version is passed the requested version, we don't want to stop,
+                // we just won't use the compaction file.
                 fileVersion = FileNames.logCompactionVersions(new Path(fs.getPath()))._1;
               } else {
                 fileVersion = FileNames.getFileVersion(new Path(fs.getPath()));
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaLogActionUtils.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaLogActionUtils.java
@@ -33,6 +33,7 @@ import io.delta.kernel.internal.lang.ListUtils;
 import io.delta.kernel.internal.replay.ActionsIterator;
 import io.delta.kernel.internal.util.FileNames;
 import io.delta.kernel.internal.util.FileNames.DeltaLogFileType;
+import io.delta.kernel.internal.util.Tuple2;
 import io.delta.kernel.types.*;
 import io.delta.kernel.utils.CloseableIterator;
 import io.delta.kernel.utils.CloseableIterator.BreakableFilterResult;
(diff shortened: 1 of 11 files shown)


kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
```diff
@@ -702,6 +730,47 @@ class SnapshotManagerSuite extends AnyFunSuite with MockFileSystemClientUtils {
 
     assert(exMsg.contains("Missing checkpoint at version 1"))
   }
+
+  test("One compaction") {
+    testWithCompactionsNoCheckpoint(
+      Seq(0, 1, 2, 3, 4, 5),
```
scottsand-db, line 736: could you please add named params here? it would make these cases a lot more readable

e.g. an example from above, that is very readable

```scala
    testWithSingularAndMultipartCheckpoint(
      deltaVersions = 15L until 25L,
      checkpointVersions = Seq(20),
      startCheckpoint = Optional.of(20),
      versionToLoad = Optional.of(22))
```

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/snapshot/LogSegmentSuite.scala
```diff
@@ -40,74 +40,114 @@ class LogSegmentSuite extends AnyFunSuite with MockFileSystemClientUtils {
...
   test("constructor -- valid case (non-empty)") {
-    new LogSegment(logPath, 12, deltasFs11To12List, checkpointFs10List, 1)
+    new LogSegment(logPath, 12, deltasFs11To12List, Collections.emptyList(), checkpointFs10List, 1)
   }
 
   test("constructor -- null arguments => throw") {
     // logPath is null
     intercept[NullPointerException] {
-      new LogSegment(null, 1, Collections.emptyList(), Collections.emptyList(), -1)
+      new LogSegment(
```
scottsand-db, line 49: do you think it's worth (not in this PR) creating a LogSegmentBuilder so we don't have to keep changing the constructor every time we add a new file type?

I think @huan233usc will be add another one for CRC files ...

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -310,18 +320,31 @@ public LogSegment getLogSegmentForVersion(Engine engine, Optional<Long> versionT
...
-    // Step 5: Partition $listedFileStatuses into the checkpoints and deltas. //
-    ////////////////////////////////////////////////////////////////////////////
-
-    final Tuple2<List<FileStatus>, List<FileStatus>> listedCheckpointAndDeltaFileStatuses =
+    //////////////////////////////////////////////////////////////////////////////////////////
+    // Step 5: Partition $listedFileStatuses into the checkpoints, deltas, and compactions. //
+    //////////////////////////////////////////////////////////////////////////////////////////
+
+    // initially partition in checkpoint | delta or compaction
+    final Tuple2<List<FileStatus>, List<FileStatus>>
```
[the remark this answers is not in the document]
nicklan: Yep, much nicer. I need to learn the java stream apis better :)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaLogActionUtils.java
```diff
@@ -224,11 +224,14 @@ public static CloseableIterator<FileStatus> listDeltaLogFilesAsIter(
     return listLogDir(engine, tablePath, startVersion)
         .breakableFilter(
             fs -> {
-              if (fileTypes.contains(DeltaLogFileType.COMMIT)
-                  && FileNames.isCommitFile(getName(fs.getPath()))) {
+              String fileName = getName(fs.getPath());
```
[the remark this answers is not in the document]
nicklan: Yep, added a bunch of tests

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -369,8 +389,31 @@ public LogSegment getLogSegmentForVersion(Engine engine, Optional<Long> versionT
 
     logDebugFileStatuses("deltasAfterCheckpoint", deltasAfterCheckpoint);
 
+    //////////////////////////////////////////////////////////////////////////////////
+    // Step 8: Grab all compactions in range [$latestCompleteCheckpointVersion + 1, //
+    //         $versionToLoad]                                                      //
+    //////////////////////////////////////////////////////////////////////////////////
+
+    final List<FileStatus> compactionsAfterCheckpoint;
```
scottsand-db, line 397: nit: Can we remove the `if (USE_COMPACTED_FILES) {` ?

If USE_COMPACTED_FILES is false, then above we would have set `listedCompactionFileStatuses` to `Collections.emptyList()`

And then here, `.stream.filter.collect` would just do nothing, right?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/FileNames.java
```diff
@@ -111,6 +131,20 @@ public static long deltaVersion(String path) {
...
+  public static Tuple2<Long, Long> logCompactionVersions(Path path) {
+    final String[] split = path.getName().split("\\.");
+    return new Tuple2<>(Long.parseLong(split[0]), Long.parseLong(split[1]));
+  }
+
+  public static Tuple2<Long, Long> logCompactionVersions(String path) {
+    return logCompactionVersions(new Path(path));
+  }
+
+  // public static boolean deltaIsCoveredBy(String deltaPath, List<String> compactionFiles) {
```
scottsand-db, line 144: should we clean this up?

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/DeltaLogActionUtilsSuite.scala
```diff
@@ -287,4 +287,117 @@ class DeltaLogActionUtilsSuite extends AnyFunSuite with MockFileSystemClientUtil
...
+  testListWithCompactions(
+    "multiple compactions, no endVersion",
+    files = deltaFileStatuses(0L to 7L) ++ compactedFileStatuses(Seq((2, 4), (5, 7))),
+    startVersion = 0,
+    endVersion = Optional.empty(),
+    expectedListedFiles = deltaFileStatuses(0L to 1L) ++ compactedFileStatuses(Seq((
+      2,
+      4))) ++ deltaFileStatuses(2L to 4L) ++ compactedFileStatuses(Seq((
+      5,
+      7))) ++ deltaFileStatuses(5L to 7L))
```
scottsand-db, line 376: Darn is the scala formatter formatting it like this? with the `2, 4` on new lines?

There's no way to make this look like

```scala
 expectedListedFiles = deltaFileStatuses(0L to 1L) ++
   compactedFileStatuses(Seq((2, 4))) ++ 
   deltaFileStatuses(2L to 4L) ++
   compactedFileStatuses(Seq((5, 7))) ++ 
   ...
```

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/snapshot/LogSegmentSuite.scala
scottsand-db: Yeah I think a builder would help a lot with the number of Collections.emptyList(), especially when Xin rebases the CRC PR on top of this ...

And probably not a "test-only" one? (Or, is that a common convention?) Probably just right in `LogSegment.java` but perhaps we could use `@VisibleForTesting` so that our prod code doesn't use it (though, it could?) i.e. it is only there for the test code to use?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/FileNames.java
nicklan: Yep that slipped through a refactor cleanup. Thanks, nice catch

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/DeltaLogActionUtilsSuite.scala
nicklan: Ohh nice. I assumed the formatted had one and only one way that it would want to format things, but yeah, I can clean it up. Thanks!

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/snapshot/LogSegmentSuite.scala
nicklan: Not sure. But you could code a "Builder" that's only in test code, if we didn't actually want to use it in prod. Like why would you put it in prod code if you don't want to use it? (I guess maybe it needs access to internal details so has to be within the class or something, not sure if that would apply here)

scottsand-db (approved):
Looks great!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/DeltaLogFile.java
```diff
@@ -40,6 +41,10 @@ public static DeltaLogFile forCommitOrCheckpoint(FileStatus file) {
     if (FileNames.isCommitFile(fileName)) {
       logType = LogType.COMMIT;
       version = FileNames.deltaVersion(fileName);
+    } else if (FileNames.isLogCompactionFile(fileName)) {
+      logType = LogType.LOG_COMPACTION;
+      // use end version, similar to a checkpoint
```
allisonport-db, line 46: nit - maybe document this either as part of `getVersion` or `LOG_COMPACTION` enum? Just to make it as obvious as possible

allisonport-db: Also should we rename this function name? `forCommitOrCheckpoint`?

allisonport-db (approved):
LGTM. Tests look great!

PR #4458, opened 2025-04-24
[Kernel] Support passing tags for generateIcebergCompatWriterV1AddAction

KaiqiJinWow:

- [ ] Spark
- [ ] Standalone
- [x] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

When the client call generateIcebergCompatWriterV1AddAction, now they can pass the tags so that it would be persist in the AddFile
## How was this patch tested?

Unit tests

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/Transaction.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/Transaction.java
@@ -256,7 +256,8 @@ public interface Transaction {
                   dataFileStatus,
                   ((DataWriteContextImpl) dataWriteContext).getPartitionValues(),
                   true /* dataChange */,
-                  Collections.emptyMap());
+                  // TODO: populate tags in generateAppendActions
+                  Collections.emptyMap() /* tags */);
           return SingleAction.createAddFileSingleAction(addFileRow.toRow());
         });
   }
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/Transaction.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/Transaction.java
@@ -40,6 +40,7 @@ import io.delta.kernel.statistics.DataFileStatistics;
 import io.delta.kernel.types.StructType;
 import io.delta.kernel.utils.*;
 import java.net.URI;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
@@ -254,7 +255,8 @@ public interface Transaction {
                   tableRoot,
                   dataFileStatus,
                   ((DataWriteContextImpl) dataWriteContext).getPartitionValues(),
-                  true /* dataChange */);
+                  true /* dataChange */,
+                  Collections.emptyMap());
           return SingleAction.createAddFileSingleAction(addFileRow.toRow());
         });
   }
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/GenerateIcebergCompatActionUtils.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/GenerateIcebergCompatActionUtils.java
@@ -39,6 +39,7 @@ import io.delta.kernel.types.StructType;
 import io.delta.kernel.utils.CloseableIterable;
 import io.delta.kernel.utils.DataFileStatus;
 import java.net.URI;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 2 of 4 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/GenerateIcebergCompatActionUtils.java
```diff
@@ -64,7 +65,8 @@ public static Row generateIcebergCompatWriterV1AddAction(
       Row transactionState,
       DataFileStatus fileStatus,
       Map<String, Literal> partitionValues,
-      boolean dataChange) {
+      boolean dataChange,
```
raveeram-db, line 68: nit: Could you please add what tags are to the method signature Kaiqi?

kernel/kernel-api/src/main/java/io/delta/kernel/Transaction.java
```diff
@@ -254,7 +255,8 @@ static CloseableIterator<Row> generateAppendActions(
                   tableRoot,
                   dataFileStatus,
                   ((DataWriteContextImpl) dataWriteContext).getPartitionValues(),
-                  true /* dataChange */);
+                  true /* dataChange */,
+                  Collections.emptyMap());
```
huan233usc, line 259: /* tags */ 

Please also add // TODO: populate tags in generateAppendActions.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/GenerateIcebergCompatActionUtils.java
huan233usc: +1, like @param tags xxxx

huan233usc (approved):
LGTM with some nit to be addressed

kernel/kernel-api/src/main/java/io/delta/kernel/Transaction.java
allisonport-db: named params super helpful! but also if we expect this to be a default case (no tags) we could add another method with a different signature with this as the default -- open to ideas

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/GenerateIcebergCompatActionUtils.java
allisonport-db: +1 please update the docs

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/CommitIcebergActionSuite.scala
```diff
@@ -432,4 +433,54 @@ class CommitIcebergActionSuite extends DeltaTableWriteSuiteBase {
...
+          assert(addFile.getTags.isPresent)
+          assert(addFile.getTags.get().getSize
+            == 1)
+          Some(ExpectedAdd(
+            addFile.getPath,
+            addFile.getSize,
+            addFile.getModificationTime,
+            addFile.getDataChange))
+        }
+      }
```
allisonport-db, line 483: I think I don't completely understand what you're trying to do here -- maybe you just copied the code from above?

1) You can probably read just the DeltaAction.ADD
2) Why are you creating an ExpectedAdd? You should probably just collect add rows --> check that there is in fact exactly 1 add row, and then check that the tags are exactly what you expect them to be (we should check that they are present AND not empty for example)

allisonport-db (changes requested):
Mostly looks good. Please fix the test though as we should actually check the tags written.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/CommitIcebergActionSuite.scala
allisonport-db: You can do this by (1) filtering just adds (2) checking length is 1 (3) use head or headOption and checking the tags are equal to the map you committed above

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/GenerateIcebergCompatActionUtils.java
KaiqiJinWow: Done, the docs is updated

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/CommitIcebergActionSuite.scala
KaiqiJinWow: Thanks for the detailed instructions, the test was updated. :)

kernel/kernel-api/src/main/java/io/delta/kernel/Transaction.java
KaiqiJinWow: I think in the future we still need to add this parameter for generateAppendActions, so it makes sense to add an empty value and leave a todo for now.

allisonport-db (approved):
LGTM!

raveeram-db (approved):
LGTM!

PR #4460, opened 2025-04-25
[Spark] Make all MERGE APIs return a DataFrame instead of Unit

larsk-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

- Ensure that all public MERGE APIs return a DataFrame with the affected rows like the SQL API does, so they are all consistent.

## How was this patch tested?

- Added some minimal testing to the connector and Python tests.
- Added full coverage to the Scala suite to ensure the results returned always match what goes into the table history (operation metrics).

Yes, the Scala `DeltaMergeBuilder.execute()` now returns a `DataFrame` instead of `Unit`, and so does the Python equivalent (instead of a `None` value).
This is a **breaking** API change, as it can lead the Scala compiler to infer different surrounding types that expect in some places.

### DIFF
--- a/project/SparkMimaExcludes.scala
+++ b/project/SparkMimaExcludes.scala
@@ -88,6 +88,7 @@ object SparkMimaExcludes {
 
       // Changes in 4.0.0
       ProblemFilters.exclude[IncompatibleResultTypeProblem]("io.delta.tables.DeltaTable.improveUnsupportedOpError"),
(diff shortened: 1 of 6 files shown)


bart-samwel (approved):
Nice improvement!

PR #4462, opened 2025-04-25
[kernel]Support schema evolution with updating cluster columns in same transaction

KaiqiJinWow:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

If we update the clustering columns and table schema at the same time, current logic is only checking the if the schema valid with existing clustering column instead of the newly passed clustering columns.

In this PR, we fixed this issue by getting the physical clustering column names and pass it when doing schema validation. 
## How was this patch tested?

Existing and newly added unit tests.


kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -66,14 +66,15 @@ public class TransactionBuilderImpl implements TransactionBuilder {
...
-  private Optional<List<Column>> clusteringColumns = Optional.empty();
+  private Optional<List<Column>> initialClusteringColumns = Optional.empty();
   private Optional<SetTransaction> setTxnOpt = Optional.empty();
   private Optional<Map<String, String>> tableProperties = Optional.empty();
   private Optional<Set<String>> unsetTablePropertiesKeys = Optional.empty();
   private boolean needDomainMetadataSupport = false;
 
   protected final TableImpl table;
   protected Optional<StructType> schema = Optional.empty();
+  private Optional<List<Column>> physicalClusteringColumns = Optional.empty();
```
allisonport-db, line 77: nit - put this next to the other private scoped variables

allisonport-db: also can you add a comment clarifying when this should/shouldn't be set? and it's relationship to `initialClusteringColumns`, how it's used, etc. It's a bit confusing how this is used and relates to the other variable so let's be sure to document it clearly

allisonport-db (commented):
Looks good thanks for doing this! Just 1 request to document the two similar variables more clearly thank you

allisonport-db (approved):
LGTM!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -296,7 +305,7 @@ protected TransactionImpl buildTransactionInternal(
 
     // We use the existing clustering columns to validate schema evolution
     Optional<List<Column>> existingClusteringCols =
```
scottsand-db, line 307: @KaiqiJinWow can you help explain this a bit for me?

so we have `initialClusteringColumns` and `existingClusteringColumns`

I think `initialClusteringColumns` really means `newClusteringColumnsInThisTxn`

but why is it that: if this is an UPDATE (not create, not replace) and `newClusteringColumnsInThisTxn`, then we set `existingClusteringCols` to EMPTY?

wouldn't the `existingClusteringCOls` during an UPDATE be `ClusteringUtils.getClusteringColumnsOptional(latestSnapshot.get());`

scottsand-db: cc @vkorukanti , @allisonport-db

KaiqiJinWow: Hi the `existingClusteringCols` is mainly used for schema check when doing UPDATE (we cannot drop a column which is a clustering column) please check https://github.com/delta-io/delta/pull/4462/files/0550aa0b6eb1635ce19006f2091a27a372140ccc#diff-5abbcf2158a0779bd57ce93b0041f6feb643c3a0bcba6d211dc0b4c062a67619R493-R496 to see what is the final clustering cols we used. 

Essentially, `initialClusteringColumns` refers to the columns passed in this transaction, but it’s an optional field.

So here is that if a new clustering column is passed, we only need to use that for the schema check. There’s no need to retrieve the existing clustering columns, as they will be overridden. In this case, we can save the time for getting the existing clustering columns

PR #4463, opened 2025-04-25
Close SparkResult to avoid memory leak

cloud-fan:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

The `SparkResult` needs to be closed after it's consumed.

## How was this patch tested?

detected by `DeltaTableSuite` with https://github.com/apache/spark/commit/510edd763b912a21883e8a0dbaa3f661fedec234 

no

### DIFF
--- a/spark-connect/client/src/main/scala-spark-master/io/delta/connect/tables/DeltaTable.scala
+++ b/spark-connect/client/src/main/scala-spark-master/io/delta/connect/tables/DeltaTable.scala
@@ -401,7 +401,12 @@ class DeltaTable private[tables](
     val extension = com.google.protobuf.Any.pack(relation)
     val sparkRelation = spark_proto.Relation.newBuilder().setExtension(extension).build()
     val result = sparkSession.newDataFrame(_.mergeFrom(sparkRelation)).collectResult()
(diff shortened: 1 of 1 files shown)


spark-connect/client/src/main/scala-spark-master/io/delta/connect/tables/DeltaTable.scala
```diff
@@ -401,7 +401,12 @@ class DeltaTable private[tables](
     val extension = com.google.protobuf.Any.pack(relation)
     val sparkRelation = spark_proto.Relation.newBuilder().setExtension(extension).build()
     val result = sparkSession.newDataFrame(_.mergeFrom(sparkRelation)).collectResult()
-    sparkSession.createDataFrame(result.toArray.toSeq.asJava, result.schema)
+    val data = try {
```
hvanhovell, line 404: If we are going to do this, then you might as well use collect() instead of collectResult(). This is BTW not the first time I am seeing this

hvanhovell: Nvm you need the schema.

tdas (approved):

PR #4476, opened 2025-04-29
[Kernel] Fix struct::get to not throw a null pointer exception

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Currently StructType::get throws a nullpointerexception if the field doesn't exist. We should at least throw a better exception.

## How was this patch tested?

Adds a test

No

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/types/StructType.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/types/StructType.java
@@ -15,6 +15,8 @@
  */
 package io.delta.kernel.types;
 
+import static io.delta.kernel.internal.util.Preconditions.checkArgument;
+
 import io.delta.kernel.annotation.Evolving;
 import io.delta.kernel.expressions.Column;
 import io.delta.kernel.internal.types.DataTypeJsonSerDe;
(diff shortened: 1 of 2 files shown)


kernel/kernel-api/src/test/scala/io/delta/kernel/types/StructTypeSuite.scala
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.delta.kernel.types
+
+import org.scalatest.funsuite.AnyFunSuite
+
+class StructTypeSuite extends AnyFunSuite {
+
+  test("get works as expected") {
```
scottsand-db, line 22: nit: the test name should give us a bit more info about what "expected" means.

i.e. `get throws exception if key doesn't exist, else returns value"

kernel/kernel-api/src/main/java/io/delta/kernel/types/StructType.java
```diff
@@ -93,7 +95,14 @@ public int indexOf(String fieldName) {
     return fieldAndOrdinal != null ? fieldAndOrdinal._2 : -1;
   }
 
+  /**
+   * @return the StructField with the given name
+   * @throws IllegalArgumentException if the field does not exist
+   */
   public StructField get(String fieldName) {
```
scottsand-db, line 102: I agree this is a strict improvement over NPE, but do we now expect every call to `StructField.get` to wrap it with a try and catch?

Could we also, in this PR, at least add a `getOpt` that returns an `Optional<String>` and this could give some more flexibility to callers?

raveeram-db (approved):

PR #4477, opened 2025-04-29
Use compacted log files during log replay

nicklan:
- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

This PR builds on #4457 and adds functionality to omit the original json files and just use the compacted files.  It introduces a new way to get the list of files from the LogSegment that returns the compacted files in place of the original jsons. 

For now we just _always_ use these files, and can plumb through an option as a follow-on if its requested.

## How was this patch tested?

* Unit tests
* Integration tests on a new table with compactions in the log

No, users do not (currently) need to be aware of this.


kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableReadsSuite.scala
```diff
@@ -571,6 +571,19 @@ class DeltaTableReadsSuite extends AnyFunSuite with TestUtils {
       expectedAnswer = (0 until 150).map(i => TestRow(i.toLong)))
   }
 
+  test("table with checkpoint and log compaction") {
```
allisonport-db, line 574: should we do something to test that we are in fact reading the compaction files for this? I think there's no way to know how we are loading this data right?

I think we already have utils for this that can save what files we read I can help find them if needed

allisonport-db: also as discussed we should (1) test with the compactions written by Kernel (2) test with compaction files with all the possible actions in it (removes, domains metadatas, etc)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
```diff
@@ -186,6 +191,69 @@ public LogSegment(
...
+                compactions.stream()
+                    .sorted(
+                        Comparator.comparing((FileStatus a) -> new Path(a.getPath()).getName())
+                            .reversed())
+                    .collect(Collectors.toList()));
+
+    this.allFilesWithCompactionsReversed = new Lazy<>(() -> loadAllFilesWithCompactionsReversed());
+  }
+
+  private List<FileStatus> loadAllFilesWithCompactionsReversed() {
```
allisonport-db, line 207: Can you point me to the corresponding code for this in delta spark?

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/snapshot/LogSegmentSuite.scala
```diff
@@ -234,4 +236,79 @@ class LogSegmentSuite extends AnyFunSuite with MockFileSystemClientUtils {
     // scalastyle:on line.size.limit
     assert(logSegment.toString === expectedToString)
   }
+
+  test("allFilesWithCompactionsReversed -- 3 - 5 in middle") {
```
allisonport-db, line 240: is it possible for log compaction files to overlap with each other? what would happen then?

allisonport-db: we can have multiple compaction files also right? let's test that? (back to back and separated?)

allisonport-db: also what about a test where the log compaction file spans the entire version range (i.e. only that should be returned)?

allisonport-db: should we include checkpoints in some of the test cases?

allisonport-db (changes requested):
Source code looks good just had some comments on testing

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
```diff
@@ -186,6 +191,69 @@ public LogSegment(
                         Comparator.comparing((FileStatus a) -> new Path(a.getPath()).getName())
                             .reversed())
                     .collect(Collectors.toList()));
+
+    this.compactionsReversed =
```
scottsand-db, line 195: Question: what are the guarantees / expectations for the compactions given to LogSegment?

Should we filter and guarantee that for all compactions, x.y.json, x is >= checkpointVersion + 1 (if present), and y is <= version?

scottsand-db: Can we also clarify the ordering of the compactions? It is lexicographical ordering, right?

So:

```
0010.0011.json
0010.9999.json
0014.0016.json
0015.0017.json
0015.0019.json
0016.0018.json
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
```diff
@@ -186,6 +191,69 @@ public LogSegment(
...
+
+    this.allFilesWithCompactionsReversed = new Lazy<>(() -> loadAllFilesWithCompactionsReversed());
+  }
+
+  private List<FileStatus> loadAllFilesWithCompactionsReversed() {
+    if (compactions.isEmpty()) {
+      return allFilesReversed.get();
+    } else {
+      final Iterator<FileStatus> deltaIt = allFilesReversed.get().iterator();
+      final Iterator<FileStatus> compactionIt = compactionsReversed.get().iterator();
```
scottsand-db, line 212: Since these are lexicographically sorted, does it make sense to process them as an iterator?

```
0010.0011.json
0010.0099.json
0014.0016.json
0015.0017.json
0015.0019.json
0016.0018.json
```

We should really just take `0010.0099.json` and ignore all the others, right?

scottsand-db: Or do we just want to implement a greedy algorithm that reads the x.y.json in order as we can?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableReadsSuite.scala
nicklan: This is now tested. I only tested adds/removes. The `LogCompactionWriterSuite` verifies that all supported actions are preserved, and we read these compactions as "normal" json files, so I believe all actual functionality is tested. I'd be quite a lot more test code to check all these actions, so I've opted not to for now. If you think it's important, let me know why, and I can look at adding them in.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
nicklan: Yep. This is already checked in the `SnapshotManager`, but I've added `checkArgument` validations here.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
nicklan: I've added a helper that uses a greedy algorithm and is (hopefully) easier to understand than this was.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
```diff
@@ -149,6 +162,28 @@ public LogSegment(
...
+
+        // Check that our compactions are in range
+        checkArgument(
+            compactions.stream()
+                .allMatch(
+                    fs -> {
+                      Tuple2<Long, Long> versions = FileNames.logCompactionVersions(fs.getPath());
+                      return versions._2 <= version;
+                    }),
+            "compactions must have end version <= version");
```
scottsand-db, line 174: Can this be moved up next to the `compactions must have start version less than end version` check?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
```diff
@@ -277,4 +343,78 @@ private String formatList(List<FileStatus> list) {
...
+      this.deltaIt = allFilesReversed.iterator();
+      this.compactionsReversed = compactionsReversed;
+    }
+
+    // We have two lists, one of deltas and one of compactions. Each is sorted in DESCENDING
+    // order. Given this, resolves as follows:
+    // - set a "hi/lo" goalpost around the next compactions
+    // - for each delta, if its version is:
+    //   - greater than the current compaction high point, include it, move to next delta
+    //   - less than (but not equal to) the current low point, skip it, move to next delta
```
scottsand-db, line 366: `the current compaction low point`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
```diff
@@ -277,4 +343,78 @@ private String formatList(List<FileStatus> list) {
...
+    //     consider. This deals with overlapping compactions in a greedy way, ensuring we
+    //     ignore any overlapping compactions.
+    List<FileStatus> resolveFiles() {
+      ArrayList<FileStatus> ret = new ArrayList<FileStatus>();
+      setHiLo();
+      while (deltaIt.hasNext()) {
+        FileStatus currentDelta = deltaIt.next();
+        long deltaVersion = FileNames.deltaVersion(currentDelta.getPath());
+        if (deltaVersion == currentCompactionLo) {
+          // we're about to cross out of the compaction. insert the compaction and advance to the
```
scottsand-db, line 379: I think this requires (fairly so) that the delta versions are contiguous. i.e. it requires that 10.json exists for us to filly accept 10.20.json into the `ret`

Q: What would change if you added the compaction at the "start" as opposed to at the end?

Could we future proof this for some crazy future world where there are gaps in the delta log (suppose a future catalogManagedV2 table feature allows for unordered backfill) ?

Not a blocker, but wanted to ask this / call it out

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
```diff
@@ -186,6 +221,27 @@ public LogSegment(
                         Comparator.comparing((FileStatus a) -> new Path(a.getPath()).getName())
                             .reversed())
                     .collect(Collectors.toList()));
+
+    this.compactionsReversed =
+        new Lazy<>(
+            () ->
+                compactions.stream()
+                    .sorted(
+                        Comparator.comparing((FileStatus a) -> new Path(a.getPath()).getName())
```
scottsand-db, line 230: Can we sort by end version instead? Not lexicographically?

It would be more optimal, for example, to process the below in reverse endVersion order

```
80.100.json (start of the reverse list)
90.95.json
81.90.json
```

Lexicographical order should always be the endVersion order, except when an x.y has a bigger "range" (num commits inside it) than others around it.

scottsand-db: In general throughout our code in Kernel, we should probably clarify / specify exactly what "order" means (lexicographical vs numerical) :/ 

Another example, btw, is due to staged commit files, where the `_staged_commits` makes them come at the bottom of any sort order (which isn't great - we will need to fix this for catalogManaged tables)

scottsand-db (commented):
Looks great! Left some questions / comments / food for thought. Heading out soon -- need to review in more detail next work week

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/snapshot/LogSegmentSuite.scala
```diff
@@ -234,4 +236,136 @@ class LogSegmentSuite extends AnyFunSuite with MockFileSystemClientUtils {
...
+    val expected = List(
+      FileStatus.of(FileNames.deltaFile(logPath, 6), 6, 6 * 10),
+      FileStatus.of(FileNames.logCompactionPath(logPath, 3, 5).toString, 3, 3 * 10),
+      FileStatus.of(FileNames.deltaFile(logPath, 2), 2, 2 * 10),
+      FileStatus.of(FileNames.deltaFile(logPath, 1), 1, 1 * 10),
+      FileStatus.of(FileNames.deltaFile(logPath, 0), 0, 0 * 10)).asJava;
+    assert(allFiles === expected);
+  }
+
+  test("allFilesWithCompactionsReversed -- 3 - 5 at start") {
```
scottsand-db, line 260: I do see some similar patterns throughout all these tests. What about consolidating it a bit?

I think we could turn each test case into a 4 line call

```java
testCompactionCase(
    deltas = 0 to 6,
    compactions = Seq((3, 5)),
    expected = "6,3-5,2,1,0")
```

And here are the helpers that claude made for me (do double check)

```java
/**
   * Test helper that simplifies LogSegment compaction tests with a concise syntax.
   * 
   * @param deltas Range or sequence of delta versions (version is taken from the max)
   * @param compactions Sequence of (start, end) compaction ranges
   * @param expected Comma-separated string of expected files in the result, e.g. "10,9,5-8,4,3"
   *                 where numbers represent delta files and ranges represent compaction files
   */
  def testCompactionCase(
      deltas: Seq[Long],
      compactions: Seq[(Long, Long)],
      expected: String): Unit = {
    
    // The version is the maximum delta version
    val version = deltas.max
    
    // Create the LogSegment
    val deltas_list = deltaFileStatuses(deltas).toList.asJava
    val compactions_list = compactedFileStatuses(compactions).toList.asJava
    val segment = new LogSegment(
      logPath, 
      version, 
      deltas_list, 
      compactions_list, 
      Collections.emptyList(), 
      1)
    
    // Parse the expected string
    val expectedFiles = parseExpectedString(expected)
    
    // Assert that allFilesWithCompactionsReversed matches expected files
    assert(segment.allFilesWithCompactionsReversed() === expectedFiles)
  }

/**
   * Parse a comma-separated string of expected files.
   * 
   * @param expected Comma-separated string like "10,9,5-8,4,3" where numbers are delta files
   *                 and ranges like "5-8" are compaction files
   * @return A Java List of FileStatus objects
   */
  private def parseExpectedString(expected: String): JList[FileStatus] = {
    expected.split(",").map(_.trim).map { item =>
      if (item.contains("-")) {
        // This is a compaction file
        val parts = item.split("-").map(_.trim.toLong)
        compactedFileStatus(parts(0), parts(1))
      } else {
        // This is a delta file
        deltaFileStatus(item.toLong)
      }
    }.toList.asJava
  }
```

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/snapshot/LogSegmentSuite.scala
```diff
@@ -234,4 +236,136 @@ class LogSegmentSuite extends AnyFunSuite with MockFileSystemClientUtils {
...
+
+    // expect to get 6, 3-5.compact, 2
+    val expected = List(
+      FileStatus.of(FileNames.deltaFile(logPath, 6), 6, 6 * 10),
+      FileStatus.of(FileNames.logCompactionPath(logPath, 3, 5).toString, 3, 3 * 10),
+      FileStatus.of(FileNames.deltaFile(logPath, 2), 2, 2 * 10)).asJava;
+    assert(allFiles === expected);
+  }
+
+  test("allFilesWithCompactionsReversed -- 3 - 5, and 7 - 9") {
```
scottsand-db, line 315: could we test some more cases of having multiple CRC?

See far I see
- no overlap (1 delta gap in between)
- overlap

Can we add
- no overlap (0 delta gap in betwee) and multiple compactions, e.g. `3-5, 6-8, 9-11`
- overlap (range contains another) e.g. `3-15, 5-9`
- one super complex case with a bunch of different overlaps and no overlap cases?

scottsand-db (changes requested):
Looks great! Logic in the LogSegment class is very clear, thank you!

Added some comments on the tests

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableReadsSuite.scala
```diff
@@ -1063,4 +1066,161 @@ class DeltaTableReadsSuite extends AnyFunSuite with TestUtils {
...
+          dataPath,
+          logPath,
+          compaction._1,
+          compaction._2,
+          0)
+        hook.threadSafeInvoke(engine)
+      }
+      engine.resetMetrics()
+
+      checkTable(
```
allisonport-db, line 1132: We are using the log compactions for three different operations right? (1) snapshot loading (P&M only) (2) log replay (Scans) (3) loading the metadata domain map

Should we test that we use it for each of these? It seems like maybe before these recent commits we actually had missed (2)?

I think this doesn't necessarily need to be too complex of a test, we can test all of these things using 1 single test table built with just a few versions (and reset the metrics in between each check).

allisonport-db: Alternatively maybe we can accomplish this instead by not converting the `actualCompactionsRead` to set and instead checking we actually read a specific compaction file twice for example when snapshot laoding + scanning

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableReadsSuite.scala
allisonport-db: Why is it a lot more code to also check the other actions? 

For example for DM, I think you can just set a domain in one of the commits and then to check it just use `snapshot.getDomain` in the check?

And then for like a metadata action, just change a table property?

allisonport-db: I would be more inclined to prefer 1 test that has a compaction file with everything (all actions) in it and checks everything but it doesn't need to be repeated with all the different cases you've added below. I think those are fairly well covered by the LogSegmentSuite

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
nicklan: > perhaps if we assume constant size of compactions, then lexicographical order is equally as good? 

This was my assumption, but I realize it _could_ be different, say if different clients prefer different intervals. So I've switched to end version as it's equal or better (at minor increase in parsing cost)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
nicklan: Well, if there are gaps I think the logic gets _way_ more complex. If we assume we handle all the missing stuff somehow, I guess the "fix" here would just be `(deltaVersion <= currentCompactionLo)`, so maybe not worth trying to "fix" things until we know what we actually need to do.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/snapshot/LogSegmentSuite.scala
nicklan: Yep, much better, thanks!

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableReadsSuite.scala
nicklan: Yep. I've moved this to a new test file now, and added a test with DM and a property change.

I didn't include removes in that test because I don't think kernel can do that just yet (easily), but that's well covered by the other tests.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableReadsSuite.scala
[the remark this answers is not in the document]
nicklan: Yep, have moved it now, good call

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/snapshot/LogSegmentSuite.scala
nicklan: These are all added now

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/snapshot/LogSegmentSuite.scala
```diff
@@ -346,4 +348,112 @@ class LogSegmentSuite extends AnyFunSuite with MockFileSystemClientUtils {
     assert(logSegment.toString === expectedToString)
   }
 
+  def deltaFileStatus(v: Long): FileStatus =
```
scottsand-db, line 351: nit: can you add this to `MockFileSystemClientUtils` ?

And refactor `deltaFileStatuses` to use `deltaFileStatus` etc?

scottsand-db (approved):
LGTM -- after please implementing the one last change to move the helper to MockFileSystemUtils so that we are standardized / can re-use it easily.

Great tests and great implementation, thanks!

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/LogCompactionSuite.scala
```diff
@@ -0,0 +1,262 @@
...
+      val hadoopFileIO = new HadoopFileIO(new Configuration())
+      val metricEngine = new MetricsEngine(hadoopFileIO)
+      val table = Table.forPath(metricEngine, tblPath)
+
+      val snapshot = table.getLatestSnapshot(metricEngine).asInstanceOf[SnapshotImpl]
+      val checkpointProp =
+        snapshot.getMetadata().getConfiguration.get(TableConfig.CHECKPOINT_POLICY.getKey)
+      assert(checkpointProp == "v2")
+
+      // this is the read that the snapshot did
```
allisonport-db, line 78: these checks for each operation are awesome!

allisonport-db (approved):
LGTM

PR #4478, opened 2025-04-29
[Kernel][InCommitTimestamp] Use ICT (if available) as the commit timestamp in getChanges

dhruvarya-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

Adds support for InCommitTimestamp in the getChanges API. If any commit that is part of the getChanges range has an inCommitTimestamp available in the CommitInfo action, that timestamp will be used as the commit timestamp instead of the file modification time.

## How was this patch tested?

Updated test cases in TableChangesSuite. Added a new golden table which has ICT enabled.
 
Yes, the timestamp returned by the getChanges function will now be different for tables with ICT enabled.

### DIFF
--- a/connectors/golden-tables/src/main/resources/golden/deltalog-getChanges-ict-enabled/_delta_log/00000000000000000000.json
+++ b/connectors/golden-tables/src/main/resources/golden/deltalog-getChanges-ict-enabled/_delta_log/00000000000000000000.json
@@ -1,4 +1,4 @@
 {"commitInfo":{"timestamp":1704392842074,"inCommitTimestamp":1704392842074,"operation":"Manual Update","operationParameters":{},"isolationLevel":"Serializable","isBlindAppend":true,"operationMetrics":{},"engineInfo":"Apache-Spark/3.5.0 Delta-Lake/3.1.0-SNAPSHOT","txnId":"95ec924a-6859-4433-8008-6d6b4a0e3ba5"}}
-{"protocol":{"minReaderVersion":3,"minWriterVersion":7,"writeFeatures":["inCommitTimestamp"]}}
+{"protocol":{"minReaderVersion":1,"minWriterVersion":7,"writerFeatures":["inCommitTimestamp"]}}
 {"metaData":{"id":"testId","format":{"provider":"parquet","options":{}},"schemaString":"{\"type\":\"struct\",\"fields\":[{\"name\":\"part\",\"type\":\"integer\",\"nullable\":true,\"metadata\":{}},{\"name\":\"id\",\"type\":\"integer\",\"nullable\":true,\"metadata\":{}}]}","partitionColumns":[],"configuration":{"delta.enableInCommitTimestamp":"true"}}}
 {"add":{"path":"fake/path/1","partitionValues":{},"size":1,"modificationTime":1,"dataChange":true}}
--- a/connectors/golden-tables/src/main/resources/golden/deltalog-getChanges-ict-enabled/_delta_log/00000000000000000000.json
+++ b/connectors/golden-tables/src/main/resources/golden/deltalog-getChanges-ict-enabled/_delta_log/00000000000000000000.json
@@ -1,4 +1,0 @@
-{"commitInfo":{"timestamp":1704392842074,"inCommitTimestamp":1704392842074,"operation":"Manual Update","operationParameters":{},"isolationLevel":"Serializable","isBlindAppend":true,"operationMetrics":{},"engineInfo":"Apache-Spark/3.5.0 Delta-Lake/3.1.0-SNAPSHOT","txnId":"95ec924a-6859-4433-8008-6d6b4a0e3ba5"}}
-{"protocol":{"minReaderVersion":1,"minWriterVersion":7,"writerFeatures":["inCommitTimestamp"]}}
-{"metaData":{"id":"testId","format":{"provider":"parquet","options":{}},"schemaString":"{\"type\":\"struct\",\"fields\":[{\"name\":\"part\",\"type\":\"integer\",\"nullable\":true,\"metadata\":{}},{\"name\":\"id\",\"type\":\"integer\",\"nullable\":true,\"metadata\":{}}]}","partitionColumns":[],"configuration":{"delta.enableInCommitTimestamp":"true"}}}
-{"add":{"path":"fake/path/1","partitionValues":{},"size":1,"modificationTime":1,"dataChange":true}}
--- a/connectors/golden-tables/src/main/resources/golden/deltalog-getChanges-ict-enabled/_delta_log/00000000000000000000.json
+++ b/connectors/golden-tables/src/main/resources/golden/deltalog-getChanges-ict-enabled/_delta_log/00000000000000000000.json
@@ -1,0 +1,4 @@
+{"commitInfo":{"timestamp":1704392842074,"inCommitTimestamp":1704392842074,"operation":"Manual Update","operationParameters":{},"isolationLevel":"Serializable","isBlindAppend":true,"operationMetrics":{},"engineInfo":"Apache-Spark/3.5.0 Delta-Lake/3.1.0-SNAPSHOT","txnId":"95ec924a-6859-4433-8008-6d6b4a0e3ba5"}}
+{"protocol":{"minReaderVersion":3,"minWriterVersion":7,"writeFeatures":["inCommitTimestamp"]}}
+{"metaData":{"id":"testId","format":{"provider":"parquet","options":{}},"schemaString":"{\"type\":\"struct\",\"fields\":[{\"name\":\"part\",\"type\":\"integer\",\"nullable\":true,\"metadata\":{}},{\"name\":\"id\",\"type\":\"integer\",\"nullable\":true,\"metadata\":{}}]}","partitionColumns":[],"configuration":{"delta.enableInCommitTimestamp":"true"}}}
+{"add":{"path":"fake/path/1","partitionValues":{},"size":1,"modificationTime":1,"dataChange":true}}
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
@@ -210,10 +210,9 @@ public class TableImpl implements Table {
               if (shouldDropProtocolColumn) {
                 batchToReturn = batchToReturn.withDeletedColumnAt(protocolIdx);
               }
-              if (shouldDropCommitInfoColumn) {
-                batchToReturn =
-                    batchToReturn.withDeletedColumnAt(
-                        batchToReturn.getSchema().indexOf("commitInfo"));
+              int commitInfoIdx = batchToReturn.getSchema().indexOf("commitInfo");
+              if (shouldDropCommitInfoColumn && commitInfoIdx != -1) {
+                batchToReturn = batchToReturn.withDeletedColumnAt(commitInfoIdx);
               }
               return batchToReturn;
             });
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
@@ -205,11 +205,13 @@ public class TableImpl implements Table {
                 }
               }
               ColumnarBatch batchToReturn = batch;
-              if (shouldDropCommitInfoColumn) {
-                batchToReturn = batchToReturn.withDeletedColumnAt(batch.getSchema().indexOf("commitInfo"));
-              }
               if (shouldDropProtocolColumn) {
                 batchToReturn = batchToReturn.withDeletedColumnAt(protocolIdx);
+              }
+              if (shouldDropCommitInfoColumn) {
+                batchToReturn =
+                    batchToReturn.withDeletedColumnAt(
+                        batchToReturn.getSchema().indexOf("commitInfo"));
               }
               return batchToReturn;
             });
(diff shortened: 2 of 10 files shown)


kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/TableChangesSuite.scala
```diff
@@ -75,87 +77,109 @@ class TableChangesSuite extends AnyFunSuite with TestUtils {
...
-        tablePath,
-        1,
-        2,
-        FULL_ACTION_SET)
-      testGetChangesVsSpark(
-        tablePath,
-        0,
-        0,
-        FULL_ACTION_SET)
+  // deltalog-getChanges is golden table from Delta Standalone test while
```
dhruvarya-db, line 80:
```suggestion
  // deltalog-getChanges is the golden table from the Delta Standalone test while
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
```diff
@@ -185,9 +185,15 @@ public CloseableIterator<ColumnarBatch> getChanges(
     // Create a new action set that always contains protocol
     Set<DeltaLogActionUtils.DeltaAction> copySet = new HashSet<>(actionSet);
     copySet.add(DeltaLogActionUtils.DeltaAction.PROTOCOL);
+    // Always include commitInfo in the action set so that we can
+    // extract the inCommitTimestamp from it when set.
```
dhruvarya-db, line 189:
```suggestion
    // extract the inCommitTimestamp from it when available.
```

connectors/golden-tables/src/main/resources/golden/deltalog-getChanges-ict-enabled/_delta_log/00000000000000000000.json
allisonport-db, line 1: If you are adding a golden table you should also add the src code to `GoldenTables.scala`. But I think since we originally added this we've moved away from using golden tables in our tests -- you can just write the test table with Spark (copy over whatever code from the golden tables file)

allisonport-db: If you don't mind & it's simple can you write the test table with spark for both the test table you're adding and the existing one?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
```diff
@@ -200,11 +206,15 @@ public CloseableIterator<ColumnarBatch> getChanges(
...
                 }
               }
+              ColumnarBatch batchToReturn = batch;
               if (shouldDropProtocolColumn) {
-                return batch.withDeletedColumnAt(protocolIdx);
-              } else {
-                return batch;
+                batchToReturn = batchToReturn.withDeletedColumnAt(protocolIdx);
               }
+              int commitInfoIdx = batchToReturn.getSchema().indexOf("commitInfo");
```
allisonport-db, line 213: nit - put this by `protocolIdx` for consistency?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
```diff
@@ -387,21 +388,56 @@ private CloseableIterator<ActionWrapper> getNextActionsIter() {
     }
   }
 
-  /** Take input (iterator<T>, boolean) and produce an iterator<T, boolean>. */
+  /**
+   * Take input (iterator<T>, boolean) and produce an iterator<T, boolean>. When the input iterator
+   * has been created from a delta file, this function will try to use the inCommitTimestamp (if
+   * available) from the CommitInfo action as the commit timestamp for the file.
+   */
```
allisonport-db, line 395: Not on you but it seems like these docs are out of date. Can you update this to something along the lines of
"Takes an input iterator of actions read from the file and metadata about the file read, and combines it to return an `Iterator<ActionWrapper>`.  The timestamp in the ActionWrapper is the inCommitTimestamp (if available)..."

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
```diff
@@ -387,21 +388,56 @@ private CloseableIterator<ActionWrapper> getNextActionsIter() {
...
       boolean isFromCheckpoint,
       long version,
       Optional<Long> timestamp) {
+    final List<ColumnarBatch> firstActionList = new ArrayList<>();
+    final int commitInfoOrdinal = deltaReadSchema.indexOf("commitInfo");
+    // For delta files, we want to use the inCommitTimestamp from commitInfo
+    // as the commit timestamp for the file.
+    // Since CommitInfo should be the first action in the delta when inCommitTimestamp is
+    // enabled, we will read the first action and try to extract the timestamp from it.
+    // We also cache the first action so that we can return it in the iterator.
```
allisonport-db, line 407: I wonder if we can just implement a default method `peek` on `CloseableIterator`? Would simplify this code

allisonport-db: If it's the first action don't we not need to check the entire column vector either? (which is done in `InCommitTimestampUtils.getCommitInfo`)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
```diff
@@ -387,21 +388,56 @@ private CloseableIterator<ActionWrapper> getNextActionsIter() {
...
+    Optional<Long> commitTimestamp = timestamp;
+    if (!isFromCheckpoint && fileReadDataIter.hasNext() && commitInfoOrdinal != -1) {
+      ColumnarBatch firstAction = fileReadDataIter.next();
```
allisonport-db, line 410: nit - naming this `firstAction` is misleading -- it's a batch of actions

allisonport-db (commented):
Just a few questions / requests to clean up the code! Thanks

Also, wondering if we plan to also update `getVersionBeforeOrAtTimestamp` and `getVersionAtOrAfterTimestamp`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
```diff
@@ -387,21 +388,56 @@ private CloseableIterator<ActionWrapper> getNextActionsIter() {
...
+      firstActionList.add(firstAction);
+      ColumnVector commitInfoVector = firstAction.getColumnVector(commitInfoOrdinal);
+      Optional<Long> ictOpt =
+          InCommitTimestampUtils.getCommitInfo(commitInfoVector)
+              .flatMap(CommitInfo::getInCommitTimestamp);
+      if (ictOpt.isPresent()) {
+        commitTimestamp = ictOpt;
+      }
+    }
+    final Iterator<ColumnarBatch> firstActionIter = firstActionList.iterator();
```
scottsand-db, line 420: I think you can use:

`Utils.singletonCloseableIterator` to create a single-element closeable iterator

Then use `CloseableIterator.combine` to combine that iterator with `fileReadDataIter`

And then you just use that final iterator to fulfilll the `CloseableIterator<ActionWrapper>() {` implementation.

Seems simpler? no if statements?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/InCommitTimestampUtils.java
```diff
@@ -75,4 +77,20 @@ private static boolean didCurrentTransactionEnableICT(
...
   }
+
+  /**
+   * Converts the given {@link ColumnVector} containing commit info to a {@link CommitInfo} object.
+   *
+   * @param commitInfoVector a ColumnVector containing commit info
+   * @return the commit info as a {@link CommitInfo} object, or an empty Optional if no commit info
+   *     is found
+   */
+  public static Optional<CommitInfo> getCommitInfo(ColumnVector commitInfoVector) {
```
scottsand-db, line 88: Why not move this to the CommitInfo class?

scottsand-db (changes requested):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/InCommitTimestampUtils.java
huan233usc: +1, putting CommitInfo in ICT utils is confusing.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
huan233usc: +1 alternative have a wrapper like

public class PeekableIterator<T> implements CloseableIterator<T> {
    private final CloseableIterator<T> delegate;
    private T next;

    public PeekableIterator(CloseableIterator<T> delegate) {
        this.delegate = delegate;
        this.next = delegate.hasNext() ? delegate.next() : null;
    }

This way we don't need to have two iterators.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
```diff
@@ -185,9 +185,15 @@ public CloseableIterator<ColumnarBatch> getChanges(
     // Create a new action set that always contains protocol
     Set<DeltaLogActionUtils.DeltaAction> copySet = new HashSet<>(actionSet);
     copySet.add(DeltaLogActionUtils.DeltaAction.PROTOCOL);
+    // Always include commitInfo in the action set so that we can
```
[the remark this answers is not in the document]
dhruvarya-db: My bad, updated

connectors/golden-tables/src/main/resources/golden/deltalog-getChanges-ict-enabled/_delta_log/00000000000000000000.json
dhruvarya-db: Copied the code from GoldenTables.scala. The table is now created on the spot

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
dhruvarya-db: Updated

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
```diff
@@ -387,21 +388,56 @@ private CloseableIterator<ActionWrapper> getNextActionsIter() {
...
       CloseableIterator<ColumnarBatch> fileReadDataIter,
       boolean isFromCheckpoint,
       long version,
       Optional<Long> timestamp) {
+    final List<ColumnarBatch> firstActionList = new ArrayList<>();
+    final int commitInfoOrdinal = deltaReadSchema.indexOf("commitInfo");
+    // For delta files, we want to use the inCommitTimestamp from commitInfo
+    // as the commit timestamp for the file.
+    // Since CommitInfo should be the first action in the delta when inCommitTimestamp is
+    // enabled, we will read the first action and try to extract the timestamp from it.
```
dhruvarya-db: > If it's the first action don't we not need to check the entire column vector either? (which is done in InCommitTimestampUtils.getCommitInfo)

@allisonport-db Thanks for catching this. Since in this case we are extracting the CommitInfo only to get the ICT, it must be the first action of the batch. I have inlined a simpler version of the extractor

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
allisonport-db: Gotcha. This will be changed/fixed in the future anyways when we revisit our expression support so fine by me for now.

allisonport-db (approved):
LGTM

huan233usc (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
```diff
@@ -185,9 +185,15 @@ public CloseableIterator<ColumnarBatch> getChanges(
     // Create a new action set that always contains protocol
```
scottsand-db, line 185: `always contains protocol` --> change the wording, so we don't have to keep adding all the new actions we want it to contain.

the goal is that it is a superset and contains additonal actions that we need to do checks / extract important information.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
```diff
@@ -200,11 +206,15 @@ public CloseableIterator<ColumnarBatch> getChanges(
                   TableFeatures.validateKernelCanReadTheTable(protocol, getDataPath().toString());
                 }
               }
+              ColumnarBatch batchToReturn = batch;
               if (shouldDropProtocolColumn) {
```
scottsand-db, line 210: why don't we have the `protocolIdx != -1` check here?

dhruvarya-db: That's a good question. Turns we don't need that check even for commitInfoIdx since the schema is constructed from `copySet`, which is guaranteed to have both these action types. Will remove

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
dhruvarya-db: Updated the comment. Is it better now?

scottsand-db (approved):

PR #4481, opened 2025-04-29
[Kernel] Initial E2E PR to support replace; excludes RTAS & column mapping support

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Initial E2E PR to support replace table operation. For now we manually block RTAS and column mapping support, those will come in later (separate) PRs. These are easy to factor out and reduce the # of tests coverage and src code changes in this PR.

## How was this patch tested?

Adds a test suite + runs that test suite with CRC enabled. Also updates other unit tests.


kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/utils/TestUtils.scala
```diff
@@ -797,27 +798,37 @@ trait TestUtils extends Assertions with SQLHelper {
...
-          .filterNot(_.isRemoved)
-          .toSet
-          .asJava),
-      "Domain metadata in checksum should match snapshot")
+    crcInfoOpt.toScala.foreach { crcInfo =>
+      // TODO: check metadata, protocol and file size.
+      assert(
+        crcInfo.getNumFiles === collectScanFileRows(snapshot.getScanBuilder.build()).size,
+        "Number of files in checksum should match snapshot")
+      if (expectEmptyTable) {
```
allisonport-db, line 817: this is a super easy additional check to add for the tableSizeInBytes and fileSizeHistogram; in the future we should probably revisit this but this was a super low-effort improvement and it make sure these are reset and 0/empty for replace table (without ctas)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaErrors.java
```diff
@@ -233,6 +233,10 @@ public static KernelException requiresSchemaForNewTable(String tablePath) {
         tablePath, "Must provide a new schema to write to a new table.");
   }
 
+  public static KernelException requireSchemaForReplaceTable(String tablePath) {
```
scottsand-db, line 236: `tablePath` is unused

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ReplaceTableTransactionBuilderImpl.java
```diff
@@ -0,0 +1,61 @@
...
+      throw new TableNotFoundException(
+          tblf.getTablePath(), "Trying to replace a table that does not exist.");
+    }
+  }
+
+  /*
+  Generally for replace table we want to reset all table state, however there are a few
+  delta-specific properties that we should preserve
+  */
+  protected static Set<String> tablePropertyKeysToPreserve =
```
scottsand-db, line 54: final?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
```diff
@@ -358,4 +358,8 @@ private void recordSnapshotErrorReport(
     SnapshotReport snapshotReport = SnapshotReportImpl.forError(snapshotContext, e);
     engine.getMetricsReporters().forEach(reporter -> reporter.report(snapshotReport));
   }
+
+  public TransactionBuilder createReplaceTableTransactionBuilder(Engine engine, String engineInfo) {
```
scottsand-db, line 362: Can you group this with the other public APIs in this class? i.e. not underneath the private helper methods?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ReplaceTableTransactionBuilderImpl.java
```diff
@@ -0,0 +1,61 @@
...
+public class ReplaceTableTransactionBuilderImpl extends TransactionBuilderImpl {
+
+  public ReplaceTableTransactionBuilderImpl(TableImpl table, String engineInfo) {
+    super(table, engineInfo, Operation.REPLACE_TABLE);
+  }
+
+  @Override
+  public Transaction build(Engine engine) {
+    try {
+      withMaxRetries(0); // We don't support conflict resolution yet so disable retries for now
```
scottsand-db, line 37: So the user can set `myBuilder.withMaxRetries(10)` and then we will just ignore it? Would that be weird or confusing? Should we override that method and either fail or log a warning?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -262,6 +267,17 @@ protected TransactionImpl buildTransactionInternal(
     if (isCreateOrReplace) {
       // For a new table definition start with an empty initial metadata
       baseMetadata = getInitialMetadata();
+      // In the case of Replace table there are a few delta-specific properties we want to preserve
+      if (latestSnapshot.isPresent()) {
```
scottsand-db, line 271: So the semantic/assumption/conclusion here is that:

If it is a replace table operation then we expect `isLatestSnapshot.isPresent` to also be true, right? i.e. that's how we are concluding that it is a replace table operation in the first place?

Worth adding a comment explaining that?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ReplaceTableTransactionBuilderImpl.java
scottsand-db: and if it is static and final, then it should be `TABLE_PROPERTY_KEYS_TO...`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -386,11 +383,25 @@ private long doCommit(
     resolvedDomainMetadatas.forEach(
         dm -> metadataActions.add(createDomainMetadataSingleAction(dm.toRow())));
 
-    try (CloseableIterator<Row> stageDataIter = dataActions.iterator()) {
+    try (CloseableIterator<Row> userStageDataIter = dataActions.iterator()) {
+
```
scottsand-db, line 387: nit: no need to have empty line here

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -386,11 +383,25 @@ private long doCommit(
...
+      CloseableIterator<Row> completeFileActionIter;
```
scottsand-db, line 388: final? This makes it clearer that we are trying to assign it in either block of the next `if` statement -- and we don't want to re-assign it later

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -386,11 +383,25 @@ private long doCommit(
...
+      if (isReplaceTable()) {
+        // For now, block CTAS with replace; this will be unblocked in a follow-up PR
```
scottsand-db, line 390: What does `block CTAS with replace`? the "with" confuses me -- do you mean "block CTAS and replace"? Do you mean "block replace table as select? RTAS??

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -752,8 +780,32 @@ private void generateClusteringDomainMetadataIfNeeded() {
...
             ClusteringUtils.getClusteringDomainMetadata(clusteringColumnsOpt.get());
-        domainsToAdd.put(clusteringDomainMetadata.getDomain(), clusteringDomainMetadata);
+        addDomain(
+            clusteringDomainMetadata.getDomain(), clusteringDomainMetadata.getConfiguration());
       }
     }
   }
+
+  /**
+   * Returns the remove file rows to completely reset the table state in the case of a replace
```
scottsand-db, line 790: `to completely reset the table state in the case of a replace (i.e. removes every active add file in the table)` 

This seems like 2 different things to me. The former seems like a superset of the latter. i.e. the former would also include resetting the schema? / protocol?

Can you just say: ` Returns the remove file rows needed to remove every active add file in the table`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -664,6 +675,23 @@ public List<DomainMetadata> getComputedDomainMetadatasToCommit() {
       }
 
       generateClusteringDomainMetadataIfNeeded();
```
scottsand-db, line 677: do we still need to invoke `generateClusteringDomainMetadataIfNeeded` if it is a replaceTable call?

scottsand-db (changes requested):
Looks great! Reviewed the prod code. Will review the test code later

vkorukanti (approved):
LGTM.

Very thorough tests. One suggestion is there a way we could put the tests near the existing suite? For example icebergCompatV2 tests with replace in the existing icebergcompatV2 suite. No need to do this in this PR.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
allisonport-db: yes! we might be replacing with a clustered table

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
allisonport-db: Makes sense sgtm

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ReplaceTableTransactionBuilderImpl.java
allisonport-db: I'm open to opinions on this one but I think it's okay to retry _less_ than what is specified (what we do if we find a conflict that we cannot resolve anyways). It's called "max" retries too :)

Also this is similar to what would happen if for example we haven't updated our conflict resolution logic yet for some operation to be more fine-grained. We always might fail conflict resolution for conflicting commits which in some way might be resolvable. This just basically means we always fail conflict resolution for replace at the moment.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
allisonport-db: yeah good call thanks

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
allisonport-db: Added a comment to hopefully clarify

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -183,13 +186,16 @@ public TransactionBuilder withDomainMetadataSupported() {
 
   @Override
   public Transaction build(Engine engine) {
+    if (operation == Operation.REPLACE_TABLE) {
+      throw new UnsupportedOperationException("REPLACE TABLE is not yet supported");
```
scottsand-db, line 190: Shouldn't this say: Please use `createReplaceTableTransactionBuilder` instead?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -752,8 +779,31 @@ private void generateClusteringDomainMetadataIfNeeded() {
...
+    return Utils.intoRows(scan.getScanFiles(engine))
+        .map(
+            scanRow -> {
+              AddFile add = new AddFile(scanRow.getStruct(InternalScanFileUtils.ADD_FILE_ORDINAL));
+              return SingleAction.createRemoveFileSingleAction(
+                  add.toRemoveFileRow(true /* dataChange */, Optional.empty()));
+            });
+  }
+
+  /** Whether this transaction is a REPLACE TABLE operation */
```
scottsand-db, line 805: nit: the function name already tells me this

scottsand-db (approved):
Looks geat! Great tests!

PR #4482, opened 2025-04-30
[Kernel]Introduce checksum full post commit hook

huan233usc:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This PR adds a fallback mechanism for checksum generation when the simple CRC information isn't available. Previously, in cases where we couldn't generate an incremental checksum, we would skip checksum generation entirely. With this change, we:
- Try to generate a simple checksum first (fast path)
- If not feasible, add a post commit hook to as a fall back, the fall back will do state construction.

## How was this patch tested?

Extends two write tests

No

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/hook/PostCommitHook.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/hook/PostCommitHook.java
@@ -45,8 +45,9 @@ public interface PostCommitHook {
     /**
      * Writes a checksum file at the version committed by the transaction. This hook is present when
      * CHECKSUM_SIMPLE is missing. It requires constructing table stats via full log replay, which
-     * can be expensive for large tables. Unlike CHECKSUM_SIMPLE, this always performs a full table
-     * state construction rather than incrementally computing from a previous CRC.
+     * can be expensive for large tables, with latency scaling based on log size. Unlike
+     * CHECKSUM_SIMPLE, this always performs a full table state construction rather than
+     * incrementally computing from a previous CRC.
      */
     CHECKSUM_FULL,
 
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/hook/PostCommitHook.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/hook/PostCommitHook.java
@@ -43,8 +43,10 @@ public interface PostCommitHook {
     CHECKSUM_SIMPLE,
 
     /**
-     * Writes a checksum file at the version committed by the transaction. This hook is present
-     * CHECKSUM_SIMPLE is missing, it requires construct table stats via log replay.
+     * Writes a checksum file at the version committed by the transaction. This hook is present when
+     * CHECKSUM_SIMPLE is missing. It requires constructing table stats via full log replay, which
+     * can be expensive for large tables. Unlike CHECKSUM_SIMPLE, this always performs a full table
+     * state construction rather than incrementally computing from a previous CRC.
      */
     CHECKSUM_FULL,
 
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/hook/PostCommitHook.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/hook/PostCommitHook.java
@@ -43,6 +43,12 @@ public interface PostCommitHook {
     CHECKSUM_SIMPLE,
 
     /**
+     * Writes a checksum file at the version committed by the transaction. This hook is present
+     * CHECKSUM_SIMPLE is missing, it requires construct table stats via log replay.
+     */
+    CHECKSUM_FULL,
+
+    /**
      * Writes a log compaction file that merges a range of commit JSON files into a single file.
      * This hook is triggered on a configurable interval (e.g., every 10 commits) and reduces the
      * number of small log files that need to be read when reconstructing the table state, thereby
--- a/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumSimpleComparisonSuite.scala
+++ b/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumSimpleComparisonSuite.scala
@@ -33,7 +33,6 @@ import io.delta.kernel.internal.TableImpl
 import io.delta.kernel.internal.actions.{AddFile, Metadata, SingleAction}
 import io.delta.kernel.internal.checksum.{ChecksumReader, CRCInfo}
 import io.delta.kernel.internal.fs.Path
-import io.delta.kernel.internal.util.FileNames
 import io.delta.kernel.internal.util.Utils.toCloseableIterator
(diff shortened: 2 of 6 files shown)


kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumSimpleComparisonSuite.scala
```diff
@@ -46,10 +47,14 @@ import org.apache.spark.sql.functions.col
...
  * for various table operations.
  */
-class ChecksumSimpleComparisonSuite extends DeltaTableWriteSuiteBase with TestUtils {
+trait ChecksumComparisonSuiteBase extends DeltaTableWriteSuiteBase with TestUtils {
 
   private val PARTITION_COLUMN = "part"
 
+  protected def getPostCommitHookType: PostCommitHookType
+
+  protected def maybeDeleteChecksum(tablePath: String, version: Long): Unit = {}
```
allisonport-db, line 56: can you document when to call this? and why we have it?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumSimpleComparisonSuite.scala
```diff
@@ -65,11 +70,13 @@ class ChecksumSimpleComparisonSuite extends DeltaTableWriteSuiteBase with TestUt
         .forEach(hook => hook.threadSafeInvoke(engine))
       spark.sql(s"CREATE OR REPLACE TABLE delta.`${sparkTablePath}` (id LONG) USING DELTA")
       assertChecksumEquals(engine, sparkTablePath, kernelTablePath, 0)
+      maybeDeleteChecksum(kernelTablePath, 0)
```
allisonport-db, line 73: it's not clear to me why we delete it here? should it actually be before each commit (except the first one)?

allisonport-db: actually should it be part of `commitSparkChangeToKernel`? removes the version prior?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumStatsSuite.scala
```diff
@@ -172,9 +173,40 @@ class ChecksumStatsSuite extends DeltaTableWriteSuiteBase {
       engine: Engine,
       dataActions: CloseableIterable[Row]): TransactionCommitResult = {
     val result = txn.commit(engine, dataActions)
-    result.getPostCommitHooks
-      .stream()
-      .forEach(hook => hook.threadSafeInvoke(engine))
+    val checksumHook = result.getPostCommitHooks.stream().filter(hook =>
+      hook.getType == getPostCommitHookType).findFirst()
+    // When result.getVersion is 0, there will only be CHECKSUM_SIMPLE.
+    assert(result.getVersion == 0 || checksumHook.isPresent)
```
allisonport-db, line 179: maybe do this conditionally on getPostCommitHookType?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumStatsSuite.scala
```diff
@@ -172,9 +173,40 @@ class ChecksumStatsSuite extends DeltaTableWriteSuiteBase {
       engine: Engine,
       dataActions: CloseableIterable[Row]): TransactionCommitResult = {
     val result = txn.commit(engine, dataActions)
-    result.getPostCommitHooks
-      .stream()
-      .forEach(hook => hook.threadSafeInvoke(engine))
+    val checksumHook = result.getPostCommitHooks.stream().filter(hook =>
```
allisonport-db, line 176: also maybe add a check that both are never present?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumStatsSuite.scala
```diff
@@ -172,9 +173,40 @@ class ChecksumStatsSuite extends DeltaTableWriteSuiteBase {
...
+    super.checkCrcCorrect(
+      engine,
+      tablePath,
+      version,
+      expectedFileCount,
+      expectedTableSize,
+      expectedFileSizeHistogram)
+    // Delete the checksum after the validation, so that the subsequence commit
+    // will generate CHECKSUM_FULL hook.
+    deleteChecksumFileForTable(tablePath.stripPrefix("file:"), Seq(version.toInt))
```
allisonport-db, line 210: again not sure, but I think maybe it makes more semantic sense to do this instead before we commit?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/utils/TestUtils.scala
```diff
@@ -793,6 +793,14 @@ trait TestUtils extends Assertions with SQLHelper {
     result
   }
 
+  def executeCrcFull(result: TransactionCommitResult, engine: Engine): TransactionCommitResult = {
```
allisonport-db, line 796: is this used anywhere?

allisonport-db (commented):
Just a few minor comments/questions then LGTM

kernel/kernel-api/src/main/java/io/delta/kernel/hook/PostCommitHook.java
```diff
@@ -42,6 +42,12 @@ enum PostCommitHookType {
      */
     CHECKSUM_SIMPLE,
 
+    /**
+     * Writes a checksum file at the version committed by the transaction. This hook is present
+     * CHECKSUM_SIMPLE is missing, it requires construct table stats via log replay.
+     */
```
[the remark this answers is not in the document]
huan233usc: >It always does a fully log replay?
Yes.

use a prior CRC if it exists & incrementally computes right in CRC full is not yet there. We will build it as an improvement

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/ChecksumSimpleComparisonSuite.scala
huan233usc: With the suggestion of do it before commit, this method is not needed

kernel/kernel-api/src/main/java/io/delta/kernel/hook/PostCommitHook.java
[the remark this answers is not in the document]
allisonport-db: Cool makes sense thanks for clarifying

vkorukanti (approved):

allisonport-db (approved):
LGTM! Tests are easier to understand :)

PR #4483, opened 2025-04-30
[Kernel][InCommitTimestamp][WIP] Support InCommitTimestamp-based time travel

dhruvarya-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java
```diff
@@ -75,32 +80,175 @@ public static Commit getActiveCommitAtTimestamp(
...
-    Commit commit =
-        lastCommitBeforeOrAtTimestamp(commits, timestamp)
-            .orElse(commits.get(0)); // This is only returned if canReturnEarliestCommit (see below)
+    Commit placeholderEarliestCommit = new Commit(earliestVersion, -1L /* timestamp */);
+    Commit ictEnablementCommit = getICTEnablementCommit(latestSnapshot, placeholderEarliestCommit);
+    Commit searchResult;
+    if (ictEnablementCommit.getTimestamp() <= timestamp) {
+      // The target commit is in the ICT range.
+      long latestSnapshotTimestamp = latestSnapshot.getTimestamp(engine);
+      if (latestSnapshotTimestamp <= timestamp) {
```
scovich, line 89: This is an illegal scenario, the query needs to fail.

Reason: It's unsafe to do timestamp as-of for timestamps later than the latest commit. We don't know when (or if) a later commit might land, and also don't know what ICT that commit might have. It's possible a later time travel request for the same timestamp might find a newer commit and we get a non-repeatable read.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java
```diff
@@ -75,32 +80,175 @@ public static Commit getActiveCommitAtTimestamp(
...
+            CommitInfo.getRequiredInCommitTimestampFromFile(
+                engine, logPath, placeholderEarliestCommit.getVersion());
+        searchResult = new Commit(placeholderEarliestCommit.getVersion(), ict);
+      } else {
+        // start non-ICT linear search over [earliestVersion, ictEnablementVersion)
+        List<Commit> commits = getCommits(engine, logPath, earliestVersion);
+        searchResult =
+            lastCommitBeforeOrAtTimestamp(commits, timestamp)
+                .orElse(
+                    commits.get(0)); // This is only returned if canReturnEarliestCommit (see below)
```
scovich, line 135: nit: weird indentation?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java
```diff
@@ -75,32 +80,175 @@ public static Commit getActiveCommitAtTimestamp(
...
+  private static Optional<Long> getInitialCommitVersionForICTSearch(
+      long searchTimestamp, long startCommitVersion, Commit endCommit, Engine engine, Path logPath)
+      throws IOException {
+    long listingStartVersion = Math.max(startCommitVersion, endCommit.getVersion() - 1000);
+    try (CloseableIterator<Commit> commits =
+        listFrom(engine, logPath, listingStartVersion)
+            .takeWhile(
+                fs ->
+                    FileNames.getFileVersionOpt(new Path(fs.getPath())).orElse(-1L)
+                        < endCommit.getVersion())
```
scovich, line 168: nit:
```suggestion
            .takeWhile(fs -> {
                long version = FileNames.getFileVersionOpt(new Path(fs.getPath())).orElse(-1L)
                version < endCommit.getVersion()
            })
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java
```diff
@@ -75,32 +80,175 @@ public static Commit getActiveCommitAtTimestamp(
...
+              searchTimestamp,
+              pivotVersion,
+              searchLeft ? lowerBoundVersion : upperBoundVersion,
+              version -> CommitInfo.getRequiredInCommitTimestampFromFile(engine, logPath, version),
+              searchLeft /* reversed */);
+      lowerBoundVersion = narrowerBounds._1;
+      upperBoundVersion = narrowerBounds._2;
+      if (upperBoundVersion == lowerBoundVersion + 1) {
+        // We have a single commit in the range.
+        return pivotCommit;
```
scovich, line 217: Why is the pivot guaranteed to lie between the newly narrowed bounds?
Are we claiming the invariant `pivotCommit.version() == lowerBoundVersion` holds in this case?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/InCommitTimestampUtils.java
```diff
@@ -101,4 +99,61 @@ private static boolean didCurrentTransactionEnableICT(
...
   }
+
+  public static Tuple2<Long, Long> greatestLowerBound(
+      long target,
+      long lowerBoundInclusive,
+      long upperBoundExclusive,
+      Function<Long, Long> indexToValueMapper) {
+    long start = lowerBoundInclusive;
+    long end = upperBoundExclusive;
+    Tuple2<Long, Long> result = null;
```
scovich, line 110: How do we guarantee we never return `null`? The caller doesn't seem to be checking for that case?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/InCommitTimestampUtils.java
```diff
@@ -101,4 +99,61 @@ private static boolean didCurrentTransactionEnableICT(
...
+      Function<Long, Long> indexToValueMapper,
+      boolean searchFromRightInclusive) {
+    final long iterationDirection = searchFromRightInclusive ? -1 : 1;
+    long lowerBoundIdx = lowerBound;
+    long upperBoundIdx = upperBound;
+    long searchStartEnd = searchFromRightInclusive ? upperBound : lowerBound;
+    long curIdx = searchStartEnd + iterationDirection;
+    for (long i = 1;
+        curIdx > lowerBound && curIdx < upperBound;
+        curIdx = Math.round(searchStartEnd + iterationDirection * (Math.pow(2, ++i) - 1))) {
```
scovich, line 139: This seems really complex. I don't think we need `pow` and `round` here? Can we not start with `i=0` and then:
```java
curIdx += (iterationDirection << i++)
```
(`-1 << 5` is `-32`, just like `1 << 5` is `32`)

scovich (commented):
Approach looks reasonable... but we'll need a lot of good unit testing to validate something this complex (code inspection by reviewers won't be enough)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java
dhruvarya-db: We explicitly handle this scenario later on here: https://github.com/delta-io/delta/pull/4483/files#diff-07b47ed7d50294a001f579299ebd0e7d2328a991babcf6ba23af3c9cacdf8cdcR148. [This ](https://github.com/delta-io/delta/pull/4483/files#diff-07b47ed7d50294a001f579299ebd0e7d2328a991babcf6ba23af3c9cacdf8cdcR89) just ensures that we have a timestamp for the exception message.

dhruvarya-db: @scovich I have addressed all the binary search-specific comments in https://github.com/delta-io/delta/pull/4581/files

dhruvarya-db:
Continued in https://github.com/delta-io/delta/pull/4581

PR #4486, opened 2025-04-30
[Spark] Add invariant checks to DML commands

tomvanbussel:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR adds invariant checks to DELETE, UPDATE, and MERGE to detect bugs in Spark and Delta, and to prevent these DML statements from committing if they suffered from the bug.

These invariants come in two flavors:
- Unreliable checks using the commit stats (i.e. number of rows in the files added and removed) and the SQL metrics. These checks are disabled by default, as Spark can overcount metrics when there are retries.
- Reliable checks based purely on the commit stats. These checks are enabled by default, but cannot detect every occurrence of a bug.

## How was this patch tested?

- Manually ran existing DML tests with all invariant checks enabled by default to confirm that the checks do not cause any issues.
- Manually ran existing DML tests with all invariant checks enabled by default with a bug introduced to confirm that the checks trigger.
- Added `DeltaCommandInvariantsSuite`.

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/commands/MergeIntoCommandBase.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/commands/MergeIntoCommandBase.scala
@@ -128,15 +128,15 @@ trait MergeIntoCommandBase extends LeafRunnableCommand
     }
   }
 
-  /** Whether this merge statement has only MATCHED clauses. */
+  /** Whether this merge statement only has MATCHED clauses. */
   protected def isMatchedOnly: Boolean = notMatchedClauses.isEmpty && matchedClauses.nonEmpty &&
     notMatchedBySourceClauses.isEmpty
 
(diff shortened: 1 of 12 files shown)


larsk-db (approved):

spark/src/main/scala/org/apache/spark/sql/delta/commands/MergeIntoCommandBase.scala
```diff
@@ -136,6 +136,12 @@ trait MergeIntoCommandBase extends LeafRunnableCommand
   protected def isInsertOnly: Boolean = matchedClauses.isEmpty && notMatchedClauses.nonEmpty &&
     notMatchedBySourceClauses.isEmpty
 
+  /** Whether this merge statement only has only delete clauses. */
```
juliuszsompolski, line 139: nit: "only has only"

juliuszsompolski (approved):

PR #4492, opened 2025-05-02
[Kernel]Filter out tombstone when building domain metadata map in log replay rather than at snapshot.getDomainMetadata

huan233usc:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

Follow up of this thread https://github.com/delta-io/delta/pull/4438#discussion_r2064161532.
Every methods that call this method will just do the filter tombstone and there are some redundant logic in TransactionImpl to handle that.

This PR makes LogReplay's domainMetadataMap doesn't contains tombstone for all cases and remove some redundant logic

## How was this patch tested?

Updated domain metadata test.

No


kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java
```diff
@@ -119,9 +120,10 @@ public StructType getSchema() {
 
   @Override
   public Optional<String> getDomainMetadata(String domain) {
-    return Optional.ofNullable(getDomainMetadataMap().get(domain))
-        .filter(dm -> !dm.isRemoved()) // only consider active domain metadatas (not tombstones)
-        .map(DomainMetadata::getConfiguration);
+    Optional<DomainMetadata> domainMetadata =
+        Optional.ofNullable(getDomainMetadataMap().get(domain));
+    domainMetadata.ifPresent(dm -> checkArgument(!dm.isRemoved()));
```
scottsand-db, line 125: do we really want to throw an exception? what are the semantics of this API?

could we, for exampl, just define them to return a non-removed domain metadata?

should callers first check if the domain is removed _before_ calling this API? how would they do that?

allisonport-db: This should return empty if it is removed -- that's the semantics outlined in the interface method docs + the current behavior

allisonport-db: But I think actually that should be an impossible state to reach? So maybe `checkArgument` is just the wrong use here -- `checkState` or something? or maybe we don't need to verify at all

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -687,13 +687,7 @@ public List<DomainMetadata> getComputedDomainMetadatasToCommit() {
...
           // removing and adding a domain with the same identifier in a single txn!
           DomainMetadata domainToRemove = snapshotDomainMetadataMap.get(domainName);
-          if (domainToRemove.isRemoved()) {
-            // If the domain is already removed we throw an error to avoid any inconsistencies or
-            // ambiguity. The snapshot read by the connector is inconsistent with the snapshot
-            // loaded here as the domain to remove no longer exists.
-            throw new DomainDoesNotExistException(
-                dataPath.toString(), domainName, readSnapshot.getVersion());
-          }
+          checkArgument(!domainToRemove.isRemoved());
```
allisonport-db, line 690: same comment here, checkArgument is probably the wrong type of check

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metadatadomain/JsonMetadataDomain.java
```diff
@@ -100,9 +102,10 @@ protected static <T> T fromJsonConfiguration(String json, Class<T> clazz) {
    */
   protected static <T> Optional<T> fromSnapshot(
       SnapshotImpl snapshot, Class<T> clazz, String domainName) {
-    return Optional.ofNullable(snapshot.getDomainMetadataMap().get(domainName))
-        .filter(domainMetadata -> !domainMetadata.isRemoved())
-        .map(domainMetadata -> fromJsonConfiguration(domainMetadata.getConfiguration(), clazz));
+    Optional<DomainMetadata> domainMetadata =
+        Optional.ofNullable(snapshot.getDomainMetadataMap().get(domainName));
+    domainMetadata.ifPresent(dm -> checkArgument(!dm.isRemoved()));
```
allisonport-db, line 107: can this use the public method? we just use `getConfiguration` anyways, this seems like some code duplication  on `Snapshot::getDomainMetadata`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -367,8 +367,7 @@ private Optional<Long> loadLatestTransactionVersion(Engine engine, String applic
 
   /**
    * Loads the domain metadata map, either from CRC info (if available) or from the transaction log.
-   * Note that when loading from CRC info, tombstones (removed domains) are not preserved, while
-   * they are preserved when loading from the transaction log.
+   * Note that when loading from CRC info, tombstones (removed domains) are not preserved.
```
allisonport-db, line 370: No need for this detail, just state this map only contains active domains and any tombstones are not included/are filtered out.

allisonport-db: Can you please also update the docs for `SnapshotImpl::getDomainMetadataMap` and maybe even `LogReplay::getDomainMetadataMap` (the public API that uses this method here)

allisonport-db (commented):
Thanks @huan233usc for doing this!! Just a few minor comments to improve our docs

kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java
huan233usc: this api will only return active domain.

and yes this is an impossible state to reach after the the change. It will be reached only if internal impl has a bug. 
Will change it to checkState.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java
```diff
@@ -155,6 +152,7 @@ public SnapshotReport getSnapshotReport() {
   /**
    * Get the domain metadata map from the log replay, which lazily loads and replays a history of
    * domain metadata actions, resolving them to produce the current state of the domain metadata.
+   * Only active domain metadata should be in this map.
```
allisonport-db, line 155: nit - "are included" or something instead of "should"; the docs should be a promise not an instruction

allisonport-db (approved):
LGTM with 1 minor comment

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -184,6 +184,9 @@ public Optional<Long> getLatestTransactionIdentifier(Engine engine, String appli
     return loadLatestTransactionVersion(engine, applicationId);
   }
 
+  /*
+   * Returns map for all active domain metadata.
+   **/
   public Map<String, DomainMetadata> getDomainMetadataMap() {
```
scottsand-db, line 190: `getActiveDomainMetadataMap` ?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -184,6 +184,9 @@ public Optional<Long> getLatestTransactionIdentifier(Engine engine, String appli
     return loadLatestTransactionVersion(engine, applicationId);
   }
 
+  /*
+   * Returns map for all active domain metadata.
+   **/
```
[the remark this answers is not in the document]
huan233usc: Fixed

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
huan233usc: Updated the naming for variable and methods

scottsand-db (approved):

PR #4498, opened 2025-05-05
[Kernel] Support writing data in REPLACE TABLE transactions (RTAS)

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Supports writing data as part of a REPLACE TABLE transaction. In https://github.com/delta-io/delta/pull/4481 we explicitly blocked this, this PR unblocks it and adds tests for it. Also blocks rowTracking tables for now as we need to spend a little bit more time investigating how to treat rowTracking for `REPLACE` and what the fields in the new adds should look like.

## How was this patch tested?

Updates our unit tests.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
@@ -326,6 +326,11 @@ public class TransactionBuilderImpl implements TransactionBuilder {
         throw new UnsupportedOperationException(
             "REPLACE TABLE is not yet supported with column mapping");
       }
+      if (newProtocol.orElse(baseProtocol).supportsFeature(TableFeatures.ROW_TRACKING_W_FEATURE)) {
+        // Block this for now to be safe, we will return to this in the future
+        throw new UnsupportedOperationException(
+            "REPLACE TABLE is not yet supported on row tracking tables");
+      }
     }
 
     return new TransactionImpl(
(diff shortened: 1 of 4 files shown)


huan233usc (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -326,6 +326,11 @@ protected TransactionImpl buildTransactionInternal(
         throw new UnsupportedOperationException(
             "REPLACE TABLE is not yet supported with column mapping");
       }
+      if (newProtocol.orElse(baseProtocol).supportsFeature(TableFeatures.ROW_TRACKING_W_FEATURE)) {
```
scottsand-db, line 329: I see 5 usages of `newProtocol.orElse(baseProtocol)` and 10 usages of `newMetadata.orElse(baseMetadata)`

In another PR, could you please create `txnProtocol` and `txnMetadata` private helper methods?

scottsand-db (approved):
Looks great! Thanks! Can you please own the TransactionBuilderImpl refactor I've asked for?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
allisonport-db: Hmm both `newProtocol` and `baseProtocol` are local variables I'm not sure the best way to refactor this now (without majorly refactoring how we build the transaction)?

PR #4504, opened 2025-05-07
[Spark] Fix `REPLACE` and `CREATE LIKE` for `CatalogOwned`

xzhseh:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
This PR fixes two subtle but critical parts for CatalogOwned enable table.

- For `CREATE LIKE`, enable CatalogOwned w/ default spark configuration.
  - This is different from the normal behavior/semantic of `CREATE LIKE`, where the source table protocol will take precedence and override any default (table feature) enablement.
- For `REPLACE`, only filter out `CatalogOwnedTableFeature` if the existing table does **_not_** have CatalogOwned enabled - otherwise there would be protocol downgrade issue if we are adding new table feature (e.g., `ClusteringTableFeature`) at the same time.
  - We also filter out any **_unintended_** `InCommitTimestampTableFeature` if the original table does not have ICT enabled, and users are not enabling ICT via explicit overrides or default ICT spark configuration.
    - The unintended `InCommitTimestampTableFeature` may be generated as part of the default CatalogOwned enablement, and we do not support enabling CatalogOwned through `REPLACE`.

## How was this patch tested?
N/A

N/A

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
@@ -805,16 +805,6 @@ trait OptimisticTransactionImpl extends DeltaTransaction
       CoordinatedCommitsUtils.getExplicitICTConfigurations(snapshot.metadata.configuration)
     // Update the metadata.
     updateMetadataForNewTable(metadata)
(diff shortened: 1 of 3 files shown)


yumingxuanguo-db (approved):
LGTM!

PR #4510, opened 2025-05-08
Use Apache Spark 4.0.1-SNAPSHOT

hvanhovell:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
This PR updates the version of Apache Spark  to 4.0.1-SNAPSHOT. This is effectively tracks the version of Spark that will be released as Spark 4 (at some point).

## How was this patch tested?
Existing tests.

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala
@@ -55,7 +55,7 @@ import org.apache.spark.sql.connector.catalog.{DelegatingCatalogExtension, Ident
 import org.apache.spark.sql.connector.catalog.TableCapability._
 import org.apache.spark.sql.connector.catalog.TableChange._
 import org.apache.spark.sql.connector.expressions.{FieldReference, IdentityTransform, Literal, NamedReference, Transform}
-import org.apache.spark.sql.connector.write.{LogicalWriteInfo, V1Write, WriteBuilder}
+import org.apache.spark.sql.connector.write.{LogicalWriteInfo, SupportsTruncate, V1Write, WriteBuilder}
 import org.apache.spark.sql.execution.datasources.{DataSource, PartitioningUtils}
 import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.sources.InsertableRelation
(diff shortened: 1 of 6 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaCatalog.scala
```diff
@@ -629,7 +629,8 @@ class DeltaCatalog extends DelegatingCatalogExtension
     /*
      * WriteBuilder for creating a Delta table.
      */
-    private class DeltaV1WriteBuilder extends WriteBuilder {
+    private class DeltaV1WriteBuilder extends WriteBuilder with SupportsTruncate {
```
hvanhovell, line 632: @cloud-fan this is needed because of https://github.com/apache/spark/commit/5054b41cffc55054918fbabe8981aa9a31f9e2cc. While I do think that the change is correct, I also think we might want to treat a StagingTable slightly different, since it only becomes an actual table once it is committed (truncation just means you don't commit it).

cloud-fan: Looking at the DS v2 doc
```
  /**
   * Configures a write to replace all existing data with data committed in the write.
   *
   * @return this write builder for method chaining
   */
  WriteBuilder truncate();
```

I think `StagingTable` should implement `SupportsTruncate` to indicate whether it's an append or overwrite. Delta uses table existence to decide append or overwrite, which is also OK but not all the v2 sources can follow it.

spark/src/test/scala/org/apache/spark/sql/delta/stats/PartitionLikeDataSkippingSuite.scala
```diff
@@ -434,7 +434,7 @@ class PartitionLikeDataSkippingColumnMappingSuite
   extends PartitionLikeDataSkippingSuiteBase with DeltaColumnMappingEnableIdMode {
   override def runAllTests: Boolean = true
 
-  test("partition-like data skipping with special characters in column names") {
+  ignore("partition-like data skipping with special characters in column names") {
```
cloud-fan, line 437: why is this test ignored?

hvanhovell: I re-enabled it.

Issue #4491, opened 2025-05-02
[Feature Request][Kernel] Support Type Widening Writing in Java Kernel


emkornfield:
#### Which Delta project/connector is this regarding?

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

### Overview

[Type widening](https://github.com/delta-io/delta/issues/2623) has been introduced to Delta to allow changing column types with rewrites.  We should add support for this in Kernel.

### Motivation

This brings Kernel inline with the existing Spark implementation.

### Further details

The scope of this work is to allow type widening for Delta and to also make sure type widening is further constrained when IcebergCompatV1 is enabled.

### Willingness to contribute

The Delta Lake Community encourages new feature contributions. Would you or another member of your organization be willing to contribute an implementation of this feature?

- [x] Yes. I can contribute this feature independently.
- [ ] Yes. I would be willing to contribute this feature with guidance from the Delta Lake community.
- [ ] No. I cannot contribute this feature at this time.

PR #4512, opened 2025-05-08
[WIP][Kernel][RFC] Add type change history to StructField

emkornfield:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

As part of supporting write support for [type widening the kernel](https://github.com/delta-io/delta/issues/4491). We run into a similar design issue as Collation.  Where type changes are stored in [FieldMetadata](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#type-change-metadata) at the nearest parent struct field, but for maps/arrays we need to construct a path separately.  

This change serves as a discussion point on two questions:
1.  Should we add a separate member at all to `StructField` and make use of similar logic in collation to collect it up to the nearest non-map/array parent?
2. What should its representation be.  In this PR I currently use a list which assumes linear transitions.  In theory during typeChange cleanup, one could remove a link, so it might be worth while to create a minimal `TypeChange` class to wrap from/to.  If we want to proceed with the new member on `StructField` I will update the PR to include the new class as well.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/types/StructField.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/types/StructField.java
@@ -60,6 +60,9 @@ public class StructField {
   private final DataType dataType;
   private final boolean nullable;
   private final FieldMetadata metadata;
+  // A list of preivious types that this field held. Only applicable to
+  // fields for BasePrimitiveType. Higher indexed elements are new types.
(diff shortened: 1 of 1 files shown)


emkornfield:
Superceded by https://github.com/delta-io/delta/pull/4519

PR #4520, opened 2025-05-09
[Kernel] Update column mapping and schema evolution code to support usage with replace table

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Update SchemaUtils and ColumnMapping with unit tests in order to support REPLACE TABLE with column mapping + fieldId re-use in PR #2. Specifically this involves the following changes (not necessarily related, but combined in this PR)

1) When a connector provides its own column mapping info in the schema pre-populated we require that it's complete (i.e. fieldId AND physicalName must be present)
2) We add an argument to our schema validation checks `allowNewNonNullableFields`. This is useful in cases where we can be sure the table state has been completely cleared, and thus new non-null fields are valid (like REPLACE).
3) We don't allow adding a new column with a fieldId less than the maxColId. For now, do this proactively for safety. In the future in the case of something like RESTORE in the future we will likely need a config to bypass this check.

## How was this patch tested?

Updates unit tests.

Also, all the changes in this PR are used by https://github.com/delta-io/delta/pull/4520 which adds a lot more E2E tests with multiple schema scenarios.

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/ColumnMapping.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/ColumnMapping.java
@@ -131,14 +131,6 @@ public class ColumnMapping {
     return field.getMetadata().getLong(COLUMN_MAPPING_ID_KEY).intValue();
   }
 
-  public static boolean hasColumnId(StructField field) {
-    return field.getMetadata().contains(COLUMN_MAPPING_ID_KEY);
-  }
-
-  public static boolean hasPhysicalName(StructField field) {
-    return field.getMetadata().contains(COLUMN_MAPPING_PHYSICAL_NAME_KEY);
-  }
-
   public static void verifyColumnMappingChange(
       Map<String, String> oldConfig, Map<String, String> newConfig, boolean isNewTable) {
     ColumnMappingMode oldMappingMode = getColumnMappingMode(oldConfig);
@@ -216,6 +208,14 @@ public class ColumnMapping {
       maxColumnId = findMaxColumnId(field, maxColumnId);
     }
     return maxColumnId;
+  }
+
+  static boolean hasColumnId(StructField field) {
+    return field.getMetadata().contains(COLUMN_MAPPING_ID_KEY);
+  }
+
+  static boolean hasPhysicalName(StructField field) {
+    return field.getMetadata().contains(COLUMN_MAPPING_PHYSICAL_NAME_KEY);
   }
 
   private static int findMaxColumnId(StructField field, int maxColumnId) {
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/ColumnMapping.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/ColumnMapping.java
@@ -131,6 +131,14 @@ public class ColumnMapping {
     return field.getMetadata().getLong(COLUMN_MAPPING_ID_KEY).intValue();
   }
 
+  public static boolean hasColumnId(StructField field) {
+    return field.getMetadata().contains(COLUMN_MAPPING_ID_KEY);
+  }
+
+  public static boolean hasPhysicalName(StructField field) {
+    return field.getMetadata().contains(COLUMN_MAPPING_PHYSICAL_NAME_KEY);
+  }
+
   public static void verifyColumnMappingChange(
       Map<String, String> oldConfig, Map<String, String> newConfig, boolean isNewTable) {
     ColumnMappingMode oldMappingMode = getColumnMappingMode(oldConfig);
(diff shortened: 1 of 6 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -101,18 +101,27 @@ public static void validateSchema(StructType schema, boolean isColumnMappingEnab
    * </ul>
    */
   public static void validateUpdatedSchema(
```
allisonport-db, line 103: Just refactored to not duplicate info across different fields (we can get this info from just the metadata)

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableSchemaEvolutionSuite.scala
```diff
@@ -978,7 +978,12 @@ class DeltaTableSchemaEvolutionSuite extends DeltaTableWriteSuiteBase with Colum
               true),
             false),
           true,
-          fieldMetadataForMapColumn(4, "map", "map", 5, 6))
```
allisonport-db, line 981: This test case was testing the wrong thing before and the change not allowing lower fieldIds highlighted it. We were not using the fieldIds here that were used in the original schema and the entire "map" column was being treated as a new column.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/util/SchemaUtilsSuite.scala
```diff
@@ -766,7 +765,7 @@ class SchemaUtilsSuite extends AnyFunSuite {
             .add(
               "renamed_id",
               IntegerType.INTEGER,
-              false,
+              true,
```
allisonport-db, line 768: don't tighten nullability for existing field, not what this is testing

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/ColumnMapping.java
```diff
@@ -131,6 +131,14 @@ public static int getColumnId(StructField field) {
     return field.getMetadata().getLong(COLUMN_MAPPING_ID_KEY).intValue();
   }
 
+  public static boolean hasColumnId(StructField field) {
+    return field.getMetadata().contains(COLUMN_MAPPING_ID_KEY);
+  }
+
+  public static boolean hasPhysicalName(StructField field) {
+    return field.getMetadata().contains(COLUMN_MAPPING_PHYSICAL_NAME_KEY);
+  }
```
amogh-jahagirdar, line 140: Any reason these need to be made public? Since these are accessors it's fine but don't see anything outside actually using these so it just widens the public surface more than we need to. (Lmk if I'm missing that it's used in another package somewhere or going to be used)

allisonport-db: I think I'm hallucinating... I distinctly remember asking myself the same question and checking and concluding that they do need to be public... but I cannot find that code anymore. So seems like they can go back to being private haha good catch

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/util/SchemaUtilsSuite.scala
```diff
@@ -970,14 +990,61 @@ class SchemaUtilsSuite extends AnyFunSuite {
...
   }
 
+  private val underMaxColIdFieldAdded = Table(
+    ("schemaBefore", "schemaWithUnderMaxFieldIdAdded"),
+    (
+      primitiveSchema,
+      primitiveSchema.add(
+        "required_field",
+        IntegerType.INTEGER,
+        true,
```
amogh-jahagirdar, line 1007: The validation for new fields being greater than the current max field ID is independent of whether the new field is required or not. I think the test is fine as written (and the logic is right in the actual implementation) but maybe want to mix in a case where the added field is not required.

amogh-jahagirdar (approved):
Fundamentally everything looks great, thank you @allisonport-db . Just a couple minor comments

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -101,18 +101,27 @@ public static void validateSchema(StructType schema, boolean isColumnMappingEnab
...
-      StructType currentSchema,
-      StructType newSchema,
-      Set<String> currentPartitionColumns,
+      Metadata currentMetadata,
+      Metadata newMetadata,
       Set<String> clusteringColumnPhysicalNames,
-      Metadata newMetadata) {
+      boolean allowNewNonNullFields) {
```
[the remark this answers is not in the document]
allisonport-db: yeah that works for me & makes sense

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/util/SchemaUtilsSuite.scala
allisonport-db: I think I just need to rename this field :) copied it over

But `true` means `nullable` so it's not actually required. I'll update the column names (and also make the fields below nullable I think I missed the others)

PR #4033, opened 2025-01-09
[PROTOCOL RFC] Variant Shredding

gene-db:
#### Which Delta project/connector is this regarding?

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [X] Other (protocol RFC)

## Description

Adds the protocol changes for the Variant shredding (see https://github.com/delta-io/delta/issues/4032) to the RFC folder.

## How was this patch tested?

N/A

## Does this PR introduce _any_ user-facing changes?

N/A

PR #4521, opened 2025-05-09
[VARIANT] Add VariantShredding-preview table feature in Delta-Spark

harshmotw-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Add the preview table feature for Variant Shredding (see https://github.com/delta-io/delta/pull/4033)

## How was this patch tested?

Unit test

No


spark/src/test/scala-spark-master/org/apache/spark/sql/delta/DeltaVariantShreddingSuite.scala
```diff
@@ -0,0 +1,91 @@
...
+    }
+    assert(DeltaConfigs.ENABLE_VARIANT_SHREDDING.key == "delta.enableVariantShredding")
+  }
+
+  test("Spark can read Delta tables with the shredding table feature") {
+    withTable("tbl") {
+      sql(s"CREATE TABLE tbl USING DELTA " +
+        s"TBLPROPERTIES('${DeltaConfigs.ENABLE_VARIANT_SHREDDING.key}' = 'true') " +
+        s"as select id i, (id + 1)::string s from range(10)")
+      assert(getProtocolForTable("tbl")
```
bart-samwel, line 62: The table has the protocol, but does the written table actually contain shredded variants? The current change will tell Delta Spark that it supports this, but ultimately that should only be true if it *actually* supports it.

harshmotw-db: I've added a test to make sure that Spark can read shredded tables with the feature.

bart-samwel (approved):
LG!

spark/src/test/scala-spark-master/org/apache/spark/sql/delta/DeltaVariantShreddingSuite.scala
```diff
@@ -0,0 +1,174 @@
...
+        }
+      }
+    }
+  }
+
+  test("Set table property to invalid value") {
+    withTable("tbl") {
+      sql("CREATE TABLE tbl(s STRING, i INTEGER) USING DELTA")
+      val deltaLog = DeltaLog.forTable(spark, TableIdentifier("tbl"))
+      assert(!deltaLog.unsafeVolatileSnapshot.protocol
```
scovich, line 156: Please use [DeltaLog.forTableWithSnapshot](https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala#L872-L874) instead of the volatile snapshot:
```suggestion
      val (deltaLog, snapshot) = DeltaLog.forTableWithSnapshot(spark, TableIdentifier("tbl"))
      assert(!snapshot.protocol
```

spark/src/test/scala-spark-master/org/apache/spark/sql/delta/DeltaVariantShreddingSuite.scala
```diff
@@ -0,0 +1,174 @@
...
+            |""".stripMargin)
+
+        sql("CREATE TABLE tbl (i long, v variant) USING DELTA " +
+          s"TBLPROPERTIES ('${DeltaConfigs.ENABLE_VARIANT_SHREDDING.key}' = 'true') " +
+          s"LOCATION '${dir.getAbsolutePath}'")
+        assert(getProtocolForTable("tbl")
+          .readerAndWriterFeatures.contains(VariantShreddingPreviewTableFeature))
+        withSQLConf(SQLConf.VARIANT_WRITE_SHREDDING_ENABLED.key -> true.toString,
+          SQLConf.VARIANT_ALLOW_READING_SHREDDED.key -> true.toString,
+          SQLConf.VARIANT_FORCE_SHREDDING_SCHEMA_FOR_TEST.key -> schema) {
```
scovich, line 132: This test verifies that spark can consume a shredded table, but I don't see any code anywhere that controls spark's decision to shred or not? 

For sure, any Delta table that lacks the variant shredding feature should _block_ shredding even if the spark conf tries to enable it (and a table that lacks the variant feature should block the data type as well). Otherwise, spark could produce invalid Delta tables that sabotage other readers who can't read variant and who (correctly, based on protocol check) thought the table was safe to consume.

Ideally, we should also blow up if we encounter shredded data on a table that doesn't support the feature -- just because we "can" read variant shredding doesn't mean we "should"

scovich (changes requested):

spark/src/test/scala-spark-master/org/apache/spark/sql/delta/DeltaVariantShreddingSuite.scala
harshmotw-db: `VARIANT_FORCE_SHREDDING_SCHEMA_FOR_TEST` is a test-only config as official support for shredded writes has not been enabled in OSS yet. I have added some code which checks the consistency of this config with the shredding table property in the commit stage. This codepath can be extended to future shredded write-related configs as well.

spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala
```diff
@@ -942,6 +943,18 @@ trait OptimisticTransactionImpl extends DeltaTransaction
     }
   }
 
+  // Make sure shredded writes are only performed if the shredding table property was set.
+  private def assertShreddingStateConsistent() = {
+    if (!DeltaConfigs.ENABLE_VARIANT_SHREDDING.fromMetaData(metadata)) {
+      val isVariantShreddingSchemaForced =
+        spark.sessionState.conf
+          .getConfString("spark.sql.variant.forceShreddingSchemaForTest", "").nonEmpty
```
harshmotw-db, line 951: @scovich I'm hard-coding the config here since it appears that it isn't present in many Spark versions being tested in this PR's tests.

scovich (approved):
LGTM

spark/src/main/scala/org/apache/spark/sql/delta/TableFeature.scala
```diff
@@ -713,6 +714,18 @@ object VariantTypeTableFeature extends ReaderWriterFeature(name = "variantType")
...
 
+object VariantShreddingPreviewTableFeature
+    extends ReaderWriterFeature(name = "variantShredding-preview")
+    with FeatureAutomaticallyEnabledByMetadata {
+  override def automaticallyUpdateProtocolOfExistingTables: Boolean = true
+
+  override def metadataRequiresFeatureToBeEnabled(
+      protocol: Protocol, metadata: Metadata, spark: SparkSession): Boolean = {
+    DeltaConfigs.ENABLE_VARIANT_SHREDDING.fromMetaData(metadata) &&
+      SchemaUtils.checkForVariantTypeColumnsRecursively(metadata.schema)
```
harshmotw-db, line 725: @bart-samwel @scovich I have added a schema check for Variant in the enablement condition.

harshmotw-db: cc @richardc-db because you referenced this PR in https://github.com/delta-io/delta/commit/22dd9c47c34b8246b83b1942ad102f9ddc008d3a

PR #4501, opened 2025-05-06
Update IcebergWriterCompatV1 with map struct key requirements

nicklan:
#### Which Delta project/connector is this regarding?

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [X] Other (proposed RFC modification)

## Description
Add a new requirement to the RFC. See changes in the PR.

## How was this patch tested?
N/A

## Does this PR introduce _any_ user-facing changes?

No

PR #4525, opened 2025-05-09
[Kernel][icebergWriterCompatV1] Add a check that map struct keys don't evolve

nicklan:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description
Add a check to forbid changes in struct types that are map keys. See [docs](https://iceberg.apache.org/docs/nightly/spark-ddl/#alter-table-add-column) which state:
> Altering a map 'key' column by adding columns is not allowed. Only map values can be updated.

We have empirically determined that drops also trigger an error, so we forbid any changes to structs that are used as map keys.

This check only triggers if IcebergWriterCompatV1 is enabled on the table. This is required by the spec ([see here](https://github.com/delta-io/delta/pull/4501))

## How was this patch tested?

Unit Tests

No


fivemeyestore:
This is a well-structured codebase. I appreciate the clean separation of concerns.

fivemserverhostings:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -648,4 +651,47 @@ protected static void validateSupportedType(DataType dataType) {
       throw unsupportedDataType(dataType);
     }
   }
+
+  /**
+   * If IcebergWriterCompatV1 is enabled, we need to ensure that map struct keys don't change. This
+   * validates that
+   */
+  private static void validateNoMapStructKeyChanges(
```
allisonport-db, line 659: I wonder if this check should be nested deeper within the schema evolution checks? Then we can use the `SchemaChanges` class @amogh-jahagirdar added to just check the "updatedColumns"

But I'm also okay with this if it's not any easier/better. Maybe it's fewer schema traversals however? Since I think this method alone involves at least 2

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -648,4 +651,47 @@ protected static void validateSupportedType(DataType dataType) {
...
+          // if the map was removed, that's okay, so only check if newMap is not null
+          DataType newDataType = newMap.getDataType();
+          if (newDataType instanceof MapType) {
+            // This check doesn't block changing the whole type, so if it's not a Map, don't bother
+            // to check anything
+            StructField currentStructKey = ((MapType) pathAndField._2.getDataType()).getKeyField();
+            // cast is safe, we filtered for this
+            StructType currentKeyType = (StructType) currentStructKey.getDataType();
+            StructField newStructKey = ((MapType) newDataType).getKeyField();
+            if (!currentKeyType.equals(newStructKey.getDataType())) {
```
allisonport-db, line 686: I think this check doesn't allow re-arranging fields (should it?), or updating the column metadata, changing nullability, etc. What's the expected behavior for these kinds of changes?

Also, I don't think this is necessarily the solution, but consider the difference in `StructType::equivalent` vs `StructType::equals`, but maybe neither of those are the comparison needed.

allisonport-db: Also should you be able to for example, widen the type, of a nested key field?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -648,4 +651,47 @@ protected static void validateSupportedType(DataType dataType) {
...
+              currentSchema,
+              true, /* recurseIntoMapOrArrayElements, need to check maps inside arrays */
+              false, /* stopOnFirstMatch */
+              sf -> {
+                DataType dataType = sf.getDataType();
+                return (dataType instanceof MapType)
+                    && ((MapType) dataType).getKeyField().getDataType() instanceof StructType;
+              });
+      Map<Integer, StructField> updatedFieldsById = fieldsById(newSchema);
+      for (Tuple2<List<String>, StructField> pathAndField : columnPathToMaps) {
```
allisonport-db, line 674: nit - We could probably nest all of this within the `filterRecursively` fx? Then the type checks and casts/code that uses them can be together

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
allisonport-db: I see in the spec RFC it says Iceberg blocks any change not just additions. If this is true I guess this is okay? If you can confirm all the above listed changes should in fact be blocked this seems fine to me

nicklan: Yes, we're aiming to just block _any_ changes for now. We can relax it in the future if we find it to be necessary. I think the basic idea is that changes could change the hash code and mean you have to recompute everything

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
nicklan: Actually that's much nicer, thanks for the suggestion I didn't see the SchemaChanges stuff

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
nicklan: I don't use `filterResusively` anymore as I'm using `SchemaChanges`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -114,8 +114,12 @@ public static void validateUpdatedSchema(
...
     validateSchema(newSchema, true /*columnMappingEnabled*/);
     validatePartitionColumns(newSchema, new ArrayList<>(currentPartitionColumns));
-    validateSchemaEvolution(currentSchema, newSchema, newMetadata, clusteringColumnPhysicalNames);
-    validateNoMapStructKeyChanges(currentSchema, newSchema, newMetadata.getConfiguration());
+    validateSchemaEvolution(
+        currentSchema,
+        newSchema,
+        newMetadata,
+        clusteringColumnPhysicalNames,
+        newMetadata.getConfiguration());
```
allisonport-db, line 122: no need to pass the configuration as we are already passing `newMetadata`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -449,7 +458,10 @@ private static void validateSchemaEvolution(
    * fields
    */
   private static void validateSchemaEvolutionById(
-      StructType currentSchema, StructType newSchema, Set<String> clusteringColumnPhysicalNames) {
+      StructType currentSchema,
+      StructType newSchema,
+      Set<String> clusteringColumnPhysicalNames,
+      Map<String, String> configuration) {
```
allisonport-db, line 464: nit - I think I'd prefer this to be more specific for now, until in the future we maybe need this for more things. Can the arg instead just be `boolean icebergWriterCompatV1Enabled`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
allisonport-db: I was actually picturing nesting this even further in the existing code, I think it can just be an extra check in `validateFieldCompatibility`. There we already have a branch where we have established a field update where we go from map type -> map type (skips a lot of this code here) and then all we need to add there is the additional check if `icebergWriterCompatV1` is enabled & keyType is struct

allisonport-db: I think this should be super concise and simple there just something like
```
    } else if (existingField.getDataType() instanceof MapType
        && newField.getDataType() instanceof MapType) {
      MapType existingMapType = (MapType) existingField.getDataType();
      MapType newMapType = (MapType) newField.getDataType();

      if (icebergWriterCompatV1Enabled && existingMapType.getKeyType() instanceof StructType && newMapType.getKeyType() instanceof StructType) {
               // Require existingMapType.getKeyType() equals newMapType.getKeyType()
      }

      validateFieldCompatibility(existingMapType.getKeyField(), newMapType.getKeyField());
      validateFieldCompatibility(existingMapType.getValueField(), newMapType.getValueField());
    }
```

nicklan: Ohh man, can't believe I missed that those checks already exist. Thanks! That's obviously much better.

allisonport-db (approved):
LGTM

PR #4531, opened 2025-05-12
[SPARK] Handle special characters correctly in validateDataSkippingType

rahulsmahadev:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Fix bug in stats collection code for handling special characters in column names

## How was this patch tested?
- unit tests

no

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala
@@ -47,6 +47,7 @@ import org.apache.spark.sql.catalyst.catalog.CatalogTable
 import org.apache.spark.sql.catalyst.expressions._
 import org.apache.spark.sql.catalyst.parser.{AbstractSqlParser, AstBuilder, ParseException, ParserUtils}
 import org.apache.spark.sql.catalyst.parser.SqlBaseParser.MultipartIdentifierListContext
+import org.apache.spark.sql.catalyst.util.quoteIfNeeded
 import org.apache.spark.sql.functions._
 import org.apache.spark.sql.functions.lit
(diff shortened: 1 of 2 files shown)


chirag-s-db (approved):
thanks!

spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala
```diff
@@ -517,7 +526,15 @@ object StatisticsCollection extends DeltaCommand {
         transformSchema(schema, Some(columnName.head)) {
           case (`prefixPath`, struct @ StructType(_), _) =>
             val columnField = struct(columnName.head)
-            validateDataSkippingType(columnAttribute.name, columnField.dataType, visitedColumns)
+            // We need to figure out if the column is top-level column
+            // or a column inside a struct, we support collecting null count stats
+            // on nested columns part of a struct.
+            val fieldInsideStruct = prefixPath.size > 0
```
rahulsmahadev, line 532: Fix this bug as well, we were not correctly handling nested columns in the root schema

spark/src/test/scala/org/apache/spark/sql/delta/stats/StatsCollectionSuite.scala
```diff
@@ -526,32 +526,6 @@ class StatsCollectionSuite
         )
       }
     }
-
-    test(s"Delta statistic column: invalid data type $invalidType in nested column") {
```
rahulsmahadev, line 530: remove this test, no longer applicable since we shouldn't throw an error https://github.com/delta-io/delta/pull/4531/files#diff-2af271647c445a4476d202c2b2a08fd22e8ea41ad1da85153812f750c52d52abL468

PR #4537, opened 2025-05-13
[PROTOCOL RFC] Add info about variant per-file stats

gene-db:
- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [x] Other (protocol RFC)

## Description

Updates the Variant shredding RFC to specify how per-file stats are encoded.

## How was this patch tested?

N/A

N/A

### DIFF
--- a/protocol_rfcs/variant-shredding.md
+++ b/protocol_rfcs/variant-shredding.md
@@ -43,3 +43,22 @@ When Variant Shredding is supported (`writerFeatures` field of a table's `protoc
 When Variant type is supported (`readerFeatures` field of a table's `protocol` action contains `variantShredding`), readers:
 - must recognize and tolerate a `variant` data type in a Delta schema
 - must recognize and correctly process a parquet schema that is either unshredded (only `metadata` and `value` struct fields) or shredded (`metadata`, `value`, and `typed_value` struct fields) when reading a Variant data type from file.
+
+> ***Update the `Per-file Statistics` section***
+
+> After the description and examples starting from: `Per-column statistics record information for each column in the file and they are encoded, mirroring the schema of the actual data. For example, given the following data schema:`
+
+For a column of type Variant, any structure or type can be encoded for the per-column statistics. In a Parquet file, the statistics for a Variant column must have a schema that follows the [Parquet Variant Shredding specification](https://github.com/apache/parquet-format/blob/master/VariantShredding.md). In JSON format, the per-column statistics for a Variant column can be any JSON structure. For example, for a table with a single Variant column (`varCol: variant`) in its data schema, the JSON per-column statistics could have the form:
+
+```
+"stats": {
+  "minValues": {
+    "varCol": {
+      "a": "min-string"
+      "b": {
+        "c": 10
+      }
+    }
+  }
+}
+```


protocol_rfcs/variant-shredding.md
```diff
@@ -43,3 +43,22 @@ When Variant Shredding is supported (`writerFeatures` field of a table's `protoc
...
 - must recognize and tolerate a `variant` data type in a Delta schema
 - must recognize and correctly process a parquet schema that is either unshredded (only `metadata` and `value` struct fields) or shredded (`metadata`, `value`, and `typed_value` struct fields) when reading a Variant data type from file.
+
+> ***Update the `Per-file Statistics` section***
+
+> After the description and examples starting from: `Per-column statistics record information for each column in the file and they are encoded, mirroring the schema of the actual data. For example, given the following data schema:`
+
+For a column of type Variant, any structure or type can be encoded for the per-column statistics. In a Parquet file, the statistics for a Variant column must have a schema that follows the [Parquet Variant Shredding specification](https://github.com/apache/parquet-format/blob/master/VariantShredding.md). In JSON format, the per-column statistics for a Variant column can be any JSON structure. For example, for a table with a single Variant column (`varCol: variant`) in its data schema, the JSON per-column statistics could have the form:
```
scovich, line 51: Should we specifically state that clients should parse the `varCol` as variant? Or is that obvious?

scovich, line 51: Is it worth stating that it's implementation-dependent/client-choice which variant leaf columns, if any, to emit stats for? For example, Delta-spark uses [specific table properties](https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/DeltaConfig.scala#L569-L603) to control the set of columns for which stats should be collected, including collecting stats only for a subset of leaf columns in a given struct. Although that functionality is not actually part of the Delta spec today, it is rather widespread. As another example, I imagine the shredding schema of a given file would also influence strongly which stats the client can easily produce?

gene-db: Hrmmm, this is an interesting point. Would that mean that the entire min Variant value is whatever we specify for `varCol`, or that the specific paths listed under `varCol` have the individual min values? We will want the "specific paths/leaves" semantics, not the entire-record semantics.

Iceberg stores a collection of paths -> min/max values, so it doesn't provide the min of the entire Variant value, but only specific paths from the Variant value. To be compatible, we want to specify that the Delta variant min/max value is not of the entire Variant value, but only for the specified paths.

anoopj: I think it should be the min value of the path, not the entire variant value. It is not clear to me what the "min" value for an entire variant field even means, semantically.

anoopj: That makes sense. A related question is whether the spec for the stats belong in the  variant spec or the variant shredding spec. FWIW, the Iceberg spec does not technically require shredding for statistics, but it is collected only through Parquet currently, so it needs to be shredded.

anoopj, line 51: Please clarify that:

- Stats should only be written for the leaves, packed into a variant representation.
- Stats should be written only if every row in the data file has the same data type for the leaf.
- The type in `minValues` and `maxValues` have to be the same.

gene-db: Updated.

gene-db: Updated to mention this is implementation specific. Shredding makes it easier to produce these stats, but shredding is not a requirement.

gene-db: Updated to clarify that this for paths, not the entire variant.

gene-db (commented):
@scovich @anoopj Thanks! I updated the PR.

protocol_rfcs/variant-shredding.md
```diff
@@ -43,3 +43,31 @@ When Variant Shredding is supported (`writerFeatures` field of a table's `protoc
...
+- Stats should be written only if every row in the data file has the same data type for the leaf.
+- The structure and types in minValues and maxValues must be the same.
+- The decision about which Variant leaf columns to emit statistics for is determined by the implementation/client.
+
+In a Parquet file, the statistics for a Variant column must have a schema that follows the [Parquet Variant Shredding specification](https://github.com/apache/parquet-format/blob/master/VariantShredding.md).
+In JSON format, the per-column statistics for a Variant column can be any JSON structure.
+For example, for a table with a single Variant column (`varCol: variant`) in its data schema, the JSON per-column statistics could have the form:
+
+```
+"stats": {
```
anoopj, line 63: This would change to encoded variant values, right?

gene-db: Updated with a JSON example, and the human readable version.

However, is the stat supposed to be named `"nullCount"`, or `"nullCounts"`? I see examples of both in the current protocol.

protocol_rfcs/variant-shredding.md
```diff
@@ -43,3 +43,31 @@ When Variant Shredding is supported (`writerFeatures` field of a table's `protoc
...
+> ***Update the `Per-file Statistics` section***
+
+> After the description and examples starting from: `Per-column statistics record information for each column in the file and they are encoded, mirroring the schema of the actual data. For example, given the following data schema:`
+
+For a column of type Variant, any structure or type can be encoded for the per-column statistics.
+- The per-column statistics for a Variant column define the min and max values for paths within the Variant, and not the entire Variant value.
+- Stats should only be written for the leaves, packed into a Variant representation.
+- Stats should be written only if every row in the data file has the same data type for the leaf.
+- The structure and types in minValues and maxValues must be the same.
+- The decision about which Variant leaf columns to emit statistics for is determined by the implementation/client.
```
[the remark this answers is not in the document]
gene-db: Updated.

Does this mean, even for the parquet form of stats, it will just be a struct of 2 binaries, and not a shredded form following the parquet spec?

gene-db (commented):
@scovich Thanks! I updated the PR.

protocol_rfcs/variant-shredding.md
scovich: The [per-file statistics](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#per-file-statistics) section of the spec would be the source of truth, no?
image

Also corroborated by [delta-spark](https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/stats/ReadsMetadataFields.scala#L36-L42)

protocol_rfcs/variant-shredding.md
```diff
@@ -43,3 +43,64 @@ When Variant Shredding is supported (`writerFeatures` field of a table's `protoc
...
+
+> ***Update the `Per-file Statistics` section***
+
+> After the description and examples starting from: `Per-column statistics record information for each column in the file and they are encoded, mirroring the schema of the actual data. For example, given the following data schema:`
+
+### Statistics for Variant Columns
+
+- The `nullCount` stat for a Variant column is a LONG representing the nullcount for the Variant column itself (nullcount stats are not captured for individual paths within the Variant).
+- In JSON, the `minValues` and `maxValues` stats for a Variant column are [binary-encoded](https://github.com/apache/parquet-format/blob/master/VariantEncoding.md) Variant values, with the `metadata` and `value` columns serialized to strings using [z85](https://rfc.zeromq.org/spec/32/) encoding (see example below).
+- In Parquet, the `minValues` and `maxValues` stats for a Variant column are Parquet Variant columns, following the [Parquet Variant specification](https://github.com/apache/parquet-format/blob/master/VariantShredding.md).
```
scovich, line 55: This links to the shredding spec, which is a subset of the variant spec being mentioned?

(it's probably good to say that shredding stats in a checkpoint parquet file is _allowed_ but not required)

scovich (approved):
LGTM

protocol_rfcs/variant-shredding.md
anoopj: +1

anoopj (approved):
LGTM

protocol_rfcs/variant-shredding.md
gene-db: Added a link to both the encoding and shredding specifications, and added a point for allowing, but not requiring, shredding.

gene-db (commented):
Thanks! Updated the PR.

anoopj:
@gene-db Can we update the RFC such that the stored statistics is a variant that is essentially a String->Primitive Value map? The map key would be the JSON path of primitive fields and the values are the bounds of the primitive field in the file?

gene-db:
@anoopj Thanks! I updated the PR with an updated description and example.

protocol_rfcs/variant-shredding.md
```diff
@@ -43,3 +43,56 @@ When Variant Shredding is supported (`writerFeatures` field of a table's `protoc
...
+  }
+  "minValues": {
+    "varCol": {
+      "$['a']" : "min-string",
+      "$['b']['c']" : 1
+    }
+  },
+  "maxValues": {
+    "varCol": {
+      "$['a']" : "variant",
```
anoopj, line 93: Perhaps make it "max-string" to be consistent and avoiding confusion with the type itself?

scovich: Except `"max"` < `"min"` ...

anoopj: Yeah. We can keep "variant" as the value.

gene-db: hahaha, yeah, that is why I had to use `"variant"` for the max.

protocol_rfcs/variant-shredding.md
```diff
@@ -43,3 +43,56 @@ When Variant Shredding is supported (`writerFeatures` field of a table's `protoc
...
+- In Parquet, the `minValues` and `maxValues` stats for a Variant column are Parquet Variant columns, following the Parquet Variant [encoding](https://github.com/apache/parquet-format/blob/master/VariantEncoding.md) and [shredding](https://github.com/apache/parquet-format/blob/master/VariantShredding.md) specifications.
+- In Parquet, the Variant `minValues` and `maxValues` stats are allowed to be shredded, but it is not required.
+- Each path in the Variant `minValues` (`maxValues`) value is the independently computed min (max) stat for the corresponding path in the file's Variant data, so e.g. `minValues.v:a` and `minValues.v:b` could come from different rows in the file.
+- Min/max stats may only be written for primitive (leaf) values, packed into a Variant representation.
+- Min/max stats may only be written for a path if that path has the same data type in every row of the data file.
```
cashmand, line 60: If a path contains VariantNull, is that considered a different data type, or can it be ignored for the purposes of computing min/max? E.g. if my data contains `{"a": 1}` and `{"a": null}`, is the min/max 1 or can it not be written?

gene-db: Yes, the only-VariantNull case be ignored too. If the "leftover" `value` min and max binary value is the VariantNull value, then we know the only untyped values are VariantNull, and we can still compute the min/max for the typed path.

PR #4546, opened 2025-05-14
[kernel] Block updating partitioned table with clustering columns

KaiqiJinWow:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

When the existing table is already a partitioned table, we cannot update it to a clustered table anymore. 
## How was this patch tested?

Unit tests.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
@@ -221,7 +221,8 @@ public class TransactionBuilderImpl implements TransactionBuilder {
     checkArgument(
         isCreateOrReplace || latestSnapshot.isPresent(),
         "Existing snapshot must be provided if not defining a new table definition");
-    latestSnapshot.ifPresent(snapshot -> validateWriteToExistingTable(engine, snapshot));
+    latestSnapshot.ifPresent(
+        snapshot -> validateWriteToExistingTable(engine, snapshot, isCreateOrReplace));
     validateTransactionInputs(engine, isCreateOrReplace);
 
     boolean enablesDomainMetadataSupport =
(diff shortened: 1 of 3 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -507,6 +507,10 @@ protected void validateWriteToExistingTable(Engine engine, SnapshotImpl snapshot
                 txnId.getAppId(), txnId.getVersion(), lastTxnVersion.get());
           }
         });
+    if (clusteringColumns.isPresent()
+        && snapshot.getMetadata().getPartitionColumns().getSize() != 0) {
+      throw new KernelException("Cannot set clustering columns on a partitioned table");
```
allisonport-db, line 512: add this to DeltaErrors? And maybe include more information (existing partition columns, clustering columns set?)

allisonport-db: Also maybe be more specific? if someone does `clusterBy(emptyList())` this should fail right? based on the docs for `withClusteringColumns` this makes it a clustered table

So maybe "cannot enable clustering on a partitioned table" is more accurate?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableClusteringSuite.scala
```diff
@@ -292,6 +292,20 @@ class DeltaTableClusteringSuite extends DeltaTableWriteSuiteBase {
     }
   }
 
+  test("update a partitioned table with clustering columns should fail") {
```
allisonport-db, line 295: test with empty list of clustering cols too?

allisonport-db (changes requested):
Just a few minor comments

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -507,6 +507,10 @@ protected void validateWriteToExistingTable(Engine engine, SnapshotImpl snapshot
                 txnId.getAppId(), txnId.getVersion(), lastTxnVersion.get());
           }
         });
+    if (operation != Operation.REPLACE_TABLE && clusteringColumns.isPresent()
```
allisonport-db, line 510: I think the better way to do this check is to add `isCreateOrReplace` to the args here -- that's what we standardize on throughout the rest of the code

allisonport-db (approved):

PR #4557, opened 2025-05-16
[Spark] Introduce CommittedTransaction to store post-commit txn info

ctring:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
An `OptimisticTransaction` is currently used to keep track of in-progress transaction and also as a container for post-commit information about the transaction. This makes the semantics messy when it comes to post-commit usage. For example, it is possible for a post-commit hook to takes in an in-progress transaction, intentionally or not. A post-commit hook has access to mutable fields and may need to check whether a field is ready to be consumed or not. 

Previously, we introduced `DeltaTransaction` to limit what can be accessed in a post-commit hook. However, that does not solve the in-progress vs. committed semantics problem and requires that if a new transaction implementation is introduced, it has to inherit all the var fields from the trait, so that it can be passed to the post-commit hooks.

This PR is part of a series of PRs to make the refactoring where:
1. [This PR] We introduce a new `CommittedTransaction` class, that is produced by every transaction implementation, to be passed to the post-commit hooks. This class is immutable and contains only the information needed by the post-commit hooks. It is extending `DeltaTransaction` for now so that it can be passed to the post-commit hooks without making too many changes in this PR.
2. We simplify the API of the post-commit hooks' `run` method so that it takes in only the `CommittedTransaction` struct.
3. We remove all var fields from `DeltaTransaction` and make `CommittedTransaction` not extend from `DeltaTransaction`.

## How was this patch tested?

Existing UTs

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/CommittedTransaction.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/CommittedTransaction.scala
@@ -1,0 +1,73 @@
+/*
+ * Copyright (2021) The Delta Lake Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.sql.delta
+
+import scala.collection.mutable
+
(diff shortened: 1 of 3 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/CommittedTransaction.scala
```diff
@@ -0,0 +1,73 @@
...
+    finalNeedsCheckpoint: Boolean,
+    finalPartitionsAddedToOpt: Option[mutable.HashSet[Map[String, String]]],
+    finalIsBlindAppend: Boolean
+)
+  extends DeltaTransaction
+{
+  override def txnExecutionTimeMs: Option[Long] = Some(finalTxnExecutionTimeMs)
+  needsCheckpoint = finalNeedsCheckpoint
+  partitionsAddedToOpt = finalPartitionsAddedToOpt
+  isBlindAppend = finalIsBlindAppend
```
ctring, line 72: To be removed in a future PR where `CommittedTransaction` no longer extend `DeltaTransaction`.

LukasRupprecht (approved):
LGTM!

PR #4559, opened 2025-05-16
Clarify that variant defaults should be null

anoopj:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [X] Protocol

## Description

Clarifies that the default value should be `null` for variant data type

## How was this patch tested?

n/a

No


scovich (approved):
Good catch.

tdas (approved):
thanks for the clarification lgtm.

PR #4563, opened 2025-05-18
[DELTA-4294] Fix time travel by timestamp with In-Commit Timestamps

### LINKED ISSUE #4294: [BUG] [Kernel] Time travel by timestamp does not take into account ICT

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

### Describe the problem

In-Commit Timestamps (ICT) is a table feature that changes who/what owns the actual timestamp of a Delta commit. Previously, it was the timestamp of the commit file status. With ICT, it is moved to the CommitInfo action which must be the first action in a Delta write.

Today, Kernel does not take into account the ICT feature during time travel by timestamp reads. This means that we may request a historical time `T` which incorrectly loads a Delta version `X` (with file status timestamp matching T) instead of loading Delta verion `Y` (which has a CommitInfo indicating that this, in fact, is the right version to load).

meatulbisht:
This PR fixes a bug where time travel by timestamp doesn't account for In-Commit Timestamps (ICT), causing incorrect version loading. The fix ensures that when ICT is enabled, time travel by timestamp uses the ICT values instead of file timestamps.

The changes include:
1. Improved handling of timestamps in DeltaHistoryManager
2. Enhanced tests in InCommitTimestampSuite to verify correct behavior

Fixes #4294

## Which Delta project/connector is this regarding?
- [x] Kernel

## Description
This PR addresses a bug in the Delta Kernel where time travel by timestamp doesn't properly account for In-Commit Timestamps (ICT). When a table has ICT enabled, the system should use those timestamps for time travel operations instead of file timestamps.

The bug was causing incorrect versions to be returned when using time travel by timestamp on tables with ICT enabled.

## How was this patch tested?
Enhanced tests in InCommitTimestampSuite to verify correct behavior, including:
- Edge case where file timestamp < ICT (clock skew)
- Common case where file timestamp > ICT (normal operation)
- Mixed tables with both ICT and non-ICT commits

All tests pass successfully.

## Does this PR introduce any user-facing changes?
No, this is a bug fix that ensures the documented behavior of time travel by timestamp works correctly with In-Commit Timestamps.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java
@@ -16,8 +16,12 @@
 package io.delta.kernel.internal;
 
 import static io.delta.kernel.internal.DeltaErrors.wrapEngineExceptionThrowsIO;
+import static io.delta.kernel.internal.TableConfig.IN_COMMIT_TIMESTAMPS_ENABLED;
+import static io.delta.kernel.internal.TableConfig.IN_COMMIT_TIMESTAMP_ENABLEMENT_VERSION;
 import static io.delta.kernel.internal.fs.Path.getName;
 
+import io.delta.kernel.Snapshot;
+import io.delta.kernel.Table;
 import io.delta.kernel.engine.Engine;
 import io.delta.kernel.exceptions.KernelException;
 import io.delta.kernel.exceptions.TableNotFoundException;
@@ -31,6 +35,7 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.util.*;
+import java.util.Iterator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


raveeram-db:
@dhruvarya-db is this similar to the support you're currently adding?

dhruvarya-db:
Thanks for the ping @raveeram-db. Yes, this addresses the same issue.

Hey @meatulbisht, thanks for raising this PR! Reading the ICT for every listed file until the right one is found could end up being very slow even for very small tables. This PR tries to address some of this slowness by using binary search: https://github.com/delta-io/delta/pull/4581. 
Since you have already worked on this issue, maybe it would be a good idea to use this PR to enhance the test coverage once https://github.com/delta-io/delta/pull/4581 is in?

meatulbisht:
Thanks for the context, @dhruvarya-db @raveeram-db, and for pointing out #4581 —  it is definitely improvement. Once that PR is merged, I’d be happy to help add/enhance the test coverage around this logic. Let’s coordinate then to ensure we cover any edge cases that might come up.
Let me know if there are any specific areas you'd like to focus on for the tests.

PR #4578, opened 2025-05-19
[Spark] Add Python Clone APIs clone, cloneAtVersion, and cloneAtTimestamp

longvu-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Add Python APIs for cloning Delta tables: `clone`, `cloneAtVersion` and `cloneAtTimestamp`.
## How was this patch tested?

Added UTs.

Yes, new Python APIs for cloning.

### DIFF
--- a/python/delta/tests/test_deltatable.py
+++ b/python/delta/tests/test_deltatable.py
@@ -1485,6 +1485,79 @@ class DeltaTableTestsMixin:
         self.assertEqual('all', metrics.zOrderStats.strategyName)
         self.assertEqual(1, metrics.zOrderStats.numOutputCubes)  # one per each affected partition
 
+        def test_clone(self):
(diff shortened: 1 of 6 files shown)


python/delta/tests/test_deltatable.py
```diff
@@ -1500,6 +1500,79 @@ def test_optimize_zorder_by_w_partition_filter(self) -> None:
         self.assertEqual('all', metrics.zOrderStats.strategyName)
         self.assertEqual(1, metrics.zOrderStats.numOutputCubes)  # one per each affected partition
 
+        def test_clone(self) -> None:  # type: ignore[no-untyped-def]
```
longvu-db, line 1503: Pycodestyle wants to write a type description for `self`... I think it's an overkill so we put ignore, note: pycodestyle told me that there should be 2 space instead of one between the code and the comment.

PR #4579, opened 2025-05-19
[Kernel][Catalog-Managed] ParsedLogData

scottsand-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

Add classes: `ParsedLogData`, `ParsedLogCompactionData`, `ParsedCheckpointData`, and `ParsedMultiPartCheckpointData`

These classes aren't used anywhere, yet. That will come in future PRs.

## How was this patch tested?

`ParsedLogDataSuite`

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedLogData.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedLogData.java
@@ -1,0 +1,182 @@
+/*
+ * Copyright (2025) The Delta Lake Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.delta.kernel.internal.files;
+
+import static io.delta.kernel.internal.util.Preconditions.checkArgument;
+
+import io.delta.kernel.data.ColumnarBatch;
+import io.delta.kernel.internal.util.FileNames;
+import io.delta.kernel.utils.FileStatus;
+import java.util.Objects;
+import java.util.Optional;
+
+public class ParsedLogData {
+
+  ///////////////////////////////////////
+  // Static enums, fields, and methods //
+  ///////////////////////////////////////
+
+  public enum ParsedLogCategory {
+    DELTA,
+    LOG_COMPACTION,
+    CHECKPOINT,
+    CHECKSUM
+  }
+
+  public enum ParsedLogType {
+    PUBLISHED_DELTA(ParsedLogCategory.DELTA),
+    RATIFIED_STAGED_DELTA(ParsedLogCategory.DELTA),
+    RATIFIED_INLINE_DELTA(ParsedLogCategory.DELTA),
+    LOG_COMPACTION(ParsedLogCategory.LOG_COMPACTION),
+    CLASSIC_CHECKPOINT(ParsedLogCategory.CHECKPOINT),
+    MULTIPART_CHECKPOINT(ParsedLogCategory.CHECKPOINT),
+    V2_CHECKPOINT(ParsedLogCategory.CHECKPOINT),
+    CHECKSUM(ParsedLogCategory.CHECKSUM);
+
+    public final ParsedLogCategory category;
+
+    ParsedLogType(ParsedLogCategory category) {
+      this.category = category;
+    }
+  }
+
+  public static ParsedLogData forFileStatus(FileStatus fileStatus) {
+    final String path = fileStatus.getPath();
+
+    if (FileNames.isLogCompactionFile(path)) {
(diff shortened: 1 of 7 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/FileNames.java
```diff
@@ -155,6 +160,22 @@ public static long checkpointVersion(String path) {
     return Long.parseLong(name.split("\\.")[0]);
   }
 
+  public static Tuple2<Integer, Integer> multiPartCheckpointPartAndNumParts(Path path) {
+    final String fileName = path.getName();
+    final Matcher matcher = MULTI_PART_CHECKPOINT_FILE_PATTERN.matcher(fileName);
+    if (!matcher.matches()) {
+      throw new IllegalArgumentException(
+          String.format("Path is not a multi-part checkpoint file: %s", fileName));
+    }
```
allisonport-db, line 169: nit this can just use checkArgument but doesn't really matter

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/files/ParsedLogDataSuite.scala
```diff
@@ -0,0 +1,378 @@
...
+
+  test("Throws on both fileStatusOpt and inlineDataOpt empty") {
+    val exMsg = intercept[IllegalArgumentException] {
+      new ParsedLogData(10, ParsedLogType.PUBLISHED_DELTA, Optional.empty(), Optional.empty())
+    }.getMessage
+    assert(exMsg.contains("Exactly one of fileStatusOpt or inlineDataOpt must be present"))
+  }
+
+  ////////////
+  // Deltas //
```
allisonport-db, line 74: should there be a test for `RATIFIED_INLINE_COMMIT`?

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/files/ParsedLogDataSuite.scala
```diff
@@ -0,0 +1,378 @@
...
+    assert(delta1 == delta1)
+    assert(delta1 == delta2)
+    assert(delta1 != delta3)
+  }
+
+  /////////////////
+  // Checkpoints //
+  /////////////////
+
+  test("Inline classic checkpoint") {
```
allisonport-db, line 119: what exactly do these two tests check?

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/files/ParsedLogDataSuite.scala
```diff
@@ -0,0 +1,378 @@
...
+    }.getMessage
+    assert(exMsg2 ==
+      "For MULTIPART_CHECKPOINT, use ParsedMultiPartCheckpointData.forInlineData() instead")
+  }
+
+  test("Correctly parses multi-part checkpoint file") {
+    val chkpt_15_1_3 = multiPartCheckpointFileStatus(15, 1, 3)
+    val parsed = ParsedLogData
+      .forFileStatus(chkpt_15_1_3)
+      .asInstanceOf[ParsedMultiPartCheckpointData]
```
allisonport-db, line 205: Do we expect to do this sort of casting throughout our code? Doesn't need to be addressed in this PR just wondering if that's how we plan to use these other more specific methods or if there is another solution

allisonport-db (commented):
Looks awesome! Mostly just had some clarifying questions

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/files/ParsedLogDataSuite.scala
scottsand-db: Added!

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/files/ParsedLogDataSuite.scala
scottsand-db: I updated the test names to be more helpful.

We want to test that we can create the types we want using the factory methods we want, and that we cannot create certain types using other certain factory methods

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/files/ParsedLogDataSuite.scala
scottsand-db: Casting will only ever need to be done _once_, and it will be done in snapshot manager, when we parse the filestatuses into ParsedLogData and then group by the category

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedLogData.java
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  ///////////////////////////////
+  // Member fields and methods //
+  ///////////////////////////////
+
+  public final long version;
+  public final ParsedLogType type;
+  public final Optional<FileStatus> fileStatusOpt;
+  public final Optional<ColumnarBatch> inlineDataOpt;
```
huan233usc, line 111: Just clarify: the assumption inline data could only contain 1 batch, however, there could be multiple rows in the batch?

scottsand-db: Yup -- it's just 1 columnarBatch (not a list of columnarBatch) -- and yes that columnarBatch can contain many rows

PR #4584, opened 2025-05-20
[Sparl] Ensure structs nested in maps get updated by name in MERGE/UPDATE expressions

larsk-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Within DML commands when we update some field values (in UPDATE or MERGE) we normally resolve struct fields by name, as opposed to the by-position behaviour that struct casts have by default. However, due to a couple of buggy test cases, this was not happening when the struct was nested somewhere inside a map (either on key or value side).

This PR changes the semantics to be more consistent, so that also within map types struct fields are updated by name.

## How was this patch tested?

Existing tests (corrected).

Yes.
Instead of updating struct fields nested within map types by position in MERGE/UPDATE expressions, we now update them by name, like we do for other nested struct fields.


johanl-db (approved):
Looks good!

I'm the one to blame for writing that buggy test...

PR #4594, opened 2025-05-21
[Spark] Stop using deprecated createTable() in CLONE analysis

johanl-db:
## Description
During analysis of CLONE command, we attempt to create a table, using a deprecated `createTable()` method.
This leads to a confusing assertion when running against unitycatalog: https://github.com/unitycatalog/unitycatalog/blob/f626a9efba14821808498e5af6d1344854951294/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala#L127

Updating the call to transform the argument and call into the preferred `createTable()` method

## How was this patch tested?
Existing clone tests

No


larsk-db (approved):
LGTM

PR #4636, opened 2025-05-27
Mark setSchema API as experimental in Javadoc

amogh-jahagirdar:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This change marks TransactionBuilder#setSchema API as experimental in the javadoc until we add the new validation.

## How was this patch tested?

None needed, just a doc change

Marks the API as experimental in the JavaDoc


kernel/kernel-api/src/main/java/io/delta/kernel/TransactionBuilder.java
```diff
@@ -55,6 +55,7 @@ public interface TransactionBuilder {
    *   <li>ToDo: Validate invalid field reorderings
    * </ul>
    *
+   * @experimental This API does not yet perform all validations required for safe schema evolution
```
raveeram-db, line 58: cc: @scottsand-db who's adding some experimental APIs for Catalog-based commits for guidance here.

PR #4637, opened 2025-05-27
Add a PostCommitHook Metrics Report

raveeram-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

Add a PostCommitHook Metrics report that captures metrics like the duration of a single hook execution and execution details. 

## How was this patch tested?

Unit tests. Additional testing is in progress

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/hook/CheckpointHook.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/hook/CheckpointHook.java
@@ -17,12 +17,13 @@ package io.delta.kernel.internal.hook;
 
 import io.delta.kernel.Table;
 import io.delta.kernel.engine.Engine;
-import io.delta.kernel.hook.PostCommitHook;
 import io.delta.kernel.internal.fs.Path;
+import io.delta.kernel.internal.metrics.hook.PostCommitHookExecutionContext;
 import java.io.IOException;
+import java.util.Collections;
 
 /** Write a new checkpoint at the version committed by the txn. */
-public class CheckpointHook implements PostCommitHook {
+public class CheckpointHook extends TimedPostCommitHook {
 
   private final Path tablePath;
   private final long checkpointVersion;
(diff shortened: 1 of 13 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/hook/TimedPostCommitHook.java
```diff
@@ -0,0 +1,95 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(TimedPostCommitHook.class);
+
+  /////////////////////////////////////////////////////////////////////////////////
+  /// Public methods.                                                           ///
+  /////////////////////////////////////////////////////////////////////////////////
+  @Override
+  public final void threadSafeInvoke(Engine engine) throws IOException {
+    PostCommitHookExecutionContext context = createHookExecutionContext();
+    long startNs = System.nanoTime();
+    Exception caughtException = null;
```
nicklan, line 42: Shouldn't we use `Optional`?

kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/engine/LoggingMetricsReporter.java
```diff
@@ -47,6 +44,11 @@ public void report(MetricsReport report) {
         logger.info(
             "TransactionReport = {}",
             MetricsReportSerializers.serializeTransactionReport((TransactionReport) report));
+      } else if (report instanceof PostCommitHookMetricsReport) {
```
nicklan, line 47: Why don't we push all the `instanceof`ing down into a `serializeMetricsReport` method? Probably cc @allisonport-db

kernel/kernel-api/src/main/java/io/delta/kernel/internal/hook/CheckpointHook.java
```diff
@@ -33,12 +34,20 @@ public CheckpointHook(Path tablePath, long checkpointVersion) {
   }
 
   @Override
-  public void threadSafeInvoke(Engine engine) throws IOException {
+  public void executeHook(Engine engine) throws IOException {
```
nicklan, line 37: This is now a breaking change because we're changing the name of this method. I think either:

a) We think engines can still invoke hooks "directly" (i.e. not via `threadSafeInvoke` on `TimedPostCommitHook`), in which case we should keep this method still called `threadSafeInvoke` and public, or

b) We _always_ want to go by `TrimedPostCommitHook`, in which case let's not have this method be public anymore.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/hook/PostCommitHookExecutionContext.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.Collections;
+import java.util.Map;
+import java.util.Optional;
+
+/** Encapsulates context for a post-commit hook execution. */
+public class PostCommitHookExecutionContext {
+
+  private final String tablePath;
+  private final PostCommitHookType hookType;
+  private final Map<String, String> hookExecutionDetails;
```
nicklan, line 31: Alternate strawman, this is just an `Object`. Hooks can put whatever they want there and then cast it back.

nicklan (commented):
Mostly looks good, thanks! Just a few comments

Issue #4573, opened 2025-05-19
[Kernel] [Tracking] Implement catalogManaged table feature in Kernel


scottsand-db:
| Description | PR | Status |
|--------|--------|--------|
| ParsedLogData | TODO | TODO |

PR #4641, opened 2025-05-27
[Kernel][CatalogManaged] Refactor LogReplay to lazy-load P&M [Note: Snapshot still constructed with P&M eagerly]

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/4641/files) to review incremental changes.
- [**stack/kernel_log_replay_refactor**](https://github.com/delta-io/delta/pull/4641) [[Files changed](https://github.com/delta-io/delta/pull/4641/files)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

As part of the `catalogManaged` table support feature (#4573), we want our incoming `ResolvedTableBuilder` (#4614) to be able to receive from the Engine/Catalog the latest Protocol and Metadata for that pinned `ResolvedTable`.

Consequently, we won't always want replay the log to get the latest P & M -- we might just want to use what was given to us!

Thus, this PR makes `LogReplay` load the P & M lazily -- that is, not in the constructor.

This has no impact on existing call sites today:
- `SnapshotManager` -- will immediately then call `logReplay.getProtocol` to create the `SnapshotImpl`
- `TransactionBuilderImpl` -- just creates an `InitialSnapshot` with an `EmptyLogReplay` -- no changes needed

## How was this patch tested?

Just a refactor. Existing UTs

No

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
@@ -53,17 +53,19 @@ import org.slf4j.LoggerFactory;
 
 /**
  * Replays a history of actions, resolving them to produce the current state of the table. The
- * protocol for resolution is as follows: - The most recent {@code AddFile} and accompanying
- * metadata for any `(path, dv id)` tuple wins. - {@code RemoveFile} deletes a corresponding
- * AddFile. A {@code RemoveFile} "corresponds" to the AddFile that matches both the parquet file URI
- * *and* the deletion vector's URI (if any). - The most recent {@code Metadata} wins. - The most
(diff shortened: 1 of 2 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -53,17 +53,19 @@
 
 /**
  * Replays a history of actions, resolving them to produce the current state of the table. The
- * protocol for resolution is as follows: - The most recent {@code AddFile} and accompanying
```
scottsand-db, line 56: just cleaning this up :)

PR #4643, opened 2025-05-28
[Infra] Use Delta Spark 4.0 for DSMP CI job instead of rc6 which is no longer available

allisonport-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Update install

## How was this patch tested?

Existing tests suffice


spark/src/main/scala/org/apache/spark/sql/delta/DeltaSharedExceptions.scala
```diff
@@ -61,8 +61,6 @@ class DeltaIllegalArgumentException(
   override def getMessageParameters: java.util.Map[String, String] = {
     DeltaThrowableHelper.getMessageParameters(errorClass, errorSubClass = null, messageParameters)
   }
-
```
allisonport-db, line 64: This was fixed in RC6 so this hack in Delta is no longer needed

raveeram-db (approved):

vkorukanti (approved):
LGTM pendin test failure fixes

PR #4644, opened 2025-05-28
[Kernel] [CatalogManaged] Load the protocol, metadata, and LogSegment only as needed

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/4644/files) to review incremental changes.
- [**stack/kernel_catalog_managed_3**](https://github.com/delta-io/delta/pull/4644) [[Files changed](https://github.com/delta-io/delta/pull/4644/files)]
  - [stack/kernel_catalog_managed_5_builder_validation](https://github.com/delta-io/delta/pull/4664) [[Files changed](https://github.com/delta-io/delta/pull/4664/files/0b4bdd9a76defb12a692789fbcacc3c4662d7ac4..3089c3e234e6f9ea451f57d25893e39920cf71b8)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Add the ability to load the protocol and metadata -- but do so only as needed. e.g. if we got the P & M in the ResolvedTableBuilder, then we don't need to perform LogReplay at all to load them.

In this PR, we _only_ eagerly provide the P & M. We should add better mock test capabilities to easily provide a P & M to the, for example, json handler in a followup PR.

## How was this patch tested?

New simple UT.

No.

### DIFF
--- a/kernel/kernel-api/src/test/scala/io/delta/kernel/internal/catalogManaged/ResolvedTableBuilderSuite.scala
+++ b/kernel/kernel-api/src/test/scala/io/delta/kernel/internal/catalogManaged/ResolvedTableBuilderSuite.scala
@@ -29,7 +29,7 @@ import org.scalatest.funsuite.AnyFunSuite
 class ResolvedTableBuilderSuite extends AnyFunSuite
     with MockFileSystemClientUtils with ActionUtils {
 
-  test("if P&M are provided LogSegment is not loaded") {
+  test("if P & M are provided then LogSegment is not loaded") {
     val testSchema = new StructType().add("c1", IntegerType.INTEGER)
     val engine = createMockFSListFromEngine(Nil)
 


kernel/kernel-api/src/main/java/io/delta/kernel/internal/table/ResolvedTableFactory.java
```diff
@@ -0,0 +1,145 @@
...
+      throw e;
+    }
+  }
+
+  private ResolvedTableInternalImpl createImpl(Engine engine, SnapshotQueryContext snapshotCtx) {
+    final Lazy<LogSegment> lazyLogSegment = getLazyLogSegment(engine, snapshotCtx);
+
+    final long version = ctx.versionOpt.orElseGet(() -> lazyLogSegment.get().getVersion());
+
+    final Lazy<LogReplay> lazyLogReplay = getLazyLogReplay(engine, lazyLogSegment);
```
scottsand-db: Will make this NOT lazy -- waiting for https://github.com/delta-io/delta/pull/4682 to be merged first

nicklan: Just so I'm clear, the Log _Segment_ will remain lazy (that's expensive since it does i/o). The _replay_ is just logic, so it can be "eager"

scottsand-db: Yup!

LogSegment --> do not compute or construct it until requirement

LogReplay --> we want to refactor it so it doesn't need to LogSegment (loaded) in the constructor -- it can just depend on a `Lazy<LogSegment>`

kernel/kernel-api/src/main/java/io/delta/kernel/ResolvedTableBuilder.java
```diff
@@ -18,18 +18,34 @@
...
   ResolvedTableBuilder atVersion(long version);
 
   // TODO: atTimestamp
 
+  /** For now, only log datas of type {@link ParsedLogType#RATIFIED_STAGED_COMMIT}s are supported */
   ResolvedTableBuilder withLogData(List<ParsedLogData> logData);
 
-  // TODO: withProtocolAndMetadata
+  // TODO: P & M must be public interfaces, not internal classes
+  ResolvedTableBuilder withProtocolAndMetadata(Protocol protocol, Metadata metadata);
```
huan233usc, line 48: Good with this, but there is an alternative: not couple P&M (I know they are mentioned almost altogether but conceptually they are two concept)
e.g.
ResolvedTableBuilder withProtocol
ResolvedTableBuilder withMetadata

huan233usc (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/ResolvedTableBuilder.java
scottsand-db: Yup! We thought about this. I believe @scovich , for example, liked requiring that they be coupled together.

If you can store one of them and give it back to us, then you should be able to store two of them.

And this just keeps things much simpler in Kernel.

PR #4645, opened 2025-05-28
[Spark] Make CommittedTransaction NOT extend from DeltaTransaction

ctring:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Continuation of https://github.com/delta-io/delta/pull/4583

An OptimisticTransaction is currently used to keep track of in-progress transaction and also as a container for post-commit information about the transaction. This makes the semantics messy when it comes to post-commit usage. For example, it is possible for a post-commit hook to takes in an in-progress transaction, intentionally or not. A post-commit hook has access to mutable fields and may need to check whether a field is ready to be consumed or not.

Previously, we introduced DeltaTransaction to limit what can be accessed in a post-commit hook. However, that does not solve the in-progress vs. committed semantics problem and requires that if a new transaction implementation is introduced, it has to inherit all the var fields from the trait, so that it can be passed to the post-commit hooks.

This PR is part of a series of PRs to make the refactoring where:

- We introduce a new CommittedTransaction class, that is produced by every transaction implementation, to be passed to the post-commit hooks.
- We simplify the API of the post-commit hooks' run method so that it takes in only the CommittedTransaction struct.
- [This PR] We remove all var fields from DeltaTransaction and make CommittedTransaction not extend from DeltaTransaction.

## How was this patch tested?
Existing UTs

 No

### DIFF
--- a/hudi/src/main/scala/org/apache/spark/sql/delta/hudi/HudiConverter.scala
+++ b/hudi/src/main/scala/org/apache/spark/sql/delta/hudi/HudiConverter.scala
@@ -226,8 +226,8 @@ class HudiConverter(spark: SparkSession)
 
     // Get the most recently converted delta snapshot, if applicable
     val prevConvertedSnapshotOpt = (lastDeltaVersionConverted, txnOpt) match {
-      case (Some(version), Some(txn)) if version == txn.snapshot.version =>
-        Some(txn.snapshot)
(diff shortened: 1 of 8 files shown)


LukasRupprecht (approved):
LGTM!

PR #4656, opened 2025-05-28
[Kernel][4.0] Block schema evolution via withSchema for 4.0

amogh-jahagirdar:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This change will block schema evolution support in 4.0. There's more work to be done to make sure this API is really safe to prevent invalid field reorderings. We aim to get this out and ready for the next release but for now to prevent users from potentially breaking their table due to lack of this validation we'll block the operation.

## How was this patch tested?
It's blocking a feature so I'm just disabling the tests which used to test this feature

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
@@ -233,6 +233,8 @@ public class TransactionBuilderImpl implements TransactionBuilder {
     latestSnapshot.ifPresent(
         snapshot -> validateWriteToExistingTable(engine, snapshot, isCreateOrReplace));
     validateTransactionInputs(engine, isCreateOrReplace);
+    checkArgument(
+        isCreateOrReplace || !schema.isPresent(), "Schema evolution is not supported currently");
 
     boolean enablesDomainMetadataSupport =
         needDomainMetadataSupport
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
@@ -233,8 +233,10 @@ public class TransactionBuilderImpl implements TransactionBuilder {
     latestSnapshot.ifPresent(
         snapshot -> validateWriteToExistingTable(engine, snapshot, isCreateOrReplace));
     validateTransactionInputs(engine, isCreateOrReplace);
-    checkArgument(
-        isCreateOrReplace || !schema.isPresent(), "Schema evolution is not supported currently");
+    if (!isCreateOrReplace && schema.isPresent()) {
+      throw new UnsupportedOperationException(
+          "Schema can only be provided for new tables. Evolution is not currently supported");
+    }
 
     boolean enablesDomainMetadataSupport =
         needDomainMetadataSupport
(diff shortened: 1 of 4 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -233,6 +233,8 @@ protected TransactionImpl buildTransactionInternal(
     latestSnapshot.ifPresent(
         snapshot -> validateWriteToExistingTable(engine, snapshot, isCreateOrReplace));
     validateTransactionInputs(engine, isCreateOrReplace);
+    checkArgument(
+        isCreateOrReplace || !schema.isPresent(), "Schema evolution is not supported currently");
```
amogh-jahagirdar, line 237: Of course if someone just removes the blanket ignore in the schema evolution tests, we'll see this failure but @allisonport-db let me know if you'd prefer an explicit test for this. It didn't seem worth it since it's just scoped for 4.0 and I didn't want to deviate too much from what exists in master.

allisonport-db: I think I'd prefer just a very simple test to make sure we are indeed blocking it -- I see some tests failing, maybe fixing those to intercept this error can be sufficient?

allisonport-db: Also minor but can we make this error message a little more specific? ie. schema evolution is unsupported -- a schema can only be provided for a new table

allisonport-db: Should we throw `UnsupportedOperationException`?

raveeram-db: +1, a simple unit test would be great

raveeram-db (approved):
Thanks for making this change!

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWritesSuite.scala
```diff
@@ -1157,6 +1157,25 @@ class DeltaTableWritesSuite extends DeltaTableWriteSuiteBase with ParquetSuiteBa
     }
   }
 
+  test("Attempting schema evolution throws unsupported operation") {
+    withTempDirAndEngine { (tablePath, engine) =>
+      val table = Table.forPath(engine, tablePath)
+      val txnBuilder = table.createTransactionBuilder(engine, testEngineInfo, CREATE_TABLE)
+      val txn = txnBuilder.withSchema(engine, testSchema).build(engine)
+      commitTransaction(txn, engine, emptyIterable())
```
allisonport-db, line 1165: nit you can use `createTable` helper

allisonport-db (approved):
Minor comment but LGTM (non blocking)

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWritesSuite.scala
allisonport-db: honestly given this is just in the release branch whatever you can ignore this

PR #4664, opened 2025-05-29
[Kernel] [CatalogManaged] More Builder validation and tests

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/4664/files) to review incremental changes.
- [**stack/kernel_catalog_managed_5_builder_validation**](https://github.com/delta-io/delta/pull/4664) [[Files changed](https://github.com/delta-io/delta/pull/4664/files)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

This PR adds some pretty straight forward builder validation and tests.

## How was this patch tested?

New UTs.

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/table/ResolvedTableBuilderImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/table/ResolvedTableBuilderImpl.java
@@ -25,7 +25,6 @@ import io.delta.kernel.internal.actions.Metadata;
 import io.delta.kernel.internal.actions.Protocol;
 import io.delta.kernel.internal.files.ParsedLogData;
 import io.delta.kernel.internal.files.ParsedLogData.ParsedLogType;
-import io.delta.kernel.internal.lang.ListUtils;
 import io.delta.kernel.internal.tablefeatures.TableFeatures;
 import io.delta.kernel.internal.util.Clock;
 import io.delta.kernel.internal.util.Tuple2;
@@ -116,7 +115,6 @@ public class ResolvedTableBuilderImpl implements ResolvedTableBuilder {
     validateProtocolRead();
     validateLogDatasOnlyRatifiedCommits();
     validateLogDatasSortedContiguous();
-    validateLogDatasEndsWithVersion();
   }
 
   private void validateProtocolAndMetadataOnlyIfVersionProvided() {
@@ -128,19 +126,6 @@ public class ResolvedTableBuilderImpl implements ResolvedTableBuilder {
   private void validateProtocolRead() {
     ctx.protocolAndMetadataOpt.ifPresent(
         x -> TableFeatures.validateKernelCanReadTheTable(x._1, ctx.unresolvedPath));
-  }
-
-  private void validateLogDatasEndsWithVersion() {
-    if (ctx.versionOpt.isPresent() && !ctx.logDatas.isEmpty()) {
-      final long expectedVersion = ctx.versionOpt.get();
-      final long actualVersion = ListUtils.getLast(ctx.logDatas).version;
-
-      checkArgument(
-          actualVersion == expectedVersion,
-          String.format(
-              "Last log data version (%d) must match provided version (%d)",
-              actualVersion, expectedVersion));
-    }
   }
 
   private void validateLogDatasOnlyRatifiedCommits() {
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/table/ResolvedTableBuilderImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/table/ResolvedTableBuilderImpl.java
@@ -90,7 +90,11 @@ public class ResolvedTableBuilderImpl implements ResolvedTableBuilder {
 
   @Override
   public ResolvedTableBuilder withProtocolAndMetadata(Protocol protocol, Metadata metadata) {
-    ctx.protocolAndMetadataOpt = Optional.of(new Tuple2<>(protocol, metadata));
+    ctx.protocolAndMetadataOpt =
+        Optional.of(
+            new Tuple2<>(
+                requireNonNull(protocol, "protocol is null"),
+                requireNonNull(metadata, "metadata is null")));
     return this;
   }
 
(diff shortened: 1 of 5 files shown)


kernel/kernel-api/src/test/scala/io/delta/kernel/internal/catalogManaged/ResolvedTableBuilderSuite.scala
```diff
@@ -18,27 +18,135 @@ package io.delta.kernel.internal.catalogManaged
...
+    val exMsg = intercept[IllegalArgumentException] {
+      TableManager.loadTable(dataPath.toString)
+        .withLogData(parsedRatifiedStagedCommits(Seq(2, 1, 0)).toList.asJava)
+        .build(emptyMockEngine)
+    }.getMessage
+
+    assert(exMsg.contains("Log data must be sorted and contiguous"))
+  }
+
+  test("withProtocolAndMetadata - null protocol throws NullPointerException") {
```
scottsand-db, line 105: `:` to be consistent with other tests

kernel/kernel-api/src/main/java/io/delta/kernel/internal/table/ResolvedTableBuilderImpl.java
```diff
@@ -0,0 +1,173 @@
...
+
+  ////////////////////////////
+  // Private Helper Methods //
+  ////////////////////////////
+
+  private void validateInputOnBuild() {
+    checkArgument(ctx.versionOpt.orElse(0L) >= 0, "version must be >= 0");
+    validateProtocolAndMetadataOnlyIfVersionProvided();
+    validateProtocolRead();
+    validateLogDatasOnlyIfVersionProvided();
```
scottsand-db, line 117: delete this

kernel/kernel-api/src/main/java/io/delta/kernel/internal/table/ResolvedTableBuilderImpl.java
```diff
@@ -106,7 +111,41 @@ public ResolvedTableInternal build(Engine engine) {
...
+
+  private void validateProtocolRead() {
+    ctx.protocolAndMetadataOpt.ifPresent(
+        x -> TableFeatures.validateKernelCanReadTheTable(x._1, ctx.unresolvedPath));
+  }
+
+  private void validateLogDatasOnlyRatifiedCommits() {
+    for (ParsedLogData logData : ctx.logDatas) {
+      checkArgument(
+          logData.type == ParsedLogType.RATIFIED_STAGED_COMMIT,
```
zachschuermann, line 134: do we (kernel) care about the differences between each kind of commit? are {ratified, normal, etc.} all ultimately treated the same or are there special considerations for each?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/table/ResolvedTableBuilderImpl.java
```diff
@@ -106,7 +111,41 @@ public ResolvedTableInternal build(Engine engine) {
 
   private void validateInputOnBuild() {
     checkArgument(ctx.versionOpt.orElse(0L) >= 0, "version must be >= 0");
-    // TODO: logData only ratified staged commits
-    // TODO: logData sorted and contiguous
+    validateProtocolAndMetadataOnlyIfVersionProvided();
+    validateProtocolRead();
+    validateLogDatasOnlyRatifiedCommits();
+    validateLogDatasSortedContiguous();
```
zachschuermann: unsure if it's worth the effort but could we encode some of these invariants into the type system? perhaps we could make bad states unrepresentable? (probably not worth targeting immediately but maybe a follow-up if it seems a plausible improvement)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/table/ResolvedTableBuilderImpl.java
scottsand-db: Good question. I'm explicitly having three distinct types of commits: `PUBLISHED_DELTA`, `RATIFIED_STAGED_COMMIT`,  and `RATIFIED_INLINE_COMMIT`.

Our code should pretty much treat them all the same -- they are all `ParsedLogData` and can either have a `file` or an `inlineData`.

Note that we definitely want "published" (007.json) to look different from staged (`_stage_commits/0007.uuuid.json)`

For now I'm having them be separate types for clarity -- e.g. right now, to have a narrow waist, we only want to support ratified staged commits, not injecting published deltas or ratified inline commits.

And as we add more support later we can easily refactor as needed.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/table/ResolvedTableBuilderImpl.java
scottsand-db: So this is the input from the user. I think our _internal_ type system should correctly represent this -- i.e. we have `LogSegment::getAllFiles` and that is guaranteed to return a contiguous list of files, but I think having a type system for the input is overkill -- it's just a `List`, and we can really quickly and easily verify if it is contiguous or not

kernel/kernel-api/src/main/java/io/delta/kernel/internal/table/ResolvedTableBuilderImpl.java
```diff
@@ -106,7 +111,41 @@ public ResolvedTableInternal build(Engine engine) {
 
   private void validateInputOnBuild() {
     checkArgument(ctx.versionOpt.orElse(0L) >= 0, "version must be >= 0");
-    // TODO: logData only ratified staged commits
-    // TODO: logData sorted and contiguous
+    validateProtocolAndMetadataOnlyIfVersionProvided();
+    validateProtocolRead();
+    validateLogDatasOnlyRatifiedCommits();
```
huan233usc: Can you add a TODO to support other type? Maybe only RatifiedCommits are supported in the short term but this validation now give me a wrong sense that resolved table is only working with the specific type.

scottsand-db: Created https://github.com/delta-io/delta/issues/4765

kernel/kernel-api/src/main/java/io/delta/kernel/internal/table/ResolvedTableBuilderImpl.java
```diff
@@ -106,7 +111,41 @@ public ResolvedTableInternal build(Engine engine) {
...
+
+  private void validateLogDatasOnlyRatifiedCommits() {
+    for (ParsedLogData logData : ctx.logDatas) {
+      checkArgument(
+          logData.type == ParsedLogType.RATIFIED_STAGED_COMMIT,
+          "Only RATIFIED_STAGED_COMMIT log data is supported, but found: " + logData);
+    }
+  }
+
+  private void validateLogDatasSortedContiguous() {
```
raveeram-db: Super nit-picky but do we want to use LogData in a plural context like this? `validateLogDataIsSortedContiguous ` and `validateLogDataContainsOnlyRatifiedCommits` reads more naturally perhaps? No strong opinion either way :)

scottsand-db: SGTM -- will update 👍

PR #4667, opened 2025-05-29
[Kernel] Introduce PeekableIterator class

huan233usc:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

Introduce PeekableIterator class, with this utils, we could check the head of a iterator without advancing.

It is prepared for some optimization like https://github.com/huan233usc/delta/pull/5

## How was this patch tested?

Unit test

No

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/utils/PeekableIterator.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/utils/PeekableIterator.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright (2025) The Delta Lake Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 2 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/utils/PeekableIterator.java
```diff
@@ -0,0 +1,64 @@
...
+ * standard Iterator contract.
+ *
+ * <p>The peek operation does not advance the iterator, so multiple calls to peek() will return the
+ * same element until next() is called.
+ *
+ * @param <T> the type of elements returned by this iterator
+ */
+public class PeekableIterator<T> implements Iterator<T> {
+  private final Iterator<T> iterator;
+  private T peekedItem = null;
```
dhruvarya-db, line 34: Would it be better to use an Optional instead? That would eliminate the need for `hasPeeked`.

huan233usc: I was thinking about optional, one case it is hard to handle is null element. Checking if the use cases in kernel contain null element.

kernel/kernel-api/src/test/scala/io/delta/kernel/PeekableIteratorSuite.scala
```diff
@@ -0,0 +1,60 @@
...
+
+    assert(peekable.peek() === 1)
+    assert(peekable.peek() === 1) // Should return same element
+    assert(peekable.next() === 1)
+    assert(peekable.next() === 2)
+    assert(peekable.peek() === 3)
+    assert(peekable.next() === 3)
+    assert(peekable.hasNext === false)
+  }
+
```
emkornfield, line 39: Maybe add a test case to ensure exceptions are propagated properly from the underlying interator.

PR #4675, opened 2025-05-30
[Spark] Avoid repeating catalog name in Insert tests

ctring:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR cleans up the repeated catalog name in Insert tests

## How was this patch tested?

Existing UTs

No


spark/src/test/scala/org/apache/spark/sql/delta/DeltaInsertIntoMissingColumnSuite.scala
```diff
@@ -35,6 +35,8 @@ class DeltaInsertIntoMissingColumnSuite extends DeltaInsertIntoTest {
     spark.conf.set(SQLConf.ANSI_ENABLED.key, "true")
   }
 
+  val catalogName = "spark_catalog"
```
LukasRupprecht, line 38: Should we move this up into the `DeltaInsertIntoTest` trait so we can directly use it in both this and the schema evolution suite?

PR #4683, opened 2025-05-30
[Spark] Fix DomainMetadata handling for RESTORE/CLONE table

chirag-s-db:
- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
For RESTORE TABLE and the CLONE command, we currently don't mark a DomainMetadata as removed when target snapshot has the DomainMetadata, but the source snapshot does not have that DomainMetadata.

The fix is to mark a DomainMetadata as removed during a RESTORE or CLONE command if:
1. The DomainMetadata is included in the list of DomainMetadata to remove for REPLACE TABLE (thereby excluding the row ID high water mark).
2. The DomainMetadata is not present in the list of DomainMetadata that would anyways be committed by the CLONE command.

## How was this patch tested?
See test changes. 

No.


chirag-s-db:
@rahulsmahadev @zedtang can you take a look at this PR?

PR #4689, opened 2025-06-02
Ensure deletion vector fileId-URL entries are preserved during cache refresh

littlegrasscao:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [x] Other (delta-sharing)

## Description

[During cache registration](https://sourcegraph.prod.databricks-corp.com/databricks-eng/runtime/-/blob/sql/core/src/main/scala/io/delta/sharing/spark/DeltaSharingLogFileSystem.scala?L676-679), we store more fileId-URL entries than what the refresh logic stores. The additional entries come from files associated with deletion vectors. However, [the refresh logic does not include these deletion vector entries](https://sourcegraph.prod.databricks-corp.com/databricks-eng/runtime/-/blob/sql/core/src/main/scala/io/delta/sharing/spark/DeltaSharingUtils.scala?L142-155). As a result, if we query a Delta table in Delta format and the table contains deletion vectors, the refresh operation will drop the corresponding fileId-URL entries for those files.

This causes long running queries seeing file id not found errors.

## How was this patch tested?

Adding new test cases on refresh functions.

No

### DIFF
--- a/sharing/src/test/scala/io/delta/sharing/spark/DeltaSharingUtilsSuite.scala
+++ b/sharing/src/test/scala/io/delta/sharing/spark/DeltaSharingUtilsSuite.scala
@@ -32,48 +32,74 @@ class DeltaSharingUtilsSuite extends SparkFunSuite with SharedSparkContext {
 
   type RefresherFunction = Option[String] => TableRefreshResult
   class SimpleTestDeltaSharingClient extends DeltaSharingClient {
-    def getAddFileStr(): String = {
-        s"""{"file":{
-           |  "id":"add_file_id1",
-           |  "expirationTimestamp":1721350999999,
-           |  "deltaSingleAction":{
-           |    "add":{
-           |      "path":"70/part-00000-df2a2a20.c000.snappy.parquet",
-           |      "partitionValues":{"col-partition":"3"},
-           |      "size":1213,
-           |      "modificationTime":1721350059000,
-           |      "dataChange":true,
-           |      "stats":"{\\"numRecords\\":20,\\"minValues\\":{\\"col-a\\":0},\\"maxValues\\":{\\"col-a\\":19},\\"nullCount\\":{\\"col-a\\":0}}",
-           |      "tags":{"INSERTION_TIME":"1721350059000000"}
-           |    }
-           |  }
-           |}}""".stripMargin
+    def getStatsStr(): String = {
+      """{
+        |  "numRecords": 20,
+        |  "minValues": { "col-a": 0 },
+        |  "maxValues": { "col-a": 19 },
+        |  "nullCount": { "col-a": 0 }
+        |}""".stripMargin
+        .replace("\n", "")
+        .replace(" ", "")
+        .replace("\"", "\\\"")
     }
 
-    def getDeletionVectorStr(): String = {
-      s"""{"file":{
-         |  "id":"add_file_id2",
-         |  "expirationTimestamp":1721350999999,
-         |  "deletionVectorFileId":"dv_file_id",
-         |  "deltaSingleAction":{
-         |    "add":{
-         |      "path":"70/part-00000-df2a2a20.c000.snappy.parquet",
-         |      "partitionValues":{"col-partition":"3"},
-         |      "size":1213,
-         |      "modificationTime":1721350059000,
-         |      "dataChange":true,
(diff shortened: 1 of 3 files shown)


sharing/src/test/scala/io/delta/sharing/spark/DeltaSharingUtilsSuite.scala
```diff
@@ -55,4 +202,55 @@ class DeltaSharingUtilsSuite extends SparkFunSuite with SharedSparkContext {
...
+      None,
+      None,
+      None,
+      None
+    )
+    val idToUrls = func(None).idToUrl
+    assert(idToUrls.size == 3)
+    assert(idToUrls.contains("add_file_id1"))
+    assert(idToUrls.contains("add_file_id2"))
+    assert(idToUrls.contains("dv_file_id"))
```
linzhou-db, line 226: We want to check the url as well? 
Or is it covered in other tests?

sharing/src/test/scala/io/delta/sharing/spark/DeltaSharingUtilsSuite.scala
```diff
@@ -20,9 +20,156 @@ import scala.reflect.ClassTag
...
+         |        "size": 1213,
+         |        "modificationTime": 1721350059000,
+         |        "dataChange": true,
+         |        "stats": "$stats",
+         |        "tags": {
+         |          "INSERTION_TIME": "1721350059000000"
+         |        },
+         |        "deletionVector": {
+         |          "storageType": "p",
+         |          "pathOrInlineDv": ":encoded:placeholder:",
```
linzhou-db: nit but: `:encoded:placeholder:` could be something like `fakeurl` to be easier to read for future?

linzhou-db (approved):
Thanks for fixing this!! 
left two comments

sharing/src/test/scala/io/delta/sharing/spark/DeltaSharingUtilsSuite.scala
littlegrasscao: updated to check urls

sharing/src/test/scala/io/delta/sharing/spark/DeltaSharingUtilsSuite.scala
littlegrasscao: updated to simpler url

sharing/src/test/scala/io/delta/sharing/spark/DeltaSharingUtilsSuite.scala
```diff
@@ -20,9 +20,156 @@ import scala.reflect.ClassTag
...
+      val dv = getDeletionVectorStr()
+      DeltaTableFiles(
+        version = 0L,
+        respondedFormat = DeltaSharingRestClient.RESPONSE_FORMAT_DELTA,
+        lines = Seq(file, dv)
+      )
+    }
+
+    override def getCDFFiles(
+                     table: Table,
```
linzhou-db: nit: should we fix the indent?

sharing/src/test/scala/io/delta/sharing/spark/DeltaSharingUtilsSuite.scala
```diff
@@ -20,9 +20,156 @@ import scala.reflect.ClassTag
...
+    override def getTableVersion(table: Table, startingTimestamp: Option[String] = None): Long = 0
+
+    override def getMetadata(
+                     table: Table,
+                     versionAsOf: Option[Long] = None,
+                     timestampAsOf: Option[String] = None): DeltaTableMetadata
+    = throw new UnsupportedOperationException
+
+    override def getFiles(
+                  table: Table,
```
linzhou-db: and here. and all other places..

sharing/src/test/scala/io/delta/sharing/spark/DeltaSharingUtilsSuite.scala
littlegrasscao: fixed indent

sharing/src/test/scala/io/delta/sharing/spark/DeltaSharingUtilsSuite.scala
littlegrasscao: fixed indent

PR #4712, opened 2025-06-05
[Spark] Fix DomainMetadata handling for REPLACE TABLE concurrent with a transaction that adds a new domain

chirag-s-db:
- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
If the winning transaction adds a new DomainMetadata for the first time in a table's history concurrent to a REPLACE TABLE operation, the REPLACE TABLE will incorrectly not mark this DomainMetadata as removed. This PR fixes this issue by marking any DomainMetadata added by the winning transaction as removed if they appear in the list of DomainMetadata to be removed and the domain does not appear in the DomainMetadata added by the REPLACE TABLE.

## How was this patch tested?
N/A - no DomainMetadata currently exist that fall into this category.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/ConflictChecker.scala
@@ -677,6 +677,11 @@ private[delta] class ConflictChecker(
     }
   }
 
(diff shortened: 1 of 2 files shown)


chirag-s-db:
@rahulsmahadev can you take a look?

rahulsmahadev (approved):

PR #4714, opened 2025-06-05
[PROTOCOL RFC][FOLLOW-UP] Variant Shredding

harshmotw-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR adds a requirement to the [Variant Shredding Protocol RFC](https://github.com/delta-io/delta/pull/4033) for writers to add the `variantShredding` table feature when enabling the `enableVariantShredding` table property. This way, writers can assume that the enablement of the table property implies the existence of the table feature, and can safely write shredded data.

## How was this patch tested?

N/A

N/A

### DIFF
--- a/protocol_rfcs/variant-shredding.md
+++ b/protocol_rfcs/variant-shredding.md
@@ -36,7 +36,8 @@ typed_value | * | (optional) This can be any Parquet type, representing the data
 ## Writer Requirements for Variant Shredding
 
 When Variant Shredding is supported (`writerFeatures` field of a table's `protocol` action contains `variantShredding`), writers:
-- must respect the `delta.enableVariantShredding` table property configuration. If `delta.enableVariantShredding=false`, a column of type `variant` must not be written as a shredded Variant, but as an unshredded Variant. If `delta.enableVariantShredding=true`, the writer can choose to shred a Variant column according to the [Parquet Variant Shredding specification](https://github.com/apache/parquet-format/blob/master/VariantShredding.md)
+- must respect the `delta.enableVariantShredding` table property configuration. If `delta.enableVariantShredding=false`, a column of type `variant` must not be written as a shredded Variant, but as an unshredded Variant. If `delta.enableVariantShredding=true`, the writer can choose to shred a Variant column according to the [Parquet Variant Shredding specification](https://github.com/apache/parquet-format/blob/master/VariantShredding.md).
+- must ensure the `variantType` table feature is present in the table protocol's `writerFeatures`  and `readerFeatures` when enabling the `delta.enableVariantShredding` table property, i.e. setting `delta.enableVariantShredding=true`.
 
 ## Reader Requirements for Variant Shredding
 


harshmotw-db:
@bart-samwel @gene-db Can you look at this PR, thanks!

protocol_rfcs/variant-shredding.md
```diff
@@ -36,7 +36,8 @@ typed_value | * | (optional) This can be any Parquet type, representing the data
...
-- must respect the `delta.enableVariantShredding` table property configuration. If `delta.enableVariantShredding=false`, a column of type `variant` must not be written as a shredded Variant, but as an unshredded Variant. If `delta.enableVariantShredding=true`, the writer can choose to shred a Variant column according to the [Parquet Variant Shredding specification](https://github.com/apache/parquet-format/blob/master/VariantShredding.md)
+- must respect the `delta.enableVariantShredding` table property configuration. If `delta.enableVariantShredding=false`, a column of type `variant` must not be written as a shredded Variant, but as an unshredded Variant. If `delta.enableVariantShredding=true`, the writer can choose to shred a Variant column according to the [Parquet Variant Shredding specification](https://github.com/apache/parquet-format/blob/master/VariantShredding.md).
+- must ensure the `variantType` table feature is present in the table protocol's `writerFeatures`  and `readerFeatures` when enabling the `delta.enableVariantShredding` table property, i.e. setting `delta.enableVariantShredding=true`.
```
gene-db: I think this should be:
```suggestion
- must ensure the `variantShredding` table feature is present in the table protocol's `writerFeatures`  and `readerFeatures` when enabling the `delta.enableVariantShredding` table property, i.e. setting `delta.enableVariantShredding=true`.
```

gene-db (commented):
Thanks!

protocol_rfcs/variant-shredding.md
```diff
@@ -36,7 +36,8 @@ typed_value | * | (optional) This can be any Parquet type, representing the data
 ## Writer Requirements for Variant Shredding
 
 When Variant Shredding is supported (`writerFeatures` field of a table's `protocol` action contains `variantShredding`), writers:
-- must respect the `delta.enableVariantShredding` table property configuration. If `delta.enableVariantShredding=false`, a column of type `variant` must not be written as a shredded Variant, but as an unshredded Variant. If `delta.enableVariantShredding=true`, the writer can choose to shred a Variant column according to the [Parquet Variant Shredding specification](https://github.com/apache/parquet-format/blob/master/VariantShredding.md)
+- must respect the `delta.enableVariantShredding` table property configuration. If `delta.enableVariantShredding=false`, a column of type `variant` must not be written as a shredded Variant, but as an unshredded Variant. If `delta.enableVariantShredding=true`, the writer can choose to shred a Variant column according to the [Parquet Variant Shredding specification](https://github.com/apache/parquet-format/blob/master/VariantShredding.md).
```
bart-samwel, line 39: Either this, or require that the setting must be exactly "true" or "false" or unset. I think requiring that might lead to more annoyances.

```suggestion
- must respect the `delta.enableVariantShredding` table property configuration. If `delta.enableVariantShredding` is not present or if it is set to any value other than `true`, a column of type `variant` must not be written as a shredded Variant, but as an unshredded Variant. If `delta.enableVariantShredding=true`, the writer can choose to shred a Variant column according to the [Parquet Variant Shredding specification](https://github.com/apache/parquet-format/blob/master/VariantShredding.md).
```

PR #4717, opened 2025-06-06
[Kernel] [Infra] Only trigger Kernel tests when relevant folders / files / dependencies have changed

scottsand-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Only trigger Kernel tests when relevant folders / files / dependencies have changed

## How was this patch tested?

CI

No

### DIFF
--- a/.github/workflows/kernel_docs.yaml
+++ b/.github/workflows/kernel_docs.yaml
@@ -27,6 +27,18 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+      - uses: technote-space/get-diff-action@v4
+        id: git-diff
+        with:
+          PATTERNS: |
+            kernel/**
+            project/**
+            connectors/golden-tables/**
+            storage/**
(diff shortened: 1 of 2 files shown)


.github/workflows/kernel_docs.yaml
```diff
@@ -27,6 +27,18 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+      - uses: technote-space/get-diff-action@v4
```
allisonport-db, line 30: I am pretty sure you need to add this to the next steps where the tests actually run
```
        if: steps.git-diff.outputs.diff
```
Did you test this out?

allisonport-db: Also you can match ! files. Not sure how you feel but maybe listing the stuff we don't want to run the tests on is in some ways safer than everything we want to run tests on. But if the file listing isn't changing that much I think it's also fine

PR #4719, opened 2025-06-06
[Kernel][Metric]Record the time taken loading CRC during snapshot construction

huan233usc:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description
Add a metric to record the time taken loading CRC during snapshot construction, also refactor the test to make helper method with less arguments

No


raveeram-db (approved):
LGTM!

PR #4727, opened 2025-06-09
[POC] [Kernel] Support for V3 deletion vectors

murali-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This POC PR introduces support for Deletion Vectors in the Kernel as a part of Iceberg V3 support. 
In order to enable this feature, we make the following updates:
- IcebergCompatV3 and WriterCompatV2 are introduced
  - This includes `TableConfig, DeltaErrors, TransactionBuilder, Generating Actions, ValidatorAndUpdator, TableFeatures, TypeWideningChecker`
- AddFile now supports DeletionVectorDescriptor
  - Also includes DataFileStatus
  - We only use the DeletionVectorDescriptor class internally (for the IRC service) so we do not refactor this class out of the `kernel.internal` namespace
  ---
#### Pending Questions
- ParquetFileWriter: how do we include DeletionVectorDescriptor?
- TransactionBuilderImpl: is the current way I have added CompatV3 and WriterCompatV2 correct?
- TypeWideningChecker: Check if the type change is supported by Iceberg V3 schema evolution.
- Any additional tests?

---

## How was this patch tested?

Unit tests.

Yes, introduces IcebergCompatV3, IcebergWriterCompatV2.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/icebergcompat/IcebergCompatV3MetadataValidatorAndUpdater.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/icebergcompat/IcebergCompatV3MetadataValidatorAndUpdater.java
@@ -129,7 +129,8 @@ public class IcebergCompatV3MetadataValidatorAndUpdater
                       || dataType instanceof TimestampNTZType
                       || dataType instanceof ArrayType
                       || dataType instanceof MapType
-                      || dataType instanceof StructType);
+                      || dataType instanceof StructType
+                      || dataType instanceof VariantType);
                 });
 
         if (!matches.isEmpty()) {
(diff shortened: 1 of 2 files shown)


kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/internal/parquet/ParquetFileWriter.java
```diff
@@ -460,11 +460,13 @@ private DataFileStatus constructDataFileStatus(String path, StructType dataSchem
...
       }
 
+      // TODO (LC-7744): Add support for DeletionVectorDescriptor
       return new DataFileStatus(
           resolvedPath,
           fileStatus.getSize(),
           fileStatus.getModificationTime(),
-          Optional.ofNullable(stats));
+          Optional.ofNullable(stats),
+Optional.empty());
```
murali-db, line 469: TODO: How do we include Deletion Vector Descriptor here?

PR #4743, opened 2025-06-12
[Kernel] use Schema iterable for assigning IDs

emkornfield:
## 🥞 Stacked PR

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/ColumnMapping.java
```diff
@@ -343,18 +343,19 @@ private static Optional<Metadata> assignColumnIdAndPhysicalName(
...
-                  isNewTable,
-                  useColumnIdForPhysicalName));
-    }
+    SchemaIterable schemaIterable = new SchemaIterable(oldSchema);
+    schemaIterable
+        .mutableStream()
+        .filter(SchemaIterable.MutableSchemaElement::isStructField)
+        .forEach(
+            mutableElement ->
+                mutableElement.updateField(
```
emkornfield: We could avoid some object churn here by modifying assignColumnIdAndPhysicalNameNameToField to only return a value if there is in fact a change.

emkornfield: done.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableSchemaEvolutionSuite.scala
```diff
@@ -360,7 +360,7 @@ class DeltaTableSchemaEvolutionSuite extends DeltaTableWriteSuiteBase with Colum
 
       val structType = table.getLatestSnapshot(engine).getSchema
       assertColumnMapping(structType.get("a"), 1)
-      assertColumnMapping(structType.get("map"), 2)
```
emkornfield, line 363: this highlights one potential problem if the new old schema was created with an old library, and then modified with the new library it would change the order of fields.

This might be a more general problem though, if schema evolution of inserting a new element in a struct prior to an existing element.  If the new schema did not have type IDs this would be treated as a rename.

This might be an argument for putting in the effort to keep the column ordering consistent (but really what we want is intent based APIs).  For now we can maybe table this PR.

emkornfield:
closing this for now.

PR #4747, opened 2025-06-12
[INFRA] Remove DSMP CI job as it no longer works (Spark 4.0 is no longer compatible with spark master)

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [X] Other (fill in here)

## Description

We enabled the DSMP job using the Spark 4.0 release (in the RC stages) a while back knowing that eventually there would be a breaking change between Spark master and Spark 4.0. This has already happened and it is no longer possible to run/test Spark master python code with pyspark. See https://github.com/delta-io/delta/pull/4643 where we try to upgrade the version to the released 4.0 but run into class not found errors.

For now, disable this broken job as there is no way to enable it. The only way to unblock this in the future is to have a pyspark nightly snapshot.

## How was this patch tested?

N/A


raveeram-db:
Should we consider archiving this (e.g. move to a different folder called _archived_jobs) in case we want to re-enable it in the future without having to search for this commit @allisonport-db ?

allisonport-db:
@raveeram-db I think we barely touch these jobs, it should be fine to go back and find it via the git history for `.github/workflows`. It also is in the 4.0 branch still which is probably what we would use for reference anyways

raveeram-db:
> @raveeram-db I think we barely touch these jobs, it should be fine to go back and find it via the git history for `.github/workflows`. It also is in the 4.0 branch still which is probably what we would use for reference anyways

makes sense!

PR #4759, opened 2025-06-13
Fix wrong schema evolution inside structs in MERGE

mkroll-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR fixes query failure on MERGE if an `UPDATE` is used together with an `INSERT` with a struct. In case that the `UPDATE` does not modify the struct, but the `INSERT` the schema evolution is not correctly propagated resulting in an error.

### Example
```sql
DROP TABLE IF EXISTS simple_target_table;

CREATE OR REPLACE TABLE simple_target_table(
  key INT, value STRUCT<a: INT>
) USING DELTA;
SET spark.databricks.delta.schema.autoMerge.enabled = true;

INSERT INTO simple_target_table VALUES (1, struct(1, 2));

CREATE OR REPLACE TEMPORARY VIEW source_data AS
SELECT key, struct(a, b) AS value
FROM VALUES (1, 11, 12), (2, 21, 22) AS t (id, a, b);

-- This will fail. Even though UPDATE doesn't touch the nested_data column
MERGE INTO
  simple_target_table AS target
USING
  source_data AS source
ON
  target.id = source.id
WHEN MATCHED THEN UPDATE SET key = 0 -- only updating a simple field, not the struct
WHEN NOT MATCHED THEN INSERT *;
```

Produces following error instead of completing successfully:
```
[DATATYPE_MISMATCH.DATA_DIFF_TYPES] Cannot resolve "CASE WHEN (_source_row_present_ IS NULL) THEN named
```

## How was this patch tested?

Added tests to verify the correct behavior. 

A flag (default: on) is introduced to disable the change `spark.databricks.delta.schemaEvolution.merge.fixNestedStructAlignment`.


johanl-db (approved):
LGTM, thanks for fixing this

PR #4762, opened 2025-06-13
prepare to sync delta commit version with iceberg snapshot sequence number

lzlfred:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

prepare the conversion transaction to sync delta commit version with iceberg snapshot sequence number


spark/src/test/scala/org/apache/spark/sql/delta/UniversalFormatSuiteBase.scala
```diff
@@ -135,6 +135,26 @@ trait UniversalFormatSuiteBase extends IcebergCompatUtilsBase
     }
   }
 
+  test("create new UniForm table via clone") {
```
lzlfred, line 138: unintended. will remove

PR #4768, opened 2025-06-14
done

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/4768/files) to review incremental changes.
- [**stack/kernel_catalog_managed_prioritize_catalog_commmits**](https://github.com/delta-io/delta/pull/4768) [[Files changed](https://github.com/delta-io/delta/pull/4768/files)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
@@ -27,6 +27,7 @@ import io.delta.kernel.internal.*;
 import io.delta.kernel.internal.annotation.VisibleForTesting;
 import io.delta.kernel.internal.checkpoints.*;
 import io.delta.kernel.internal.files.ParsedLogData;
+import io.delta.kernel.internal.files.ParsedLogData.ParsedLogCategory;
 import io.delta.kernel.internal.files.ParsedLogData.ParsedLogType;
 import io.delta.kernel.internal.fs.Path;
 import io.delta.kernel.internal.lang.Lazy;
@@ -39,7 +40,9 @@ import io.delta.kernel.internal.util.Tuple2;
 import io.delta.kernel.utils.FileStatus;
 import java.util.*;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -613,6 +604,46 @@ public LogSegment getLogSegmentForVersion(
   // getLogSegment utils //
   /////////////////////////
 
+  private List<ParsedLogData> getAllDeltasAfterCheckpointWithCatalogPriority(
+      List<ParsedLogData> allPublishedDeltas,
+      List<ParsedLogData> parsedLogDatas,
```
huan233usc, line 623: Do we enforce that parsedLogDatas is continuous, and sorted by version?

scottsand-db: We do -- `ResolvedTableBuilderImpl` does this 👍

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -613,6 +604,46 @@ public LogSegment getLogSegmentForVersion(
   // getLogSegment utils //
   /////////////////////////
 
+  private List<ParsedLogData> getAllDeltasAfterCheckpointWithCatalogPriority(
```
nicklan, line 621: Doc comment please

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -613,6 +604,46 @@ public LogSegment getLogSegmentForVersion(
...
+    }
+
+    final long firstRatified = allRatifiedCommitsAfterCheckpoint.get(0).version;
+    final long lastRatified = ListUtils.getLast(allRatifiedCommitsAfterCheckpoint).version;
+
+    return Stream.of(
+            allPublishedDeltasAfterCheckpoint.stream().filter(x -> x.version < firstRatified),
+            allRatifiedCommitsAfterCheckpoint.stream(),
+            allPublishedDeltasAfterCheckpoint.stream().filter(x -> x.version > lastRatified))
+        .flatMap(Function.identity())
```
nicklan, line 657: weird that there's no `flatten` function

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
huan233usc: Maybe consider dropping a note here as it is not obvious if I just read this piece of code?

huan233usc (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -491,11 +466,16 @@ public LogSegment getLogSegmentForVersion(
...
     }
 
+    final Lazy<Optional<ParsedLogData>> lazyDeltaAtCheckpointVersionOpt =
+        new Lazy<>(
+            () ->
+                allPublishedDeltas.stream()
+                    .filter(x -> x.version == latestCompleteCheckpointVersion)
+                    .findFirst());
+
     // Check that, for a checkpoint at version N, there's a delta file at N, too.
```
nicklan, line 476: What error case are we checking for here? I know this check already existed, but is it a common failure mode that there's a checkpoint without the delta file?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -596,13 +580,20 @@ public LogSegment getLogSegmentForVersion(
...
-        ListUtils.getLast(listedDeltaFileStatuses).getModificationTime();
+        allDeltasAfterCheckpoint.isEmpty()
+            ? lazyDeltaAtCheckpointVersionOpt.get().get().getFileStatus().getModificationTime()
+            : ListUtils.getLast(allDeltasAfterCheckpoint).getFileStatus().getModificationTime();
 
     return new LogSegment(
         logPath,
         newVersion,
-        allDeltasAfterCheckpoint,
+        allDeltasAfterCheckpoint.stream()
```
nicklan, line 594: do we want to have `LogSegment` take `ParsedLogData` rather that `FileStatus` at some point?

scottsand-db: Yup! We want to use `ParsedLogData` in our end-to-end Cloud List --> LogSegment Construction --> LogReplay loop.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
scottsand-db: I think we are just checking if the table is corrupt -- if there is a 10.checkpoint but not 10.json, that is bad.

I don't think this is a common scenario at all.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
scottsand-db: Yup :/ Would love to upgrade off of Java 8!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
nicklan: Okay yeah. This comes up somewhat often because there's an infinite # of ways that a table can be corrupt, and we can't catch them all, so should we jam all that we can think of in, or just catch the ones that are "easy" to check.

This is borderline for me since we are doing some extra work to check this, but it's not much, so likely okay.

Final question though, would it not be a valid for vacuum to remove the `.json` file and leave the checkpoint?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -257,8 +260,8 @@ public LogSegment getLogSegmentForVersion(Engine engine, Optional<Long> versionT
   }
 
   /**
-   * Recall: Right now, we are only supporting sorted and contiguous log datas of type {@link
-   * ParsedLogType#RATIFIED_STAGED_COMMIT}s.
+   * [delta-io/delta#4765]: Right now, we are only supporting sorted and contiguous log datas of
```
nicklan, line 263: Can we make an issue (maybe a good first one) to split this up into multiple functions. It's _way_ too long as is :)

nicklan (approved):
lgtm!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
scottsand-db: 👍  https://github.com/delta-io/delta/issues/4787

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
scottsand-db: > Final question though, would it not be a valid for vacuum to remove the .json file and leave the checkpoint?

What is and is not valid for VACUUM to clean up is not super crisply defined in the protocol.

For example, if there is 10.checkpoint but no 10.json, can you load version 10? What is the timestamp of that version?

I'd love to sync and clarify this expected behavior in the protocol? cc @scovich

PR #4780, opened 2025-06-17
[Kernel] [CatalogManaged] [UC] Simple UCCatalogManagedClient

scottsand-db:
## 🥞 Stacked PR

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Create a new jar `delta-unity` and add a `UCCatalogManagedClient` inside of it, along with a simple test.

More advanced tests will come soon (in future PRs).

## How was this patch tested?

New simple unit test.

No.

### DIFF
--- a/unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
+++ b/unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
@@ -56,6 +56,8 @@ public class UCCatalogManagedClient {
     this.ucClient = Objects.requireNonNull(ucClient, "ucClient is null");
   }
 
+  // TODO: [delta-io/delta#4817] loadTable API that takes in a UC TableInfo object
+
   /**
    * Loads a Kernel {@link ResolvedTable} at a specific version.
    *
@@ -70,7 +72,7 @@ public class UCCatalogManagedClient {
     Objects.requireNonNull(tablePath, "tablePath is null");
     checkArgument(version >= 0, "version must be non-negative");
 
-    logger.info("[{}] Loading table at version {}", ucTableId, version);
+    logger.info("[{}] Resolving table at version {}", ucTableId, version);
     final GetCommitsResponse response = getRatifiedCommitsFromUC(ucTableId, tablePath, version);
     validateLoadTableVersionExists(ucTableId, version, response.getLatestTableVersion());
     final List<ParsedLogData> logData =
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedLogData.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedLogData.java
@@ -29,7 +29,7 @@ import java.util.Optional;
  * Represents a Delta Log "file" - the actual content may be materialized to disk (with a file
  * status) or stored inline (as a columnar batch).
  */
-// TODO: Move this to be a public API
+// TODO: [delta-io/delta#4816] Move this to be a public API
 public class ParsedLogData {
 
   ///////////////////////////////////////


unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
```diff
@@ -0,0 +1,128 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * Client for interacting with Unity Catalog (UC) catalog-managed Delta tables.
+ *
+ * @see UCClient
+ * @see ResolvedTable
+ */
+@Experimental
+public class UCCatalogManagedClient {
```
scottsand-db, line 50: An alternatieve idea is that we treat this entity, instead, like just a UC version of the `TableManager`, like so:

```java
public class UnityCatalogTableManager {
  private UCCatalogManager() {}

  public static ResolvedTable loadTable(
      UCClient ucClient,
      Engine engine,
      String ucTableId,
      String tablePath,
      long version) {
    ....
  }
}
```

To be clear, though: This `UnityCatalogTableManager` should  **not** expose a `loadTable: ResolvedTableBuilder` because: That API is for table managers like this! This table manager must invoke those APIs correctly, e.g. to pass in the parsed log data. This table manager must **not** pass on such an API to the engine.

raveeram-db: What would be the advantages of an additional `TableManager` here Scott? Just a Util/static factory class?

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
```diff
@@ -0,0 +1,128 @@
...
+   * Loads a Kernel {@link ResolvedTable} at a specific version.
+   *
+   * @param engine The Delta Kernel {@link Engine} to use for loading the table.
+   * @param version The version of the table to load.
+   * @throws IOException if there's an error during the commit process, such as network issues.
+   * @throws UCCommitCoordinatorException if there's an error specific to Unity Catalog such as the
+   *     table not being found.
+   */
+  public ResolvedTable loadTable(Engine engine, long version)
+      throws UCCommitCoordinatorException, IOException {
```
raveeram-db: Do we need to pass through the UCCommitCoordinatorException as a checked exception or could that be handled here?

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
```diff
@@ -0,0 +1,128 @@
...
+    Objects.requireNonNull(engine, "engine is null");
+
+    logger.info("[{}] Loading table at version {}", ucTableId, version);
+
+    final GetCommitsResponse response = getRatifiedCommitsFromUC(version);
```
raveeram-db: Style Nit: Is this spacing intentional?

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
```diff
@@ -0,0 +1,128 @@
...
+
+    validateLoadTableVersionExists(version, response.getLatestTableVersion());
+
+    final List<ParsedLogData> logData =
```
raveeram-db, line 78: Could we also log the time taken for these operations? (Perhaps makes sense to have a consolidated timer-wrapper lambda we could reuse at some point)

We can consider a metrics report at a later time, but just logging would be great

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
```diff
@@ -0,0 +1,128 @@
...
+          String.format(
+              "[%s] Cannot load table version %s as the latest version ratified by UC is %s",
+              ucTableId, tableVersionToLoad, maxRatifiedVersion));
+    }
+  }
+
+  @VisibleForTesting
+  static List<ParsedLogData> getSortedKernelLogDataFromRatifiedCommits(
+      String ucTableId, List<Commit> commits) {
+    final List<ParsedLogData> result =
```
raveeram-db, line 131: time taken here would be useful as well

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
```diff
@@ -0,0 +1,128 @@
...
+                  final io.delta.kernel.utils.FileStatus kernelFS =
+                      FileStatus.of(
+                          hadoopFS.getPath().toString(),
+                          hadoopFS.getLen(),
+                          hadoopFS.getModificationTime());
+                  return ParsedLogData.forFileStatus(kernelFS);
+                })
+            .collect(Collectors.toList());
+
+    logger.debug("[{}] Created ParsedLogData from ratified commits: {}", ucTableId, result);
```
raveeram-db, line 144: Do we want to log the full result or the count?

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
```diff
@@ -0,0 +1,128 @@
...
+
+  @VisibleForTesting
+  static List<ParsedLogData> getSortedKernelLogDataFromRatifiedCommits(
+      String ucTableId, List<Commit> commits) {
+    final List<ParsedLogData> result =
+        commits.stream()
+            .sorted(Comparator.comparingLong(Commit::getVersion))
+            .map(
+                commit -> {
+                  final org.apache.hadoop.fs.FileStatus hadoopFS = commit.getFileStatus();
```
raveeram-db: Could this be factored out into a Utils/helper?

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
scottsand-db: Great question, Rakesh. So far, this seems to come down to a matter of syntax:

```java
Note: param a is really the UC client (shared for all tables for that metastore)
Note: params b, c, d, e, f is per-table params

// Option A: Class
val ucCatalogManagedClient = new UCCatalogManagedClient(a, b, c);
val resolvedTable = ucCatalogManagedClient.loadTable(d, e, f);

// Option B: Static function
val resolvedTable = UcCatalogManager.loadTable(a, b, c, d, e, f);

// Option C
val ucCatalogManagedClient = new UCCatalogManagedClient(a);
val resolvedTable = ucCatalogManagedClient.loadTable(b, c, d, e, f);
```

edit: After chatting with Nick, I am leaning towards option C

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
scottsand-db: the new line spacing? yup -- but I am not married to it. Let me know if there's a different style you'd prefer, I'm easy and open to feedback here!

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
scottsand-db: 👍

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
raveeram-db: Thanks Scott, Option C makes sense to me as well. Option B can be added later if necessary based on usage patterns as well.

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
raveeram-db: Ah just to align with the archetype we have in other methods, could remove the new lines perhaps?

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
scottsand-db: This is a `logger.debug` -- IMO it's fine to log the full result for a debug? What do you think?

raveeram-db: Yeah should perhaps be ok. Just that it could get large - but makes sense in a debug context and we can revisit if needed 👍

raveeram-db (approved):
LGTM!

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
```diff
@@ -0,0 +1,169 @@
...
+
+  /**
+   * Loads a Kernel {@link ResolvedTable} at a specific version.
+   *
+   * @param engine The Delta Kernel {@link Engine} to use for loading the table.
+   * @param ucTableId The Unity Catalog table ID, which is a unique identifier for the table in UC.
+   * @param tablePath The path to the Delta table in the underlying storage system.
+   * @param version The version of the table to load.
+   */
+  public ResolvedTable loadTable(Engine engine, String ucTableId, String tablePath, long version) {
```
zachschuermann, line 69: is needing tablePath currently just a limitation of UC? shouldn't we only need table id?

scottsand-db: Yeah it's a requirement of UC, today. I've followed up to ask why we need it

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
```diff
@@ -0,0 +1,169 @@
...
+    Objects.requireNonNull(engine, "engine is null");
+    Objects.requireNonNull(ucTableId, "ucTableId is null");
+    Objects.requireNonNull(tablePath, "tablePath is null");
+    checkArgument(version >= 0, "version must be non-negative");
+
+    logger.info("[{}] Loading table at version {}", ucTableId, version);
```
nicklan: nit:
```suggestion
    logger.info("[{}] Resolving table at version {}", ucTableId, version);
```

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
```diff
@@ -0,0 +1,169 @@
...
+package io.delta.unity;
+
+import static io.delta.kernel.internal.util.Preconditions.checkArgument;
+
+import io.delta.kernel.ResolvedTable;
+import io.delta.kernel.TableManager;
+import io.delta.kernel.annotation.Experimental;
+import io.delta.kernel.engine.Engine;
+import io.delta.kernel.internal.annotation.VisibleForTesting;
+import io.delta.kernel.internal.files.ParsedLogData;
```
nicklan, line 26: Importing `internal` classes to me indicates we've not got the right API yet. Or do we expect that an OSS catalog would have to import and use these classes too? 

Basically it seems like we should strive to have a catalog client that only uses truly "public" kernel interfaces.

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
```diff
@@ -0,0 +1,169 @@
...
+      String ucTableId, String tablePath, long version) {
+    logger.info(
+        "[{}] Invoking the UCClient to get ratified commits at version {}", ucTableId, version);
+
+    return timeOperation(
+        "UCClient.getCommits",
+        ucTableId,
+        () -> {
+          try {
+            return ucClient.getCommits(
```
nicklan, line 101: So this seems like we've gotten away with just one rpc. Do we just rely on this call returning everything we need? Or do we expect the engine will have made the other call earlier (e.g. to get the table path)?

If the latter, maybe we should have a method that wraps that up for them?

nicklan (commented):
nice, looks great. left a few comments

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
scottsand-db: > Importing internal classes to me indicates we've not got the right API yet.

Agreed! `Public ParsedLogData API` is on the backlog of refactors for this CCv2 work (https://github.com/delta-io/delta/issues/4573). Want me to implement than now / sooner rather than later?

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
scottsand-db: > Or do we expect the engine will have made the other call earlier (e.g. to get the table path)?

This. The Engine uses the official UC client SDK to resolve the table name -> path, get the UC table ID, get the credentials, etc, and then must use that information to correctly invoke the `UCCatalogManagedClient.loadTable`

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
scottsand-db: We have the same issue with our Protocol and Metadata in our ResolvedTableBuilder -- I also need to make a public versions of those internal classes, but have just been putting it off to focus on velocity.

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
scottsand-db: > If the latter, maybe we should have a method that wraps that up for them?

Sounds good. This will just take in the https://github.com/unitycatalog/unitycatalog/blob/main/api/Models/TableInfo.md

Can I add that in a followup? Or would you like me to include it now?

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
nicklan: No, this is fine for now, but let's add a TODO in the code with a link to the issue

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
nicklan: SG thanks, whatever is easier for you is fine

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
scottsand-db: Created https://github.com/delta-io/delta/issues/4816

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
scottsand-db: Created https://github.com/delta-io/delta/issues/4817

nicklan (approved):
lgtm, thanks!

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
scottsand-db: Btw to come back to this -- to me the problem here is that UCCommitCoordinatorException is a checked exception -- IMO it should be an unchecked exception, wdyt?

scottsand-db: And btw we also need to eventually deprecate the UCCommitCoordinatorException anyways and use UCCatalogManagedException

PR #4785, opened 2025-06-18
Add a flag enabling more general widening during schema evolution

tmarzec:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR introduces a new internal flag `spark.databricks.delta.typeWidening.allowAutomaticWidening` with values `never`, `always`, `same_family_type` (default). It controls the extent of type widening during automatic schema evolution and is only applicable for tables with `delta.enableTypeWidening` enabled.

By default, automatic schema evolution enables column widening between two types of the same type family (like `short` -> `int` -> `long` or `float` -> `double`). Setting the new flag to always allows widenings across type families (like `int` -> `double`). Using the value `same_family_type` follows the default behavior of type widening during schema evolution, and using `never` disables any widenings.

In addition to updating the permitted schema widenings, this PR adds event logging for situations where the current type widening behavior during schema evolution prevents a widening that would be allowed with a different configuration. The logged settings include the value of `uniformIcebergCompatibleOnly` and the new `allowAutomaticWidening` flag.

## How was this patch tested?

With new tests covering `always` and `never`, and existing tests covering `same_family_type`.

Using the new flag impacts the type widening behavior during schema evolution. But the flag itself is internal and not exposed to users.

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/delta/typewidening/TypeWideningInsertSchemaEvolutionBasicSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/typewidening/TypeWideningInsertSchemaEvolutionBasicSuite.scala
@@ -53,7 +53,8 @@ class TypeWideningInsertSchemaEvolutionBasicSuite
  */
 trait TypeWideningInsertSchemaEvolutionBasicTests
   extends DeltaInsertIntoTest
-  with TypeWideningTestCases {
+  with TypeWideningTestCases
+  with DeltaExcludedBySparkVersionTestMixinShims {
   self: QueryTest with TypeWideningTestMixin with DeltaDMLTestUtils =>
 
   import testImplicits._
@@ -105,52 +106,45 @@ trait TypeWideningInsertSchemaEvolutionBasicTests
     }
   }
 
-  for {
-    testCase <- alterTableOnlySupportedTestCases
-  } {
-    test(s"INSERT - logs for missed opportunity for conversion " +
-      s"${testCase.fromType.sql} -> ${testCase.toType.sql}") {
-      append(testCase.initialValuesDF)
+  testSparkMasterOnly(s"INSERT - logs for missed opportunity for conversion") {
+    val testCase = alterTableOnlySupportedTestCases.head
 
-      val events = Log4jUsageLogger.track {
-        withSQLConf(SQLConf.STORE_ASSIGNMENT_POLICY.key -> StoreAssignmentPolicy.LEGACY.toString) {
-          testCase.additionalValuesDF
-          .write
-          .format("delta")
-          .mode("append")
-          .option("mergeSchema", "true")
-          .insertInto(s"delta.`$tempPath`")
-          }
+    append(testCase.initialValuesDF)
+
+    val events = Log4jUsageLogger.track {
+      withSQLConf(SQLConf.STORE_ASSIGNMENT_POLICY.key -> StoreAssignmentPolicy.LEGACY.toString) {
+        testCase.additionalValuesDF
+        .write
+        .format("delta")
+        .mode("append")
+        .option("mergeSchema", "true")
+        .insertInto(s"delta.`$tempPath`")
       }
-
-      assert(readDeltaTable(tempPath).schema("value").dataType === testCase.fromType)
-      assert(events.exists(event => event.metric == "tahoeEvent" &&
-        event.tags.get("opType") == Option("delta.typeWidening.missedAutomaticWidening")))
     }
+
+    assert(readDeltaTable(tempPath).schema("value").dataType === testCase.fromType)
+    assert(events.exists(event => event.metric == "tahoeEvent" &&
+      event.tags.get("opType") == Option("delta.typeWidening.missedAutomaticWidening")))
   }
 
-  for {
-    testCase <- supportedTestCases
-  } {
-    test(s"INSERT - no logs for lack of missed opportunity for conversion " +
-      s"${testCase.fromType.sql} -> ${testCase.toType.sql}") {
-      append(testCase.initialValuesDF)
+  test(s"INSERT - no logs for lack of missed opportunity for conversion") {
+    val testCase = supportedTestCases.head
+    append(testCase.initialValuesDF)
 
-      val events = Log4jUsageLogger.track {
-        withSQLConf(SQLConf.STORE_ASSIGNMENT_POLICY.key -> StoreAssignmentPolicy.LEGACY.toString) {
-          testCase.additionalValuesDF
-            .write
-            .format("delta")
-            .mode("append")
-            .option("mergeSchema", "true")
-            .insertInto(s"delta.`$tempPath`")
-        }
+    val events = Log4jUsageLogger.track {
+      withSQLConf(SQLConf.STORE_ASSIGNMENT_POLICY.key -> StoreAssignmentPolicy.LEGACY.toString) {
+        testCase.additionalValuesDF
+        .write
+        .format("delta")
+        .mode("append")
+        .option("mergeSchema", "true")
+        .insertInto(s"delta.`$tempPath`")
       }
-
-      assert(readDeltaTable(tempPath).schema("value").dataType === testCase.toType)
-      assert(!events.exists(event => event.metric == "tahoeEvent" &&
-        event.tags.get("opType") == Option("delta.typeWidening.missedAutomaticWidening")))
     }
+
+    assert(readDeltaTable(tempPath).schema("value").dataType === testCase.toType)
+    assert(!events.exists(event => event.metric == "tahoeEvent" &&
+      event.tags.get("opType") == Option("delta.typeWidening.missedAutomaticWidening")))
   }
 
   for {


spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMode.scala
```diff
@@ -54,19 +56,60 @@ object TypeWideningMode {
...
    * Type changes that are eligible to be applied automatically during schema evolution are allowed.
    * Can be restricted to only type changes supported by Iceberg.
    */
-  case class TypeEvolution(uniformIcebergCompatibleOnly: Boolean) extends TypeWideningMode {
-    override def getWidenedType(fromType: AtomicType, toType: AtomicType): Option[AtomicType] =
-        Option.when(TypeWidening.isTypeChangeSupportedForSchemaEvolution(
-          fromType = fromType, toType = toType, uniformIcebergCompatibleOnly))(toType)
+  case class TypeEvolution(
+                            uniformIcebergCompatibleOnly: Boolean,
+                            allowAutomaticWidening: String) extends TypeWideningMode {
```
johanl-db: Prefer using the enum type instead of string to prevent callers from passing a wrong value:
```suggestion
                            allowAutomaticWidening: AllowAutomaticWideningMode.Value) extends TypeWideningMode {
```

You can add a helper method to make it easier for callers to build a value from the spark conf, e.g.:
```
  object AllowAutomaticWideningMode extends Enumeration {
    val NEVER, SAME_FAMILY_TYPE, ALWAYS = Value

    def fromConf(conf: SQLConf): Value =
      withName(conf.get(DeltaSqlConf.DELTA_ALLOW_AUTOMATIC_WIDENING))
  }
```
 DeletionVectorCachePolicy.withName(conf.get(entry.key, entry.defaultValueString))

spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMode.scala
```diff
@@ -54,19 +56,60 @@ object TypeWideningMode {
...
 
   /** All supported type widening changes are allowed. */
-  case object AllTypeWidening extends TypeWideningMode {
+  case object AllTypeWidening
+    extends TypeWideningMode {
     override def getWidenedType(fromType: AtomicType, toType: AtomicType): Option[AtomicType] =
-      Option.when(TypeWidening.isTypeChangeSupported(fromType = fromType, toType = toType))(toType)
+      Option.when(TypeWidening.isTypeChangeSupported(
+        fromType = fromType,
+        toType = toType))(toType)
```
johanl-db: Undo formatting change

spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMode.scala
```diff
@@ -54,19 +56,60 @@ object TypeWideningMode {
...
 
   /**
    * Type changes that are eligible to be applied automatically during schema evolution are allowed.
    * Can be restricted to only type changes supported by Iceberg.
    */
-  case class TypeEvolution(uniformIcebergCompatibleOnly: Boolean) extends TypeWideningMode {
-    override def getWidenedType(fromType: AtomicType, toType: AtomicType): Option[AtomicType] =
-        Option.when(TypeWidening.isTypeChangeSupportedForSchemaEvolution(
-          fromType = fromType, toType = toType, uniformIcebergCompatibleOnly))(toType)
+  case class TypeEvolution(
```
johanl-db, line 73: Update the comment above to mention the set of eligible type changes can be configured with `allowAutomaticTypeWidening`

spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMode.scala
```diff
@@ -54,19 +56,60 @@ object TypeWideningMode {
...
+                            uniformIcebergCompatibleOnly: Boolean,
+                            allowAutomaticWidening: String) extends TypeWideningMode {
+    private def logMissedWidening(fromType: AtomicType, toType: AtomicType): Unit = {
+      // Check if widening is possible under the least restricting conditions.
+      if (TypeEvolution(uniformIcebergCompatibleOnly = false,
+        allowAutomaticWidening = AllowAutomaticWideningMode.ALWAYS.toString)
+        .canWiden(fromType, toType)) {
```
johanl-db: nit: indent is off, slightly more readable as:
```suggestion
      val allowAllTypeEvolution = TypeEvolution(
        uniformIcebergCompatibleOnly = false, 
        allowAutomaticWidening = AllowAutomaticWideningMode.ALWAYS
      )
      if (allowAllTypeEvolution.canWiden(fromType, toType)) {
```

spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMode.scala
```diff
@@ -54,19 +56,60 @@ object TypeWideningMode {
...
+
+    override def getWidenedType(fromType: AtomicType, toType: AtomicType): Option[AtomicType] = {
+      val result = Option.when(canWiden(fromType, toType))(toType)
+
+      if (result.isEmpty) {
+        logMissedWidening(fromType = fromType, toType = toType)
+      }
+
+      result
+    }
```
johanl-db: nit: Move this up in TypeEvolution, it's the most important part

spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMode.scala
```diff
@@ -54,19 +56,60 @@ object TypeWideningMode {
...
+      } else if (allowAutomaticWidening == AllowAutomaticWideningMode.SAME_FAMILY_TYPE.toString) {
+        TypeWidening.isTypeChangeSupportedForSchemaEvolution(
+          fromType = fromType, toType = toType, uniformIcebergCompatibleOnly)
+      } else {
+        false
+      }
+    }
+
+    override def getWidenedType(fromType: AtomicType, toType: AtomicType): Option[AtomicType] = {
+      val result = Option.when(canWiden(fromType, toType))(toType)
```
johanl-db: (I'm really nitpicking here)
```suggestion
      Option.when(canWiden(fromType, toType))(toType).orElse {
        logMissedWidening(fromType = fromType, toType = toType)
        None
      }
```

spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala
```diff
@@ -1345,6 +1345,23 @@ trait DeltaSQLConfBase {
...
+    val NEVER, SAME_FAMILY_TYPE, ALWAYS = Value
+  }
+
+  val DELTA_ALLOW_AUTOMATIC_WIDENING =
+    buildConf("typeWidening.allowAutomaticWidening")
+      .doc("Controls the scope of enabled widening conversions in automatic schema widening " +
+        "during schema evolution. This flag is guarded by the flag \"delta.enableTypeWidening\"" +
+        "All supported widenings are enabled with \"always\" selected, which allows some " +
+        "conversions between integer types and floating numbers. The value \"same_family_type\" " +
+        "fallbacks to the default behavior of the guarding flag. \"never\" allows no widenings.")
```
johanl-db:
```suggestion
      .doc("Controls the scope of enabled widening conversions in automatic schema widening " +
        "during schema evolution. This flag is guarded by the flag 'delta.enableTypeWidening' +
        "All supported widenings are enabled with 'always' selected, which allows some " +
        "conversions between integer types and floating numbers. The value 'same_family_type' " +
        "fallbacks to the default behavior of the guarding flag. 'never' allows no widenings.")

File: spark/src/test/scala/org/apache/spark/sql/delta/schema/SchemaUtilsSuite.scala [hunk 8]
```diff
@@ -36,7 +36,7 @@ import org.scalatest.GivenWhenThen
 
 import org.apache.spark.sql.{AnalysisException, Column, DataFrame, QueryTest, Row}
 import org.apache.spark.sql.catalyst.{InternalRow, TableIdentifier}
-import org.apache.spark.sql.catalyst.analysis.{caseInsensitiveResolution, caseSensitiveResolution, UnresolvedAttribute}
+import org.apache.spark.sql.catalyst.analysis.{UnresolvedAttribute, caseInsensitiveResolution, caseSensitiveResolution}
```
johanl-db: Undo

spark/src/test/scala/org/apache/spark/sql/delta/typewidening/TypeWideningMergeIntoSchemaEvolutionSuite.scala
```diff
@@ -48,10 +49,54 @@ trait TypeWideningMergeIntoSchemaEvolutionTests
     extends MergeIntoSQLTestUtils
     with MergeIntoSchemaEvolutionMixin
     with TypeWideningTestCases {
-  self: QueryTest with TypeWideningTestMixin with DeltaDMLTestUtils =>
+  self: QueryTest
```
johanl-db: Do instead:
```
trait TypeWideningMergeIntoSchemaEvolutionTests
    extends MergeIntoSQLTestUtils
    with MergeIntoSchemaEvolutionMixin
    with TypeWideningTestCases
    with DeltaExcludedBySparkVersionTestMixinShims {
```

That way you don't have to also mix that trait in the class TypeWideningMergeIntoSchemaEvolutionSuite.

Note: `self: <trait> =>` essentially declares a dependency on `<trait>` without actually extending it. Meaning any class that extends the current trait must also extend `<trait>`
In practice here it doesn't make a lot of difference whether it's directly mixed in here or just declared as a dependency and mixed in the main class.

I tend to declare as dependency when the trait has a side effect (e.g. `TypeWideningTestMixin` turns on type widening by default`), so that the trait with side effects must be explicitly mixed in the base class, making the side effect more visible.

spark/src/test/scala/org/apache/spark/sql/delta/typewidening/TypeWideningInsertSchemaEvolutionSuite.scala
```diff
@@ -58,6 +59,100 @@ trait TypeWideningInsertSchemaEvolutionTests
   import testImplicits._
   import scala.collection.JavaConverters._
 
+  for {
```
johanl-db, line 63: This suite is getting really large and long running, can you split it in two suites.
You can e.g. put all tests that rely on `testInserts` on one side in e.g. `TypeWideningInsertSchemaEvolutionExtendedSuite` and the other tests in `TypeWideningInsertSchemaEvolutionBasicSuite`

spark/src/test/scala/org/apache/spark/sql/delta/typewidening/TypeWideningStreamingSourceSuite.scala
```diff
@@ -542,6 +543,75 @@ trait TypeWideningStreamingSourceTests
     }
   }
 
+  testSparkMasterOnly("always automatic in delta source writing to a delta sink") {
```
johanl-db: We can remove these two tests, it's actually testing the write to Delta sink on not read from Delta source - which isn't not governed by schema evolution: we always allow reading all supported type changes

spark/src/test/scala/org/apache/spark/sql/delta/typewidening/TypeWideningMergeIntoSchemaEvolutionSuite.scala
tmarzec: >Do instead: ...

I initially did exactly that, but then the build fails due to illegal inheritance:
```
illegal inheritance; superclass SparkFunSuite
 is not a subclass of the superclass QueryTest
 of the mixin trait DeltaExcludedBySparkVersionTestMixinShims
    with DeltaExcludedBySparkVersionTestMixinShims {
```

I didn't want to go deeper into analyzing this problem so I went ahead with whatever worked.

I'll look into this a bit more, this should also be an interesting Scala learning experience.

spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMode.scala
tmarzec: Done :+1:

spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMode.scala
tmarzec: Done :+1:

spark/src/main/scala/org/apache/spark/sql/delta/TypeWideningMode.scala
tmarzec: Done :+1:

spark/src/test/scala/org/apache/spark/sql/delta/typewidening/TypeWideningMergeIntoSchemaEvolutionSuite.scala
tmarzec: I changed it to:
```
trait TypeWideningMergeIntoSchemaEvolutionTests
    extends DeltaExcludedBySparkVersionTestMixinShims
    with MergeIntoSQLTestUtils
    with MergeIntoSchemaEvolutionMixin
    with TypeWideningTestCases {
```
and it works :muscle:

spark/src/test/scala/org/apache/spark/sql/delta/typewidening/TypeWideningInsertSchemaEvolutionBasicSuite.scala
```diff
@@ -50,14 +51,108 @@ class TypeWideningInsertSchemaEvolutionSuite
...
+      }
+
+      assert(readDeltaTable(tempPath).schema("value").dataType === testCase.fromType)
+    }
+  }
+
+  for {
+    testCase <- alterTableOnlySupportedTestCases
+  } {
+    test(s"INSERT - logs for missed opportunity for conversion " +
```
johanl-db: For this logging test and the one below: only run it for single type change, no need to cover all of them

PR #4786, opened 2025-06-18
[Draft] Kernel Pagination Support

mmmyr:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
mmmyr, line 1: maybe I should not changed this file. I don't really use the code change here.

PR #4792, opened 2025-06-19
[Spark] Better exception for clones without stats that enable row tracking

cstavr:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Improve exception message when customer attempt to clone a table without `numRecords` statistics and at the same time enable Row Tracking.

## How was this patch tested?

Updated existing test.

No


tomvanbussel (approved):
LGTM!

PR #4796, opened 2025-06-21
[Spark] Block VACUUM on CC tables regardless of table type

xzhseh:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
This PR blocks running VACUUM command on CC tables for Delta-Spark clients.

The reason is that any CC table is expected to already be a UC Managed table, so we can simply check the table's protocol to verify this.

## How was this patch tested?
Through existing UT.

N/A


Omar-Saleh (approved):
LGTM

PR #4798, opened 2025-06-22
[Spark]Added table property `hierarchicalClusteringColumns`

jiahao-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Added table property `hierarchicalClusteringColumns` and refactored column parser utils. 
The reason we need this is that we want to be able to specific columns to use for hierarchical clustering. We want Liquid to prioritize clustering the data files based on those columns if presented.

## How was this patch tested?

Through existing test suite.

No


itsbilal (approved):
LGTM

PR #1234, opened 2022-06-24
Add a SQL conf to allow arbitrary table properties (disable checks)

allisonport-db:
Adds SQL conf `ALLOW_ARBITRARY_TABLE_PROPERTIES` that disables our enforcement of valid table properties. This added as a solution to https://github.com/delta-io/delta/issues/1129.
    
Adds tests to `DeltaConfigSuite`

PR #4802, opened 2025-06-23
[Kernel] [Refactor] Update ParquetHandler::readParquetFiles interface to return a different data type

mmmyr:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description
Update ParquetHandler::readParquetFiles interface. Change the element type of returned CloseableIterator from ColumnarBatch to FileReadResult. 

Yes.

### DIFF
--- a/kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
+++ b/kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
@@ -800,7 +800,7 @@ trait SidecarIteratorProvider extends VectorTestUtils {
     Utils.singletonCloseableIterator(new FileReadResult(batch, v2CheckpointFileName))
   }
 
-  // TODO: extend FileReadResult for JSON read result
+  // TODO: [delta-io/delta#4849] extend FileReadResult for JSON read result
   def singletonSidecarJsonIterator(sidecars: Seq[FileStatus]): CloseableIterator[ColumnarBatch] = {
     val batch = buildSidecarBatch(sidecars)
     Utils.singletonCloseableIterator(batch)


kernel/kernel-api/src/main/java/io/delta/kernel/engine/ParquetReadResult.java
```diff
@@ -0,0 +1,21 @@
...
+public class ParquetReadResult {
+
+  private final ColumnarBatch data;
+  private final String fileName;
+
+  public ParquetReadResult(ColumnarBatch data, String fileName) {
+    this.data = data;
+    this.fileName = fileName;
+  }
+  public ColumnarBatch getData(){
```
scottsand-db: nit: newline

kernel/kernel-api/src/main/java/io/delta/kernel/engine/ParquetReadResult.java
```diff
@@ -0,0 +1,21 @@
+package io.delta.kernel.engine;
+
+import io.delta.kernel.data.ColumnarBatch;
+
+public class ParquetReadResult {
+
+  private final ColumnarBatch data;
+  private final String fileName;
```
scottsand-db: why just the file `name` instead of the file `path` ?

This means you would not be able to distinguish between `/a/b/c/foo.parquet` and `/d/e/f/foo.parquet`, right?

kernel/kernel-api/src/main/java/io/delta/kernel/engine/ParquetReadResult.java
```diff
@@ -0,0 +1,21 @@
...
+
+  public ParquetReadResult(ColumnarBatch data, String fileName) {
+    this.data = data;
+    this.fileName = fileName;
+  }
+  public ColumnarBatch getData(){
+    return this.data;
+  }
+
+  public  String getFileName(){
```
scottsand-db: nit: double space after `public`

scottsand-db: nit: no space after `()`

kernel/kernel-api/src/main/java/io/delta/kernel/engine/ParquetReadResult.java
mmmyr: Yes. In checkpoint file case, we only need file name (because all checkpoint files are in same directory). I'm not sure about data files.

mmmyr: I can change it to Path type (i guess this makes more sense).

kernel/kernel-api/src/main/java/io/delta/kernel/engine/ParquetReadResult.java
```diff
@@ -0,0 +1,23 @@
+package io.delta.kernel.engine;
+
+import io.delta.kernel.data.ColumnarBatch;
+import io.delta.kernel.internal.fs.Path;
```
scottsand-db: the Path is **internal**, we don't want it in our public API

This can be a string

kernel/kernel-api/src/main/java/io/delta/kernel/engine/ParquetReadResult.java
```diff
@@ -0,0 +1,22 @@
...
+  private final ColumnarBatch data;
+  private final String filePath;
+
+  public ParquetReadResult(ColumnarBatch data, String filePath) {
+    this.data = data;
+    this.filePath = filePath;
+  }
+
+  public ColumnarBatch getData() {
+    return this.data;
```
scottsand-db: in java, we typically only say `this.` in the constructor. In this getter, it is very clear what `return data;` means

kernel/kernel-api/src/main/java/io/delta/kernel/engine/ParquetReadResult.java
```diff
@@ -0,0 +1,22 @@
+package io.delta.kernel.engine;
+
+import io.delta.kernel.data.ColumnarBatch;
+
+public class ParquetReadResult {
```
scottsand-db: javadocs please: classdocs and methoddocs

kernel/kernel-api/src/main/java/io/delta/kernel/engine/ParquetHandler.java
```diff
@@ -56,12 +56,12 @@ public interface ParquetHandler {
...
    *     is still responsible apply the predicate on the data returned by this method.
-   * @return an iterator of {@link ColumnarBatch}s containing the data in columnar format. It is the
-   *     responsibility of the caller to close the iterator. The data returned is in the same as the
-   *     order of files given in {@code scanFileIter}.
+   * @return an iterator of {@link ParquetReadResult}s containing the data in columnar format along
+   *     with metadata. It is the responsibility of the caller to close the iterator. The data
+   *     returned is in the same as the order of files given in {@code scanFileIter}.
    * @throws IOException if an I/O error occurs during the read.
    */
-  CloseableIterator<ColumnarBatch> readParquetFiles(
```
huan233usc, line 64: Can you add a new one and mark this one deprecated. 

I think `readParquetFiles` is called in multiple places so probably you don't want to change all of caller in the same PR

kernel/kernel-api/src/main/java/io/delta/kernel/engine/ParquetReadResult.java
```diff
@@ -0,0 +1,22 @@
...
+
+  private final ColumnarBatch data;
+  private final String filePath;
+
+  public ParquetReadResult(ColumnarBatch data, String filePath) {
+    this.data = data;
```
huan233usc: can you add null check? Object.requireNonNull

kernel/kernel-api/src/main/java/io/delta/kernel/engine/ParquetReadResult.java
mmmyr: Added.

kernel/kernel-api/src/main/java/io/delta/kernel/engine/ParquetReadResult.java
mmmyr: Added.

kernel/kernel-api/src/main/java/io/delta/kernel/engine/ParquetReadResult.java
```diff
@@ -0,0 +1,60 @@
...
+   */
+  public ParquetReadResult(ColumnarBatch data, String filePath) {
+    this.data = Objects.requireNonNull(data, "data must not be null");
+    this.filePath = Objects.requireNonNull(filePath, "filePath must not be null");
+  }
+
+  /**
+   * Returns the {@link ColumnarBatch} of data that was read from the file.
+   *
+   * @return the columnar data
```
scottsand-db: these two lines say the same thing; let's just include the top line as the primary content of the @return and only have one line; let's note have duplication in our method docs

kernel/kernel-api/src/main/java/io/delta/kernel/engine/FileReadResult.java
```diff
@@ -0,0 +1,52 @@
...
+ * <p>Encapsulates both the data read (as a {@link ColumnarBatch}) and the full path of the file *
+ * from which the data was read.
+ */
+public class FileReadResult {
+
+  private final ColumnarBatch data;
+  private final String filePath;
+
+  /**
+   * Constructs a {@code ParquetReadResult} object with the given data and file path.
```
scottsand-db: nit: old name `ParquetReadResult`

mmmyr: Fixed.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
```diff
@@ -775,31 +776,43 @@ class SnapshotManagerSuite extends AnyFunSuite with MockFileSystemClientUtils {
...
+  // For Parquet (ColumnarBatch + filePath)
+  def singletonSidecarParquetIterator(sidecars: Seq[FileStatus])
+      : CloseableIterator[FileReadResult] = {
+    val batch = buildSidecarBatch(sidecars)
+    val filePath = if (sidecars.nonEmpty) sidecars.head.getPath else "<no_file>"
+    val result = new FileReadResult(batch, filePath)
+    Utils.singletonCloseableIterator(result)
+  }
+
+  // For JSON (only ColumnarBatch)
```
huan233usc: Can you create a issue(TODO) to extend it for JSON? I feel like in general file path is useful

huan233usc (approved):

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
mmmyr: Sure, will do.

kernel/kernel-api/src/main/java/io/delta/kernel/engine/FileReadResult.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+package io.delta.kernel.engine;
+
+import io.delta.kernel.data.ColumnarBatch;
+import java.util.Objects;
+
+/**
+ * Represents the result of reading a batch of data in a file.
+ *
+ * <p>Encapsulates both the data read (as a {@link ColumnarBatch}) and the full path of the file *
```
scottsand-db: line ends with `*`

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
```diff
@@ -775,31 +776,43 @@ class SnapshotManagerSuite extends AnyFunSuite with MockFileSystemClientUtils {
...
+    override def getColumnVector(ordinal: Int): ColumnVector = ordinal match {
+      case 0 => stringVector(sidecars.map(_.getPath)) // path
+      case 1 => longVector(sidecars.map(_.getSize): _*) // size
+      case 2 => longVector(sidecars.map(_.getModificationTime): _*) // modification time
+    }
+
+    override def getSize: Int = sidecars.length
+  }
+
+  // For Parquet (ColumnarBatch + filePath)
```
scottsand-db: IMO this comment and thte JSON comment are superfluous -- if/when we refactor the JSON code, there is a chance we forget to refactor that JSON comment. we can delete this comment

btw -- method comments should use the form `/** my comment */

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
```diff
@@ -775,31 +776,43 @@ class SnapshotManagerSuite extends AnyFunSuite with MockFileSystemClientUtils {
...
+  def singletonSidecarParquetIterator(sidecars: Seq[FileStatus])
+      : CloseableIterator[FileReadResult] = {
+    val batch = buildSidecarBatch(sidecars)
+    val filePath = if (sidecars.nonEmpty) sidecars.head.getPath else "<no_file>"
```
scottsand-db, line 797: so we are mocking reading these sidecar files from a v2 parquet checkpoint manifest, right?

`10.uuid.checkpoint.parquet` ?

so shouldn't the "file path" be the checkpoint file path, and not the file path of the inner contents of that parquet file? i.e. i don't think it should be the first sidecar path? for example, why not the 2nd or 3rd sidecar path?

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
```diff
@@ -775,31 +776,43 @@ class SnapshotManagerSuite extends AnyFunSuite with MockFileSystemClientUtils {
...
+      case 0 => stringVector(sidecars.map(_.getPath)) // path
+      case 1 => longVector(sidecars.map(_.getSize): _*) // size
+      case 2 => longVector(sidecars.map(_.getModificationTime): _*) // modification time
+    }
+
+    override def getSize: Int = sidecars.length
+  }
+
+  // For Parquet (ColumnarBatch + filePath)
+  def singletonSidecarParquetIterator(sidecars: Seq[FileStatus])
```
scottsand-db: can you please remind me what exactly we are mocking here?

we want an iterator of sidecar files, is this representing that we have read a parquet v2 checkpoint manifest?

i.e. we have read a `10.uuid.checkpoint.parquet` and it had `N` sidecars inside of it? and we want a single closeable iterator element that contains a batch representing all of those `N` sidecar files?

in this case ... shouldn't the "path" of the parquet file we are reading be the parent `10.uuid.checkpoint.parquet` file? i.e. not the sidecar file path? e.g. why the 1st sidecar file and why not the 2nd?

mmmyr: You are right, I double checked the code, `SidecarFile.READ_SCHEMA` represents the format of a row in a batch from v2-checkpoint file. Sorry I should have read this code more carefully. Thanks for pointing out!

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
```diff
@@ -775,31 +784,41 @@ class SnapshotManagerSuite extends AnyFunSuite with MockFileSystemClientUtils {
...
+    override def getColumnVector(ordinal: Int): ColumnVector = ordinal match {
+      case 0 => stringVector(sidecars.map(_.getPath)) // path
+      case 1 => longVector(sidecars.map(_.getSize): _*) // size
+      case 2 => longVector(sidecars.map(_.getModificationTime): _*) // modification time
+    }
+
+    override def getSize: Int = sidecars.length
+  }
+
+  def singletonSidecarParquetIterator(sidecars: Seq[FileStatus], V2CheckpointFileName: String)
```
scottsand-db: variable name should start lowercase: `v2CheckpointFileName`

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
```diff
@@ -775,31 +784,41 @@ class SnapshotManagerSuite extends AnyFunSuite with MockFileSystemClientUtils {
...
+      : CloseableIterator[FileReadResult] = {
+    val batch = buildSidecarBatch(sidecars)
+    val filePath = V2CheckpointFileName
```
scottsand-db: why `filePath = v2CheckpointFileName`? why do we need another variable `filePath`?

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
```diff
@@ -775,31 +784,41 @@ class SnapshotManagerSuite extends AnyFunSuite with MockFileSystemClientUtils {
...
+    val result = new FileReadResult(batch, filePath)
```
scottsand-db: nit: this isn't the result of the method, the result of the method is the singleton iterator

`fileReadResult` would be a better name

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
```diff
@@ -135,6 +136,14 @@ class SnapshotManagerSuite extends AnyFunSuite with MockFileSystemClientUtils {
     topLevelFileTypes.foreach { topLevelFileType =>
       val v2Checkpoints =
         v2CheckpointFileStatuses(v2CheckpointSpec, topLevelFileType)
+
+      val v2CheckpointFileName = if (v2CheckpointSpec.nonEmpty) {
```
scottsand-db: isn't `v2Checkpoints` a list of v2 checkpoint files statuses? are we able to use that? do we need to re-make a new name?

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
mmmyr: filePath won't be used in this test case, but `FileReadResult` class requires a non-empty filePath for constructor.

mmmyr: We can also remove that restriction, only ask for non-empty ColumnarBatch.

scottsand-db: I think there's a miscommunication here :) 

Your current code:

```scala
val filePath = V2CheckpointFileName
new FileReadResult(batch, filePath)
```

and I am asking, why can't it just be 
```scala
new FileReadResult(batch, V2CheckpointFileName)
```

i.e. I am asking: why do you need to declare a **new** variable `filePath`? there is already a variable `V2CheckpointFileName` passed into this method?`def singletonSidecarParquetIterator(sidecars: Seq[FileStatus], V2CheckpointFileName: String)`

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
nicklan: Yeah, I think each sidecar needs its own saved filename, we can't just use the same file for each.

I think it's worth getting this right here because this is going to be core to testing the pagination as I understand it.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
nicklan: I think the point was that below you can just do:
```scala
    val result = new FileReadResult(batch, v2checkpointFileName)
```

And you don't need to introduce a new variable.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
```diff
@@ -775,31 +784,41 @@ class SnapshotManagerSuite extends AnyFunSuite with MockFileSystemClientUtils {
...
+
+    override def getSize: Int = sidecars.length
+  }
+
+  def singletonSidecarParquetIterator(sidecars: Seq[FileStatus], V2CheckpointFileName: String)
+      : CloseableIterator[FileReadResult] = {
+    val batch = buildSidecarBatch(sidecars)
+    val filePath = V2CheckpointFileName
+    val result = new FileReadResult(batch, filePath)
+    Utils.singletonCloseableIterator(result)
```
nicklan: We can just shorten this a bit:

```suggestion
    Utils.singletonCloseableIterator(new FileReadResult(batch, filePath))
```

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
```diff
@@ -161,10 +163,15 @@ class SnapshotManagerSuite extends AnyFunSuite with MockFileSystemClientUtils {
...
+      val v2CheckpointFileName = if (expectedV2Checkpoint.nonEmpty) {
+        expectedV2Checkpoint.head.getPath
+      } else {
+        "no-v2-checkpoint"
+      }
       val logSegment = snapshotManager.getLogSegmentForVersion(
         createMockFSListFromEngine(
           listFromProvider(deltas ++ compactions ++ checkpointFiles)("/"),
-          new MockSidecarParquetHandler(expectedSidecars),
+          new MockSidecarParquetHandler(expectedSidecars, v2CheckpointFileName),
```
scottsand-db: could we just _not_ create a `new MockSidecarParquetHandler(expectedSidecars, ...)` if expectedSidecars is empty?

and then we wouldn't need a `"no-v2-checkpoint"`

mmmyr: Yep we can just use BaseMockParquetHandler to create engine if expectedSidecars is empty

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
```diff
@@ -160,11 +161,15 @@ class SnapshotManagerSuite extends AnyFunSuite with MockFileSystemClientUtils {
...
+      val mockSidecarParquetHandler = if (expectedSidecars.nonEmpty) {
+          new MockSidecarParquetHandler(expectedSidecars, expectedV2Checkpoint.head.getPath)
+        } else {
+          new BaseMockParquetHandler {}
+        }
       val logSegment = snapshotManager.getLogSegmentForVersion(
         createMockFSListFromEngine(
           listFromProvider(deltas ++ compactions ++ checkpointFiles)("/"),
-          new MockSidecarParquetHandler(expectedSidecars),
+          mockSidecarParquetHandler,
```
scottsand-db, line 173: do we have to add an empty parquet handler? can we just _not_ add it?

mmmyr:
```
/**
   * Create a mock [[Engine]] to mock the [[FileSystemClient.listFrom]] calls using
   * the given contents. The contents are filtered depending upon the list from path prefix.
   */
  def createMockFSListFromEngine(
      contents: Seq[FileStatus],
      parquetHandler: ParquetHandler,
      jsonHandler: JsonHandler): Engine = {
    mockEngine(
      fileSystemClient =
        new MockListFromFileSystemClient(listFromProvider(contents)),
      parquetHandler = parquetHandler,
      jsonHandler = jsonHandler)
  }
```
Hmmm I don't get it. Based on my understanding, we must have an engine right, which needs a parquet handler to create. Do you mean just pass null here or sth?

scottsand-db: ah, thanks for clarifying! I was misreading this, sgtm

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/SnapshotManagerSuite.scala
```diff
@@ -775,31 +781,39 @@ class SnapshotManagerSuite extends AnyFunSuite with MockFileSystemClientUtils {
...
+    override def getSize: Int = sidecars.length
+  }
+
+  def singletonSidecarParquetIterator(sidecars: Seq[FileStatus], v2CheckpointFileName: String)
+      : CloseableIterator[FileReadResult] = {
+    val batch = buildSidecarBatch(sidecars)
+    Utils.singletonCloseableIterator(new FileReadResult(batch, v2CheckpointFileName))
+  }
+
+  // TODO: extend FileReadResult for JSON read result
```
scottsand-db: Can you please link to the issue? `// TODO: [delta-io/delta#1234] Extend FileReadResult for JSON read result`

mmmyr: Added.

scottsand-db (approved):

nicklan (approved):
lgtm! thanks

PR #4804, opened 2025-06-23
[Pagination] New Kernel Pagination API (no implementation)

mmmyr:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

Add new Pagination APIs.

Yes, it adds new Pagination API.


kernel/kernel-api/src/main/java/io/delta/kernel/PaginatedScan.java
```diff
@@ -0,0 +1,48 @@
...
+public interface PaginatedScan extends Scan {
+
+  /**
+   * Get an iterator of Scan files for the current page.
+   *
+   * @param engine {@link Engine} instance to use in Delta Kernel.
+   * @return iterator of {@link FilteredColumnarBatch}s for the current page.
+   */
+  @Override
+  CloseableIterator<FilteredColumnarBatch> getScanFiles(Engine engine);
```
scottsand-db, line 50: Isn't this defined in the `Scan`? Is there a need to re-define it here?

kernel/kernel-api/src/main/java/io/delta/kernel/PaginatedScan.java
```diff
@@ -0,0 +1,48 @@
...
+
+  /**
+   * Returns a page token representing the current position in the iterator. If the iterator is
+   * partially consumed, the token corresponds to the next unconsumed element, allowing the scan to
```
scottsand-db: > If the iterator is partially consumed, the token corresponds to the next unconsumed element

Probably mention (in your own words): the page token assumes that the current batch has been fully consumed, that's why it points to the start of the "next" batch

kernel/kernel-api/src/main/java/io/delta/kernel/PaginatedScan.java
```diff
@@ -0,0 +1,48 @@
...
+   * be resumed from that point in a subsequent paginated request.
+   *
+   * <p>If the scan has been fully consumed (i.e., no more pages remain), this method returns {@code
+   * Optional.empty()}.
+   *
+   * <p>The returned token may also include metadata for validation purposes, such as detecting
```
scottsand-db: "may"? does it sometimes *not* include metadata for validation purposes? would it be more accurate to say, "the returned token includes metadat for ..."

kernel/kernel-api/src/main/java/io/delta/kernel/PaginatedScan.java
```diff
@@ -0,0 +1,48 @@
...
+  /**
+   * Returns the log replay hashsets as a {@link ColumnarBatch}. This data can optionally be cached
+   * by the caller and injected into subsequent page requests to improve performance.
+   *
+   * <p>Each column in the {@link ColumnarBatch} represents a specific attribute of the log replay
+   * hashsets.
+   *
+   * @return a {@link ColumnarBatch} containing the log replay hashsets.
+   */
+  ColumnarBatch getLogReplayHashsets();
```
scottsand-db: Let's leave this out for now :)

kernel/kernel-api/src/main/java/io/delta/kernel/ScanBuilder.java
```diff
@@ -65,4 +68,17 @@ public interface ScanBuilder {
...
+   *
+   * @param pageSize Maximum number of AddFiles to return in this page.
+   * @param pageToken Optional page token representing the current scan position; null to start from
+   *     the beginning.
+   * @param logReplayHashsets Optional tombstone hashsets to filter out deleted files; null if not
+   *     applicable.
+   * @return A {@link PaginatedScan} configured for pagination.
+   */
+  PaginatedScan buildPaginated(
+      long pageSize, @Nullable Row pageToken, @Nullable ColumnarBatch logReplayHashsets);
```
scottsand-db: let's not use `@Nullable`

let's ignore `logReplayHashsets` for now entirely

if the page token is optional, let's clearly indicate that using java `Optional`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanBuilderImpl.java
```diff
@@ -85,4 +88,10 @@ public Scan build() {
         dataPath,
         snapshotReport);
   }
+
+  @Override
+  public PaginatedScan buildPaginated(
+      long pageSize, Row pageToken, ColumnarBatch logReplayHashsets) {
+    return null;
```
scottsand-db: for now, let's just throw `UnsupportedOperationException("not implemented")`

kernel/kernel-api/src/main/java/io/delta/kernel/PaginatedScan.java
```diff
@@ -0,0 +1,48 @@
+package io.delta.kernel;
+
+import io.delta.kernel.data.ColumnarBatch;
+import io.delta.kernel.data.FilteredColumnarBatch;
+import io.delta.kernel.data.Row;
+import io.delta.kernel.engine.Engine;
+import io.delta.kernel.utils.CloseableIterator;
+import java.util.Optional;
+
+/** Extension of {@link Scan} that supports pagination. */
```
scottsand-db: Can you describe this a bit more? What _exactly_ does Pagination look like? How should callers consume this API to achieve pagination?

kernel/kernel-api/src/main/java/io/delta/kernel/PaginatedScan.java
mmmyr: Well, the only difference is if we re-define it here, all class that implements PaginatedScan MUST implement it. If we don't re-define it here, it's optional for any implementation of PaginatedScan to override this method : ).

kernel/kernel-api/src/main/java/io/delta/kernel/PaginatedScan.java
mmmyr: Okay

kernel/kernel-api/src/main/java/io/delta/kernel/PaginatedScan.java
scottsand-db: > If we don't re-define it here, it's optional for any implementation of PaginatedScan to override this class : ).

Override the class? or this method?

Doesn't Scan define this method? Don't all implementors of Scan need to override this?

So you are saying: if PaginatedScan interface does not override and define this method, then I can create a new PaginatedScanImpl that does not define the `getScanFiles` method, and that will compile?

mmmyr: And another reason I redefine it here: I want to document clearly this PaginatedScan::getScanFiles is different from Scan::getScanFiles. (it only retrieves ScanFiles in current page instead of retrieve all files)

scottsand-db: > I want to document clearly this PaginatedScan::getScanFiles is different from Scan::getScanFiles

This makes sense. Thanks! SGTM 👍

mmmyr: > So you are saying: if PaginatedScan interface does not override and define this method, then I can create a new PaginatedScanImpl that does not define the `getScanFiles` method, and that will compile?

You are right, it still needs to override getScanFiles. I guess the best reason for re-defining here is we can document this API more clearly.

kernel/kernel-api/src/main/java/io/delta/kernel/PaginatedScan.java
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * <p>This interface allows consumers to retrieve scan results in discrete, ordered pages rather
+ * than all at once. This is particularly useful for large datasets where materializing the full
+ * result set would be expensive in terms of memory or compute resources.
+ *
+ * <p>Pagination is achieved via a combination of {@code pageSize} and {@code pageToken}. The {@code
+ * pageSize} controls how many Parquet data files are returned in each page, while the {@code
+ * pageToken} encodes the location of next batch to read and is used to resume the scan from exactly
+ * where the last page ended. For the first page, the {@code pageToken} should be {@code
+ * Optional.empty()} or empty.
```
scottsand-db: nit: "should be empty or empty"?

kernel/kernel-api/src/main/java/io/delta/kernel/PaginatedScan.java
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Extension of {@link Scan} that supports pagination.
+ *
+ * <p>This interface allows consumers to retrieve scan results in discrete, ordered pages rather
+ * than all at once. This is particularly useful for large datasets where materializing the full
+ * result set would be expensive in terms of memory or compute resources.
+ *
+ * <p>Pagination is achieved via a combination of {@code pageSize} and {@code pageToken}. The {@code
+ * pageSize} controls how many Parquet data files are returned in each page, while the {@code
```
scottsand-db: how many `scan files` are returned in each page; scan files are more than just a parquet file, it also included deletion vector information

kernel/kernel-api/src/main/java/io/delta/kernel/PaginatedScan.java
```diff
@@ -0,0 +1,66 @@
...
+ * pageToken} encodes the location of next batch to read and is used to resume the scan from exactly
+ * where the last page ended. For the first page, the {@code pageToken} should be {@code
+ * Optional.empty()} or empty.
+ *
+ * <p>Consumers typically use {@link PaginatedScan} in a loop: they call {@code getScanFiles()} to
+ * retrieve an iterator over the current page's scan files, After consuming the iterator, then call
```
scottsand-db: `files, After consuming the iterator, then call` --> can you please fix the wording/grammar here a bit?

kernel/kernel-api/src/main/java/io/delta/kernel/PaginatedScan.java
```diff
@@ -0,0 +1,66 @@
...
+ * {@code getCurrentPageToken()} to retrieve a token to pass into the next page request. This allows
+ * users to scan the dataset incrementally, resuming from where they left off.
+ *
+ * <p>Implementations must ensure that the {@code pageToken} is opaque and self-describing;
```
scottsand-db: You don't need to talk about implementations -- we are the only implementation :) it's not like we are expecting connectors or other consumers to implement this -- this is just an API that we (kernel) are giving to connectors

kernel/kernel-api/src/main/java/io/delta/kernel/PaginatedScan.java
```diff
@@ -0,0 +1,66 @@
...
+   * that the current batch has been fully consumed. As a result, the token points to the start of
+   * the next batch.
+   *
+   * <p>If the scan has been fully consumed (i.e., no more pages remain), this method returns {@code
+   * Optional.empty()}.
+   *
+   * <p>The returned token also includes metadata for validation purposes, such as detecting changes
+   * in query parameters or log segments between requests.
+   *
+   * <p><strong>Usage note:</strong> Callers should invoke {@code getCurrentPageToken()} only after
```
scottsand-db: nit: this doesn't need to be `strong`: you are basically saying, stringly, "you **should** do X" and then in the next sentence you are saying "actually it is not required to do X"

kernel/kernel-api/src/main/java/io/delta/kernel/ScanBuilder.java
```diff
@@ -65,4 +67,14 @@ public interface ScanBuilder {
 
   /** @return Build the {@link Scan instance} */
   Scan build();
+
+  /**
+   * Build a Paginated Scan with a required page size and an optional page token.
+   *
+   * @param pageSize Maximum number of AddFiles to return in this page.
```
scottsand-db: ScanFiles

PR #4805, opened 2025-06-24
[Kernel] ParsedLogData refactor -- Remove Enums and create child classes

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/4805/files) to review incremental changes.
- [**stack/kernel_parsed_log_data_refactor**](https://github.com/delta-io/delta/pull/4805) [[Files changed](https://github.com/delta-io/delta/pull/4805/files)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Refactor ParsedLogData hierarchy to use classes instead of enums for compile-time type safety:

**Type system improvements:**
- Removed enum-based type distinctions and replaced with class-based polymorphism
- Now have compile-time guarantees about which type of log data we're using (ParsedDeltaData, ParsedCheckpointData, etc.)
- Updated SnapshotManager to use List<ParsedDeltaData> instead of List<ParsedLogData> where appropriate

**Validation improvements:**
- Added checkArgument validation to all ParsedXXX.forFileStatus() methods to catch type mismatches early
- Each subclass now validates its specific file type (commit, checkpoint, checksum, etc.)

**Checkpoint comparison refactoring:**
- Removed instanceof code smell by implementing polymorphic compareToSameType() method
- Fixed compareTo vs equals inconsistency
- Added comprehensive javadocs documenting checkpoint priority and comparison behavior

**Documentation and testing:**
- Added comprehensive javadocs to all ParsedLogData classes
- Enhanced test suite with validation tests, inline data tests, and cross-type equality tests
- Fixed toString test issues and added regex matching for dynamic values

## How was this patch tested?

- Updated existing unit tests with new validation scenarios
- Added comprehensive test coverage for all ParsedXXX classes
- Verified type safety changes don't break existing functionality

No

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedCheckpointData.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedCheckpointData.java
@@ -26,36 +26,32 @@ import java.util.Optional;
 public abstract class ParsedCheckpointData extends ParsedLogData
     implements Comparable<ParsedCheckpointData> {
 
+  /**
+   * Enum representing checkpoint type priorities for comparison when multiple checkpoint types
+   * exist at the same version. Higher ordinal values indicate higher priority.
+   */
+  protected enum CheckpointTypePriority {
+    CLASSIC, // priority 0 - least preferred
+    MULTIPART, // priority 1 - better than classic
+    V2 // priority 2 - most preferred
+  }
+
   protected ParsedCheckpointData(
       long version, Optional<FileStatus> fileStatusOpt, Optional<ColumnarBatch> inlineDataOpt) {
     super(version, fileStatusOpt, inlineDataOpt);
   }
 
   /**
-   * Returns the priority of this checkpoint type for Kernel's checkpoint preference ranking. Higher
-   * values indicate higher priority (preferred checkpoints).
-   *
-   * <p>This priority is only used to break ties when multiple checkpoint types exist at the same
-   * version. The ordering is based on safety and parallelizability of reading:
-   *
-   * <ul>
-   *   <li>V2 (priority 2): Best and safest checkpoint format
-   *   <li>MultiPart (priority 1): Better than classic due to parallelizable reading
-   *   <li>Classic (priority 0): Least preferred due to lower performance (single file)
-   * </ul>
+   * Returns the checkpoint type priority used as a tiebreaker when multiple checkpoint types exist
+   * at the same version.
    */
-  protected abstract int getCheckpointTypePriority();
+  protected abstract CheckpointTypePriority getCheckpointTypePriority();
 
   /**
-   * Compares two checkpoints of the same type and priority. Subclasses should implement
-   * type-specific comparison logic.
+   * Compares two checkpoints of the same version and same type. Subclasses should implement
+   * type-specific comparison logic for final tiebreaking.
    */
   protected abstract int compareToSameType(ParsedCheckpointData that);
-
-  @Override
-  public String getParentCategoryName() {
-    return "Checkpoint";
-  }
 
   @Override
   public Class<? extends ParsedLogData> getParentCategoryClass() {
@@ -85,10 +81,10 @@ public abstract class ParsedCheckpointData extends ParsedLogData
     }
 
     // 2. Compare types by priority (V2 > MultiPart > Classic)
-    int thisTypePriority = this.getCheckpointTypePriority();
-    int thatTypePriority = that.getCheckpointTypePriority();
+    CheckpointTypePriority thisTypePriority = this.getCheckpointTypePriority();
+    CheckpointTypePriority thatTypePriority = that.getCheckpointTypePriority();
     if (thisTypePriority != thatTypePriority) {
-      return Integer.compare(thisTypePriority, thatTypePriority);
+      return thisTypePriority.compareTo(thatTypePriority);
     }
 
     // 3. Use type-specific comparison when version and type are the same
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedClassicCheckpointData.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedClassicCheckpointData.java
@@ -51,8 +51,8 @@ public final class ParsedClassicCheckpointData extends ParsedCheckpointData {
   }
 
   @Override
-  protected int getCheckpointTypePriority() {
-    return 0; // (V2 > MultiPart > Classic)
+  protected CheckpointTypePriority getCheckpointTypePriority() {
+    return CheckpointTypePriority.CLASSIC;
   }
 
   @Override
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedLogData.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedLogData.java
@@ -83,10 +83,18 @@ public abstract class ParsedLogData {
     this.inlineDataOpt = inlineDataOpt;
   }
 
+  /**
+   * Returns true if this log data is stored as a file on disk. When false, the data is stored
+   * inline.
+   */
   public boolean isFile() {
     return fileStatusOpt.isPresent();
   }
 
+  /**
+   * Returns true if this log data is stored inline as a ColumnarBatch. When false, the data is
+   * stored as a file on disk.
+   */
   public boolean isInline() {
     return inlineDataOpt.isPresent();
   }
@@ -109,8 +117,15 @@ public abstract class ParsedLogData {
     return inlineDataOpt.get();
   }
 
-  public abstract String getParentCategoryName();
+  /**
+   * Returns a human-readable name for the parent category class. Used as a helper utility for
+   * debugging and print statements.
+   */
+  public String getParentCategoryName() {
+    return getParentCategoryClass().getSimpleName();
+  }
 
+  /** Returns the parent category class used for grouping collections of ParsedLogData instances. */
   public abstract Class<? extends ParsedLogData> getParentCategoryClass();
 
   /** Protected method for subclasses to override to add output to {@link #toString}. */
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedV2CheckpointData.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedV2CheckpointData.java
@@ -51,8 +51,8 @@ public final class ParsedV2CheckpointData extends ParsedCheckpointData {
   }
 
   @Override
-  protected int getCheckpointTypePriority() {
-    return 2; // (V2 > MultiPart > Classic)
+  protected CheckpointTypePriority getCheckpointTypePriority() {
+    return CheckpointTypePriority.V2;
   }
 
   @Override
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedMultiPartCheckpointData.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedMultiPartCheckpointData.java
@@ -67,8 +67,8 @@ public final class ParsedMultiPartCheckpointData extends ParsedCheckpointData {
   }
 
   @Override
-  protected int getCheckpointTypePriority() {
-    return 1; // (V2 > MultiPart > Classic)
+  protected CheckpointTypePriority getCheckpointTypePriority() {
+    return CheckpointTypePriority.MULTIPART;
   }
 
   @Override


kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedLogData.java
```diff
@@ -93,16 +88,18 @@ public static ParsedLogData forFileStatus(FileStatus fileStatus) {
...
-      throw new IllegalArgumentException(
-          "For PUBLISHED_DELTA|RATIFIED_STAGED_COMMIT, use ParsedLogData.forFileStatus() instead");
-    } else if (type == ParsedLogType.LOG_COMPACTION) {
-      throw new IllegalArgumentException(
-          "For LOG_COMPACTION, use ParsedLogCompactionData.forInlineData() instead");
-    } else if (type.category == ParsedLogCategory.CHECKPOINT) {
-      return ParsedCheckpointData.forInlineData(version, type, inlineData);
+    switch (type.category) {
+      case LOG_COMPACTION:
+        throw new IllegalArgumentException(
```
nicklan: Mmm. Sorry I missed this before. I feel like we should rather design the API so this can't happen. My initial thought is, we have `ParsedCheckpointData`, so could we just remove the `type` argument, and then you just have to call on the type you want directly? Perhaps method names change to indicate what type you're getting too.

nicklan (commented):
generally looks great! I have one api question

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedLogData.java
scottsand-db: We could do that, yeah, but we do not have a `ParsedCheckSumData` class today -- we would have to create that.

scottsand-db: I can go ahead and create that, SGTM!

scottsand-db: update: we also don't have a `ParsedDeltaData` class today (though, I think I'd like to add one so it can contain the additional commitTimestamp info that is exclusive to delta files and commits)

for now, I'll just add these as static factories on ParsedLogData for the delta and checksum types

scottsand-db: I decided just to rip off the bandaid and do a larger refactor; now there is one `ParsedLogData.forFileStatus` method that will delegate to children, and there is no `ParsedLogData.forInlineData` method, and instead each child class implements that method themselves.

As part of this, since we are now having so many classes, that means we do not need the `type` enum. And since we don't need that, we can also just use the `class` for the `category` enum (which we use in SnapshotManager to groupBy)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedCheckpointData.java
```diff
@@ -16,56 +16,23 @@
...
 
-import static io.delta.kernel.internal.util.Preconditions.checkArgument;
-
 import io.delta.kernel.data.ColumnarBatch;
-import io.delta.kernel.internal.util.FileNames;
 import io.delta.kernel.utils.FileStatus;
 import java.util.Optional;
 
-public class ParsedCheckpointData extends ParsedLogData
+public abstract class ParsedCheckpointData extends ParsedLogData
```
emkornfield, line 26: nit: class level javadoc?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedCheckpointData.java
```diff
@@ -16,56 +16,23 @@
...
-  public static ParsedCheckpointData forInlineData(
-      long version, ParsedLogType type, ColumnarBatch inlineData) {
-    if (type == ParsedLogType.MULTIPART_CHECKPOINT) {
-      throw new IllegalArgumentException(
-          "For MULTIPART_CHECKPOINT, use ParsedMultiPartCheckpointData.forInlineData() instead");
-    }
-
-    return new ParsedCheckpointData(version, type, Optional.empty(), Optional.of(inlineData));
-  }
+  protected abstract int getCheckpointTypePriority();
```
emkornfield: nit: document contract for this (are higher numbers considered higher priority or lower priority)?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedCheckpointData.java
```diff
@@ -100,11 +66,11 @@ public int compareTo(ParsedCheckpointData that) {
    * instead of going to the cloud store to read a file status.
    */
   protected int getTieBreaker(ParsedCheckpointData that) {
-    if (this.isInline() && that.isMaterialized()) {
+    if (this.isInline() && that.isFile()) {
       return 1; // Prefer this
```
emkornfield: nit: existing code but why?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedCheckpointData.java
```diff
@@ -100,11 +66,11 @@ public int compareTo(ParsedCheckpointData that) {
    * instead of going to the cloud store to read a file status.
    */
   protected int getTieBreaker(ParsedCheckpointData that) {
```
emkornfield: nit: existing code but it is a little bit strange that this is protected (and not private or final), it seems to imply a subclass could override this and then ordering would change depending on whether this or that was compared first.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedChecksumData.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package io.delta.kernel.internal.files;
+
+import io.delta.kernel.data.ColumnarBatch;
+import io.delta.kernel.internal.util.FileNames;
+import io.delta.kernel.utils.FileStatus;
+import java.util.Optional;
+
+public class ParsedChecksumData extends ParsedLogData {
```
emkornfield: nit: javadoc.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedClassicCheckpointData.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package io.delta.kernel.internal.files;
+
+import io.delta.kernel.data.ColumnarBatch;
+import io.delta.kernel.internal.util.FileNames;
+import io.delta.kernel.utils.FileStatus;
+import java.util.Optional;
+
+public class ParsedClassicCheckpointData extends ParsedCheckpointData {
```
emkornfield: nit: java doc.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedClassicCheckpointData.java
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  private ParsedClassicCheckpointData(
+      long version, Optional<FileStatus> fileStatusOpt, Optional<ColumnarBatch> inlineDataOpt) {
+    super(version, fileStatusOpt, inlineDataOpt);
+  }
+
+  @Override
+  protected int getCheckpointTypePriority() {
+    return 0; // Classic (0) < MultiPart (1) < V2 (2). Classic has the lowest priority.
```
emkornfield: why? Is this part of the protocol (sorry need to refresh myself on the RFC)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedDeltaData.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package io.delta.kernel.internal.files;
+
+import io.delta.kernel.data.ColumnarBatch;
+import io.delta.kernel.internal.util.FileNames;
+import io.delta.kernel.utils.FileStatus;
+import java.util.Optional;
+
+public class ParsedDeltaData extends ParsedLogData {
```
emkornfield: nit: Java doc

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedLogData.java
```diff
@@ -171,14 +111,13 @@ public boolean equals(Object o) {
     }
     ParsedLogData that = (ParsedLogData) o;
     return version == that.version
```
emkornfield, line 142: hmm, it would be good to clarify the contract for equality of these classes (and an argument for why they make sense).  Still ramping up here but given that subclassing was decided, it isn't clear to me that they are all interchangeable?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedV2CheckpointData.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package io.delta.kernel.internal.files;
+
+import io.delta.kernel.data.ColumnarBatch;
+import io.delta.kernel.internal.util.FileNames;
+import io.delta.kernel.utils.FileStatus;
+import java.util.Optional;
+
+public class ParsedV2CheckpointData extends ParsedCheckpointData {
```
emkornfield: nit: class javadoc.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
```diff
@@ -635,7 +636,7 @@ private List<ParsedLogData> getAllDeltasAfterCheckpointWithCatalogPriority(
 
     final List<ParsedLogData> allRatifiedCommitsAfterCheckpoint =
         parsedLogDatas.stream()
-            .filter(x -> x.type == ParsedLogType.RATIFIED_STAGED_COMMIT)
+            .filter(x -> x instanceof ParsedDeltaData && x.isFile())
```
emkornfield, line 609: nit: this pattern seems to appear at least twice, should it be a named concept (i.e. a method in this class or a method on ParsedLogData)?  If it appears a third time I would feel more strongly about it).  (instanceof checks are sometime a code smell as well.).

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedCheckpointData.java
```diff
@@ -76,22 +43,21 @@ public int compareTo(ParsedCheckpointData that) {
       return Long.compare(version, that.version);
     }
 
-    // Compare types.
-    if (type != that.type) {
-      return Integer.compare(type.ordinal(), that.type.ordinal());
+    // Compare types by priority.
+    int thisTypePriority = this.getCheckpointTypePriority();
```
emkornfield: nit: per comment on equality below, this is existing code but it is odd to have a compareTo method that does not return 0, when `equals()` return true.  As a follow-up it might make sense to extract a priority comparator instead of making this a object level property.

emkornfield: Using comparator could also potentially make the code more succinct (if performance isn't critical).

e.g. `Comparator.comparingLong(ParsedCheckpointData::getVersion).andThenComparingInt(ParsedCheckpointData::getTypePriority)....`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedCheckpointData.java
```diff
@@ -76,22 +43,21 @@ public int compareTo(ParsedCheckpointData that) {
...
-      case CLASSIC_CHECKPOINT:
-      case V2_CHECKPOINT:
-        return getTieBreaker(that);
-      case MULTIPART_CHECKPOINT:
-        final ParsedMultiPartCheckpointData thisCasted = (ParsedMultiPartCheckpointData) this;
-        final ParsedMultiPartCheckpointData thatCasted = (ParsedMultiPartCheckpointData) that;
-        return thisCasted.compareToMultiPart(thatCasted);
-      default:
-        throw new IllegalStateException("Unexpected type: " + type);
+    if (this instanceof ParsedMultiPartCheckpointData
```
emkornfield: nit: this seems like a code smell, if we are checking for subclasses of this class to apply custom logic?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedLogData.java
```diff
@@ -155,9 +97,7 @@ public ColumnarBatch getInlineData() {
     return inlineDataOpt.get();
   }
 
-  public ParsedLogCategory getCategory() {
-    return type.category;
-  }
+  public abstract Class<? extends ParsedLogData> getParentCategoryClass();
```
emkornfield, line 129: existing code but javadoc, parent class specifically seems like a concept that should be explained?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedLogData.java
```diff
@@ -30,106 +30,48 @@
  * status) or stored inline (as a columnar batch).
  */
 // TODO: Move this to be a public API
-public class ParsedLogData {
-
-  ///////////////////////////////////////
-  // Static enums, fields, and methods //
-  ///////////////////////////////////////
-
-  public enum ParsedLogCategory {
```
emkornfield, line 39: it seems like keeping the Category might be cleaner then relying on class hierarchy?  (Generally relying on class hierarchy for business logic can lead to tigher coupling then one might want)

emkornfield: e.g. its not clear to me if all the subclasses here would need to be public if some more accessors where added, and class hierarchy is not involved?

scottsand-db: > it seems like keeping the Category might be cleaner then relying on class hierarchy?

Yeah, that's a fair thought, and that's why I initially started with the Category enum.

To me, the issue is: we now have several classes that are needed to provide additional information than what is in the parent class.

ParsedLogData: version, file or inline
ParsedDeltaData: eventually will add the commit timestmap
Checkpoints: need to be able to be compared to other checkpoints
MultiPartCheckpoint: has part and num parts
LogCompaction: has startVersion

Now we have a lot of classes -- and it seemed reasonabe that we just make a few more classes, too, for Checksum, ClassicCheckpoint, and V2Checkpoint.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
scottsand-db: Great question, Micah.

To me, the **type** is "Delta" and where the Delta metadata content is located is either in the file or in the inline data.

Other possible "Delta" types are: published delta, staged ratified delta, staged inline delta -- but here, in this line, we don't care **what** type of Delta it is -- we just care that it **is** a Delta and it's metadata content is **in** a File.

Is the `x instance` that much worse than `enum.type == X` ?

emkornfield: > Is the x instance that much worse than enum.type == X

No, thats why it is a nit.  The advantage using the type separate from class hierarchy is that you can refactor the class hierarchy without impacting logic here.

There is still the second part of this comment, on whether we should consolidate the concept of `x instanceof ParsedDeltaData && x.isFile()` into something appropriately named.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedLogData.java
emkornfield: We can keep as is, I worry a little bit around evolution here but it shouldn't block the review.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/SnapshotManager.java
scottsand-db: > The advantage using the type separate from class hierarchy is that you can refactor the class hierarchy without impacting logic here.

This is a great point. Let me think on this. Appreciate the detailed review and feedback!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedCheckpointData.java
```diff
@@ -16,97 +16,104 @@
...
+   *   <li>Classic (priority 0): Least preferred due to lower performance (single file)
+   * </ul>
+   */
+  protected abstract int getCheckpointTypePriority();
+
+  /**
+   * Compares two checkpoints of the same type and priority. Subclasses should implement
+   * type-specific comparison logic.
+   */
+  protected abstract int compareToSameType(ParsedCheckpointData that);
```
allisonport-db, line 54: sorry what does this comparison yield? more fine-grained priority?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedCheckpointData.java
```diff
@@ -16,97 +16,104 @@
...
-    } else if (this.isMaterialized() && that.isInline()) {
-      return -1; // Prefer that
-    } else if (this.isMaterialized() && that.isMaterialized()) {
+  protected final int compareByDataSource(ParsedCheckpointData that) {
+    if (this.isInline() && that.isFile()) {
+      return 1; // Prefer this (inline data)
+    } else if (this.isFile() && that.isInline()) {
+      return -1; // Prefer that (inline data)
+    } else if (this.isFile() && that.isFile()) {
+      // Both are files - use path as arbitrary but deterministic tiebreaker
```
allisonport-db, line 109: why not just also return 0 since we do that below?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedDeltaData.java
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * <ul>
+ *   <li><b>Published commits</b>: Normal delta files in {@code _delta_log/} directory like {@code
+ *       00000000000000000001.json}
+ *   <li><b>Staged commits</b>: Files in {@code _delta_log/_staged_commits/} with UUID naming like
+ *       {@code 00000000000000000001.uuid-1234.json}
+ *   <li><b>Inline commits</b>: Content stored directly by the catalog, not as files
+ * </ul>
+ */
+public final class ParsedDeltaData extends ParsedLogData {
```
allisonport-db, line 39: Do we need to distinguish between ratified and published here at all? Since formerly this was done with the ParsedLogType

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedLogData.java
```diff
@@ -26,110 +26,64 @@
...
+    } else if (FileNames.isMultiPartCheckpointFile(path)) {
+      return ParsedMultiPartCheckpointData.forFileStatus(fileStatus);
     } else {
       throw new IllegalArgumentException("Unknown log file type: " + path);
     }
-
-    return new ParsedLogData(version, type, Optional.of(fileStatus), Optional.empty());
-  }
-
-  public static ParsedLogData forInlineData(
```
allisonport-db, line 94: are we removing this since it's not currently used? or any particular reason to remove this but keep the above method?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedLogData.java
```diff
@@ -155,9 +109,9 @@ public ColumnarBatch getInlineData() {
     return inlineDataOpt.get();
   }
 
-  public ParsedLogCategory getCategory() {
-    return type.category;
-  }
+  public abstract String getParentCategoryName();
```
allisonport-db: maybe useful to document these categories here?

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
```diff
@@ -229,7 +230,7 @@ private Map<String, String> getRequiredTablePropertiesForCreate(String ucTableId
    * loading a Delta table.
    */
   @VisibleForTesting
-  static List<ParsedLogData> getSortedKernelLogDataFromRatifiedCommits(
+  static List<ParsedLogData> getSortedKernelParsedDeltaDataFromRatifiedCommits(
```
allisonport-db, line 233: should the return type be updated to ParsedDeltaData?

allisonport-db (commented):
A few questions/comments

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedCheckpointData.java
allisonport-db: Also: realized the answer to my question is yes -- maybe we can improve the docs to make this more clear?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedCheckpointData.java
scottsand-db: Here do we have information that could let us make a deterministic decision. We should always try and operate deterministically when we can. That improves debugability.

Below, we don't have any information to help us make a deterministic decision.

Also, this is what CheckpointInstance does today -- a class we will want to deprecate. We should maintain that behavior.

https://github.com/delta-io/delta/blob/39ce7209a1e12ac91ed8cebaeca57188da666356/kernel/kernel-api/src/main/java/io/delta/kernel/internal/checkpoints/CheckpointInstance.java#L158

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedDeltaData.java
scottsand-db: So far I don't think so! Can add it if we need it.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedLogData.java
scottsand-db: Kernel only supports file-based data, not materialized so far. So `ParsedLogData::forInlineData` was only used in tests.

Next, this method was taking in the enum type -- but that's what this PR is removing! It just doesn't belong at this level in the API. Any time you are operation with logData and want to pass it in to Kernel, you should know what type it is, and invoke the right API.

When you are dealing with files --> the file name tells you the type! When you are dealing with inline content -- the content itself doesn't tell you the type!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedLogData.java
scottsand-db: I can add a comment that this is useful just for a simple string for printing out. Can include the types of names, too.

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
scottsand-db: Great question. So, `SnapshotBuilder::withLogData` needs to take in ParsedLogData. It cannot just take in a List of ParsedDeltaData.

If we went with your suggestion, somewhere we would need to convert that list of ParsedDeltaData to a list of ParsedLogData.

By Keeping this as ParsedLogData ... we basically just do that conversion inline in this mapping. Keeps things a bit simpler

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedLogData.java
```diff
@@ -26,110 +26,64 @@
...
         "Exactly one of fileStatusOpt or inlineDataOpt must be present");
     checkArgument(version >= 0, "version must be non-negative");
     this.version = version;
-    this.type = type;
     this.fileStatusOpt = fileStatusOpt;
     this.inlineDataOpt = inlineDataOpt;
   }
 
-  public boolean isMaterialized() {
+  public boolean isFile() {
```
emkornfield, line 90: document (in particular the alternative is `inline`?)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedLogData.java
emkornfield: I think this still applies?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedCheckpointData.java
```diff
@@ -16,97 +16,104 @@
...
-    }
+  /**
+   * Returns the priority of this checkpoint type for Kernel's checkpoint preference ranking. Higher
+   * values indicate higher priority (preferred checkpoints).
+   *
+   * <p>This priority is only used to break ties when multiple checkpoint types exist at the same
+   * version. The ordering is based on safety and parallelizability of reading:
+   *
+   * <ul>
+   *   <li>V2 (priority 2): Best and safest checkpoint format
```
huan233usc: Wondering if we could define a enum like below that are comparable or just add ranking ability to ParseLogData enum? If we add 4th type of checkpoint in the future, we need to update the code/comments everywhere.

public enum CheckpointType implements Comparable<CheckpointType> {
    CLASSIC(0, "Classic checkpoint)"),
    MULTIPART(1, "Multipart checkpoint"),
    V2(2, "V2 checkpoint");

    private final int priority;
    private final String description;

    CheckpointType(int priority, String description) {
        this.priority = priority;
        this.description = description;
    }

    public int priority() { return priority; }

    @Override
    public int compareTo(CheckpointType o) {
        return Integer.compare(this.priority, o.priority);
    }
}

emkornfield (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedChecksumData.java
```diff
@@ -0,0 +1,64 @@
...
+        FileNames.isChecksumFile(fileStatus.getPath()),
+        "Expected a checksum file but got %s",
+        fileStatus.getPath());
+
+    final String path = fileStatus.getPath();
+    final long version = FileNames.checksumVersion(path);
+    return new ParsedChecksumData(version, Optional.of(fileStatus), Optional.empty());
+  }
+
+  public static ParsedChecksumData forInlineData(long version, ColumnarBatch inlineData) {
```
huan233usc, line 46: Love this one, it could potentially unblocks lots of optimizations

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedCheckpointData.java
huan233usc: Oh, it seems that one of the intention of this pr is kinda remove enum, so maybe have one at checkpoint level to simplify the checkpoint comparision? wdyt

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedCheckpointData.java
scottsand-db: If you look at the ParsedCheckpointData::compareTo method -- you will see that we first compare by version, then we compare by type, and then if the types are equal we have to defer to child-class code to compare within that type! for example a multi-part checkpoint with 4 parts is prioritized over a multi-part checkpoint of 2 parts

This just lets us have 1 compareTo method in the parent class that lets children inject custom compareTo logic, too

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedLogData.java
scottsand-db: Hi Micah. Thanks for your question. Can you elaborate on what you mean by the contract for equality?

- A Delta is not equal to a LogCompaction which is not equal to a Checksum
- A Delta is equal to another Delta if they are at the same version and have either the same file status or inline data

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedCheckpointData.java
scottsand-db: Added!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedLogData.java
emkornfield: That sounds right to me, I think maybe in the previous iteration more things might have been considered equal (or maybe it was implicit that each priority represents is a different type, so types could not be equal)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedDeltaData.java
allisonport-db: I feel like this will be useful in the future, since we want to prefer ratified staged commits over published right? But we can add it then when we need it

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedLogData.java
allisonport-db: Makes sense! thanks

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedLogData.java
allisonport-db: I think it might be useful somewhere to have a list of the possible categories, kind of what was stored in `ParsedLogCategory` before; not a blocker though!

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
allisonport-db: I did not realize List are not upcast automatically :/ makes sense!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/files/ParsedDeltaData.java
scottsand-db: Agreed. Can improve this semantic later. For now, we very clearly know: (a) we LISTed and got back some files (so, not ratified), and (b) maybe we were INJECTED with some parsed log data (so, ratified).

That's the current semantic. Can iterate and improve in followup PRs!

PR #4814, opened 2025-06-24
done first pass

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/4814/files) to review incremental changes.
- [**stack/kernel_catalog_managed_writes_1**](https://github.com/delta-io/delta/pull/4814) [[Files changed](https://github.com/delta-io/delta/pull/4814/files)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)


kernel/kernel-api/src/main/java/io/delta/kernel/ResolvedTable.java
```diff
@@ -38,4 +39,6 @@ public interface ResolvedTable {
   StructType getSchema();
 
   ScanBuilder getScanBuilder();
+
+  Committer getCommitter();
```
scottsand-db, line 70: Committer probably should not always be required -- what about read-only integrations?

I can make this Optional.

cc @zachschuermann

scottsand-db: Update: Kernel should just provide simple, default committers (e.g. for path-based tables)
and if a connector is read-only, just pass a commit that no-ops or throws

raveeram-db, line 70: Nit: Overall for this class can we add some class and method comments Scott? And how it's expected to be used

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitFailedException.java
```diff
@@ -0,0 +1,49 @@
...
+ *  | retryable | conflict  | meaning                                                         |
+ *  |   no      |   no      | something bad happened (e.g. auth failure)                      |
+ *  |   no      |   yes     | permanent transaction conflict (e.g. multi-table commit failed) |
+ *  |   yes     |   no      | transient error (e.g. network hiccup)                           |
+ *  |   yes     |   yes     | physical conflict (allowed to rebase and retry)                 |
+ * </pre>
+ */
+public class CommitFailedException extends Exception {
+
+  public boolean retryable;
```
raveeram-db: Do we want to make these mutable?

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitPayload.java
```diff
@@ -0,0 +1,99 @@
...
+   * being changed.
+   */
+  public final Optional<Metadata> newMetadataOpt;
+
+  public CommitPayload(
+      long version,
+      String logPath,
+      CloseableIterator<Row> finalizedActions,
+      CommitInfo commitInfo,
+      Optional<Protocol> readProtocolOpt,
```
raveeram-db: should we also allow for a generic bag of properties to be plumbed through here?

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitFailedException.java
scottsand-db: we do not! good catch! I just copied this from the existing CommitFailedException class in delta-storage, today. good attention to detail, thanks!

kernel/kernel-api/src/main/java/io/delta/kernel/ResolvedTable.java
scottsand-db: Will add in another PR 👍 Btw, I've only been purposefully skimping on this sort of documentation because I have been trying to deliver the "MVP" deadline.

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitPayload.java
scottsand-db: Good question. I'd rather ship a slimmer API and expand upon it as new requirements come in.

kernel/kernel-api/src/main/java/io/delta/kernel/ResolvedTable.java
scottsand-db: https://github.com/delta-io/delta/pull/4822

scottsand-db: Update: I think this would make more sense to on the Transaction class?

nicklan: I think we can discuss this today, I think we don't want to have to thread things through in kernel such that we would put this in the txn class

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitPayload.java
```diff
@@ -0,0 +1,99 @@
...
+  public final CloseableIterator<Row> finalizedActions;
+
+  /** The {@link CommitInfo} that is being written as part of this commit. */
+  public final CommitInfo commitInfo;
+
+  /**
+   * The {@link Protocol} that was read at the beginning of the commit. Empty if a new table is
+   * being created.
+   */
+  public final Optional<Protocol> readProtocolOpt;
```
nicklan: Can you remind me why we need the old/new p and ms?

scottsand-db: Good question! For example, to detect that the `catalogManaged` table feature has been enabled on the table.

scottsand-db: And, certainly, you want the new Protocol and Metadata so you can send that to the catalog, and then it always knows what's the latets P & M, and it can give that back on the next read request to efficiently construct a ResolvedTable without having to go to the file system to resolve the latest P & M

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitFailedException.java
```diff
@@ -0,0 +1,49 @@
...
+ *  | retryable | conflict  | meaning                                                         |
+ *  |   no      |   no      | something bad happened (e.g. auth failure)                      |
+ *  |   no      |   yes     | permanent transaction conflict (e.g. multi-table commit failed) |
+ *  |   yes     |   no      | transient error (e.g. network hiccup)                           |
+ *  |   yes     |   yes     | physical conflict (allowed to rebase and retry)                 |
+ * </pre>
+ */
+public class CommitFailedException extends Exception {
+
+  public final boolean retryable;
```
raveeram-db: `private final` ?

scottsand-db: I'm treating this as a `data` class -- and I'm purposefully avoiding the duplicate/boilerplate of `private` member fields and `getters`

Want me to _not_ do that?

raveeram-db: Yes, would be good to align with general Java convention here - Will also allow for extending the getters with additional logging/functionality without breaking clients (especially since this is a public class)

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitPayload.java
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * Contains all information required to commit changes to a Delta table.
+ *
+ * <p>The {@code CommitPayload} encapsulates the complete state needed for a Delta commit operation,
+ * including all Delta actions, metadata changes, and contextual information. This payload is passed
+ * from Kernel to {@link Committer} implementations, which use the information to perform
+ * catalog-specific or filesystem-specific commit operations.
+ *
+ * <p><strong>Important:</strong> The actions iterator should only be consumed once. Committer
```
scottsand-db: Engine implementations

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitPayload.java
```diff
@@ -0,0 +1,124 @@
...
+    return version;
+  }
+
+  /** The path to the Delta log directory. */
+  public String getLogPath() {
+    return logPath;
+  }
+
+  /** All Delta actions (data and metadata) that are part of this commit. */
+  public CloseableIterator<Row> getFinalizedActions() {
```
scottsand-db: mention again that this can only be consumed once

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitFailedException.java
```diff
@@ -0,0 +1,62 @@
...
+ * <pre>
+ *  | retryable | conflict  | meaning                                                         |
+ *  |   no      |   no      | something bad happened (e.g. auth failure)                      |
+ *  |   no      |   yes     | permanent transaction conflict (e.g. multi-table commit failed) |
+ *  |   yes     |   no      | transient error (e.g. network hiccup)                           |
+ *  |   yes     |   yes     | physical conflict (allowed to rebase and retry)                 |
+ * </pre>
+ */
+@Experimental
+public class CommitFailedException extends Exception {
```
emkornfield, line 33: Is there a base Kernel Exception class that this should extend?

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
```diff
@@ -0,0 +1,107 @@
...
+
+  private final long version;
+  private final String logPath;
+  private final CommitInfo commitInfo;
+  private final Optional<Protocol> readProtocolOpt;
+  private final Optional<Metadata> readMetadataOpt;
+  private final Optional<Protocol> newProtocolOpt;
+  private final Optional<Metadata> newMetadataOpt;
+
+  public CommitMetadata(
```
emkornfield, line 43: aside: I wonder if we've discussed using something like [Immutables](https://immutables.github.io/) to reduce boiler plate

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
```diff
@@ -0,0 +1,107 @@
...
+   * The {@link Protocol} that was read at the beginning of the commit. Empty if a new table is
+   * being created.
+   */
+  public Optional<Protocol> getReadProtocolOpt() {
+    return readProtocolOpt;
+  }
+
+  /**
+   * The {@link Metadata} that was read at the beginning of the commit. Empty if a new table is
+   * being created.
```
emkornfield, line 86: are reads required for blind inserts?

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
```diff
@@ -0,0 +1,107 @@
...
+    this.newProtocolOpt = requireNonNull(newProtocolOpt, "newProtocolOpt is null");
+    this.newMetadataOpt = requireNonNull(newMetadataOpt, "newMetadataOpt is null");
+  }
+
+  /** The version of the Delta table this commit is targeting. */
+  public long getVersion() {
+    return version;
+  }
+
+  /** The path to the Delta log directory. */
```
emkornfield: Does this always end in _delta_log?

kernel/kernel-api/src/main/java/io/delta/kernel/commit/Committer.java
```diff
@@ -0,0 +1,56 @@
...
+   *   <li>Calling catalog commit APIs with the staged commit location (or inline content) and
+   *       additional metadata (such as the commit Protocol and Metadata)
+   *   <li>Publishing ratified catalog commits into the Delta log
+   * </ol>
+   *
+   * @return CommitResponse containing the resultant commit
+   * @throws CommitFailedException if the commit operation fails.
+   */
+  CommitResponse commit(
+      Engine engine, CloseableIterator<Row> finalizedActions, CommitMetadata commitMetadata)
```
emkornfield, line 64: nit: parameter docs for finalizedAction would be useful (at least a pointer to what a Row is supposed to contain for them (or are they opaque)?

kernel/kernel-api/src/main/java/io/delta/kernel/transaction/TransactionV2.java
```diff
@@ -0,0 +1,58 @@
...
+import io.delta.kernel.Transaction;
+import io.delta.kernel.annotation.Experimental;
+import io.delta.kernel.commit.CommitContext;
+import io.delta.kernel.commit.CommitMetadata;
+import io.delta.kernel.data.Row;
+import io.delta.kernel.engine.Engine;
+import io.delta.kernel.utils.CloseableIterator;
+import java.util.Map;
+
+/** Represents a transaction to mutate a Delta table. */
```
emkornfield: maybe document difference from Transaction

emkornfield (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
scovich: Blind inserts don't read data, but they still have to read P&M to know what table features and schema they're dealing with.

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitFailedException.java
scottsand-db: There is https://github.com/delta-io/delta/blob/master/kernel/kernel-api/src/main/java/io/delta/kernel/exceptions/KernelException.java which is a `RuntimeException`.

We want this to explicitly be a checked exception.

In theory we could, rename `KernelException` to `KernelRuntimeException` and create a new checked KernelException?

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
scottsand-db: Yes! Anything I should change/tweak because of that fact?

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
scottsand-db: I'm open to that! Any risk/harm/cost that you can think of? e.g. how this impacts our published JAR?

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
emkornfield: maybe just an example of the path in docs (this might start getting redundant though ...)

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
emkornfield: I have to admit I've only used something similar in the context of builds where it doesn't matter, some more research would be necessary to see if there are any downsides.

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitFailedException.java
emkornfield: I think it is OK to refactor later if we need a common base-class later

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
```diff
@@ -0,0 +1,107 @@
...
+  private final Optional<Metadata> newMetadataOpt;
+
+  public CommitMetadata(
+      long version,
+      String logPath,
+      CommitInfo commitInfo,
+      Optional<Protocol> readProtocolOpt,
+      Optional<Metadata> readMetadataOpt,
+      Optional<Protocol> newProtocolOpt,
+      Optional<Metadata> newMetadataOpt) {
```
huan233usc, line 50: Not related to this PR. Maybe we could have CommitMetadata.Builder class in the future, given constructor has very long list of arguments.

kernel/kernel-api/src/main/java/io/delta/kernel/transaction/TransactionV2.java
```diff
@@ -0,0 +1,84 @@
...
+   * protocol and table features enabled on the table. The engine should use this at the data writer
+   * task to transform logical data into physical data that goes in data files using {@link
+   * Transaction#transformLogicalData(Engine, Row, CloseableIterator, Map)}.
+   *
+   * @return the internal state {@link Row}
+   */
+  Row getTransactionState();
+
+  /**
+   * Get the {@link CommitContext} that can be used only for the very first commit to the table.
```
raveeram-db: Nit: `for the very first commit to the table` -> `for the very first commit attempt to the table`
would perhaps be more clear?

raveeram-db (approved):
looks good! Just a nit.

kernel/kernel-api/src/main/java/io/delta/kernel/transaction/TransactionV2.java
scottsand-db: Yup, great feedback. Implemented and clarified further.

PR #4831, opened 2025-06-25
[UniForm] Change Iceberg format-version for CompatV1/V2 to 2

harperjiang:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This change allows IcebergCompatV1/V2 to generate Iceberg format-version = 2, to be consistent with the Apache Iceberg default.
## How was this patch tested?
UT

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/IcebergCompat.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/IcebergCompat.scala
@@ -42,7 +42,7 @@ import org.apache.spark.sql.types._
 
 object IcebergCompatV1 extends IcebergCompatBase(
   version = 1,
-  icebergFormatVersion = 1,
+  icebergFormatVersion = 2,
   config = DeltaConfigs.ICEBERG_COMPAT_V1_ENABLED,
   tableFeature = IcebergCompatV1TableFeature,
   requiredTableProperties = Seq(RequireColumnMapping),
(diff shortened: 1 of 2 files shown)


KaiqiJinWow (approved):

spark/src/main/scala/org/apache/spark/sql/delta/IcebergCompat.scala
```diff
@@ -42,7 +42,7 @@ import org.apache.spark.sql.types._
 
 object IcebergCompatV1 extends IcebergCompatBase(
   version = 1,
-  icebergFormatVersion = 1,
+  icebergFormatVersion = 2,
```
scottsand-db, line 45: @harperjiang does this mean IcebergCompatV1 is only for Iceberg Version 2?

Can we please update the protocol to clarify?

cc @dengsh12

harperjiang: IcebergCompatV1 is not only for Iceberg Version 2, and works for both Iceberg Spec 1 and 2.
It's just Delta-Spark's UniForm reference implementation will generate Spec 2 by default. 
As UniForm is not part of the Protocol, there is no good place to put this.

dengsh12: Thanks for response! I wonder if we want to clarify in protocol: IcebergCompatV1 works for both Iceberg Spec 1 and 2, IcebergCompatV2 works for ?(I'm unsure if it works for both V1+V2 or V2 only), and IcebergCompatV3 works for V3?

scottsand-db: That seems like a very helpful addition to the protocol

PR #4835, opened 2025-06-25
[Kernel] [CatalogManaged] [UC] InMemoryUCClient

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/4835/files) to review incremental changes.
- [**stack/uc_kernel_catalog_managed_2_in_memory_test_client_b**](https://github.com/delta-io/delta/pull/4835) [[Files changed](https://github.com/delta-io/delta/pull/4835/files)]
  - [stack/uc_kernel_catalog_managed_3](https://github.com/delta-io/delta/pull/4838) [[Files changed](https://github.com/delta-io/delta/pull/4838/files/e9eeef9043b562c6e7eac81f95da0b3e94f44a9c..c3662f9ddfb192244e89f9c7d15c56dbe063cb45)]

---------
- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Create an `InMemoryUCClient` and a simple test suite

## How was this patch tested?

New test suite

No

### DIFF
--- a/unity/src/test/scala/io/delta/unity/InMemoryUCClient.scala
+++ b/unity/src/test/scala/io/delta/unity/InMemoryUCClient.scala
@@ -1,0 +1,102 @@
+/*
+ * Copyright (2025) The Delta Lake Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.delta.unity
+
+import java.lang.{Long => JLong}
+import java.net.URI
+import java.util.Optional
+import java.util.concurrent.ConcurrentHashMap
+
+import scala.collection.JavaConverters._
+import scala.collection.mutable.ArrayBuffer
+
+import io.delta.storage.commit.{Commit, CommitFailedException, GetCommitsResponse}
+import io.delta.storage.commit.actions.{AbstractMetadata, AbstractProtocol}
+import io.delta.storage.commit.uccommitcoordinator.{InvalidTargetTableException, UCClient}
+
+object InMemoryUCClient {
+  class TableData {
+    private var maxRatifiedVersion = -1L
+    private val commits: ArrayBuffer[Commit] = ArrayBuffer.empty
+
+    def getMaxRatifiedVersion: Long = maxRatifiedVersion
+
+    def getCommits: List[Commit] = commits.toList
+
+    def appendCommit(commit: Commit): Unit = {
+      val expectedCommitVersion = maxRatifiedVersion + 1
+
+      if (commit.getVersion != expectedCommitVersion) {
+        throw new CommitFailedException(
+          false, /* retryable */
+          false, /* conflict */
+          s"Expected commit version $expectedCommitVersion but got ${commit.getVersion}")
+      }
+
+      commits += commit
+      maxRatifiedVersion += 1
+    }
+  }
+}
+
+class InMemoryUCClient(ucMetastoreId: String) extends UCClient {
+
+  import InMemoryUCClient._
+
+  /** Map from UC_TABLE_ID to TABLE_DATA */
+  val tables = new ConcurrentHashMap[String, TableData]()
+
+  override def getMetastoreId: String = ucMetastoreId
+
+  override def commit(
+      tableId: String,
+      tableUri: URI,
+      commit: Optional[Commit],
+      lastKnownBackfilledVersion: Optional[JLong],
+      disown: Boolean,
+      newMetadata: Optional[AbstractMetadata],
+      newProtocol: Optional[AbstractProtocol]): Unit = {
+    throw new UnsupportedOperationException("Not implemented yet")
+  }
+
+  override def getCommits(
+      tableId: String,
+      tableUri: URI,
+      startVersion: Optional[JLong],
+      endVersion: Optional[JLong]): GetCommitsResponse = {
+    val tableData = getTableDataElseThrow(tableId)
+
+    val commits = tableData
+      .getCommits
+      .filter { x =>
+        startVersion.orElse(0) <= x.getVersion &&
+        x.getVersion <= endVersion.orElse(Long.MaxValue)
+      }
+
+    new GetCommitsResponse(commits.asJava, tableData.getMaxRatifiedVersion)
+  }
+
+  override def close(): Unit = {}
+
+  private def getTableDataElseThrow(tableId: String): TableData = {
+    if (!tables.containsKey(tableId)) {
+      throw new InvalidTargetTableException(s"Table not found: $tableId")
+    }
+
+    tables.get(tableId)
+  }
+}
--- a/unity/src/test/scala/io/delta/unity/InMemoryUCClient.scala
+++ b/unity/src/test/scala/io/delta/unity/InMemoryUCClient.scala
@@ -94,19 +94,50 @@ class InMemoryUCClient(ucMetastoreId: String) extends UCClient {
   import InMemoryUCClient._
 
   /** Map from UC_TABLE_ID to TABLE_DATA */
-  val tables = new ConcurrentHashMap[String, TableData]()
+  private val tables = new ConcurrentHashMap[String, TableData]()
 
   override def getMetastoreId: String = ucMetastoreId
+
+  /** Convenience method for tests to commit with default parameters. */
+  def commitWithDefaults(
+      tableId: String,
+      tableUri: URI,
+      commit: Optional[Commit],
+      lastKnownBackfilledVersion: Optional[JLong] = Optional.empty(),
+      disown: Boolean = false,
+      newMetadata: Optional[AbstractMetadata] = Optional.empty(),
+      newProtocol: Optional[AbstractProtocol] = Optional.empty()): Unit = {
+    this.commit(
+      tableId,
+      tableUri,
+      commit,
+      lastKnownBackfilledVersion,
+      disown,
+      newMetadata,
+      newProtocol)
+  }
 
   override def commit(
       tableId: String,
       tableUri: URI,
-      commit: Optional[Commit],
+      commit: Optional[Commit] = Optional.empty(),
       lastKnownBackfilledVersion: Optional[JLong],
       disown: Boolean,
       newMetadata: Optional[AbstractMetadata],
       newProtocol: Optional[AbstractProtocol]): Unit = {
-    throw new UnsupportedOperationException("Not implemented yet")
+    Seq(
+      (lastKnownBackfilledVersion.isPresent, "lastKnownBackfilledVersion"),
+      (disown, "disown"),
+      (newMetadata.isPresent, "newMetadata"),
+      (newProtocol.isPresent, "newProtocol")).foreach { case (isUnsupported, name) =>
+      if (isUnsupported) {
+        throw new UnsupportedOperationException(s"$name not supported yet in InMemoryUCClient")
+      }
+    }
+
+    if (!commit.isPresent) return
+
+    createTableIfNotExists(tableId).appendCommit(commit.get())
   }
 
   override def getCommits(
@@ -120,6 +151,14 @@ class InMemoryUCClient(ucMetastoreId: String) extends UCClient {
   }
 
   override def close(): Unit = {}
+
+  private[unity] def getTablesCopy: Map[String, TableData] = {
+    tables.asScala.toMap
+  }
+
+  private def createTableIfNotExists(tableId: String): TableData = {
+    tables.computeIfAbsent(tableId, _ => new TableData)
+  }
 
   /** Retrieves table data for the given table ID or throws an exception if not found. */
   private def getTableDataElseThrow(tableId: String): TableData = {
--- a/unity/src/test/scala/io/delta/unity/InMemoryUCClient.scala
+++ b/unity/src/test/scala/io/delta/unity/InMemoryUCClient.scala
@@ -163,9 +163,7 @@ class InMemoryUCClient(ucMetastoreId: String) extends UCClient {
 
   /** Retrieves table data for the given table ID or throws an exception if not found. */
   private def getTableDataElseThrow(tableId: String): TableData = {
-    Option(tables.get(tableId)) match {
-      case Some(tableData) => tableData
-      case None => throw new InvalidTargetTableException(s"Table not found: $tableId")
-    }
+    Option(tables.get(tableId))
+      .getOrElse(throw new InvalidTargetTableException(s"Table not found: $tableId"))
   }
 }
--- a/unity/src/test/scala/io/delta/unity/InMemoryUCClient.scala
+++ b/unity/src/test/scala/io/delta/unity/InMemoryUCClient.scala
@@ -137,7 +137,7 @@ class InMemoryUCClient(ucMetastoreId: String) extends UCClient {
 
     if (!commit.isPresent) return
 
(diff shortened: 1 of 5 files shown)


unity/src/test/scala/io/delta/unity/InMemoryUCClientSuite.scala
```diff
@@ -0,0 +1,103 @@
...
+import io.delta.storage.commit.uccommitcoordinator.InvalidTargetTableException
+
+import org.scalatest.funsuite.AnyFunSuite
+
+/** Unit tests for [[InMemoryUCClient]]. */
+class InMemoryUCClientSuite extends AnyFunSuite with UCCatalogManagedTestUtils {
+
+  private val fakeURI = new URI("s3://bucket/table")
+
+  private def getInMemoryUCClientWithCommitsForTableId(
```
scottsand-db: move to UCCatalogManagedTestUtils

unity/src/test/scala/io/delta/unity/InMemoryUCClient.scala
```diff
@@ -0,0 +1,111 @@
...
+ * Provides a mock implementation of UCClient that stores all table data in memory. This is  useful
+ * for unit tests that need to simulate Unity Catalog operations without connecting to an actual UC
+ * service.
+ */
+class InMemoryUCClient(ucMetastoreId: String) extends UCClient {
+
+  import InMemoryUCClient._
+
+  /** Map from UC_TABLE_ID to TABLE_DATA */
+  val tables = new ConcurrentHashMap[String, TableData]()
```
emkornfield: I'm a little confused as to the semantics intended relative to thread-safety for this class.  Perphaps some comments on this and TableData to spell out guarantees with respect to thread-safety would be useful?

unity/src/test/scala/io/delta/unity/InMemoryUCClient.scala
```diff
@@ -0,0 +1,111 @@
...
+        x.getVersion <= endVersion.orElse(Long.MaxValue)
+      }
+
+    new GetCommitsResponse(commits.asJava, tableData.getMaxRatifiedVersion)
+  }
+
+  override def close(): Unit = {}
+
+  private def getTableDataElseThrow(tableId: String): TableData = {
+    if (!tables.containsKey(tableId)) {
```
emkornfield: If the intent is to keep thread safety, I'd suggest inverting this case and calling .get and checking for null instead of two different calls without a mutex to the same data-structure?

unity/src/test/scala/io/delta/unity/InMemoryUCClient.scala
```diff
@@ -0,0 +1,111 @@
...
+import scala.collection.mutable.ArrayBuffer
+
+import io.delta.storage.commit.{Commit, CommitFailedException, GetCommitsResponse}
+import io.delta.storage.commit.actions.{AbstractMetadata, AbstractProtocol}
+import io.delta.storage.commit.uccommitcoordinator.{InvalidTargetTableException, UCClient}
+
+object InMemoryUCClient {
+
+  /** Internal data structure to track table state including commits and version information. */
+  class TableData {
```
emkornfield, line 40: with the use of ConcurrentHashMap below, it isn't clear to me if maybe some of these methods should be synchronized in some fasionh.

unity/src/test/scala/io/delta/unity/InMemoryUCClient.scala
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * Provides a mock implementation of UCClient that stores all table data in memory. This is  useful
+ * for unit tests that need to simulate Unity Catalog operations without connecting to an actual UC
+ * service.
+ */
+class InMemoryUCClient(ucMetastoreId: String) extends UCClient {
+
+  import InMemoryUCClient._
+
+  /** Map from UC_TABLE_ID to TABLE_DATA */
```
emkornfield, line 96: nit: what is UC_TABLE_ID?  maybe just tableId?

unity/src/test/scala/io/delta/unity/InMemoryUCClient.scala
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  override def getCommits(
+      tableId: String,
+      tableUri: URI,
+      startVersion: Optional[JLong],
+      endVersion: Optional[JLong]): GetCommitsResponse = {
+    val tableData = getTableDataElseThrow(tableId)
+
+    val commits = tableData
```
emkornfield: if the intent is to have a a thread-safe version here, consider moving this method into tableData? (even without this it might be more natural to have the logic there).

unity/src/test/scala/io/delta/unity/InMemoryUCClientSuite.scala
```diff
@@ -0,0 +1,103 @@
...
+    val tableId = "non-existent-table"
+
+    val exception = intercept[InvalidTargetTableException] {
+      client.getCommits(tableId, new URI("s3://bucket/table"), Optional.empty(), Optional.empty())
+    }
+    assert(exception.getMessage.contains(s"Table not found: $tableId"))
+  }
+
+  test("getCommits returns all commits if no startVersion or endVersion filter") {
+    val client = getInMemoryUCClientWithCommitsForTableId("tableId", 0L to 5L)
```
emkornfield: nit: comment literal parameters (0L to 5L)?

unity/src/test/scala/io/delta/unity/InMemoryUCClient.scala
```diff
@@ -0,0 +1,111 @@
...
+      disown: Boolean,
+      newMetadata: Optional[AbstractMetadata],
+      newProtocol: Optional[AbstractProtocol]): Unit = {
+    throw new UnsupportedOperationException("Not implemented yet")
+  }
+
+  override def getCommits(
+      tableId: String,
+      tableUri: URI,
+      startVersion: Optional[JLong],
```
emkornfield, line 146: nit: I guess this contract is already decided upon, but it might be clearer to create a minimal Range class that more clearly defines inclusive/exclusive values for Long

unity/src/test/scala/io/delta/unity/UCCatalogManagedClientSuite.scala
```diff
@@ -34,7 +34,23 @@ class UCCatalogManagedClientSuite extends AnyFunSuite with UCCatalogManagedTestU
...
+      ucCatalogManagedClient.loadTable(null, "ucTableId", "tablePath", 1L)
+    }
+    assertThrows[NullPointerException] {
+      ucCatalogManagedClient.loadTable(defaultEngine, null, "tablePath", 1L)
+    }
+    assertThrows[NullPointerException] {
+      ucCatalogManagedClient.loadTable(defaultEngine, "ucTableId", null, 1L)
+    }
+    assertThrows[IllegalArgumentException] {
+      ucCatalogManagedClient.loadTable(defaultEngine, "ucTableId", "tablePath", -1L)
```
emkornfield: is 0 a valid input for the last argument?  if so should the test cases above use 0.

unity/src/test/scala/io/delta/unity/UCCatalogManagedClientSuite.scala
```diff
@@ -34,7 +34,23 @@ class UCCatalogManagedClientSuite extends AnyFunSuite with UCCatalogManagedTestU
     }
   }
 
-  // TODO: loadTable throws on invalid input. We need a non-null UCClient in order to test this.
+  test("loadTable throws on invalid input") {
+    val ucClient = new InMemoryUCClient("ucMetastoreId")
+    val ucCatalogManagedClient = new UCCatalogManagedClient(ucClient)
+
+    assertThrows[NullPointerException] {
+      ucCatalogManagedClient.loadTable(null, "ucTableId", "tablePath", 1L)
```
emkornfield: nit: consider commenting literal parameters with parameter name?

emkornfield (changes requested):
Biggest concern is nailing down what is intended with relation to thread-safety.

unity/src/test/scala/io/delta/unity/UCCatalogManagedClientSuite.scala
scottsand-db: Yes! I'd love to, but our scala formatter complains about that :( You can try it out yourself, it won't let you do `.loadTable(null /* param */, xxx /* param */)`

unity/src/test/scala/io/delta/unity/InMemoryUCClient.scala
scottsand-db: Hm. This is the `InMemoryUCClient`, as opposed to `InMemoryGenericClient` --> i think it's fair to reference/use the `UC_TABLE_ID`? representing the unity catalog table id for our table data? With that context, does this now make sense? Let me know!

unity/src/test/scala/io/delta/unity/UCCatalogManagedClientSuite.scala
emkornfield: :( I wonder if it allows `.loadTable(/* param */ null , /* param */ xxx )` (which actually reads more naturally but might not be convention in the code in general).

unity/src/test/scala/io/delta/unity/InMemoryUCClient.scala
scottsand-db: I'm pushing my code soon -- I hope using the existing startVersion and endVersion is sufficiently clear?

unity/src/test/scala/io/delta/unity/UCCatalogManagedClientSuite.scala
scottsand-db: that works for the 1st param (after the `(`) but does not work for other params, (after a `,`)

emkornfield: OK, thanks for trying.

unity/src/test/scala/io/delta/unity/InMemoryUCClient.scala
emkornfield: I think the thing that was slightly suprising to me is that endVersion appears to be inclusive (which is not necessarily my expectation for most APIs).  If this is deeply entrenched we can ignore this for now (hopefully the parent interface is well documented about inclusive/exclusive).

unity/src/test/scala/io/delta/unity/InMemoryUCClient.scala
emkornfield: Sure, just reading this code it stood out as something that I couldn't understand from local context.

unity/src/test/scala/io/delta/unity/InMemoryUCClientSuite.scala
scottsand-db: added!

scottsand-db:
@emkornfield -- thanks for the comments, I've responded and updated the code, PTAL!

unity/src/test/scala/io/delta/unity/InMemoryUCClient.scala
```diff
@@ -0,0 +1,170 @@
...
+    new GetCommitsResponse(filteredCommits.asJava, tableData.getMaxRatifiedVersion)
+  }
+
+  override def close(): Unit = {}
+
+  private[unity] def getTablesCopy: Map[String, TableData] = {
+    tables.asScala.toMap
+  }
+
+  private def createTableIfNotExists(tableId: String): TableData = {
```
scottsand-db: getOrCreateTableIfNotExists

unity/src/test/scala/io/delta/unity/InMemoryUCClientSuite.scala
```diff
@@ -0,0 +1,144 @@
...
+
+  private def testGetCommitsFiltering(
+      allVersions: Seq[Long],
+      startVersionOpt: Optional[JLong],
+      endVersionOpt: Optional[JLong],
+      expectedVersions: Seq[Long]): Unit = {
+    val client = getInMemoryUCClientWithCommitsForTableId("tableId", allVersions)
+    val response = client.getCommits("tableId", fakeURI, startVersionOpt, endVersionOpt)
+    val actualVersions = response.getCommits.asScala.map(_.getVersion).toSeq
+    val expectedVersions = expectedVersions
```
scottsand-db: delete

unity/src/test/scala/io/delta/unity/InMemoryUCClientSuite.scala
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  private def testGetCommitsFiltering(
+      allVersions: Seq[Long],
+      startVersionOpt: Optional[JLong],
+      endVersionOpt: Optional[JLong],
+      expectedVersions: Seq[Long]): Unit = {
+    val client = getInMemoryUCClientWithCommitsForTableId("tableId", allVersions)
+    val response = client.getCommits("tableId", fakeURI, startVersionOpt, endVersionOpt)
+    val actualVersions = response.getCommits.asScala.map(_.getVersion).toSeq
```
scottsand-db: delete .toSeq

raveeram-db (approved):

unity/src/test/scala/io/delta/unity/InMemoryUCClient.scala
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  /** Retrieves the table data for the given table ID, creating it if it does not exist. */
+  private def getOrCreateTableIfNotExists(tableId: String): TableData = {
+    tables.computeIfAbsent(tableId, _ => new TableData)
+  }
+
+  /** Retrieves table data for the given table ID or throws an exception if not found. */
+  private def getTableDataElseThrow(tableId: String): TableData = {
+    Option(tables.get(tableId)) match {
```
emkornfield:
```suggestion
    Option(tables.get(tableId)).getOrElse(throw new ...)
```
I think something like this might work and might be more succinct (not sure if there is clear style guidance here, feel free to ignore).

unity/src/test/scala/io/delta/unity/InMemoryUCClientSuite.scala
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  test("getCommits throws InvalidTargetTableException for non-existent table") {
+    val client = new InMemoryUCClient("ucMetastoreId")
+    val tableId = "non-existent-table"
+
+    val exception = intercept[InvalidTargetTableException] {
+      client.getCommits(tableId, new URI("s3://bucket/table"), Optional.empty(), Optional.empty())
+    }
+    assert(exception.getMessage.contains(s"Table not found: $tableId"))
```
emkornfield:
```suggestion
    assert(exception.getMessage.contains("Table not found: non-existent-table"))
```

nit: sorry if i missed this, but try not to replicate business logic in tests.

scottsand-db: Hey! Could you actually explain this a bit more to me? I'd like to understand this best practice more

i.e. how this was replicating business logic? and why that was bad or why printing out `non-existent-table` is better?

emkornfield: Its replicating the substitution of the table name to construct the error message.  printing out non-existent-table means no additional steps are needed to verify what the output looks like.  So I can say yes the error message looks good or not.

One could argue that you want to show the table ID shows up in the error message.  This could be handled by splitting up the assert into two parts (either the full message hard-coded + an assert the message contains `tableId` or just two asserts that the message contains both the prefix and suffix).

https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html is I think is the most recent place I've seen it but its come up as a reasonable practice at a few places I've worked and tend to think it makes the behavior being tested easier to understand in a lot of cases (and potentially avoid bugs).

scottsand-db: oh, neat! that page really clarifies this feedback. SGTM!

unity/src/test/scala/io/delta/unity/InMemoryUCClient.scala
scottsand-db: simplified, yup! thanks!

PR #4836, opened 2025-06-25
[Kernel] [Pagination] New Pagination API

mmmyr:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/4836/files) to review incremental changes.
- [**stack/pagination-api**](https://github.com/delta-io/delta/pull/4836) [[Files changed](https://github.com/delta-io/delta/pull/4836/files)]
  - [stack/read-page1](https://github.com/delta-io/delta/pull/4837) [[Files changed](https://github.com/delta-io/delta/pull/4837/files/9d95b4dc82006d2883eb11cee98caeb6899aa4e1..ab5ebd100df54e220cabfbc89eb94a520f3c2db3)]
    - [stack/skipJSON](https://github.com/delta-io/delta/pull/4844) [[Files changed](https://github.com/delta-io/delta/pull/4844/files/ab5ebd100df54e220cabfbc89eb94a520f3c2db3..971a98d84f12031099aa35a77e39a8e87a69b05f)]
      - [stack/multi-CP](https://github.com/delta-io/delta/pull/4845) [[Files changed](https://github.com/delta-io/delta/pull/4845/files/971a98d84f12031099aa35a77e39a8e87a69b05f..40c4d7d7e0924a7d8225b6a44db7ddbf60b70de6)]
        - [stack/sidecar](https://github.com/delta-io/delta/pull/4846) [[Files changed](https://github.com/delta-io/delta/pull/4846/files/40c4d7d7e0924a7d8225b6a44db7ddbf60b70de6..8cbb470df1ab57e1fd8c95c3ecd4c9f0db466390)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This PR introduces new API for Kernel Pagination Support.

## How was this patch tested?

No implementation yet, no tests. 

Yes.


kernel/kernel-api/src/main/java/io/delta/kernel/PaginatedScan.java
```diff
@@ -0,0 +1,75 @@
...
+ * than all at once. This is particularly useful for large datasets where materializing the full
+ * result set would be expensive in terms of memory or compute resources.
+ *
+ * <p>Pagination is achieved via a combination of {@code pageSize} and {@code pageToken}. The {@code
+ * pageSize} controls how many Scan files are returned in each page, while the {@code pageToken}
+ * encodes the location of next batch to read and is used to resume the scan from exactly where the
+ * last page ended. For the first page, the {@code pageToken} should be {@code Optional.empty()}.
+ *
+ * <p>Consumers typically use {@link PaginatedScan} in a loop: they call {@code getScanFiles()} to
+ * retrieve an iterator over the current page's scan files, After consuming the iterator, users
```
scottsand-db: `files, After` --> should be period

kernel/kernel-api/src/main/java/io/delta/kernel/PaginatedScan.java
```diff
@@ -0,0 +1,75 @@
...
+   *
+   * @param engine {@link Engine} instance to use in Delta Kernel.
+   * @return iterator of {@link FilteredColumnarBatch}s for the current page.
+   */
+  @Override
+  CloseableIterator<FilteredColumnarBatch> getScanFiles(Engine engine);
+
+  /**
+   * Returns a page token representing the current position in the ScanFiles iterator, allowing the
+   * scan to be resumed from that point in a subsequent paginated request. If the iterator is
```
scottsand-db: `in a subsequent request`

kernel/kernel-api/src/main/java/io/delta/kernel/PaginatedScan.java
```diff
@@ -0,0 +1,75 @@
...
+   *
+   * <p>Note: Since {@code getScanFiles()} returns an iterator over batches, this page token assumes
+   * that the current batch has been fully consumed. As a result, the token points to the start of
+   * the next batch.
+   *
+   * <p>If the scan has been fully consumed (i.e., no more pages remain), this method returns {@code
+   * Optional.empty()}.
+   *
+   * <p>The returned token also includes metadata for validation purposes, such as detecting changes
+   * in query parameters or log segments between requests.
```
scottsand-db: `log segments` is not quite a public API term. let's just say `such as detecting changes in query parameters or underlying delta log content between requests`

kernel/kernel-api/src/main/java/io/delta/kernel/PaginatedScan.java
```diff
@@ -0,0 +1,75 @@
...
+   *
+   * <p>Callers should invoke {@code getCurrentPageToken()} only after finishing consumption of the
+   * current iterator. Actually it is not required to consume the entire iterator; the returned page
```
scottsand-db: I don't like the phrasing of "Callers should do X. Actually they don't have to do X".

You could instead phrase this as "Callers should do X. If they do not do X, then Y"

kernel/kernel-api/src/main/java/io/delta/kernel/ScanBuilder.java
```diff
@@ -65,4 +67,14 @@ public interface ScanBuilder {
 
   /** @return Build the {@link Scan instance} */
   Scan build();
+
+  /**
+   * Build a Paginated Scan with a required page size and an optional page token.
+   *
+   * @param pageSize Maximum number of Scan Files to return in this page.
+   * @param pageToken Optional page token representing the current scan position; empty to start
```
huan233usc, line 75: nit: start position?

scottsand-db: `empty to start from the beginning` --> `to retrieve the first page in a paginated scan, pass in an {@link Optional#empty}`

kernel/kernel-api/src/main/java/io/delta/kernel/PaginatedScan.java
```diff
@@ -0,0 +1,75 @@
...
+  CloseableIterator<FilteredColumnarBatch> getScanFiles(Engine engine);
+
+  /**
+   * Returns a page token representing the current position in the ScanFiles iterator, allowing the
+   * scan to be resumed from that point in a subsequent request. If the iterator is partially
+   * consumed, the token corresponds to the next unconsumed element.
+   *
+   * <p>Note: Since {@code getScanFiles()} returns an iterator over batches, this page token assumes
+   * that the current batch has been fully consumed. As a result, the token points to the start of
+   * the next batch.
```
huan233usc: should it be called getNextPageToken then?

"If the scan has been fully consumed" does it mean the iterator returned are all consumed, or this scan has already reach the end of all the actions?

mmmyr: >"If the scan has been fully consumed" does it mean the iterator returned are all consumed, or this scan has already reach the end of all the actions?

It means this scan has reach the end of all actions (no more pages to read). Sorry for the confusion, I'll change the wording.

mmmyr: > should it be called getNextPageToken then?

getCurrentPageToken means the page token callers get depends on **when** they call this function. If they call it in the half way of consuming the iterator, the returned page token points to the start position of next batch in the iterator.

nicklan: I think the intro sentence is wrong here (and contradicts what's just above).

> this page token assumes that the current batch has been fully consumed.

It does _not_.

> the token points to the start of the next batch.

Also not right? Should be something like: "the token points to the next batch that would be returned from the iterator"

kernel/kernel-api/src/main/java/io/delta/kernel/PaginatedScan.java
```diff
@@ -0,0 +1,75 @@
...
+ * pageSize} controls how many Scan files are returned in each page, while the {@code pageToken}
+ * encodes the location of next batch to read and is used to resume the scan from exactly where the
+ * last page ended. For the first page, the {@code pageToken} should be {@code Optional.empty()}.
+ *
+ * <p>Consumers typically use {@link PaginatedScan} in a loop: they call {@code getScanFiles()} to
+ * retrieve an iterator over the current page's scan files. After consuming the iterator, users
+ * should call {@code getCurrentPageToken()} to retrieve a token to pass into the next page request.
+ * This allows users to scan the dataset incrementally, resuming from where they left off.
+ */
+public interface PaginatedScan extends Scan {
```
nicklan, line 39: I'm confused. What page token do I get if I do:

```java
PaginatedScan scan = [get paginated scan];
CloseableIterator<FilteredColumnarBatch> batches1 = scan.getScanFiles(engine);
batches1.next();
batches1.next();
CloseableIterator<FilteredColumnarBatch> batches2 = scan.getScanFiles(engine);
batches2.next();

Optional<Row> token = scan.getCurrentPageToken(); // <- how does it know which iterator to consider
```

mmmyr: This won't be allowed. In PaginatedScan::getScanFiles will call ScanImpl::getScanFiles(), and getScanFiles() can only be called once for each Scan. PaginatedScan should have same behavior. See [code pointer](https://github.com/delta-io/delta/blob/5dbc68044c9e62ef394a544b4491311b80e89526/kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java#L120C3-L125C30) here.

mmmyr: Sorry for the confusion. I'll make it clear in the documentation here : )

kernel/kernel-api/src/main/java/io/delta/kernel/PaginatedScan.java
mmmyr: >Should be something like: "the token points to the next batch that would be returned from the iterator"

Yes. "batch" here refers to each element (an instance of FilteredColumnarBatch) returned from the iterator. 
>this page token assumes that the current batch has been fully consumed.

What I'm trying to say here is: page token always assume the current FilteredColumnarBatch returned from iterator is fully consumed, and it will always points to the start position of next FilteredColumnarBatch. 

Will update documentation to make this clear.

kernel/kernel-api/src/main/java/io/delta/kernel/PaginatedScan.java
mmmyr: Btw, I checked documentation of `Scan` interface, it seems there's no documentation about that getScanFiles can only be called once for each Scan instance (but it's indeed enforced in Scan implementation). We either need to: update the scan method docs to make this clear (I prefer this), or change the existing scan implementation to remove this enforcement. 
```
  public CloseableIterator<FilteredColumnarBatch> getScanFiles(
      Engine engine, boolean includeStats) {
    if (accessedScanFiles) {
      throw new IllegalStateException("Scan files are already fetched from this instance");
    }
    accessedScanFiles = true;
```

nicklan: Yeah, I see that now. Thanks for the code pointer.

I think you're going to find the code getting hairy if we do it this way though. Now you have to keep a pointer to the iterator in the `PaginatedScan` iff it's been created, and then you'll have to poll the iterator when you call `getCurrentPageToken` to see where it is.

Could we move the token method onto the iterator (would need a new iterator type), which would make it much more clear that the token is associated with the state of the iterator.

This would also open up the possibility of _not_ preventing calling `getScanFiles` twice, since I don't see why you _shouldn't_ be able to do that (even if it's maybe a bit silly)

mmmyr: Yep, the paginated iterator already has a getPageToken() function, so we don’t really need to have getPageToken() in PaginatedScan — it currently just delegates to the iterator anyway.

The only uncertainty I have is that this would require exposing a new PaginatedIterator class (to callers). But if you're okay with that, I’m happy to go ahead and update the API accordingly : ) .

kernel/kernel-api/src/main/java/io/delta/kernel/PaginatedAddFilesIterator.java
```diff
@@ -0,0 +1,10 @@
+package io.delta.kernel;
+
+import io.delta.kernel.data.FilteredColumnarBatch;
+import io.delta.kernel.data.Row;
+import io.delta.kernel.utils.CloseableIterator;
+
+public interface PaginatedAddFilesIterator extends CloseableIterator<FilteredColumnarBatch> {
```
raveeram-db: Let's add a comment describing this interface

kernel/kernel-api/src/main/java/io/delta/kernel/ScanBuilder.java
```diff
@@ -65,4 +67,14 @@ public interface ScanBuilder {
...
+   *     from the beginning.
+   * @return A {@link PaginatedScan} configured for pagination.
+   */
+  PaginatedScan buildPaginated(long pageSize, Optional<Row> pageToken);
```
raveeram-db, line 79: Nit: Maybe just call it buildPaginatedScan?

raveeram-db (approved):
Overall API makes sense to me, nice work!

kernel/kernel-api/src/main/java/io/delta/kernel/PaginatedAddFilesIterator.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * An iterator over {@link FilteredColumnarBatch}, each representing a batch of Scan Files in a
+ * paginated scan. This iterator also exposes the page token that can be used to resume the scan
+ * from the exact position current page ends in a subsequent request.
+ *
+ * <p>This interface extends {@link CloseableIterator} and should be closed when the iteration is
+ * complete.
+ */
+public interface PaginatedAddFilesIterator extends CloseableIterator<FilteredColumnarBatch> {
```
scottsand-db: nit: we should not leak `AddFile`s. This is a PaginatedIterator over `ScanFile`s

kernel/kernel-api/src/main/java/io/delta/kernel/ScanBuilder.java
```diff
@@ -65,4 +67,14 @@ public interface ScanBuilder {
...
+
+  /**
+   * Build a Paginated Scan with a required page size and an optional page token.
+   *
+   * @param pageSize Maximum number of Scan Files to return in this page.
+   * @param pageToken Optional page token representing the current scan position; empty to start
+   *     from the beginning.
+   * @return A {@link PaginatedScan} configured for pagination.
+   */
+  PaginatedScan buildPaginatedScan(long pageSize, Optional<Row> pageToken);
```
scottsand-db: I like `buildPaginated`. above, for Scan, we call it `build`, not `buildScan`

kernel/kernel-api/src/main/java/io/delta/kernel/PaginatedScan.java
```diff
@@ -0,0 +1,48 @@
...
+ * result set would be expensive in terms of memory or compute resources.
+ *
+ * <p>Pagination is achieved via a combination of {@code pageSize} and {@code pageToken}. The {@code
+ * pageSize} controls how many Scan files are returned in each page, while the {@code pageToken}
+ * encodes the location of next batch to read and is used to resume the scan from exactly where the
+ * last page ended. For the first page, the {@code pageToken} should be {@code Optional.empty()}.
+ *
+ * <p>Consumers typically use {@link PaginatedScan} in a loop: they call {@code getScanFiles()} to
+ * retrieve an iterator over the current page's scan files. After consuming the iterator, users
+ * should call {@code getCurrentPageToken()} to retrieve a token to pass into the next page request.
```
scottsand-db: `getCurrentPageToken` -- you should probably include this as `{@link PaginatedScanFilesIterator#getCurrentPageToken}`

kernel/kernel-api/src/main/java/io/delta/kernel/PaginatedScanFilesIterator.java
```diff
@@ -0,0 +1,47 @@
...
+   * the underlying log files.
+   *
+   * <p>The page token represents the position of current iterator at the time it's called. If the
+   * iterator is only partially consumed, the returned token will always point to the beginning of
+   * the next unconsumed {@link FilteredColumnarBatch}. This method will return Option.empty() if
+   * all data in the Scan is consumed (no more non-empty pages remain).
+   *
+   * <p>Note: Each page is expected to fully include each {@link FilteredColumnarBatch}. That is,
+   * pages do not split batches — the token always refers to the start of the next full batch, never
+   * to a position within a batch.
```
scottsand-db: This is an implementation detail that we can exclude from the API documentation

scottsand-db (approved):

PR #4847, opened 2025-06-27
[Spark] Fix DeltaDataFrameWriterV2Suite.Replace: fail if table does not exist

juliuszsompolski:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

CannotReplaceMissingTableException extends AnalysisException.
Spark has rewrites that sometimes change AnalysisException into ExtendedAnalysisException to add more context.
For some reason, it started happening here. Rewriting it into ExtendedAnalysisException makes it lose the CannotReplaceMissingTableException, but it retains the correct error class and SQL state. Make the test check the error class and SQL state instead.

## How was this patch tested?

Test runs.

No.


juliuszsompolski:
cc @ala

ala (approved):
LGTM.

PR #4848, opened 2025-06-27
[Kernel] [Pagination] New Page Token Class

mmmyr:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Introduce a new page token class. 

## How was this patch tested?
PageTokenSuite.scala

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
@@ -1,0 +1,41 @@
+package io.delta.kernel.internal.replay;
+
+import io.delta.kernel.data.Row;
+import io.delta.kernel.types.*;
+
+public class PageToken {
+  /** Variables to know where last page ends (current page starts) */
+  private final String startingFileName;
+
+  private final long rowIndex;
+
+  /** TODO: Variables for validating query params */
+  public PageToken(String startingFileName, long rowIndex) {
+    this.startingFileName = startingFileName;
+    this.rowIndex = rowIndex;
+  }
+
+  public static PageToken fromRow(Row row) {
+    String fileName = row.getString(0);
+    long rowIdx = row.getLong(1);
+    return new PageToken(fileName, rowIdx);
+  }
+
+  /** Convert PageToken to a Kernel Row object. */
+  public Row toRow() {
+    StructType schema =
+        new StructType().add("fileName", StringType.STRING).add("rowIndex", LongType.LONG);
+    // return Utils.newRow(schema, startingFileName, rowIndex, logSegmentHash);
+    // TODO: make this schema into a Row Type
+    return null;
+  }
+
+  public String getStartingFileName() {
+    return startingFileName;
+  }
+
+  public long getRowIndex() {
+    return rowIndex;
+  }
+
+}
--- a/kernel/kernel-api/src/test/scala/io/delta/kernel/internal/PageTokenSuite.scala
+++ b/kernel/kernel-api/src/test/scala/io/delta/kernel/internal/PageTokenSuite.scala
@@ -1,0 +1,6 @@
+import io.delta.kernel.test.MockFileSystemClientUtils
+import org.scalatest.funsuite.AnyFunSuite
+
+class PageTokenSuite extends AnyFunSuite with MockFileSystemClientUtils {
+
+}


kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
```diff
@@ -0,0 +1,149 @@
...
+  private final long tableVersion;
+  private final long predicateHash;
+  private final long logSegmentHash;
+
+  /** Schema for PageToken Row representation */
+  public static final StructType PAGE_TOKEN_SCHEMA =
+      new StructType()
+          .add("logFileName", StringType.STRING)
+          .add("rowIndexInFile", LongType.LONG)
+          .add("sidecarIndex", LongType.LONG)
```
huan233usc: Is this an optional field?

huan233usc (approved):
LGTM with some minor

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
```diff
@@ -0,0 +1,151 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+
+/** Page Token Class for Pagination Support */
+public class PageToken {
+  /** Variables to know where last page ends (current page starts) */
+  private final String startingFileName; // starting Log File Name (also last read log file name)
+
+  private final long rowIndex; // the index of last row in the last consumed batch
```
scottsand-db: nit: can we use `/** adsf */` comments for variables?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
```diff
@@ -0,0 +1,151 @@
...
+import io.delta.kernel.types.*;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+
+/** Page Token Class for Pagination Support */
+public class PageToken {
+  /** Variables to know where last page ends (current page starts) */
+  private final String startingFileName; // starting Log File Name (also last read log file name)
```
scottsand-db: does this apply to sidecars? only json and mulit-part-checkpoints / log compactions?

when we are reading a sidecar, what is this value?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
```diff
@@ -0,0 +1,151 @@
...
+
+  private final long rowIndex; // the index of last row in the last consumed batch
+  private final Optional<Long> sidecarIndex; // optional index of sidecar checkpoint file consumed
+
+  /** Variables for validating query params */
```
scottsand-db: based on where this comment is placed, it seems like it applies to the variables below, especially since there's an empty line between this variable and tablePath

You can do `// ===== Variables for validating query params =====`

scottsand-db: and apply this feedback, too, to the `Variables to know where last page ends` comment above, too please

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
```diff
@@ -0,0 +1,151 @@
...
+  private final String kernelVersion;
+
+  private final String tablePath;
+  private final long tableVersion;
+  private final long predicateHash;
+  private final long logSegmentHash;
+
+  /** Schema for PageToken Row representation */
+  public static final StructType PAGE_TOKEN_SCHEMA =
```
scottsand-db: nit: static fields come at the top of the class

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
```diff
@@ -0,0 +1,151 @@
...
+    this.rowIndex = rowIndex;
+    this.sidecarIndex = sidecarIndex;
+    this.kernelVersion = requireNonNull(kernelVersion, "kernelVersion must not be null");
+    this.tablePath = requireNonNull(tablePath, "tablePath must not be null");
+    this.tableVersion = tableVersion;
+    this.predicateHash = predicateHash;
+    this.logSegmentHash = logSegmentHash;
+  }
+
+  public static PageToken fromRow(Row row) {
```
scottsand-db: nit: this can come at the top, too. the order is: static methods, static variables/constants, instance variables, constructor, public, protected, private

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
```diff
@@ -0,0 +1,151 @@
...
+import java.util.Objects;
+import java.util.Optional;
+
+/** Page Token Class for Pagination Support */
+public class PageToken {
+  /** Variables to know where last page ends (current page starts) */
+  private final String startingFileName; // starting Log File Name (also last read log file name)
+
+  private final long rowIndex; // the index of last row in the last consumed batch
+  private final Optional<Long> sidecarIndex; // optional index of sidecar checkpoint file consumed
```
scottsand-db: can you help me to understand what `consumed` means here? can we please clarify out "consumed" / "was the last file" / "is the current" / "is the next" file semantics and terminology?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
scottsand-db: Do you mean `of the last consumed batch` as opposed to `in`? i.e. `in` makes me think: the value of this will only ever be from 0 to batchSize

Also -- I don't think we need to specify the unit here (i.e. batches) -- this is just the row index in the last file that we ended at in the last page?

btw -- suppose our batch size is 1000 and we have a json file 007.json with 3 batches and we finished the page at the end of the 2nd batch. is our row index `1999` (first batch is 0 to 999, second batch is 1000 to 1999) or is our row index 2000?

scottsand-db (changes requested):
Good tests! Left comments and questions on the semantics of PageToken

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
mmmyr: It's the file path of sidecar file. I feel like this info is easy to retrieve, and might be useful (for debugging etc.),  I thought it made sense to include it in the page token. I can also do a check in ActionIterator when adding sidecars to filelist - if the name of sidecar and the sidecar index matches.

scottsand-db: That's helpful, thanks! Can we update the variable documentation to explain this? e.g. how should this variable be used in relation to `sidecarIndex`.

And would `startingLogFileName` be a bit more accurate variable name?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
mmmyr: Great callout! You're right — our row number isn’t directly tied to the batch size. It’s simply the index of the last row returned within the last file read during pagination. So, when we return a row index in the pagination token, it’s relative to that file — not to the overall dataset or batch size.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
```diff
@@ -0,0 +1,164 @@
...
+
+  public static PageToken fromRow(Row row) {
+    requireNonNull(row);
+    if (!PAGE_TOKEN_SCHEMA.equals(row.getSchema())) {
+      throw new IllegalArgumentException(
+          "Invalid Page Token: input row schema does not match expected PageToken schema. "
+              + "Expected: "
+              + PAGE_TOKEN_SCHEMA
+              + ", Got: "
+              + row.getSchema());
```
scottsand-db: I think all of this could fit on one line.
You could also use `String.format` here, too.
Did you want the different schemas printed on different lines? Are you missing some `\n` here?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
```diff
@@ -0,0 +1,164 @@
...
+          .add("sidecarIndex", LongType.LONG)
+          .add("kernelVersion", StringType.STRING)
+          .add("tablePath", StringType.STRING)
+          .add("tableVersion", LongType.LONG)
+          .add("predicateHash", LongType.LONG)
+          .add("logSegmentHash", LongType.LONG);
+
+  // ===== Variables to mark where the last page ended (and the current page starts) =====
+  /**
+   * The name of the log file where the current page starts. This is the same as the last log file
```
scottsand-db: will this always be the case?
what if we are able to detect that this columnar batch is the **last** columnar batch in the file?
will we then set this `startingLogFileName` to the **next** file?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
```diff
@@ -0,0 +1,164 @@
...
+   * read in the previous page.
+   */
+  private final String startingLogFileName;
+  /**
```
scottsand-db: nit: newline

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
```diff
@@ -0,0 +1,164 @@
...
+   * The index of the last row that was returned from the starting log file during the previous
+   * page. This row index is relative to the file. The current page should begin from the row
+   * immediately after this row index.
+   */
+  private final long lastReturnedRowIndex;
```
scottsand-db, line 85: why is it the **last row index** and not the **next row index** ?

if I read 1000 rows in json file 007.json, which is rows 0 to 999 inclusive, do we return 999 or 1000? I would think we should return 1000?

mmmyr: In my prototype, I returned 999, and I believe this makes more sense than returning the next row index. For example, suppose 007.json contains 1,000 rows (indexed 0 to 999), and the last page returns rows 0 to 999. If we return "007.json" + row index 1000, it's a bit odd—row index 1000 doesn't actually exist in that file.

Conceptually, I feel like it's more accurate to return the last row index read, along with the last file name read. This way, we're recording the precise position where the last page ended, and we can unambiguously start the next page from the row that comes immediately after. I'll change the variable name "startingLogFileName" into "lastReadLogFileName", and also change "startingSidecarFileIdx" to "lastReadSidecarFileIdx".

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
mmmyr: Right now it's always the case. You are right, probably it's better to say: the last log file read in the previous page.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
mmmyr: We can either record:
1. last file & last row index of the previous page. 
2. starting file & starting row index of the current page. 

But we won't be able to know the real file name of next batch to read until callers tries to get the next batch. If we want to return the start position of next page, we want an accurate starting file name + starting row index.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
```diff
@@ -0,0 +1,160 @@
...
+        row.getLong(1), // rowIndexInFile
+        Optional.ofNullable(row.isNullAt(2) ? null : row.getLong(2)), // sidecarIndex
+        row.getString(3), // kernelVersion
+        row.getString(4), // tablePath
+        row.getLong(5), // tableVersion
+        row.getLong(6), // predicateHash
+        row.getLong(7)); // logSegmentHash
+  }
+
+  /** Schema for PageToken Row representation */
```
scottsand-db: nit: superfluous comment :) you can delete this

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
```diff
@@ -0,0 +1,160 @@
...
+  public static final StructType PAGE_TOKEN_SCHEMA =
+      new StructType()
+          .add("logFileName", StringType.STRING)
```
scottsand-db: can you please explicitly add the nullability here? `, false /* nullable */)` etc

and make the nullability for the `sidecarIndex` to true?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
```diff
@@ -0,0 +1,160 @@
...
+
+  public static PageToken fromRow(Row row) {
+    requireNonNull(row);
+    checkArgument(
+        PAGE_TOKEN_SCHEMA.equals(row.getSchema()),
+        String.format(
+            "Invalid Page Token: input row schema does not match expected PageToken schema.\nExpected: %s\nGot: %s",
+            PAGE_TOKEN_SCHEMA, row.getSchema()));
+
+    for (int i = 0; i < PAGE_TOKEN_SCHEMA.length(); i++) {
```
scottsand-db, line 44: a simple comment here would help:

above: `// Check #1: Correct schema`
here: `// Check #2: All required fields are present`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
```diff
@@ -0,0 +1,160 @@
...
+          .add("rowIndexInFile", LongType.LONG)
+          .add("sidecarIndex", LongType.LONG)
+          .add("kernelVersion", StringType.STRING)
+          .add("tablePath", StringType.STRING)
+          .add("tableVersion", LongType.LONG)
+          .add("predicateHash", LongType.LONG)
+          .add("logSegmentHash", LongType.LONG);
+
+  // ===== Variables to mark where the last page ended (and the current page starts) =====
+  /** The last log file read in the previous page. */
```
scottsand-db, line 77: nit: newline between the header block and here

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
scottsand-db: >  I'll change the variable name "startingLogFileName" into "lastReadLogFileName", and also change "startingSidecarFileIdx" to "lastReadSidecarFileIdx".

SGTM. Thanks for being very clear on this detail and semantic!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
```diff
@@ -0,0 +1,160 @@
...
+  private final String lastReadLogFileName;
+
+  /**
+   * The index of the last row that was returned from the starting log file during the previous
```
scottsand-db: from the `starting log file`? or from the `last read log file` during the previous page? Let's keep our comments and variable names consistent

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
```diff
@@ -0,0 +1,160 @@
...
+   * page. This row index is relative to the file. The current page should begin from the row
+   * immediately after this row index.
+   */
+  private final long lastReturnedRowIndex;
+
+  /**
+   * Optional index of the last sidecar checkpoint file read in the previous page. If present, it
```
scottsand-db: `index` in / realtive to what?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
```diff
@@ -0,0 +1,160 @@
...
+      String lastReadLogFileName,
+      long lastReturnedRowIndex,
+      Optional<Long> lastReadSidecarFileIdx,
+      String kernelVersion,
+      String tablePath,
+      long tableVersion,
+      long predicateHash,
+      long logSegmentHash) {
+    this.lastReadLogFileName =
+        requireNonNull(lastReadLogFileName, "lastReadLogFileName must not be null");
```
scottsand-db: instead of `must not be` you can just say `x is null` --> will this make our lines shorter and let this fit onto one line? much cleaner :)

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/PageTokenSuite.scala
```diff
@@ -0,0 +1,104 @@
...
+  rowData.put(2, TEST_SIDECAR_INDEX.orElse(null))
+  rowData.put(3, TEST_KERNEL_VERSION)
+  rowData.put(4, TEST_TABLE_PATH)
+  rowData.put(5, TEST_TABLE_VERSION.asInstanceOf[Object])
+  rowData.put(6, TEST_PREDICATE_HASH.asInstanceOf[Object])
+  rowData.put(7, TEST_LOG_SEGMENT_HASH.asInstanceOf[Object])
+
+  val expectedRow = new GenericRow(PageToken.PAGE_TOKEN_SCHEMA, rowData)
+
+  test("Test PageToken.fromRow with valid data") {
```
scottsand-db: nit: you don't need to say `test("Test` --> we know this is a test :)

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/PageTokenSuite.scala
```diff
@@ -0,0 +1,104 @@
...
+    TEST_KERNEL_VERSION,
+    TEST_TABLE_PATH,
+    TEST_TABLE_VERSION,
+    TEST_PREDICATE_HASH,
+    TEST_LOG_SEGMENT_HASH)
+
+  private val rowData: Map[Integer, Object] = new HashMap()
+  rowData.put(0, TEST_FILE_NAME)
+  rowData.put(1, TEST_ROW_INDEX.asInstanceOf[Object])
+  rowData.put(2, TEST_SIDECAR_INDEX.orElse(null))
```
scottsand-db: let's not have business logic `orElse(null)` here.

is it null? make it null
is it not null? then just use that value

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/PageTokenSuite.scala
```diff
@@ -0,0 +1,104 @@
...
+    assert(pageToken.equals(expectedPageToken))
+  }
+
+  test("Test PageToken.toRow with valid data") {
+    val row = expectedPageToken.toRow
+    assert(row.getSchema.equals(PageToken.PAGE_TOKEN_SCHEMA))
+
+    assert(row.getString(0) == TEST_FILE_NAME)
+    assert(row.getLong(1) == TEST_ROW_INDEX)
+    assert(Optional.of(if (row.isNullAt(2)) null else row.getLong(2)) == TEST_SIDECAR_INDEX)
```
scottsand-db: let's not have business logic here `isNullAt`

if it is null, require and assert that it is null

if not, require and assert that it is the value we are expecting

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/PageTokenSuite.scala
```diff
@@ -0,0 +1,104 @@
...
+    assert(row.getLong(7) == TEST_LOG_SEGMENT_HASH)
+  }
+
+  test("E2E: PageToken round-trip: toRow -> fromRow") {
+    val row = expectedPageToken.toRow
+    val reconstructedPageToken = PageToken.fromRow(row)
+    assert(reconstructedPageToken.equals(expectedPageToken))
+  }
+
+  test("PageToken.fromRow throws exception when input row has invalid schema") {
```
scottsand-db: can we test:
- wrong field name
- wrong data type
- also explicitly test the sidecar-can-be-null case

scottsand-db (changes requested):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
scottsand-db: can we be sure to update these with the latest decided variable names? e.g. lastReturnedRowIndex

mmmyr: Yep, I will ping you on slack when I finish updating this PR with latest variable name (and also pagination context one at the same time).

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
mmmyr: ordering of sidecars in the V2 manifest Checkpoint file

scottsand-db: > ordering of sidecars in the V2 manifest Checkpoint file

Can you include this information in the javadoc comment please?

mmmyr: Yep, I already did. will push soon.

scottsand-db (approved):
Super clean code, clean comments, and great tests! Nice!

PR #4853, opened 2025-06-29
add number of true rows

mmmyr:
## 🥞 Stacked PR

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
@@ -37,10 +37,20 @@ import java.util.Optional;
 public class FilteredColumnarBatch {
   private final ColumnarBatch data;
   private final Optional<ColumnVector> selectionVector;
+  // Optional cached count of true rows
+  private Optional<Integer> numOfSelectedRows;
 
   public FilteredColumnarBatch(ColumnarBatch data, Optional<ColumnVector> selectionVector) {
     this.data = data;
     this.selectionVector = selectionVector;
+    this.numOfSelectedRows = Optional.empty();
+  }
+
+  public FilteredColumnarBatch(
+      ColumnarBatch data, Optional<ColumnVector> selectionVector, int numOfTrueRows) {
+    this.data = data;
+    this.selectionVector = selectionVector;
+    this.numOfSelectedRows = Optional.of(numOfTrueRows);
   }
 
   /**
@@ -107,5 +117,25 @@ public class FilteredColumnarBatch {
       @Override
       public void close() {}
     };
+  }
+
+  public Integer getNumOfSelectedRows() {
+    if (numOfSelectedRows.isPresent()) {
+      return numOfSelectedRows.get();
+    }
+    if (!selectionVector.isPresent()) {
+      // TODO: what to return if selection vector is empty?
+      return data.getSize(); // all rows are selected
+    }
+    int trueRowCount = 0;
+    for (int rowId = 0; rowId < data.getSize(); rowId++) {
+      if (selectionVector.get().isNullAt(rowId)) {
+        boolean isSelected =
+            !selectionVector.get().isNullAt(rowId) && selectionVector.get().getBoolean(rowId);
+        if (isSelected) trueRowCount++;
+      }
+    }
+    numOfSelectedRows = Optional.of(trueRowCount);
+    return trueRowCount;
   }
 }
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
@@ -37,20 +37,23 @@ import java.util.Optional;
 public class FilteredColumnarBatch {
   private final ColumnarBatch data;
   private final Optional<ColumnVector> selectionVector;
+  private final String filePath;
   // Optional cached count of true rows
   private Optional<Integer> numOfSelectedRows;
 
-  public FilteredColumnarBatch(ColumnarBatch data, Optional<ColumnVector> selectionVector) {
+  public FilteredColumnarBatch(ColumnarBatch data, Optional<ColumnVector> selectionVector, String filePath) {
     this.data = data;
     this.selectionVector = selectionVector;
     this.numOfSelectedRows = Optional.empty();
+    this.filePath = filePath;
   }
 
   public FilteredColumnarBatch(
-      ColumnarBatch data, Optional<ColumnVector> selectionVector, int numOfTrueRows) {
+      ColumnarBatch data, Optional<ColumnVector> selectionVector, String filePath, int numOfTrueRows) {
     this.data = data;
     this.selectionVector = selectionVector;
     this.numOfSelectedRows = Optional.of(numOfTrueRows);
+    this.filePath = filePath;
   }
 
   /**
@@ -62,6 +65,10 @@ public class FilteredColumnarBatch {
    */
   public ColumnarBatch getData() {
     return data;
+  }
+
+  public String getFilePath() {
+    return filePath;
   }
 
   /**
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
@@ -41,7 +41,8 @@ public class FilteredColumnarBatch {
   // Optional cached count of true rows
   private Optional<Integer> numOfSelectedRows;
 
-  public FilteredColumnarBatch(ColumnarBatch data, Optional<ColumnVector> selectionVector, String filePath) {
+  public FilteredColumnarBatch(
+      ColumnarBatch data, Optional<ColumnVector> selectionVector, String filePath) {
     this.data = data;
     this.selectionVector = selectionVector;
     this.numOfSelectedRows = Optional.empty();
@@ -49,7 +50,10 @@ public class FilteredColumnarBatch {
   }
 
   public FilteredColumnarBatch(
-      ColumnarBatch data, Optional<ColumnVector> selectionVector, String filePath, int numOfTrueRows) {
+      ColumnarBatch data,
+      Optional<ColumnVector> selectionVector,
+      String filePath,
+      int numOfTrueRows) {
     this.data = data;
     this.selectionVector = selectionVector;
     this.numOfSelectedRows = Optional.of(numOfTrueRows);
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
@@ -283,7 +283,8 @@ public class ScanImpl implements Scan {
                 () -> predicateEvaluator.eval(next.getData(), next.getSelectionVector()),
                 "Evaluating the partition expression %s",
                 predicateOnScanFileBatch);
-        return new FilteredColumnarBatch(next.getData(), Optional.of(newSelectionVector), next.getFilePath());
+        return new FilteredColumnarBatch(
+            next.getData(), Optional.of(newSelectionVector), next.getFilePath());
       }
 
       @Override
@@ -347,7 +348,9 @@ public class ScanImpl implements Scan {
                   filterToEval);
 
           return new FilteredColumnarBatch(
-              filteredScanFileBatch.getData(), Optional.of(newSelectionVector),filteredScanFileBatch.getFilePath());
+              filteredScanFileBatch.getData(),
+              Optional.of(newSelectionVector),
+              filteredScanFileBatch.getFilePath());
         });
   }
 
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
@@ -283,7 +283,7 @@ public class ScanImpl implements Scan {
                 () -> predicateEvaluator.eval(next.getData(), next.getSelectionVector()),
                 "Evaluating the partition expression %s",
                 predicateOnScanFileBatch);
-        return new FilteredColumnarBatch(next.getData(), Optional.of(newSelectionVector));
+        return new FilteredColumnarBatch(next.getData(), Optional.of(newSelectionVector), next.getFilePath());
       }
 
       @Override
@@ -347,7 +347,7 @@ public class ScanImpl implements Scan {
                   filterToEval);
 
           return new FilteredColumnarBatch(
-              filteredScanFileBatch.getData(), Optional.of(newSelectionVector));
+              filteredScanFileBatch.getData(), Optional.of(newSelectionVector),filteredScanFileBatch.getFilePath());
         });
   }
 
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActiveAddFilesIterator.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActiveAddFilesIterator.java
@@ -275,7 +275,9 @@ public class ActiveAddFilesIterator implements CloseableIterator<FilteredColumna
                           .createSelectionVector(selectionVectorBuffer, 0, addsVector.getSize()),
                   "Create selection vector for selected scan files"));
     }
-    next = Optional.of(new FilteredColumnarBatch(scanAddFiles, selectionColumnVector, _next.getFilePath()));
+    next =
+        Optional.of(
+            new FilteredColumnarBatch(scanAddFiles, selectionColumnVector, _next.getFilePath()));
   }
 
   public static String getAddFilePath(ColumnVector addFileVector, int rowId) {
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActiveAddFilesIterator.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActiveAddFilesIterator.java
@@ -275,7 +275,7 @@ public class ActiveAddFilesIterator implements CloseableIterator<FilteredColumna
                           .createSelectionVector(selectionVectorBuffer, 0, addsVector.getSize()),
                   "Create selection vector for selected scan files"));
     }
-    next = Optional.of(new FilteredColumnarBatch(scanAddFiles, selectionColumnVector));
+    next = Optional.of(new FilteredColumnarBatch(scanAddFiles, selectionColumnVector, _next.getFilePath()));
   }
 
   public static String getAddFilePath(ColumnVector addFileVector, int rowId) {
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
@@ -19,6 +19,7 @@ package io.delta.kernel;
 import io.delta.kernel.annotation.Evolving;
 import io.delta.kernel.data.*;
 import io.delta.kernel.engine.Engine;
+import io.delta.kernel.engine.FileReadResult;
 import io.delta.kernel.expressions.Predicate;
 import io.delta.kernel.internal.InternalScanFileUtils;
 import io.delta.kernel.internal.actions.DeletionVectorDescriptor;
@@ -133,7 +134,7 @@ public interface Scan {
    * @throws IOException when error occurs while reading the data.
    */
   static CloseableIterator<FilteredColumnarBatch> transformPhysicalData(
-      Engine engine, Row scanState, Row scanFile, CloseableIterator<ColumnarBatch> physicalDataIter)
+      Engine engine, Row scanState, Row scanFile, CloseableIterator<FileReadResult> physicalDataIter)
       throws IOException {
     return new CloseableIterator<FilteredColumnarBatch>() {
       boolean inited = false;
@@ -171,7 +172,9 @@ public interface Scan {
       @Override
       public FilteredColumnarBatch next() {
         initIfRequired();
-        ColumnarBatch nextDataBatch = physicalDataIter.next();
+        FileReadResult fileReadResult = physicalDataIter.next();
+        ColumnarBatch nextDataBatch = fileReadResult.getData();
+        String filePath = fileReadResult.getFilePath();
 
         DeletionVectorDescriptor dv =
             InternalScanFileUtils.getDeletionVectorDescriptorFromRow(scanFile);
@@ -223,7 +226,7 @@ public interface Scan {
                 "Column mapping mode is not yet supported: " + columnMappingMode);
         }
 
-        return new FilteredColumnarBatch(nextDataBatch, selectionVector);
+        return new FilteredColumnarBatch(nextDataBatch, selectionVector,filePath);
       }
     };
   }
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
@@ -134,7 +134,10 @@ public interface Scan {
    * @throws IOException when error occurs while reading the data.
    */
   static CloseableIterator<FilteredColumnarBatch> transformPhysicalData(
-      Engine engine, Row scanState, Row scanFile, CloseableIterator<FileReadResult> physicalDataIter)
+      Engine engine,
+      Row scanState,
+      Row scanFile,
+      CloseableIterator<FileReadResult> physicalDataIter)
       throws IOException {
     return new CloseableIterator<FilteredColumnarBatch>() {
       boolean inited = false;
@@ -226,7 +229,7 @@ public interface Scan {
                 "Column mapping mode is not yet supported: " + columnMappingMode);
         }
 
-        return new FilteredColumnarBatch(nextDataBatch, selectionVector,filePath);
+        return new FilteredColumnarBatch(nextDataBatch, selectionVector, filePath);
       }
     };
   }
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/Transaction.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/Transaction.java
@@ -189,7 +189,8 @@ public interface Transaction {
             }
             data = data.withDeletedColumnAt(partitionColIndex);
           }
-          return new FilteredColumnarBatch(data, filteredBatch.getSelectionVector(), filteredBatch.getFilePath());
+          return new FilteredColumnarBatch(
+              data, filteredBatch.getSelectionVector(), filteredBatch.getFilePath());
         });
   }
 
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/Transaction.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/Transaction.java
@@ -189,7 +189,7 @@ public interface Transaction {
             }
             data = data.withDeletedColumnAt(partitionColIndex);
           }
-          return new FilteredColumnarBatch(data, filteredBatch.getSelectionVector());
+          return new FilteredColumnarBatch(data, filteredBatch.getSelectionVector(), filteredBatch.getFilePath());
         });
   }
 
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionWrapper.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionWrapper.java
@@ -27,9 +27,13 @@ public class ActionWrapper {
   /* Timestamp of the commit file if isFromCheckpoint=false */
   private final Optional<Long> timestamp;
 
-  //TODO: remove isFromCheckpointFile?; but how to deal with V2 manifest file?
+  // TODO: remove isFromCheckpointFile?; but how to deal with V2 manifest file?
   ActionWrapper(
-      ColumnarBatch data, boolean isFromCheckpoint, long version, Optional<Long> timestamp, String filePath) {
+      ColumnarBatch data,
+      boolean isFromCheckpoint,
+      long version,
+      Optional<Long> timestamp,
+      String filePath) {
     this.columnarBatch = data;
     this.isFromCheckpoint = isFromCheckpoint;
     this.version = version;
@@ -53,5 +57,7 @@ public class ActionWrapper {
     return timestamp;
   }
 
-  public String getFilePath() {return filePath;}
+  public String getFilePath() {
+    return filePath;
+  }
 }
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionWrapper.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionWrapper.java
@@ -23,15 +23,18 @@ public class ActionWrapper {
   private final ColumnarBatch columnarBatch;
   private final boolean isFromCheckpoint;
   private final long version;
+  private final String filePath;
   /* Timestamp of the commit file if isFromCheckpoint=false */
   private final Optional<Long> timestamp;
 
+  //TODO: remove isFromCheckpointFile?; but how to deal with V2 manifest file?
   ActionWrapper(
-      ColumnarBatch data, boolean isFromCheckpoint, long version, Optional<Long> timestamp) {
+      ColumnarBatch data, boolean isFromCheckpoint, long version, Optional<Long> timestamp, String filePath) {
     this.columnarBatch = data;
     this.isFromCheckpoint = isFromCheckpoint;
     this.version = version;
     this.timestamp = timestamp;
+    this.filePath = filePath;
   }
 
   public ColumnarBatch getColumnarBatch() {
@@ -49,4 +52,6 @@ public class ActionWrapper {
   public Optional<Long> getTimestamp() {
     return timestamp;
   }
+
+  public String getFilePath() {return filePath;}
 }
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
@@ -337,8 +337,8 @@ public class ActionsIterator implements CloseableIterator<ActionWrapper> {
             // If the checkpoint file is a UUID or classic checkpoint, read the top-level
             // checkpoint file and any potential sidecars. Otherwise, look for any other
             // parts of the current multipart checkpoint.
-            CloseableIterator<ColumnarBatch> dataIter =
-                getActionsIterFromSinglePartOrV2Checkpoint(nextFile, fileName);
+            CloseableIterator<FileReadResult> dataIter =
+                getActionsIterFromSinglePartOrV2Checkpoint(nextFile, fileName).map(batch -> new FileReadResult(batch, nextFile.getPath()));
             long version = checkpointVersion(nextFilePath);
             return combine(dataIter, true /* isFromCheckpoint */, version, Optional.empty());
           }
@@ -350,13 +350,12 @@ public class ActionsIterator implements CloseableIterator<ActionWrapper> {
             // optimizations like reading multiple files in parallel.
             CloseableIterator<FileStatus> checkpointFiles =
                 retrieveRemainingCheckpointFiles(nextLogFile);
-            CloseableIterator<ColumnarBatch> dataIter =
+            CloseableIterator<FileReadResult> dataIter =
                 wrapEngineExceptionThrowsIO(
                     () ->
                         engine
                             .getParquetHandler()
-                            .readParquetFiles(checkpointFiles, deltaReadSchema, checkpointPredicate)
-                            .map(FileReadResult::getData),
+                            .readParquetFiles(checkpointFiles, deltaReadSchema, checkpointPredicate),
                     "Reading checkpoint sidecars [%s] with readSchema=%s and predicate=%s",
                     checkpointFiles,
                     deltaReadSchema,
@@ -380,17 +379,16 @@ public class ActionsIterator implements CloseableIterator<ActionWrapper> {
     // version with actions read from the JSON file for further optimizations later
     // on (faster metadata & protocol loading in subsequent runs by remembering
     // the version of the last version where the metadata and protocol are found).
-    final CloseableIterator<ColumnarBatch> dataIter =
+    final CloseableIterator<FileReadResult> dataIter =
         wrapEngineExceptionThrowsIO(
             () ->
                 engine
                     .getJsonHandler()
                     .readJsonFiles(
-                        singletonCloseableIterator(nextFile), deltaReadSchema, Optional.empty()),
+                        singletonCloseableIterator(nextFile), deltaReadSchema, Optional.empty()).map(batch -> new FileReadResult(batch, nextFile.getPath())),
             "Reading JSON log file `%s` with readSchema=%s",
             nextFile,
             deltaReadSchema);
-
     return combine(
         dataIter,
         false /* isFromCheckpoint */,
@@ -406,7 +404,7 @@ public class ActionsIterator implements CloseableIterator<ActionWrapper> {
    * of the file.
    */
   private CloseableIterator<ActionWrapper> combine(
-      CloseableIterator<ColumnarBatch> fileReadDataIter,
+      CloseableIterator<FileReadResult> fileReadDataIter,
       boolean isFromCheckpoint,
       long version,
       Optional<Long> timestamp) {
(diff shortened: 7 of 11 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
```diff
@@ -37,10 +37,27 @@
...
     this.selectionVector = selectionVector;
+    this.numOfSelectedRows = Optional.empty();
+    this.filePath = filePath;
+  }
+
+  public FilteredColumnarBatch(
+      ColumnarBatch data,
+      Optional<ColumnVector> selectionVector,
+      String filePath,
+      int numOfTrueRows) {
```
mmmyr: Change line 54: into a required field

mmmyr: If there is no selection vecotor -> number of true rows should also be emtpy.

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
```diff
@@ -108,4 +129,24 @@ public Row next() {
       public void close() {}
     };
   }
+
+  public Integer getNumOfSelectedRows() {
```
mmmyr: should be int.

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
mmmyr: getPreComputedNumSelectedRows -> this might return Optional.empty()

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
```diff
@@ -37,10 +37,23 @@
...
+  // Optional cached count of true rows in the selection vector
+  private Optional<Integer> numSelectedRows;
 
   public FilteredColumnarBatch(ColumnarBatch data, Optional<ColumnVector> selectionVector) {
+    this(data, selectionVector, Optional.empty(), Optional.empty());
+  }
+
+  public FilteredColumnarBatch(
+      ColumnarBatch data,
+      Optional<ColumnVector> selectionVector,
```
scottsand-db, line 56: I thought we agreed that this would take in a non-optional selection vector? and that the `numSelectedRows` would also be nonOptional?

mmmyr: In ActiveAddFilesIterator, the selection vector can be empty if all rows are selected. So we can't make this constructor take in a non-optional selection vector. I think a better approach is: 1. if selection vector is empty -> all rows are selected and we return the total number of rows directly; 2. if selection vector isn't empty -> we directly return the optional value of numSelectedRows.

mmmyr:
```
Optional<ColumnVector> selectionColumnVector = Optional.empty();
    if (atLeastOneUnselected) {
      selectionColumnVector =
          Optional.of(
              wrapEngineException(
                  () ->
                      engine
                          .getExpressionHandler()
                          .createSelectionVector(selectionVectorBuffer, 0, addsVector.getSize()),
                  "Create selection vector for selected scan files"));
    }
    next =
        Optional.of(
            new FilteredColumnarBatch(
                scanAddFiles,
                selectionColumnVector,
                Optional.of(_next.getFilePath()),
                Optional.of(numSelectedRows)));
```

mmmyr: But we can make numSelectedRows in constructor non-optional here. Even if selection vector is empty, numSelectedRows has value (should equal to size of batch).

scottsand-db: SGTM, thanks for explaining!

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
```diff
@@ -37,10 +37,23 @@
 public class FilteredColumnarBatch {
   private final ColumnarBatch data;
   private final Optional<ColumnVector> selectionVector;
+  private final Optional<String> filePath;
+  // Optional pre-computed count of true rows in the selection vector
+  private Optional<Integer> numSelectedRows;
```
scottsand-db: nit: in general, for **internal code** when you have a detailed comment that explains what the variable is, try and just see if you can make the variable name more self-descriptive, and then we might not need the comment at all.

e.g. if this was `numPreComputedSelectedRows` -- i don't think we would need the comment then at all?

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
mmmyr:
```
public FilteredColumnarBatch(
      ColumnarBatch data,
      Optional<ColumnVector> selectionVector,
      Optional<String> filePath,
      int numSelectedRows)
```

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
```diff
@@ -108,4 +125,22 @@ public Row next() {
       public void close() {}
     };
   }
+
+  /**
+   * Returns the pre-computed number of selected rows if available.
+   *
+   * <p>If a selection vector is not present, it implies all rows are selected, and the method
```
scottsand-db: I think some simpler method docs could be something along the lines of:

```java
<p>If present, the number of selected rows was pre-computed and is retuned with no additional computation cost. This will happen in two cases: First, when the selection vector is empty and so we know the number of selected rows is equal to the batch size; and Second when the number of selected rows is .....
```

Basically, focus on what we are returning, instead of focusing on the selection vector

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
scottsand-db: yeah it seems like we can take in numSelectedRows to be non-optional

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
```diff
@@ -37,10 +37,23 @@
...
+  }
+
+  public FilteredColumnarBatch(
+      ColumnarBatch data,
+      Optional<ColumnVector> selectionVector,
+      Optional<String> filePath,
+      Optional<Integer> numSelectedRows) {
     this.data = data;
     this.selectionVector = selectionVector;
+    this.numSelectedRows = numSelectedRows;
```
scottsand-db: are there any weird cases we should check for here?

what if the batch size is 1000, the selection vector is **empty**, but we get a numSelectedRows of 50?

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
```diff
@@ -37,10 +37,26 @@
 public class FilteredColumnarBatch {
   private final ColumnarBatch data;
   private final Optional<ColumnVector> selectionVector;
+  private final Optional<String> filePath;
+  private final Optional<Integer> preComputedNumSelectedRows;
 
   public FilteredColumnarBatch(ColumnarBatch data, Optional<ColumnVector> selectionVector) {
+    this(data, selectionVector, null, null);
```
scottsand-db: so this is raising some alarm bells for me -- why is working with `null` better?

I _think_ what you want to accomplish is:

- one constructor that takes in (data, selectionVectorOpt)
- another constructor that takes in (data, selectionVectorOpt, filePath, preComputedNumSelectedRows)

and I _think_ you are running into the problem of: you are trying to force these to use the same constructor code? i.e. you are having the 1st constructor call `this` ??

but why all this complexity? just have two super small, super simple constructors? and no nulls? and super clear semantics?

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
```diff
@@ -108,4 +128,23 @@ public Row next() {
       public void close() {}
     };
   }
+
+  /**
+   * @return an {@link Optional} containing the pre-computed number of selected rows, if available.
```
scottsand-db, line 152: Great!

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
```diff
@@ -108,4 +128,23 @@ public Row next() {
...
+   *     <ul>
+   *       <li>When the selection vector is absent, which implies that all rows are selected — in
+   *           this case, the number of selected rows is equal to the batch size.
+   *       <li>When the number of selected rows was explicitly pre-computed and passed in.
+   *     </ul>
+   *     <p>If empty, the caller must compute the number of selected rows manually from the
+   *     selection vector.
+   */
+  public Optional<Integer> getPreComputedNumSelectedRows() {
+    if (!selectionVector.isPresent()) {
```
scottsand-db: I would have expected that the first param this method checks is `preComputedNumSelectedRows`?

if it is present, just return it!

if it is not present, then check selection vector. if it is empty, return a new optional of data.size, else return null

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
mmmyr: Yep, that makes sense. My initial thinking was to avoid duplicating the assignment logic (like setting data and selectionVector) so that if we ever need to update something, it only needs to change one constructor. But I agree that having two simple and explicit constructors

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
scottsand-db: actually ... can't we just know this ahead of time? in the constructor, we just create it correct and we can return it here with no `if`s at all?

- `public FilteredColumnarBatch(ColumnarBatch data, Optional<ColumnVector> selectionVector) {` --> if selection vector is empty, then `preComputedNumSelectedRows` is equal to Optional.of(data.getSize());
- public FilteredColumnarBatch(ColumnarBatch data, Optional<ColumnVector> selectionVector, filePath, preComputedNumSelectedRows) --> just return it here!

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
scottsand-db: That can be achieved by having 2 constructors and 1 public static factory if youd like

Though, this isn't very complex, and we can always add that later

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
mmmyr: Great call out! I'll do that. Thanks!

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
mmmyr: Added to TODO

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
```diff
@@ -37,10 +37,29 @@
...
+
+  public FilteredColumnarBatch(
+      ColumnarBatch data,
+      Optional<ColumnVector> selectionVector,
+      String filePath,
+      Integer numSelectedRows) {
+    this.data = data;
+    this.selectionVector = selectionVector;
+    this.filePath = Optional.of(filePath);
+    assert selectionVector.isPresent() || numSelectedRows == data.getSize()
```
scottsand-db: you can use `checkArgument` here

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
```diff
@@ -37,10 +37,29 @@
...
+    this.filePath = Optional.empty();
+    this.preComputedNumSelectedRows =
+        !selectionVector.isPresent() ? Optional.of(data.getSize()) : Optional.empty();
+  }
+
+  public FilteredColumnarBatch(
+      ColumnarBatch data,
+      Optional<ColumnVector> selectionVector,
+      String filePath,
+      Integer numSelectedRows) {
```
scottsand-db: do we want to be able to accept `null`? would `int` be better?

mmmyr: Yep int works here, i forget to change. Sorry

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
scottsand-db: and i think this is missing a null check?

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
```diff
@@ -37,10 +39,30 @@
...
+      ColumnarBatch data,
+      Optional<ColumnVector> selectionVector,
+      String filePath,
+      int numSelectedRows) {
+    this.data = data;
+    this.selectionVector = selectionVector;
+    this.filePath = Optional.of(filePath);
+    checkArgument(
+        selectionVector.isPresent() || numSelectedRows == data.getSize(),
+        "Invalid precomputedNumSelectedRows: must be null or equal to batch size when selectionVector is empty.");
```
scottsand-db: nit: out of date comment --> `null`

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
```diff
@@ -37,10 +39,30 @@
...
+    this.filePath = Optional.empty();
+    this.preComputedNumSelectedRows =
+        !selectionVector.isPresent() ? Optional.of(data.getSize()) : Optional.empty();
+  }
+
+  public FilteredColumnarBatch(
+      ColumnarBatch data,
+      Optional<ColumnVector> selectionVector,
+      String filePath,
+      int numSelectedRows) {
```
scottsand-db: can `numSelectedRows` be less than 0?

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
```diff
@@ -37,10 +39,30 @@
...
+    this.data = data;
+    this.selectionVector = selectionVector;
+    this.filePath = Optional.of(filePath);
+    checkArgument(
```
scottsand-db, line 61: and this warrants a tiny `FilteredColumnarBatchSuite` btw

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
mmmyr: I'll add a check

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
mmmyr: Okay I'll add one

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
mmmyr: If numSelectedRows is 'int' type, we don't need a null check? Just make sure it is larger than or equal to 0 should be fine?

mmmyr: If filePath is null, it will throw exception when calling Optional.of()

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
```diff
@@ -37,10 +39,30 @@
 public class FilteredColumnarBatch {
   private final ColumnarBatch data;
   private final Optional<ColumnVector> selectionVector;
+  private final Optional<String> filePath;
```
huan233usc, line 42: Can we consider not coupling filePath to ColumnBatch here? FilteredColumnarBatch seems to be a very generic class

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -118,7 +123,7 @@ public CloseableIterator<FilteredColumnarBatch> getScanFiles(Engine engine) {
    * @return the surviving scan files as {@link FilteredColumnarBatch}s
    */
   public CloseableIterator<FilteredColumnarBatch> getScanFiles(
-      Engine engine, boolean includeStats) {
+      Engine engine, boolean includeStats, boolean isPaginated) {
```
mmmyr: This will be PaginationContext eventually.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActiveAddFilesIterator.java
```diff
@@ -275,7 +285,12 @@ private void prepareNext() {
                           .createSelectionVector(selectionVectorBuffer, 0, addsVector.getSize()),
                   "Create selection vector for selected scan files"));
     }
-    next = Optional.of(new FilteredColumnarBatch(scanAddFiles, selectionColumnVector));
+    if (isPaginated) {
+      next =
+          Optional.of(
+              new PaginationFilteredColumnarBatch(
+                  scanAddFiles, selectionColumnVector, _next.getFilePath(), numSelectedRows));
+    } else next = Optional.of(new FilteredColumnarBatch(scanAddFiles, selectionColumnVector));
```
mmmyr: In PaginatedIteratorImpl, we can have sth like: PaginatedFileteredColumnarBatch batch = originalIterator.next() ... , then getFilePath on batch.

kernel/kernel-api/src/main/java/io/delta/kernel/data/PaginationFilteredColumnarBatch.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.Optional;
+
+public class PaginationFilteredColumnarBatch extends FilteredColumnarBatch {
+  private final Optional<String> filePath;
+  private final Optional<Integer> preComputedNumSelectedRows;
+
+  public PaginationFilteredColumnarBatch(
+      ColumnarBatch data,
+      Optional<ColumnVector> selectionVector,
+      String filePath,
```
scottsand-db: does not need to be optional now, eh? this is an internal class we are producing that is only for paginated log replay

kernel/kernel-api/src/main/java/io/delta/kernel/data/PaginationFilteredColumnarBatch.java
```diff
@@ -0,0 +1,43 @@
+package io.delta.kernel.data;
```
scottsand-db: and move to internal, eh?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/data/PaginationFilteredColumnarBatch.java
```diff
@@ -0,0 +1,53 @@
...
+        "Invalid precomputedNumSelectedRows: must be equal to batch size when selectionVector is empty.");
+    this.preComputedNumSelectedRows = numSelectedRows;
+  }
+
+  public String getFilePath() {
+    return filePath;
+  }
+
+  /**
+   * @return an {@link Optional} containing the pre-computed number of selected rows, if available.
```
raveeram-db: looks like the comment here is a bit out of date?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/LogReplay.java
```diff
@@ -215,7 +215,8 @@ public CloseableIterator<FilteredColumnarBatch> getAddFilesAsColumnarBatches(
       Engine engine,
       boolean shouldReadStats,
       Optional<Predicate> checkpointPredicate,
-      ScanMetrics scanMetrics) {
+      ScanMetrics scanMetrics,
+      boolean isPagianted) {
```
raveeram-db: Spelling Nit: `isPaginated `

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/PaginationFilteredColumnarBatchSuite.scala
```diff
@@ -0,0 +1,94 @@
...
+import org.scalatest.matchers.should.Matchers
+
+class PaginationFilteredColumnarBatchSuite extends AnyFunSuite with VectorTestUtils with Matchers {
+
+  test("constructor should succeed when selectionVector is present and numSelectedRows is valid") {
+    val data = createTestColumnarBatch(5)
+    val selectionVector = Optional.of(booleanVector(Seq(true, false, true, false, true)))
+    val batch = new PaginationFilteredColumnarBatch(data, selectionVector, "/test/path", 3)
+
+    batch.getFilePath shouldBe "/test/path"
```
raveeram-db: Nit: Should we use the classic style assertions to be consistent with the rest of the code base instead of these Matcher DSLs

raveeram-db (approved):
Overall looks good, just a few nits.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -105,6 +105,18 @@ public CloseableIterator<FilteredColumnarBatch> getScanFiles(Engine engine) {
...
+  /**
+   * Get an iterator of data files in this version of scan that survived the predicate pruning.
+   *
+   * @param engine the {@link Engine} instance to use
+   * @param includeStats whether to read and include the JSON statistics
+   * @return data in {@link ColumnarBatch} batch format. Each row correspond to one survived file.
+   */
+  public CloseableIterator<FilteredColumnarBatch> getScanFiles(
+      Engine engine, boolean includeStats) {
+    return getScanFiles(engine, includeStats, false);
```
scottsand-db: nit: do we need this pass-through method today? can't we just call `getScanFiles(engine, includeStats, false);` directly?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/data/PaginationFilteredColumnarBatch.java
```diff
@@ -0,0 +1,53 @@
...
+  private final int preComputedNumSelectedRows;
+
+  public PaginationFilteredColumnarBatch(
+      ColumnarBatch data,
+      Optional<ColumnVector> selectionVector,
+      String filePath,
+      int numSelectedRows) {
+    super(data, selectionVector);
+    this.filePath = filePath;
+    checkArgument(
```
scottsand-db: nit: I think the convention here would be to do the `checkArgument`s first and then do the `this.filePath = filepath` assignment

btw, great assertions!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/data/PaginationFilteredColumnarBatch.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+public class PaginationFilteredColumnarBatch extends FilteredColumnarBatch {
+  private final String filePath;
+  private final int preComputedNumSelectedRows;
+
+  public PaginationFilteredColumnarBatch(
+      ColumnarBatch data,
+      Optional<ColumnVector> selectionVector,
+      String filePath,
+      int numSelectedRows) {
```
scottsand-db: `preComputedNumSelectedRows`

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/PaginationFilteredColumnarBatchSuite.scala
```diff
@@ -0,0 +1,94 @@
...
+  test("constructor should throw IllegalArgumentException " +
+    "when selectionVector is empty and numSelectedRows != batch size") {
+    val data = createTestColumnarBatch(5)
+    val selectionVector = Optional.empty[ColumnVector]()
+    val exception = intercept[IllegalArgumentException] {
+      new PaginationFilteredColumnarBatch(data, selectionVector, "/test/path", 3)
+    }
+
+    exception.getMessage should include("Invalid precomputedNumSelectedRows")
+    exception.getMessage should include(
```
scottsand-db: tip: you can do `val exMsg = intercept[XXX} { stuff }.getMessage` to make these lines shorter and fit on one line

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/PaginationFilteredColumnarBatchSuite.scala
```diff
@@ -0,0 +1,94 @@
...
+    val exception = intercept[IllegalArgumentException] {
+      new PaginationFilteredColumnarBatch(data, selectionVector, "/test/path", 5)
+    }
+
+    exception.getMessage should include("Invalid precomputedNumSelectedRows")
+    exception.getMessage should include(
+      "no larger than batch size")
+  }
+
+  private def createTestColumnarBatch(size: Int): ColumnarBatch = {
```
scottsand-db: can you ask cursor if there are any other places already in the code where we create simple columnar batches? to avoid defining more code here?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -102,7 +102,7 @@ public ScanImpl(
    */
   @Override
   public CloseableIterator<FilteredColumnarBatch> getScanFiles(Engine engine) {
-    return getScanFiles(engine, false);
+    return getScanFiles(engine, false, false);
```
huan233usc: nit: Add document
getScanFiles(engine, includeStats/*isPaginated*/ , false /*isPaginated*/)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActiveAddFilesIterator.java
```diff
@@ -275,7 +285,12 @@ private void prepareNext() {
                           .createSelectionVector(selectionVectorBuffer, 0, addsVector.getSize()),
                   "Create selection vector for selected scan files"));
     }
-    next = Optional.of(new FilteredColumnarBatch(scanAddFiles, selectionColumnVector));
+    if (isPaginated) {
```
scottsand-db: do we have to do the `if .. else` here? can't we always return a `PaginationFilteredColumnarBatch`?

(a) we have already done the work to know numSelectedRows++
(b) we already know the `_next.getFilePath()`

Seems like we are just returning a class with two extra fields? im fine with that

You could even update this API to return a `PaginationFilteredColumnarBatch` -- consumers won't care, because a `PaginationFilteredColumnarBatch` is a `FilteredColumnarBatch`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/data/PaginationFilteredColumnarBatch.java
```diff
@@ -0,0 +1,58 @@
...
+import io.delta.kernel.data.ColumnVector;
+import io.delta.kernel.data.ColumnarBatch;
+import io.delta.kernel.data.FilteredColumnarBatch;
+import java.util.Optional;
+
+/**
+ * An extension of {@link FilteredColumnarBatch} that adds metadata for pagination use, such as the
+ * originating file path and a pre-computed count of selected rows.
+ */
+public class PaginationFilteredColumnarBatch extends FilteredColumnarBatch {
```
scottsand-db: is anything here specific to `Pagination`?

could we call this `ExtendedFilteredColumnarBatch` or `FilteredColumnarBatchWithMetadata`?

cc @nicklan for naming input

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActiveAddFilesIterator.java
scottsand-db: cc @nicklan just an FYI in case you disagree

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/PaginationFilteredColumnarBatchSuite.scala
```diff
@@ -0,0 +1,64 @@
...
+    val batch = new PaginationFilteredColumnarBatch(data, selectionVector, "/test/path", 3)
+
+    assert(batch.getFilePath == "/test/path")
+    assert(batch.getPreComputedNumSelectedRows == 3)
+    assert(batch.getData == data)
+    assert(batch.getSelectionVector == selectionVector)
+  }
+
+  test(
+    "constructor should succeed when " +
```
scottsand-db: nit: usually strings take up the full line length, and only then do we do `+` and do a new line with more string content

kernel/kernel-api/src/main/java/io/delta/kernel/internal/data/PaginationFilteredColumnarBatch.java
mmmyr: I'm open to both. Right now only Pagination use these info. But the naming "PaginationFilteredColumnarBatch" is indeed a little bit weird.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActiveAddFilesIterator.java
mmmyr: It makes sense to me. But I feel like if we decide to do that, the name "PaginationFilteredColumnarBatch" is quite confusing. Maybe change to FilteredColumnarBatchWithMetadata?

scottsand-db: xx @huan233usc -- any thoughts?

huan233usc: imo if so then we should have a more general naming `FilteredColumnarBatchWithMetadata`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -115,10 +115,11 @@ public CloseableIterator<FilteredColumnarBatch> getScanFiles(Engine engine) {
    *
    * @param engine the {@link Engine} instance to use
    * @param includeStats whether to read and include the JSON statistics
+   * @param isPaginated whether pagination is enabled
    * @return the surviving scan files as {@link FilteredColumnarBatch}s
    */
   public CloseableIterator<FilteredColumnarBatch> getScanFiles(
-      Engine engine, boolean includeStats) {
+      Engine engine, boolean includeStats, boolean isPaginated) {
```
nicklan: I think it would be a more clear API if we had:
```
public CloseableIterator<FilteredColumnarBatch> getScanFiles(Engine engine) // get a non-paginated scan
public CloseableIterator<FilteredColumnarBatch> getPaginatedScanFiles(Engine engine) // get a paginated scan
```

I guess both would include the stats, I'm unclear on why we need the stats arg?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/data/PaginationFilteredColumnarBatch.java
nicklan: Do we need a new class? Can't we just add an `Optional<String> filePath` and `Optional<int>` to the existing one?


Or actually, since we don't see to do any extra work to compute these in the case that we've said `paginated == true`, why not just add these as private members to the existing class?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActiveAddFilesIterator.java
nicklan: To my above comment, why do we need a new class at all? Let's just add these as fields to the existing one.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
scottsand-db: > I'm unclear on why we need the stats arg?

cc @allisonport-db

kernel/kernel-api/src/main/java/io/delta/kernel/internal/data/PaginationFilteredColumnarBatch.java
huan233usc: I was a bit worry about the `filePath`. FilteredColumnarBatch may not necessary come from file, and there were no assumption that all data from `FilteredColumnarBatch` must from the same file. (If we want to coalesce small batch to a larger one)

scottsand-db: Hey Nick -- we certainly can add these to the eixsting FilteredColumnarBatch -- that is what was implemented at first actually

https://github.com/delta-io/delta/pull/4853/files/dd6f5f034775b01f8d1f31e49f6e3005abebbd52

What's a bit odd with that approach is now FilteredColumnarBatch sometimes does and sometimes does not have certain info (filePath, preComputedNumSelectedRows

It also might tempt other creators of this API to pass in those fields -- but really this is a narrow API only required for pagination (as of now)

Food for thought. -- Given this, still want to stick with just the existing FilteredColumnarBatch? and add a few more constructor params?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
mmmyr: Hey Nick, by default, includeStats is false. 
```
/**
   * Get an iterator of data files in this version of scan that survived the predicate pruning.
   *
   * @return data in {@link ColumnarBatch} batch format. Each row correspond to one survived file.
   */
  @Override
  public CloseableIterator<FilteredColumnarBatch> getScanFiles(Engine engine) {
    return getScanFiles(engine, false);
  }
```

mmmyr: I think includeStats is only true when testing. When users from outside call current Scan::getScanFiles, includeStats is false.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/PaginationFilteredColumnarBatchSuite.scala
mmmyr: Just to confirm — I should only break the string with + if it goes over the line limit (100 chars for our repo), and otherwise keep it on one line?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/data/PaginationFilteredColumnarBatch.java
mmmyr: Hey Nick, we want to keep FilteredColumnarBatch class untouched. This class is public, and there's no requirement that all batch data MUST come from a file or MUST come from the same file, so it might be confusing to include a filePath field inside that class (number of selected rows is reasonable tho). Therefore, we want to introduce an **Internal** class PaginationFilteredColumnarBatch to carry this pagination-need information. : )

mmmyr: If we decide to stick with existing FilteredColumnarBatch, we still need a separate constructor for Pagination-use FilteredColumnarBatch. Because FilteredColumnarBatch appears multiple times, and it seems that other usages aren't necessarily have file path (and many of them don't have pre-computed number of selected rows either).

nicklan: Got it, okay. Thanks for the context, will review with that in mind.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
nicklan: If it's testing only then I'd really prefer not to have it in a public API.

Thoughts on getting rid of it, and then having the two apis I suggested above? Happy for someone to take removing the `includeStats` arg as a follow-up and you could just have the existing API plus `getPaginatedScanFiles`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/data/PaginationFilteredColumnarBatch.java
nicklan: If we keep this then I think `FilteredColumnarBatchWithMetadata` makes sense then.

I'm confused about the file thing though. It's this just one batch, that _does_ come from a single file? Do we somehow combine batches together that come from different files? I would have thought that something like and `ActiveAddFilesIterator` would have batches from different files, but that each batch would be from one file.

Thanks for the link to the old way of doing it, that seems.... simpler. :) I think it would be okay for the filePath to be `Optional.empty` if the batch doesn't come from a file for some reason.

mmmyr: Right now seems that we have never created any columnar batch with data from different files, but @huan233usc mentioned that someone might want to do it in the future? There is no enforcement that all data from a same batch will come from the same file. 

I agree with the idea if batch doesn't come from file (or multiple files), we can just let filePath to be Optional.empty.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
mmmyr: Well, the code change is a bit confusing here. We eventually need a new method `Scan::getScanFile(Engine, Optional<PaginationContext> paginationContext)`,  because getting ScanFiles in pagination Scan shares existing codepath with getScanFiles today. Pagination Context will be passed down to all iterators. Both Action Iterator & Active AddFiles Iterator needs to take in an optional Pagination Context as a new parameter for their constructors.

mmmyr: So even if we can get rid of the` includestats ` here, we still need a new method: `Scan::getScanFile(Engine, Optional paginationContext) `along with `Scan::getScanFiles` today.

mmmyr: But outside users should only call `Scan::getScanFiles(engine) ` or `PaginatedScan::getScanFiles(engine)`

scottsand-db: > If it's testing only then I'd really prefer not to have it in a public API.

Hey Nick, just to clarify, `getScanFiles(engine, false);` is a public API in this `ScanImpl` class and is "public" for internal Kernel codes and testss, but it is not a public API defined in the `Scan` interface.

mmmyr: https://github.com/delta-io/delta/pull/4860#discussion_r2180578934
Xin mentioned one potential refactoring we could do here. However, for kernel pagination support, I prefer to keep the current approach with minimal code changes. We can leave a TODO for refactoring and handle that in a separate PR. I can create an issue about this refactor.

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
```diff
@@ -37,10 +39,35 @@
...
+  private final Optional<Integer> preComputedNumSelectedRows;
 
+  // TODO: use static factory for constructors
   public FilteredColumnarBatch(ColumnarBatch data, Optional<ColumnVector> selectionVector) {
     this.data = data;
     this.selectionVector = selectionVector;
+    this.filePath = Optional.empty();
+    this.preComputedNumSelectedRows =
+        !selectionVector.isPresent() ? Optional.of(data.getSize()) : Optional.empty();
```
nicklan, line 51: Is this right? We could compute the number of rows if we wanted to right? Like it's `data.getSize() - selectionVector.numRowsDeleted()`? (I don't think `numRowsDeleted` exists, but something in that spirit)

If we don't care maybe it's not worth it, but if the selection vector knows how many rows its going to remove this is almost "for free" to compute so we might as well.

mmmyr: Yes users are able to compute the number of selected rows if they want to. But this computation is O(n), we must iterate over all rows in the selection to see if it's selected or not. It do have some cost.

mmmyr:
```
int trueRowCount = 0;
    for (int rowId = 0; rowId < data.getSize(); rowId++) {
      if (selectionVector.get().isNullAt(rowId)) {
        boolean isSelected =
            !selectionVector.get().isNullAt(rowId) && selectionVector.get().getBoolean(rowId);
        if (isSelected) trueRowCount++;
      }
    }
    return trueRowCount;
  }
```
This is how users can get number of selected rows by using selection vector

kernel/kernel-api/src/main/java/io/delta/kernel/data/FilteredColumnarBatch.java
```diff
@@ -54,6 +81,10 @@ public ColumnarBatch getData() {
     return data;
   }
 
+  public Optional<String> getFilePath() {
```
huan233usc, line 92: Can you add Java doc, especially about when this field would be empty.

huan233usc (approved):
LGTM, with one minor request of a java doc

scottsand-db (approved):

PR #4866, opened 2025-07-02
[Spark] Safely log unresolved data skipping filters from DeltaScan

chirag-s-db:
- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
When logging DeltaScans, we currently naively convert the rewritten partition-like data filters (which may be unresolved) to SQL. This is unsafe, as unresolved expressions might throw an error when converted to SQL (if the conversion to SQL accesses a child expression's data type). Wrap the SQL conversion to catch any potential errors to prevent this issue.

## How was this patch tested?
N/A.

No.


chirag-s-db:
@rahulsmahadev Can you take a look?

rahulsmahadev (approved):
Looks good to me!

PR #4873, opened 2025-07-03
[Spark] Remove parallelism in Row Tracking Backfill

tomvanbussel:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR removes the parallelism from the row tracking backfill. This parallelism is currently unused, and it makes it more difficult to add performance optimizations.

As part of this change an new backfill observer was added, as we can no longer rely on injecting a thread factory to inject transaction observers.

## How was this patch tested?

Existing tests to ensure that nothing breaks.

No


larsk-db (approved):
LGTM

PR #4878, opened 2025-07-03
[kernel] support passing baseRowId and defaultRowCommitVersion when generate IcebergCompatWriterV3AddAction 

KaiqiJinWow:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

Support passing `baseRowId` and `defaultRowCommitVersion` when generating IcebergWriterCompatV3 addFile/removeFile.

## How was this patch tested?

Unit tests

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/GenerateIcebergCompatActionUtils.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/GenerateIcebergCompatActionUtils.java
@@ -208,7 +208,9 @@ public final class GenerateIcebergCompatActionUtils {
             tableRoot,
             fileStatus,
             partitionValues,
-            dataChange);
+            dataChange,
+            Optional.empty() /* baseRowId */,
+            Optional.empty() /* defaultRowCommitVersion */);
     return SingleAction.createRemoveFileSingleAction(removeFileRow);
   }
 
@@ -264,7 +266,9 @@ public final class GenerateIcebergCompatActionUtils {
             tableRoot,
             fileStatus,
             partitionValues,
-            dataChange);
+            dataChange,
+            Optional.empty() /* baseRowId */,
+            Optional.empty() /* defaultRowCommitVersion */);
     return SingleAction.createRemoveFileSingleAction(removeFileRow);
   }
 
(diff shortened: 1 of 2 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/GenerateIcebergCompatActionUtils.java
```diff
@@ -353,7 +361,30 @@ public static Row convertRemoveDataFileStatus(
...
         dataFileStatus.getSize(),
         dataFileStatus.getStatistics(),
-        physicalSchema);
+        physicalSchema,
+        Optional.empty() /* baseRowId */,
+        Optional.empty() /* defaultRowCommitVersion */);
+  }
+
+  @VisibleForTesting
+  public static Row convertRemoveDataFileStatus(
```
huan233usc: I guess we could just update existing method convertRemoveDataFileStatus? Don't see any strong reason to have a separate method.

KaiqiJinWow: updated

kernel/kernel-api/src/main/java/io/delta/kernel/Transaction.java
```diff
@@ -267,7 +268,9 @@ static CloseableIterator<Row> generateAppendActions(
                   ((DataWriteContextImpl) dataWriteContext).getPartitionValues(),
                   true /* dataChange */,
                   // TODO: populate tags in generateAppendActions
-                  Collections.emptyMap() /* tags */);
+                  Collections.emptyMap() /* tags */,
+                  Optional.empty() /* baseRowId */,
```
raveeram-db, line 272: Nit: would be good to just have one param for all RowId related state if possible - e.g `RowIdContext` that encapsulates the `baseRowId` and `RowCommitVersion`?

KaiqiJinWow: HI @raveeram-db , since this internal [function](https://github.com/delta-io/delta/pull/4878/files#diff-df301960eb1b0caa9a1166fad87668ebe3863a29989f7978df3592ad839cb74dL87-L97) `createAddFileRow` also accepts `baseRowId` and `rowCommitVersion` separately, I think it’s reasonable to expose these two here as well — does that sound okay to you?

raveeram-db: Ah okay, and I guess we won't have other related state we need to plumb through beyond these 2. SGTM

PR #4887, opened 2025-07-06
read page one + basic test cases

mmmyr:
## 🥞 Stacked PR

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -34,21 +46,68 @@ public PaginatedScanFilesIteratorImpl(
...
+      System.out.println("numTotalAddFiles: " + batch.getData().getColumnVector(0).getSize());
+      System.out.println("numOfRows: " + rowNum);
+      System.out.println("numScanFilesReturned: " + numScanFilesReturned);
+    }
   }
 
   @Override
   public FilteredColumnarBatch next() {
-    return null;
+    if (closed) {
```
scottsand-db: I think we can just return false?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -34,21 +46,68 @@ public PaginatedScanFilesIteratorImpl(
...
+      prepareNext();
+    }
+    return nextBatch.isPresent();
   }
 
-  @Override
-  public boolean hasNext() {
-    return false;
+  private void prepareNext() {
+    if (numScanFilesReturned >= pageSize) {
```
scottsand-db: Might as well do another check here: if `nextBatch.isPresent` just `return;`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -34,21 +46,68 @@ public PaginatedScanFilesIteratorImpl(
...
+      return;
+    }
+    if (originalIterator.hasNext()) {
```
scottsand-db: tip: reduce and eliminate indentation when you can: you could do `if (!originalIterator.hasNext()) { return; }` and then all the code within the brackets can be shifted left (no identation). this makes the code more readable

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -34,21 +46,68 @@ public PaginatedScanFilesIteratorImpl(
...
+      FilteredColumnarBatch batch = originalIterator.next();
+      String fileName = batch.getFilePath().get();
```
scottsand-db: how do we _know_ it has a file name? please check if -- we certainly **expect** and **require** that it does, so check if it does not and then throw an exception (IllegalStateException)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -34,21 +46,68 @@ public PaginatedScanFilesIteratorImpl(
...
+      if (!fileName.equals(currentLogFilePath)) {
+        currentLogFilePath = fileName;
+        System.out.println("fileName " + fileName);
```
scottsand-db: `logger`; apply elsewhere too plz

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -34,21 +46,68 @@ public PaginatedScanFilesIteratorImpl(
...
+        currentRowIdxInFile = 0; // row idx starts from 1
```
scottsand-db: `row idx starts from 1` --> doesn't row index start from 0? and that is why you are assigning the variable to 0?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -34,21 +46,68 @@ public PaginatedScanFilesIteratorImpl(
...
+      }
+      long numActiveAddFiles = batch.getPreComputedNumSelectedRows().get();
```
scottsand-db: `numActiveAddFiles` in what? in what context? probably `numActiveAddFilesInBatch`

but it's not just "active" -- we have also applied data skipping here, too, right? so perhaps `numSelectedAddFilesInBatch`

scottsand-db: also check that this `.get` exists, too

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -34,21 +46,68 @@ public PaginatedScanFilesIteratorImpl(
...
-  public void close() throws IOException {}
+  public void close() throws IOException {
+    closed = true;
+    Utils.closeCloseables(originalIterator);
+  }
+
+  public Row getCurrentPageToken() {
+    // return first page token
+    return new PageToken(
+            currentLogFilePath, currentRowIdxInFile, Optional.empty(), null, null, 100, 100, 100)
```
scottsand-db: named params, `null /* asdf */, null /* qwer */`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -34,21 +46,68 @@ public PaginatedScanFilesIteratorImpl(
...
 
   @Override
-  public void close() throws IOException {}
+  public void close() throws IOException {
+    closed = true;
+    Utils.closeCloseables(originalIterator);
+  }
+
+  public Row getCurrentPageToken() {
+    // return first page token
```
scottsand-db: can you help me understand what this comment is meant to convey?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -34,21 +46,68 @@ public PaginatedScanFilesIteratorImpl(
     this.pageSize = paginationContext.getPageSize();
   }
 
+  // Bug: we can't call hasNext() twice
```
scottsand-db: is this comment still applicable?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -18,14 +18,26 @@
...
 
+  private static final Logger logger = LoggerFactory.getLogger(PaginatedScanFilesIteratorImpl.class);
   private final CloseableIterator<FilteredColumnarBatch> originalIterator;
-  private final long pageSize; // max num of files to return in this page
+  private final long pageSize;
+
+  private long numScanFilesReturned = 0;
+  private String currentLogFilePath = null;
+  private long currentRowIdxInFile = 0;
+  private Optional<FilteredColumnarBatch> nextBatch = Optional.empty();
```
scottsand-db: so we have `currentLogFilePath` and `currentRowIdxInFile` (**"current"**) but then we have `nextBatch` (**"next"**) -- would `currentBatch` be a better name?

scottsand-db (changes requested):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
mmmyr: In ActiveAddFileIterator & actions iterator, we throw exception in next() (and also in hasNext() ) when the iterator is closed. I feel like we should keep the iterator behavior consistent? 

Also, this is next() method (not hasNext() ) , I can't really return false here : ) .

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
mmmyr: Ic. If row index starts from 0, when creating page token, the returned row index will be "number of row scanned -1". because right now I calculate row index in this way: rowIdxInLastFile += rowNum;

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
scottsand-db: Sorry -- i meant to comment this for `hasNext`, not next.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
scottsand-db: Sorry, I'm confused. Please setup a sync with me to discuss?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
mmmyr: In hasNext(), both actions iterator & active add files throws exception too when the iterator is detected closed. Should I make paginated iterator do the same thing or just return false?

scottsand-db: SGTM -- we can throw, to stick with the existing behavior. thanks for explaining/highlighting that for me!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -51,16 +58,56 @@ public Row getCurrentPageToken() {
...
+
+    FilteredColumnarBatch batch = filteredScanFilesIter.next();
+    String fileName = batch.getFilePath().get();
+    if (!fileName.equals(lastLogFileName)) {
+      lastLogFileName = fileName;
+      System.out.println("filePath " + fileName);
+      rowIdxInLastFile = 0; // row idx starts from 1
+    }
+    long numActiveAddFiles = batch.getPreComputedNumSelectedRows().get();
+    long rowNum = batch.getData().getSize();
```
scottsand-db: would `numRows` be a more accurate variable name?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
mmmyr: I'll set up some time to sync in the later afternoon.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -46,21 +55,66 @@ public PaginatedScanFilesIteratorImpl(
...
+
+    if (!filteredScanFilesIter.hasNext()) return; // base iterator is empty
+
+    FilteredColumnarBatch batch = filteredScanFilesIter.next();
+    checkArgument(batch.getFilePath().isPresent(), "file path doesn't exist!");
+    String filePath = batch.getFilePath().get();
+    if (!filePath.equals(currentLogFileName)) {
+      currentLogFileName = filePath;
+      System.out.println("filePath " + filePath);
+      currentRowIdxInFile = 0; // row idx starts from 1
```
mmmyr: numReturnedRows = last returned row index +1 
Set the initial value to -1 here. worth give an example in code here, when I declared last returned row index. 
read 3 batches , each has 10 rows -> this will be 29 (which is index of last returned one).

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -15,52 +15,142 @@
...
 
 /** Implementation of {@link PaginatedScanFilesIterator} */
 public class PaginatedScanFilesIteratorImpl implements PaginatedScanFilesIterator {
 
-  private final CloseableIterator<FilteredColumnarBatch> filteredScanFilesIter;
+  private static final Logger logger =
+      LoggerFactory.getLogger(PaginatedScanFilesIteratorImpl.class);
+
+  /**
+   * Filtered ScanFiles iterator retrieved from baseScan but without batches from skipped log files
```
scottsand-db: `but without batches from skipped log files` --> worth adding a simple example to really drive this point home! 

`For example, if the previous page fully comsumed sidecar files A and B, and half consumed sidecar C, we expect this iterator to exclude sidecar files A and B`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -15,52 +15,142 @@
...
+   */
+  private String lastReadLogFileName = null;
+
+  /**
+   * The index of the last row returned from the last log file that was read.
+   *
+   * <p>For example, if the last page contains 3 batches from the same log file, and each batch has
+   * 10 rows, this value will be 29 (since row indices start at 0).
+   *
+   * <p>This value corresponds to the one saved in the page token if present;
```
scottsand-db: nit, ending your sentence with `;`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -15,52 +15,142 @@
...
   private final long pageSize;
 
+  /** Total number of ScanFiles returned in current page */
+  private long numScanFilesReturned;
+
+  /**
+   * The name of the last log file that was read. This value corresponds to the one saved in the
+   * page token if present.
+   */
+  private String lastReadLogFileName = null;
```
scottsand-db: btw `name`? not `path`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -15,52 +15,142 @@
...
+    return currentBatch.isPresent();
+  }
+
+  private void prepareNext() {
+    if (currentBatch.isPresent()) return;
+    if (numScanFilesReturned >= pageSize) return;
+    if (!baseFilteredScanFilesIter.hasNext()) return;
+
+    FilteredColumnarBatch batch = baseFilteredScanFilesIter.next();
+    checkArgument(batch.getFilePath().isPresent(), "file path doesn't exist!");
```
scottsand-db, line 132: you are comparing `filePath` against `fileName` -- something seems off/wrong here. `/a/b/c/d/e/foo.json` will never be equal to `foo.json`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -15,52 +15,142 @@
...
+    if (!batch.getFilePath().get().equals(lastReadLogFileName)) {
+      lastReadLogFileName = batch.getFilePath().get();
+      lastReturnedRowIndex = -1;
+      logger.info("filePath {}", lastReadLogFileName);
+    }
+    checkArgument(
```
scottsand-db, line 133: we should do this above, right? why do the work in the `if` statement if this check will fail right after?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -15,52 +15,142 @@
...
+    }
+    return currentBatch.isPresent();
+  }
+
+  private void prepareNext() {
+    if (currentBatch.isPresent()) return;
+    if (numScanFilesReturned >= pageSize) return;
+    if (!baseFilteredScanFilesIter.hasNext()) return;
+
+    FilteredColumnarBatch batch = baseFilteredScanFilesIter.next();
```
scottsand-db: nit (my personal preference): `final`, for all of your `final` variables that should never be re-assigned. e.g. for `numRowsInBatch`, etc

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,268 @@
...
+    with ExpressionTestUtils with SQLHelper with DeltaTableWriteSuiteBase {
+
+  private val logger = LoggerFactory.getLogger(classOf[PaginatedScanSuite])
+  // TODO: test call hasNext() twice
+
+  test("Read First Page: basic pagination with single JSON file") {
+    withTempDir { tempDir =>
+      /**
+       *   Creates a Delta table with 10 Parquet data files; each file contains 10 rows
+       * * 00000000000000000000.json contains 10 AddFile actions — one per file
```
scottsand-db: nit: extra `*` -- `* * 000`

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,268 @@
...
+       */
+      spark.range(0, 100, 1, 10).write.format("delta").save(tempDir.getCanonicalPath)
+
+      val snapshot = latestSnapshot(tempDir.getCanonicalPath)
```
scottsand-db: Can you please use https://github.com/delta-io/delta/blob/e02462297a57996ebf28cecc577da7c21e88e992/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableReadsSuite.scala#L69

e.g. `tableManager.getResolvedTableAdapterAtLatest` -- this is an adapter that lets us test both the ResolvedTable API and the SnapshotAPI -- they can both give us the `scanBuilder` which is all we really need

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,268 @@
...
+class PaginatedScanSuite extends AnyFunSuite with TestUtils
+    with ExpressionTestUtils with SQLHelper with DeltaTableWriteSuiteBase {
+
+  private val logger = LoggerFactory.getLogger(classOf[PaginatedScanSuite])
+  // TODO: test call hasNext() twice
+
+  test("Read First Page: basic pagination with single JSON file") {
+    withTempDir { tempDir =>
+      /**
+       *   Creates a Delta table with 10 Parquet data files; each file contains 10 rows
```
scottsand-db: do we need to care / know that each of the parquet files contains 10 rows?
all we care is that we return the right scan files, right? I say this because seeing `10` prtaquet files and then `10 rows` (both the number 10, but we only care about the first number) is a bit confusing?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,268 @@
...
+      val snapshot = latestSnapshot(tempDir.getCanonicalPath)
+      val pageSize = 8 // TODO: change different page sizes
+      val paginatedScan = snapshot.getScanBuilder
+        .buildPaginated(pageSize, Optional.empty() /* page token opt */ )
+        .asInstanceOf[PaginatedScanImpl]
+
+      /**
+       * Creates a custom engine with batch size 5;
+       * 00000000000000000000.json contains 3 batches, 10 active AddFiles in total
+       * Batch 1: 5 rows, 2 selected AddFiles
```
scottsand-db: You should explain why this is only 2 addfiles -- it is because of the CommitInfo, Protocol, and Metadata actions. I had to think for a seccond to realize this :) 

Could you even make this clearer in the code? e.g. what happens if we change how many non-addfile rows are written in the first commit? we want to make our test robust, not brittle, to change

```scala
val numNonAddFileRows = 3 // CommitInfo, Protocol, and Metadata
...
```

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,268 @@
...
+        s"New Page Token:lastReadRowIndex should be 5, got $lastReturnedRowIndex")
+
+      logger.info(s"New PageToken: lastReadLogFileName  = $lastReadLogFileName")
+      logger.info(s"New PageToken: lastReadRowIndex = $lastReturnedRowIndex")
+
+      paginatedIter.close()
+    }
+  }
+
+  test("Read First Page: basic pagination with one checkpoint file " +
```
scottsand-db: I see lots of similar test setup and set validation between these three tests -- can we considate these into helper methods?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,268 @@
...
+        s"First page should contain 10 files, got $totalFileCountsReturned")
+
+      assert(
+        fileCounts.length == 3,
+        s"First page should contain 3 batches, got ${fileCounts.length}")
+
+      logger.info(s"Total num batches returned in page one = ${fileCounts.length}")
+      logger.info(s"Total num Parquet Files fetched in page one = $totalFileCountsReturned")
+
+      /** Verify we get correct page token for next page */
```
scottsand-db: nit: for comments within code like this, use `//` not `/**`

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,268 @@
...
+      val lastReturnedRowIndex = PageToken.fromRow(paginatedIter.getCurrentPageToken)
+        .getLastReturnedRowIndex
+
+      assert(
+        lastReadLogFileName.endsWith("00000000000000000000.json"),
+        s"New Page Token: LastReadLogFilename should be 00000000000000000000.json, " +
+          s"got $lastReadLogFileName")
+
+      assert(
+        lastReturnedRowIndex == 12,
```
scottsand-db: to me, `12` is a magic number. where did this come from? what does it mean? if we change other parts of our code, and this assert starts to fail, how do we know what this value should be updated to?

https://en.wikipedia.org/wiki/Magic_number_(programming)

scottsand-db (changes requested):
Looks great! Requested some changes

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,268 @@
...
+import org.scalatest.funsuite.AnyFunSuite
+import org.slf4j.{Logger, LoggerFactory};
+
+class PaginatedScanSuite extends AnyFunSuite with TestUtils
+    with ExpressionTestUtils with SQLHelper with DeltaTableWriteSuiteBase {
+
+  private val logger = LoggerFactory.getLogger(classOf[PaginatedScanSuite])
+  // TODO: test call hasNext() twice
+
+  test("Read First Page: basic pagination with single JSON file") {
```
scottsand-db: Could we also add a test that we read full batches, even if that slightly goes over the page size? You can add this in a followup PR

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,268 @@
...
+      logger.info(s"Total num Parquet Files fetched in page one = $totalFileCountsReturned")
+
+      /** Verify we get correct page token for next page */
+      val lastReadLogFileName = PageToken.fromRow(paginatedIter.getCurrentPageToken)
+        .getLastReadLogFileName
+      val lastReturnedRowIndex = PageToken.fromRow(paginatedIter.getCurrentPageToken)
+        .getLastReturnedRowIndex
+
+      assert(
+        lastReadLogFileName.endsWith("00000000000000000000.json"),
```
scottsand-db: nit: just the assert with no helper message is fine -- it'll make your tests a bit more succinct

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
mmmyr: In the page token, we currently use the name lastReadLogFileName. I think it makes sense to stay consistent and use the same name here as well.

Also, the full directory info in the file path isn’t particularly useful here—aside from identifying whether it's from a sidecar directory. To save memory in the page token, I was thinking of extracting just the file name from the path. 

Thoughts?

mmmyr: You are right, now it's indeed file path (because FilteredColumnarBatch saves full file path). Thanks for reminding! : )

mmmyr: Actually we will eventually save the table path in page token (for validation purpose), I feel like we don't need to save full file path in page token?

scottsand-db: I think using the `path` is simplest. It's not going to be 100x longer than the `name` -- and we are only saving one field.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
mmmyr: Yep it won't change. Thanks for reminding!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
mmmyr: Then we will need to update the name in page token. Should I do it in this PR or a separate one?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
mmmyr: Sorry for the confusion caused by naming, lastReadLogFileName is lastReadLogFilePath (since filtered columnar batch saves full path), and we don't do any string extraction here.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
mmmyr: Yep we don't care about that. We only care about how many rows and how many active AddFiles in log files. Will update the comment.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
mmmyr: Ic. Comments for variables are /** xxx */, and  comments within code are //

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
scottsand-db: This PR is fine

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
mmmyr: Yep simply changing page size can test this case. We need tests for same table but different page sizes , I'll think about how to do this cleanly.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -15,52 +15,148 @@
...
+    logger.info("total numScanFilesReturned: {}", numScanFilesReturned);
+    logger.info("numSelectedAddFilesInBatch: {}", numSelectedAddFilesInBatch);
+    logger.info("numRowsInBatch: {}", numRowsInBatch);
   }
 
   @Override
   public FilteredColumnarBatch next() {
-    throw new UnsupportedOperationException("Not implemented");
+    if (closed) {
+      throw new IllegalStateException("Can't call `next` on a closed iterator.");
```
huan233usc: https://github.com/delta-io/delta/blob/2095e31a7ba44600a5d3318d790274d36ff3380b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/Preconditions.java#L73 use this utils?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -15,52 +15,148 @@
...
+    return currentBatch.isPresent();
+  }
+
+  private void prepareNext() {
+    if (currentBatch.isPresent()) return;
+    if (numScanFilesReturned >= pageSize) return;
+    if (!baseFilteredScanFilesIter.hasNext()) return;
+
+    final FilteredColumnarBatch batch = baseFilteredScanFilesIter.next();
+    checkArgument(batch.getFilePath().isPresent(), "file path doesn't exist!");
```
scottsand-db, line 132: Can you please add a comment here for the reader: we expect these two values to be set/present, they are set in X location.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -15,52 +15,148 @@
...
+    }
+    final FilteredColumnarBatch ret = currentBatch.get();
+    currentBatch = Optional.empty();
+    return ret;
   }
 
   @Override
   public void close() throws IOException {
-    throw new UnsupportedOperationException("Not implemented");
+    closed = true;
```
scottsand-db: what if they call close twice? is this safe/does this work as expected?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,343 @@
...
+class PaginatedScanSuite extends AnyFunSuite with TestUtils
+    with ExpressionTestUtils with SQLHelper with DeltaTableWriteSuiteBase {
+
+  private val logger = LoggerFactory.getLogger(classOf[PaginatedScanSuite])
+
+  private object TestConstants {
+    val DEFAULT_BATCH_SIZE = 5
+  }
+
+  val tableManager: AbstractTableManagerAdapter = getTableManagerAdapter
```
scottsand-db, line 38: `TestUtilsWithTableManagerAPIs` -- could you please use this trait instead of `TestUtils`

https://github.com/delta-io/delta/blob/a7fc9bda3c36430572d7efb22e062c15b26a01ae/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableReadsSuite.scala#L67C10-L67C39

scottsand-db: `TestUtils` is really the legacy API, so right now this is really just using Snapshot, not ResolvedTable

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,343 @@
...
+import org.scalatest.funsuite.AnyFunSuite
+import org.slf4j.{Logger, LoggerFactory};
+
+class PaginatedScanSuite extends AnyFunSuite with TestUtils
+    with ExpressionTestUtils with SQLHelper with DeltaTableWriteSuiteBase {
+
+  private val logger = LoggerFactory.getLogger(classOf[PaginatedScanSuite])
+
+  private object TestConstants {
+    val DEFAULT_BATCH_SIZE = 5
```
scottsand-db: nit: maybe overkill? you can just do the private `val` at the top of the class?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,343 @@
...
+      "delta.kernel.default.parquet.reader.batch-size",
+      TestConstants.DEFAULT_BATCH_SIZE.toString)
+    DefaultEngine.create(hadoopConf)
+  }
+
+  private def createPaginatedScan(
+      tablePath: String,
+      tableVersion: Optional[Long],
+      pageSize: Long,
+      pageToken: Optional[io.delta.kernel.data.Row] = Optional.empty()): PaginatedScanImpl = {
```
scottsand-db: any reason we cannot just import `Row`? and just say `[Row]` here?

nit: `tableVersionOpt`, `pageTokenOpt`

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,343 @@
...
+        tableManager.getResolvedTableAdapterAtLatest(customEngine, tablePath)
+      }
+    }
+
+    resolvedTableAdapter.getScanBuilder()
+      .buildPaginated(pageSize, pageToken)
+      .asInstanceOf[PaginatedScanImpl]
+  }
+
+  // Helper function to collect all batches from a paginated iterator for current page
```
scottsand-db: nit: comments for methods (outside/above them like this) use `/** */`

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,343 @@
...
+  private def collectPaginatedBatches(
+      paginatedIter: io.delta.kernel.PaginatedScanFilesIterator)
```
scottsand-db: any reason we cannot just import `PaginatedScanFilesIterator ` directly?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,343 @@
...
+      : collection.mutable.Buffer[FilteredColumnarBatch] = {
```
scottsand-db: do you expect consumers of this method to further mutate and append to this buffer? or is it more realistic/helpful/clear to return an immutable `Seq`?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,343 @@
...
+    validatePageResults(returnedBatchesInPage, expectedFileCount, expectedBatchCount, testName)
+    validatePageToken(paginatedIter, expectedLogFileName, expectedRowIndex, testName)
+
+    paginatedIter.close()
+  }
+
+  private def runPaginationTestCasesForFirstPage(
+      tablePath: String,
+      tableVersion: Optional[Long],
+      testCases: Seq[(String, Int, Int, Int, String, Int)]): Unit = {
```
scottsand-db: you can turn this into a scala case class, instead of a long tuple -- that should make things much cleaner and more succinct

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,343 @@
...
+  private def runSinglePaginationTestCase(
+      testName: String,
+      tablePath: String,
+      tableVersion: Optional[Long],
+      pageSize: Long,
+      expectedFileCount: Int,
+      expectedBatchCount: Int,
+      expectedLogFileName: String,
+      expectedRowIndex: Int): Unit = {
+
```
scottsand-db: nit: we can avoid this empty line at the start of methods

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,343 @@
...
+      .getLastReturnedRowIndex
+
+    assert(lastReadLogFilePath.endsWith(expectedLogFilePath))
+    assert(lastReturnedRowIndex == expectedRowIndex)
+
+    logger.info(s"$testName: New PageToken: lastReadLogFileName = $lastReadLogFilePath")
+    logger.info(s"$testName: New PageToken: lastReadRowIndex = $lastReturnedRowIndex")
+  }
+
+  private def runSinglePaginationTestCase(
```
scottsand-db, line 120: is this only applicable to "first page" tests? would a better name be `runPaginationTestCaseForFirstPage`?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,343 @@
...
+      val JSONFile0 = "00000000000000000000.json"
+      val testCases = Seq(
+        (
+          "Page size 1",
+          1 /* page size */,
+          2 /* expectedFileCount */,
+          1 /* expectedBatchCount */,
+          JSONFile0,
+          4),
+        ("Page size 2", 2, 2, 1, JSONFile0, 4),
```
scottsand-db: a case class here with named params would probably be clearer

connectors/golden-tables/src/main/resources/golden/kernel-pagination-allJSONs/.part-00000-520b63c8-7004-4fb3-9a7d-6b1ee913d1ac-c000.snappy.parquet.crc
scottsand-db, line 1: (1) Could we name these sorts of folders `kernel-pagination-all-jsons`
(2) Could we reduce the number of parquet files written? And remove the `.parquet.crc` files, too?
(3) Why do we need `-all-jsons` and also `one-json` --- couldn't we just do time travel to the `0.json` in the `-all-jsons` table?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,343 @@
...
+      if (tableVersion.isPresent) {
+        tableManager.getResolvedTableAdapterAtVersion(customEngine, tablePath, tableVersion.get())
+      } else {
+        tableManager.getResolvedTableAdapterAtLatest(customEngine, tablePath)
+      }
+    }
+
+    resolvedTableAdapter.getScanBuilder()
+      .buildPaginated(pageSize, pageToken)
+      .asInstanceOf[PaginatedScanImpl]
```
scottsand-db: Can you help me to understand: Why do we need to return the implementation? and not just the public API? Is there any internal API we need to have access to in these tests?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
mmmyr: Ic. Thanks for clarifying!

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
mmmyr: Ic. I'll ask Cursor / GPT how to do it. Thanks : )

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
mmmyr: Yes. Make sense. Thanks!

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
mmmyr: Nope, simply use PaginatedScan is enough. thanks for pointing out.

connectors/golden-tables/src/main/resources/golden/kernel-pagination-allJSONs/.part-00000-520b63c8-7004-4fb3-9a7d-6b1ee913d1ac-c000.snappy.parquet.crc
mmmyr: Got it! But I don't want to write too few Parquet files, since the page size ultimately limits how many data files are returned — and we want to test a range of page sizes effectively :)

Other suggestions were great! Thank you so much !

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
mmmyr: Prob caused by intelliJ auto fix, I should double check, sorry.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
mmmyr: Actually this helper function can also be used to test second page result, but I'll use the name runPaginationTestCaseForFirstPage now (and might change it in following PRs).

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -15,52 +15,153 @@
...
+    return new PageToken(
+            lastReadLogFilePath,
+            lastReturnedRowIndex,
+            Optional.empty() /* sidecar file index */,
+            Meta.KERNEL_VERSION,
+            "fake/table/path" /* table path */,
+            -1 /* table version */,
+            -1 /* predicate hash */,
+            -1 /* log segment hash */)
+        .toRow();
```
mmmyr: TODO: we don't need to use magical number "-1" and "fake/table/path" if we get #4905 merged first.

connectors/golden-tables/src/main/resources/golden/kernel-pagination-all-jsons/_delta_log/.00000000000000000000.crc.crc
scottsand-db, line 1: can you move these to the kernel resources folder instead btw?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,353 @@
...
+  test("Pagination with single JSON file") {
+    withGoldenTable("kernel-pagination-all-jsons") { tablePath =>
+      /**
+       *  Log Segment List:
+       *  00000000000000000000.json contains 2 batches, 5 active AddFiles in total
+       *  Batch 1: 5 rows, 2 selected AddFiles
+       *  Batch 2: 3 rows, 3 selected AddFiles
+       */
+
+      val JSONFile0 = "00000000000000000000.json"
```
scottsand-db: nit: `JSON` to start a variable seems a little weird, since it is not `JSON_FILE_0` (which is used for static final values) -- I would suggest `jsonFile0`

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,353 @@
...
+      // for example, batch size goes over page size limit
+      val testCases = Seq(
+        SinglePageRequestTestCase(
+          "Page size 1",
+          pageSize = 1,
+          expectedFileCount = 2,
+          expectedBatchCount = 1,
+          expectedLogFileName = JSONFile0,
+          expectedRowIndex = 4),
+        SinglePageRequestTestCase("Page size 2", 2, 2, 1, JSONFile0, 4),
```
scottsand-db: do we need to pass in the testCaseName string? since it seems we can just create it internally based off of the page size?

Also -- it seems like your one scala test `test(...)` is actually doing many test cases inside of it.

We should invert this.

That is, there should be a Seq[SinglePageRequestTestCase], and for each one, we invoke `test(...)` and _there_ you can give it a proper test case name

And then if there are failures, it is really clear via the scala test framework which test failured.

Basically, many small tests is better than one big test

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,353 @@
...
+        SinglePageRequestTestCase("Page size 4", 4, 5, 2, JSONFile0, 7),
+        SinglePageRequestTestCase("Page size 7", 7, 5, 2, JSONFile0, 7),
+        SinglePageRequestTestCase("Page size 20", 20, 5, 2, JSONFile0, 7))
```
scottsand-db: so, I think `20, 5, 2, JSONFile0, 7` is just too hard to read/reason/understand as a reviewer.

Perhaps you can rename the case class member variables to be shorter?

```scala
SinglePageRequestTestCase(
    pageSize = 1, expfileCount = 2, expBatchCount = 1, expLastLogFile = jsonFile0, expRowIdx = 4)
```

mmmyr: I renamed class, but seems it's hard to fit this into one line : ( . You can take a look.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
mmmyr: Yep changed. There are 23 tests in total right now, combination of different tables & different page size.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,437 @@
...
+          expectedFileCount = testCase.expFileCnt,
+          expectedBatchCount = testCase.expBatchCnt,
+          expectedLogFileName = testCase.expLogFile,
+          expectedRowIndex = testCase.expRowIdx)
+      }
+    }
+  }
+
+  // Multiple JSON files test cases
+  val multipleJsonTestCases = Seq(
```
scottsand-db: does the scalafmt enforce this formatting? on multi-line? can we not do:

```scala
  val multipleJsonTestCases = Seq(
    SinglePageRequestTestCase(
      pageSize = 1, expFileCnt = 4, expBatchCnt = 1, expLogFile = JSON_FILE_2, expRowIdx = 4),
```

If not ... we can turn off scalafmt just to make these test case definitions use fewer lines

`// scalastyle:off line.size.limit .... // scalastyle:on line.size.limit`

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/utils/TestUtils.scala
```diff
@@ -183,6 +184,16 @@ trait AbstractTestUtils extends Assertions with SQLHelper {
     testFunc(tablePath)
   }
 
+  def withKernelStaticTable(tableName: String)(testFunc: String => Unit): Unit = {
```
scottsand-db: Any reason we cannot use 

https://github.com/delta-io/delta/blob/f9afb73cf68bf631b03749928a88e2fd8c39246d/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/utils/TestUtils.scala#L800

```scala
  /**
   * Returns a URI encoded path of the resource.
   */
  def getTestResourceFilePath(resourcePath: String): String = {
    val resource = ResourceLoader.classLoader.getResource(resourcePath)
    if (resource == null) {
      throw new FileNotFoundException("resource not found")
    }
    resource.getFile
  }
```

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,437 @@
...
+
+  // File name constants for better readability
+  private val JSON_FILE_0 = "00000000000000000000.json"
+  private val JSON_FILE_1 = "00000000000000000001.json"
+  private val JSON_FILE_2 = "00000000000000000002.json"
+  private val JSON_FILE_11 = "00000000000000000011.json"
+  private val JSON_FILE_12 = "00000000000000000012.json"
+  private val CHECKPOINT_FILE_10 = "00000000000000000010.checkpoint.parquet"
+
+  // Single JSON file test cases
```
scottsand-db, line 153: nit: superfluous comment

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/utils/TestUtils.scala
mmmyr: Oops I didn't find it just now. Thanks!

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,437 @@
...
+    val paginatedIter = paginatedScan.getScanFiles(customEngine)
+    val returnedBatchesInPage = collectPaginatedBatches(paginatedIter)
+
+    validatePageResults(returnedBatchesInPage, expectedFileCount, expectedBatchCount, testName)
+    validatePageToken(paginatedIter, expectedLogFileName, expectedRowIndex, testName)
+
+    paginatedIter.close()
+  }
+
+  // TODO: test call hasNext() twice
```
scottsand-db, line 140: btw, please make sure you track this TODO. have it in some project plan or some github issue so we do not forget it 👍

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
mmmyr: Sorry, forget to remove comments after using Cursor lol.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,437 @@
...
+         */
+        runSinglePaginationTestCase(
+          testName = s"Single JSON file - page size ${testCase.pageSize}",
+          tablePath = tablePath,
+          tableVersionOpt = Optional.of(0L),
+          pageSize = testCase.pageSize,
+          expectedFileCount = testCase.expFileCnt,
+          expectedBatchCount = testCase.expBatchCnt,
+          expectedLogFileName = testCase.expLogFile,
+          expectedRowIndex = testCase.expRowIdx)
```
scottsand-db: why not just pass in the test case? wouldn't that be **way** more succinct?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,437 @@
...
+      expFileCnt: Int,
+      expBatchCnt: Int,
+      expLogFile: String,
+      expRowIdx: Int)
+
+  /** Custom engine with customized batch size */
+  private val customEngine: DefaultEngine = {
+    val BATCH_SIZE = 5
+    val hadoopConf = new org.apache.hadoop.conf.Configuration()
+    hadoopConf.set("delta.kernel.default.json.reader.batch-size", BATCH_SIZE.toString)
```
scottsand-db: i don't think this `BATCH_SIZE` variable is that helpful btw

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,437 @@
...
+      expBatchCnt = 2,
+      expLogFile = JSON_FILE_0,
+      expRowIdx = 7))
+
+  singleJsonTestCases.foreach { testCase =>
+    test(s"Single JSON file - page size ${testCase.pageSize}") {
+      withKernelStaticTable("kernel-pagination-all-jsons") { tablePath =>
+        /**
+         *  Log Segment List:
+         *  00000000000000000000.json contains 2 batches, 5 active AddFiles in total
```
scottsand-db: probably worth mentioning that this assumes, and requires, that the batch size is 5

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
scottsand-db: worth mentioning in a comment that we must NOT change this batch size -- it is hardcoded and impacts every single test!

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
mmmyr: Hmmm I can try. But if turning off scalafmt, can we pass checks before merge?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,437 @@
...
+  // File name constants for better readability
+  private val JSON_FILE_0 = "00000000000000000000.json"
+  private val JSON_FILE_1 = "00000000000000000001.json"
+  private val JSON_FILE_2 = "00000000000000000002.json"
+  private val JSON_FILE_11 = "00000000000000000011.json"
+  private val JSON_FILE_12 = "00000000000000000012.json"
+  private val CHECKPOINT_FILE_10 = "00000000000000000010.checkpoint.parquet"
+
+  // Single JSON file test cases
+  val singleJsonTestCases = Seq(
```
scottsand-db: oh, you can actually just delete this `val singleJsonTestCases` and just declare the `Seq` list and then do the `.foreach { testCase => ...`

and then yeah, keep the comment for what this is for. `// ===== Single JSON file test cases =====`

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/utils/TestUtils.scala
mmmyr: To be honest, TestUtils is getting pretty long and it's a bit hard to find the useful helpers in there. I'm happy to take on the refactor, but not sure if I'll have time to get to it before the internship ends hmmm.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
mmmyr: yep but in case we might want to change it in the future ? and also avoid magical number when  creating engine?

mmmyr: Yep1

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,437 @@
...
+    SinglePageRequestTestCase(
+      pageSize = 1,
```
scottsand-db, line 164: Great test cases! IMO each one needs a 1 sentenece explanation to make it super clear what/why/how is happening.

For example:

```scala
// Kernel is asked to read the 1st page of size 1. Kernel reads the first full batch, so returns 2
// AddFiles and ends at the 5th row (index 4)
```

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
scottsand-db: Can you also move this to the right under the `// ===== Single JSON file test cases =====` declaration above?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/utils/TestUtils.scala
scottsand-db: No need to do that refactor during your internship :) Don't worry about it -- we have other and more important things to focus on

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,455 @@
...
+   *
+   *  Note: batch size is set to 5
+   *  Batch 1: 5 rows, 2 selected AddFiles
+   *  Batch 2: 3 rows, 3 selected AddFiles
+   */
+  Seq(
+    // Kernel is asked to read the 1st page of size 1. Kernel reads the 1st
+    // full batch, so returns 2 AddFiles and ends at the 5th row (index 4).
+    // Note: Kernel should always return full batches, so return full batch one.
+    SinglePageRequestTestCase(
```
scottsand-db, line 163: nit: again, are we able to format this like so?

```scala
    SinglePageRequestTestCase(
      pageSize = 2, expFileCnt = 2, expBatchCnt = 1, expLogFile = JSON_FILE_0, expRowIdx = 4),
```

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -0,0 +1,455 @@
...
+      val tablePath = getTestResourceFilePath("kernel-pagination-single-checkpoint")
+      runSinglePaginationTestCase(
+        testName = s"Single checkpoint and JSON files - page size ${testCase.pageSize}",
+        tablePath = tablePath,
+        tableVersionOpt = Optional.empty(),
+        testCase)
+    }
+  }
+
+  test("Pagination with multi-checkpoint files") {
```
scottsand-db: can you remove these empty tests? and we can add them alter in future PRs?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
mmmyr: I tried but get corrected by "build/sbt kernelGroup:scalafmtAll". Let's just keep it like this now, I might want to add more parameters to single test case here for testing more pages, and I will see if there's any way to look the final test suite look better : ).

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -15,53 +15,156 @@
...
+    Row pageTokenRow =
+        new PageToken(
+                lastReadLogFilePath,
+                lastReturnedRowIndex,
+                Optional.empty() /* sidecar file index */,
+                Meta.KERNEL_VERSION,
+                paginationContext.getTablePath() /* table path */,
+                paginationContext.getTableVersion() /* table version */,
+                -1 /* predicate hash */,
+                -1 /* log segment hash */)
```
huan233usc, line 109: Add todo?

huan233usc (approved):

kernel/kernel-defaults/src/test/resources/kernel-pagination-all-jsons/_delta_log/.00000000000000000000.crc.crc
scottsand-db, line 1: can we delete the .json.crc files? fewer files in the repo, eh?

mmmyr: Yep, sure.

scottsand-db (approved):

PR #4890, opened 2025-07-07
[Spark] Allow overriding StructField metadata in mergeDataTypes

aldenlau-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Currently, `SchemaMergingUtils.mergeDataTypes` will always use the existing metadata when merging a new `StructField` with the same name as an existing field. This change provides a new flag `overrideMetadata` that allows the update schema to override the metadata in the current schema.
## How was this patch tested?

New unit test in `SchemaUtilsSuite`.scala

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/schema/SchemaMergingUtils.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/schema/SchemaMergingUtils.scala
@@ -184,6 +184,8 @@ object SchemaMergingUtils {
    * @param caseSensitive Whether we should keep field mapping case-sensitively.
    *                      This should default to false for Delta, which is case insensitive.
    * @param allowOverride Whether to let incoming type override the existing type if unmatched.
+   * @param overrideMetadata Whether to let metadata of new fields override the existing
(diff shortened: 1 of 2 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/schema/SchemaMergingUtils.scala
```diff
@@ -192,7 +194,8 @@ object SchemaMergingUtils {
       keepExistingType: Boolean,
       typeWideningMode: TypeWideningMode,
       caseSensitive: Boolean,
-      allowOverride: Boolean): DataType = {
+      allowOverride: Boolean,
+      overrideMetadata: Boolean = false): DataType = {
```
johanl-db: nit: for consistency, don't provide a default value here, but pass it when calling `mergeDataTypes` in `mergeSchemas`
```suggestion
      overrideMetadata: Boolean): DataType = {
```

johanl-db (approved):

PR #4897, opened 2025-07-08
[Kernel] Enable row tracking in replace table transactions

lbhm:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This PR enables `REPLACE TABLE` on tables with row tracking and adds test for this scenario as discussed with @johanl-db.

## How was this patch tested?

I added unit tests that check whether replace table successfully runs in various scenarios with row tracking involved. I exhaustively enumerate the following scenarios:
- Row tracking enabled before table is replaced: yes/no
- Row tracking enabled as part of the replace transaction: yes/no
- Old table is empty: yes/no
- New table is empty: yes/no

This PR will allow users to run `REPLACE TABLE` commands on tables with row tracking.

### DIFF
--- a/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/RowTrackingSuite.scala
+++ b/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/RowTrackingSuite.scala
@@ -21,6 +21,7 @@ import java.util.Optional
 import scala.collection.JavaConverters._
 import scala.collection.immutable.Seq
 
+import io.delta.kernel.Table
 import io.delta.kernel.data.{FilteredColumnarBatch, Row}
 import io.delta.kernel.defaults.internal.parquet.ParquetSuiteBase
 import io.delta.kernel.engine.Engine
@@ -53,6 +54,54 @@ class RowTrackingSuite extends DeltaTableWriteSuiteBase with ParquetSuiteBase {
       extraProps: Map[String, String] = Map.empty): Unit = {
     val tableProps = Map(TableConfig.ROW_TRACKING_ENABLED.getKey -> "true") ++ extraProps
     createEmptyTable(engine, tablePath, schema = schema, tableProperties = tableProps)
+  }
+
+  /**
+   * Creates a table at the given path, optionally with initial data and row tracking enabled or
+   * disabled. Then replaces the table with new (or no) data and optionally enables/disables row
+   * tracking after the replace.
+   *
+   * @param engine The Delta Kernel engine to use.
+   * @param tablePath The path to the Delta table.
+   * @param schema The schema to use for the table.
+   * @param extraProps Additional table properties.
+   * @param initialData Data to append after initial table creation (before replace).
+   * @param replaceData Data to append as part of the replace transaction.
+   * @param enableRowTrackingBeforeReplace Whether to enable row tracking before the replace.
+   * @param enableRowTrackingAfterReplace Whether to enable row tracking after the replace.
+   */
+  private def createAndReplaceTable(
+      engine: Engine,
+      tablePath: String,
+      schema: StructType = testSchema,
+      extraProps: Map[String, String] = Map.empty,
+      initialData: Seq[(Map[String, Literal], Seq[FilteredColumnarBatch])] = Seq.empty,
+      replaceData: Seq[(Map[String, Literal], Seq[FilteredColumnarBatch])] = Seq.empty,
+      enableRowTrackingBeforeReplace: Boolean = false,
+      enableRowTrackingAfterReplace: Boolean = true): Unit = {
+    val initialTableProps = Map(
+        TableConfig.ROW_TRACKING_ENABLED.getKey -> enableRowTrackingBeforeReplace.toString
+    ) ++ extraProps
+
+    // Create an empty table with row tracking enabled or disabled
+    createEmptyTable(engine, tablePath, schema = schema, tableProperties = initialTableProps)
+
+    // Optionally fill the table with initial data if provided
+    if (initialData.nonEmpty) {
+      appendData(engine, tablePath, data = initialData)
+    }
+
+    // Create a REPLACE transaction
+    val replaceTableProps = Map(
+      TableConfig.ROW_TRACKING_ENABLED.getKey -> enableRowTrackingAfterReplace.toString
+    ) ++ extraProps
+    val txnBuilder = Table.forPath(engine, tablePath).asInstanceOf[TableImpl]
+      .createReplaceTableTransactionBuilder(engine, testEngineInfo)
+      .withSchema(engine, schema)
+      .withTableProperties(engine, replaceTableProps.asJava)
+    val txn = txnBuilder.build(engine)
+
+    commitTransaction(txn, engine, getAppendActions(txn, replaceData))
   }
 
   private def verifyBaseRowIDs(
@@ -416,6 +465,7 @@ class RowTrackingSuite extends DeltaTableWriteSuiteBase with ParquetSuiteBase {
       verifyDefaultRowCommitVersion(engine, tablePath, expectedDefaultRowCommitVersion)
       verifyHighWatermark(engine, tablePath, expectedHighWatermark)
     }
+
     verifyRowTrackingStates()
 
     // Create txn1 but don't commit it yet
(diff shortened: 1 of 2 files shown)


kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/RowTrackingSuite.scala
```diff
@@ -765,4 +813,62 @@ class RowTrackingSuite extends DeltaTableWriteSuiteBase with ParquetSuiteBase {
...
+          enableBefore = enableBefore,
+          enableAfter = enableAfter)
+
+        val snapshot =
+          TableImpl.forPath(engine, tablePath).getLatestSnapshot(engine).asInstanceOf[SnapshotImpl]
+
+        // Assert that row tracking is enabled/disabled as expected
+        assertMetadataProp(snapshot, TableConfig.ROW_TRACKING_ENABLED, enableAfter)
+
+        // Assert that materialized row tracking columns are present when row tracking is enabled
```
johanl-db, line 861: Can you also check that all `add` action in the latest snapshot all have `bas_row_id` and `default_row_commit_version` assigned?  Which will be trivially true if there are no files in the case where no data is written after replace

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/RowTrackingSuite.scala
```diff
@@ -55,6 +56,52 @@ class RowTrackingSuite extends DeltaTableWriteSuiteBase with ParquetSuiteBase {
...
+   * @param engine The Delta Kernel engine to use.
+   * @param tablePath The path to the Delta table.
+   * @param schema The schema to use for the table.
+   * @param extraProps Additional table properties.
+   * @param initialData Data to append after initial table creation (before replace).
+   * @param replaceData Data to append as part of the replace transaction.
+   * @param enableBefore Whether to enable row tracking before the replace.
+   * @param enableAfter Whether to enable row tracking after the replace.
+   */
+  private def createAndReplaceTable(
```
johanl-db: nit: I would just inline this method in the test, unless you already intend to reuse it mostly as is for future tests.

In case you don't inline it: probably better to remove all default parameters (except maybe extraProps) since most of them aren't used anyway, better to be explicit about the test inputs passed

johanl-db (approved):

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/RowTrackingSuite.scala
lbhm: I added a check for baseRowIds and defaultCommitVersions. Is that sufficient?

lbhm: Resolving this based on our discussion.

KaiqiJinWow (commented):
Hi @lbhm @johanl-db, quick question:
After a REPLACE TABLE operation, do we preserve the highWaterMark value from before the replacement? This is the behavior we follow in Delta Spark.
Would it be possible to add a test case to verify this?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/RowTrackingSuite.scala
```diff
@@ -765,4 +767,93 @@ class RowTrackingSuite extends DeltaTableWriteSuiteBase with ParquetSuiteBase {
...
+        // Get the latest snapshot of the table after the replace operation
+        val snapshot =
+          TableImpl.forPath(engine, tablePath).getLatestSnapshot(engine).asInstanceOf[SnapshotImpl]
+
+        // Assert that row tracking is enabled/disabled as expected
+        assertMetadataProp(snapshot, TableConfig.ROW_TRACKING_ENABLED, enableAfter)
+
+        // Assert that materialized row tracking columns are present when row tracking is enabled
+        if (enableAfter) {
+          val config = snapshot.getMetadata.getConfiguration
```
allisonport-db: Should these confs be preserved or different for (enabledBefore, enabledAfter)? Can we check this?

allisonport-db (commented):
Just 1 minor comment

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/RowTrackingSuite.scala
lbhm: The configs will be different unless row tracking is disabled before and after the replace.

If it is enabled before and after the replace, the current implementation will assign new column names for materialized row IDs and commit versions. Is this intended behavior @allisonport-db @johanl-db or do we need to fix this?

lbhm:
I added two new checks to address your comments.

@KaiqiJinWow I am correct to assume that your scenario only applies if:
- row tracking was enabled before
- row tracking is enabled after
- the tabled contained data before the replace (an empty table has no materialized high watermark)

@allisonport-db Please see my response to your comment and verify whether the current behavior is intended.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/RowTrackingSuite.scala
allisonport-db: Personally no idea on intended behavior for row tracking here was asking if anyone else knew :) it seems fine to me that they change but definitely not the expert on this

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/RowTrackingSuite.scala
```diff
@@ -765,4 +767,119 @@ class RowTrackingSuite extends DeltaTableWriteSuiteBase with ParquetSuiteBase {
...
+        // This only applies if row tracking is enabled before and after the replace operation
+        // and if there is initial data present
+        if (enableBefore && enableAfter && initialData.nonEmpty) {
+          val beforeHighWaterMark: Optional[Long] =
+            RowTrackingMetadataDomain.fromSnapshot(beforeSnapshot).map(_.getRowIdHighWaterMark)
+          val afterHighWaterMark: Optional[Long] =
+            RowTrackingMetadataDomain.fromSnapshot(afterSnapshot).map(_.getRowIdHighWaterMark)
+          if (replaceData.nonEmpty) {
+            // If replace data is provided, the high watermark should be incremented
+            assert(beforeHighWaterMark.get() < afterHighWaterMark.get())
```
KaiqiJinWow: can we also assert its actual value after replace, which should be 299?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/RowTrackingSuite.scala
```diff
@@ -765,4 +767,119 @@ class RowTrackingSuite extends DeltaTableWriteSuiteBase with ParquetSuiteBase {
...
+          } else {
+            // If no replace data, the high watermark should remain the same
+            assert(beforeHighWaterMark.get() == afterHighWaterMark.get())
```
KaiqiJinWow, line 851: also assert this value should be 99?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/RowTrackingSuite.scala
lbhm: I just discussed this with @johanl-db. He agreed that the configs should generally differ if row tracking was enabled before and/or after the replace. I added a check that verifies this and a comment explaining the behavior.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/RowTrackingSuite.scala
lbhm: I extended the check to verify that the `afterHighWaterMark` either equals `numInitialRows - 1` or `numInitialRows + numReplaceRows - 1` depending on whether the replace table has any data.

Just FYI: The concrete numbers in this case would be 599 in the first case and 2599 in the second case since `dataBatches1` and `dataBatches2` have 600 and 2000 elements, respectively.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/RowTrackingSuite.scala
lbhm: See my comment above

KaiqiJinWow (approved):
LGTM!

PR #4900, opened 2025-07-08
[Kernel] Fix CRC listing to correctly terminate at the readVersion

allisonport-db:
Adds a test (mirrors the other tests in the suite just ensures a later CRC does exist). Confirmed that the test fails without the fix.

Also cherry-picked https://github.com/delta-io/delta/commit/ab1fa99f78631ff4e6592445d9d5d8b3e650674c as it's needed for the CI.

### DIFF
--- a/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/LogReplayEngineMetricsSuite.scala
+++ b/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/LogReplayEngineMetricsSuite.scala
@@ -446,15 +446,14 @@ class LogReplayEngineMetricsSuite extends AnyFunSuite with TestUtils {
   }
 
   test("checksum not found at the read version, but found at a previous version; " +
-    "crc exists after version queried") {
+    "checksum exists after version queried") {
     withTempDirAndMetricsEngine { (path, engine) =>
       // copy the golden table with crc files to the temp dir delete the checksum files
       // for version 5
       copyTable(getTestResourceFilePath("stream_table_optimize"), path)
-      Seq(5L).foreach { version =>
-        val crcFile = new File(path, f"_delta_log/$version%020d.crc")
-        assert(Files.deleteIfExists(crcFile.toPath))
-      }
+      val crcFile = new File(path, f"_delta_log/${5L}%020d.crc")
+      assert(Files.deleteIfExists(crcFile.toPath))
+      // Now table has CRC present for versions 0, 1, 2, 3, 4, 6 (missing for version 5)
 
       loadPandMCheckMetrics(
         Table.forPath(engine, path)
--- a/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/LogReplayEngineMetricsSuite.scala
+++ b/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/LogReplayEngineMetricsSuite.scala
@@ -445,6 +445,32 @@ class LogReplayEngineMetricsSuite extends AnyFunSuite with TestUtils {
     }
   }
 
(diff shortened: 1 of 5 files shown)


kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/LogReplayEngineMetricsSuite.scala
```diff
@@ -445,6 +445,32 @@ class LogReplayEngineMetricsSuite extends AnyFunSuite with TestUtils {
     }
   }
 
+  test("checksum not found at the read version, but found at a previous version; " +
```
scottsand-db, line 448: Any reason we cannot just cherry-pick https://github.com/delta-io/delta/commit/d95d2a867ca8247075fbbf155d1a951554557941 ?

allisonport-db: Requires other changes not available (we don't even have that suite present and I'm assuming not the helpers either)

scottsand-db: Makes sense then, thanks!

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/LogReplayEngineMetricsSuite.scala
```diff
@@ -445,6 +445,32 @@ class LogReplayEngineMetricsSuite extends AnyFunSuite with TestUtils {
...
+    "crc exists after version queried") {
```
scottsand-db: nit: this test case mentions both `checksum` and `crc` -- can we just pick one terminology?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/LogReplayEngineMetricsSuite.scala
```diff
@@ -445,6 +445,32 @@ class LogReplayEngineMetricsSuite extends AnyFunSuite with TestUtils {
...
+    withTempDirAndMetricsEngine { (path, engine) =>
+      // copy the golden table with crc files to the temp dir delete the checksum files
+      // for version 5
+      copyTable(getTestResourceFilePath("stream_table_optimize"), path)
+      Seq(5L).foreach { version =>
+        val crcFile = new File(path, f"_delta_log/$version%020d.crc")
+        assert(Files.deleteIfExists(crcFile.toPath))
+      }
```
scottsand-db: could you include a comment that informs the reader what the exected state of the table is now?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/LogReplayEngineMetricsSuite.scala
```diff
@@ -445,6 +445,32 @@ class LogReplayEngineMetricsSuite extends AnyFunSuite with TestUtils {
     }
   }
 
+  test("checksum not found at the read version, but found at a previous version; " +
+    "crc exists after version queried") {
+    withTempDirAndMetricsEngine { (path, engine) =>
+      // copy the golden table with crc files to the temp dir delete the checksum files
+      // for version 5
+      copyTable(getTestResourceFilePath("stream_table_optimize"), path)
+      Seq(5L).foreach { version =>
```
scottsand-db: nit: do we need this `Seq(5L).foreach { version =>`? Don't we just _know_ that the version is 5L?

huan233usc (approved):

PR #4903, opened 2025-07-09
[Spark] Add missing clock parameter

felipepessoto:
- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Fix a small bug where clock parameter is not sent to the overloaded method.

## How was this patch tested?
Existing unit tests

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala
@@ -819,7 +819,7 @@ object DeltaLog extends DeltaLogging {
   /** Helper for creating a log for the table. */
   def forTable(spark: SparkSession, tableName: TableIdentifier, clock: Clock): DeltaLog = {
     if (DeltaTableIdentifier.isDeltaPath(spark, tableName)) {
-      forTable(spark, new Path(tableName.table))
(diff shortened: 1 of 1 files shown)


felipepessoto:
@andreaschat-db @vkorukanti could you check this small PR too, please?

spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala
```diff
@@ -819,7 +819,7 @@ object DeltaLog extends DeltaLogging {
   /** Helper for creating a log for the table. */
   def forTable(spark: SparkSession, tableName: TableIdentifier, clock: Clock): DeltaLog = {
     if (DeltaTableIdentifier.isDeltaPath(spark, tableName)) {
-      forTable(spark, new Path(tableName.table))
+      forTable(spark, new Path(tableName.table), clock)
```
andreaschat-db, line 822: This looks to be already fixed in master?
https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala#L822

felipepessoto: Oh ok. Somebody already merged this fix

PR #4904, opened 2025-07-09
[Kerne] [Pagination] Read page with page token

mmmyr:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/4904/files) to review incremental changes.
- [**stack/multi-part**](https://github.com/delta-io/delta/pull/4904) [[Files changed](https://github.com/delta-io/delta/pull/4904/files)]
  - [stack/test-only](https://github.com/delta-io/delta/pull/4945) [[Files changed](https://github.com/delta-io/delta/pull/4945/files/6488aff125c8430f7ced56699d64ad75e31f1487..150ab817595734f5e89b46bcee84943caf378b46)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

Kernel is able to read a new page and return correct results with the given page token. 

## How was this patch tested?

PaginatedScanSuite.scala. 

Yes.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
@@ -234,7 +234,8 @@ public class PaginatedScanFilesIteratorImpl implements PaginatedScanFilesIterato
             && lastReturnedRowIndex + numRowsInBatch <= tokenLastReturnedRowIndexOpt.get()) {
           lastReturnedRowIndex += numRowsInBatch;
           logger.info(
-              "Pagination: skipping batch from a partially consumed file : {}, last row index is {}",
+              "Pagination: skipping batch from a partially consumed file : {}, "
+                  + "last row index is {}",
               batchFilePath,
               lastReturnedRowIndex);
           continue;


kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
```diff
@@ -112,13 +112,25 @@ public ActionsIterator(
...
+        .filter(this::paginatedFilter)
+        .collect(Collectors.toList()));
     this.deltaReadSchema = deltaReadSchema;
     this.checkpointReadSchema = checkpointReadSchema;
     this.actionsIter = Optional.empty();
     this.schemaContainsAddOrRemoveFiles = LogReplay.containsAddOrRemoveFileActions(deltaReadSchema);
-    this.paginationContextOpt = paginationContextOpt;
+  }
+
+  private boolean paginatedFilter(DeltaLogFile nextLogFile) {
```
scottsand-db: This method needs

1. javadocs (at the method level) -- at a high level, what does this do?
2. logging for the special cases -- when we are the first page, when we skip a file, etc.
3. in-line comments explaining what we are doing and why. why always returne true for .json files? what does `nextFilePath.compareTo(lastReadLogFilePathOpt.get()) <= 0` do and why?

scottsand-db: if you made this public and static and marked it as @VisibleForTesting, would that make this method easier to test?

Food for thought -- perhaps it is still really easy to unit test without that.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -95,8 +96,11 @@ public PaginatedScanFilesIteratorImpl(
 
   @Override
   public Optional<Row> getCurrentPageToken() {
-    // TODO: add checks for last page (to return Optional.empty)
     // TODO: replace hash value of predicate and log segment
+    if(!baseFilteredScanFilesIter.hasNext()) {
+    //  System.out.println("no pages are left");
```
scottsand-db: nit: remove

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
mmmyr: Yep a unit test specific for this function would be helpful. We only have E2E test now, which is a little bit hard to see what exact files we skip reading.

mmmyr: And unit test for this method should be easy to create.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -133,17 +137,41 @@ private void prepareNext() {
...
         "pre-computed number of selected rows doesn't exist!");
-    if (!batch.getFilePath().get().equals(lastReadLogFilePath)) {
-      lastReadLogFilePath = batch.getFilePath().get();
+
+    // ====== get batch metadata (filepath, row index) ==========
+    final String batchFilePath = batch.getFilePath().get();
+    final long numSelectedAddFilesInBatch = batch.getPreComputedNumSelectedRows().get();
+    final long numRowsInBatch = batch.getData().getSize();
+
+    if (lastReadLogFilePath == null || batchFilePath.compareTo(lastReadLogFilePath) < 0) {
```
scottsand-db: 1. this needs internal comments to explain what is happening and why
2. this also needs logging
3. for example, I don't know what or why we are doing `batchFilePath.compareTo(lastReadLogFilePath) < 0)` -- i think this means, if `batchFilePath` is "less than" `lastReadLogFilePath`, but I don't really know what this means for our logic?
---> can you give some examples? like are we comparing 0004.json to 0005.json? are we comparing 00010.checkpoint to 00011.json? what happens if we are comparing aaaa.sidecar vs bbbb.sidecar? or zzzz.sidecar vs aaaa.sidecar? i.e. sidecar lexicographical order is not guaranteed?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -133,17 +137,41 @@ private void prepareNext() {
...
       logger.info("filePath {}", lastReadLogFilePath);
     }
-    final long numSelectedAddFilesInBatch = batch.getPreComputedNumSelectedRows().get();
-    final long numRowsInBatch = batch.getData().getSize();
+    lastReturnedRowIndex += numRowsInBatch; // calculate the row index of the last row in current batch
+
+    // ====== decide if current batch should be emitted ==========
+    Optional<String> tokenFilePathOpt = paginationContext.getLastReadLogFilePath();
+    // If the batch is from a previously returned file, skip this batch
+    if(tokenFilePathOpt.isPresent() && batchFilePath.compareTo(tokenFilePathOpt.get()) > 0) {
```
scottsand-db: btw, I don't really know what `batchFilePath.compareTo(tokenFilePathOpt.get()) > 0` is supposed to semantically mean? if this was renamed to a nice helper method that took in two file paths A and B and the method was named `isBatchPathBeforeTokenPath(...)` that would be more helpful

I think we are assuming that file name alone lets us tell if files are frmo previous pages or not? but again UUID-named sidecars concern me here ...

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
scottsand-db: can we also state our exact assumptions and requirements? e.g. we assume and require that files are returned in a certain order, right? and let's take that further with examples: we assume that JSONs are consumed like X, we assume that multi-part-checkpoints are consumed like Y, and we assume something else for sidecars?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
mmmyr: I'll create a stack PR on top of this one (about sidecar) asap, and we can see how sidecar files will be handled (along with regular checkpoint files).

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
```diff
@@ -112,13 +112,53 @@ public ActionsIterator(
...
+   *
+   * @param nextLogFile the log file to evaluate
+   * @return {@code true} to include the file; {@code false} to skip it
+   */
+  // TODO: add logging and unit tests for this method
+  private boolean paginatedFilter(DeltaLogFile nextLogFile) {
+    if (!paginationContextOpt.isPresent()) return true;
+    Optional<String> lastReadLogFilePathOpt = paginationContextOpt.get().getLastReadLogFilePath();
+    if (!lastReadLogFilePathOpt.isPresent()) return true; // reading first page
+    if (nextLogFile.getLogType() == DeltaLogFile.LogType.V2_CHECKPOINT_MANIFEST)
```
scottsand-db: nit: please apply kernel formatting -- we require this have `{ }` brackets

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
```diff
@@ -112,13 +112,53 @@ public ActionsIterator(
...
+   * <ul>
+   *   <li>If pagination is not enabled (i.e., {@code paginationContextOpt} is not present), return
+   *       {@code true}.
+   *   <li>If the pagination context is present but doesn't include a last read log file path,
+   *       return {@code true} (indicates reading the first page).
+   *   <li>If the file is a JSON log file, return {@code true} — we never skip JSON files as they're
+   *       needed to build hash sets.
+   *   <li>If the file is a V2 checkpoint manifest, return {@code true} — these should never be
+   *       skipped.
+   *   <li>If the file is a checkpoint file and comes after the last log file recorded in the page
```
scottsand-db, line 146: any checkpoint file? does this include sidecars?

scottsand-db: I _think_ what you are doing is: you are handling multi-part checkpoint file skipping here, and you are handling sidecar skipping below. That seems like a _very_ important implementation detail that we should call out

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
```diff
@@ -299,11 +339,17 @@ public ColumnarBatch extractSidecarsFromBatch(
...
     ColumnVector sidecarVector = columnarBatch.getColumnVector(sidecarIndex);
+    int sidecarCnt = -1; // sidecar index start from 0
     for (int i = 0; i < columnarBatch.getSize(); i++) {
       SidecarFile sidecarFile = SidecarFile.fromColumnVector(sidecarVector, i);
       if (sidecarFile == null) {
         continue;
       }
+      sidecarCnt++;
+      if (paginationContextOpt.get().getLastReadSidecarFileIdx().isPresent()
+          && sidecarCnt < paginationContextOpt.get().getLastReadSidecarFileIdx().get()) {
```
scottsand-db: nice! seems like a _very_ important place to log, eh?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -57,11 +57,28 @@ public class PaginatedScanFilesIteratorImpl implements PaginatedScanFilesIterato
...
 
   /**
-   * The name of the last log file that was read. This value corresponds to the one saved in the
-   * page token if present.
+   * The name of the last log file that was read during pagination.
+   *
+   * <p>This value is used to track which log file the current scan is processing.
+   *
+   * <p>Initialization: - If the pagination token includes a log file path, this value is
+   * initialized from it. - If the pagination token does not include a log file path (i.e., the
```
scottsand-db: I see you are using `-` for different points. As you can see, the java formatter doesn't like this. Can you please use `<ul><li>...`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
mmmyr: Yep, I think I will log: 1. the name of sidecar / checkpoint files to skip (along with its index in v2 manifest file if it's sidecar file) 2. the total number of skipped files.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -91,17 +109,27 @@ public PaginatedScanFilesIteratorImpl(
     this.baseFilteredScanFilesIter = baseFilteredScanFilesIter;
     this.pageSize = paginationContext.getPageSize();
     this.paginationContext = paginationContext;
+    if (paginationContext.getLastReadSidecarFileIdx().isPresent()) {
+      lastSidecarIndex = paginationContext.getLastReadSidecarFileIdx().get();
+    }
+    if (paginationContext.getLastReadLogFilePath().isPresent()) {
+      lastReadLogFilePath = paginationContext.getLastReadLogFilePath().get();
```
scottsand-db, line 126: Nit: Can you please initialize member variables in the same order in which they were declared above?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +148,125 @@ public boolean hasNext() {
...
+      final FilteredColumnarBatch batch = baseFilteredScanFilesIter.next();
+      // FilePath and pre-computed number of selected rows are expected to be present; both are
+      // computed and set in ActiveAddFilesIterator (when building FilteredColumnarBatch from
+      // ActionWrapper)
+      checkArgument(batch.getFilePath().isPresent(), "file path doesn't exist!");
+      checkArgument(
+          batch.getPreComputedNumSelectedRows().isPresent(),
+          "pre-computed number of selected rows doesn't exist!");
+
+      final String batchFilePath = batch.getFilePath().get(); // which file this batch comes from
```
scottsand-db: `// which file this batch comes from` -- I don't think we need this comment

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +148,125 @@ public boolean hasNext() {
...
+
+      // Skip batch if it;s from a fully consumed file (all data have been included in previous
```
scottsand-db: nit: can you move this comment to inside the `if` block? that makes it even tigher coupled to the code

nit: ` it;s`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +148,125 @@ public boolean hasNext() {
...
+      // pages).
+      if (isFromFullyConsumedFile(batchFilePath, tokenFilePathOpt, tokenSidecarIndexOpt)) {
```
scottsand-db: nit: is _what_ from fully consumed file? probably `isBatchFromFullyConsumedFile`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +148,125 @@ public boolean hasNext() {
...
-    // ActionWrapper)
-    checkArgument(batch.getFilePath().isPresent(), "file path doesn't exist!");
-    checkArgument(
-        batch.getPreComputedNumSelectedRows().isPresent(),
-        "pre-computed number of selected rows doesn't exist!");
-    if (!batch.getFilePath().get().equals(lastReadLogFilePath)) {
-      lastReadLogFilePath = batch.getFilePath().get();
-      lastReturnedRowIndex = -1;
-      logger.info("filePath {}", lastReadLogFilePath);
+      // Reset row index if this batch is from a new file not yet seen in previous pages.
```
scottsand-db: nit: you should move this comment to be right next to `lastReturnedRowIndex = -1;`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +148,125 @@ public boolean hasNext() {
...
+   * stored in the pagination context.
+   */
+  // TODO: add logging
   private void prepareNext() {
     if (currentBatch.isPresent()) return;
-    if (numScanFilesReturned >= pageSize) return;
     if (!baseFilteredScanFilesIter.hasNext()) return;
+    if (numScanFilesReturned >= pageSize) return;
+
+    Optional<String> tokenFilePathOpt = paginationContext.getLastReadLogFilePath();
```
scottsand-db: `tokenLastReadFilePathOpt`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +148,125 @@ public boolean hasNext() {
...
+    Optional<Long> tokenSidecarIndexOpt = paginationContext.getLastReadSidecarFileIdx();
```
scottsand-db: `tokenLastReadSidecarFileIdx`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +148,125 @@ public boolean hasNext() {
...
+
+    while (baseFilteredScanFilesIter.hasNext() && numScanFilesReturned < pageSize) {
```
scottsand-db, line 189: I think the `while` loop here can be cleaned up a bit, wdyt?

```java
    while (baseFilteredScanFilesIter.hasNext() && numScanFilesReturned < pageSize) {
      final FilteredColumnarBatch batch = baseFilteredScanFilesIter.next();
    
      validateBatch(batch);

      final String batchFilePath = batch.getFilePath().get(); // which file this batch comes from

      // Case 1: <describe with example>
      if (isBatchFromFullyConsumedFile(batchFilePath, tokenFilePathOpt, tokenSidecarIndexOpt)) {
        continue;
      }

      // Case 2: <describe with example>
      if (isBatchFromLastFileInToken(...)) {
        // skip if we have already consumed this batch
      }

      // Case 3: <describe with example>
      if (isBatchFromNewFile(..)) {
        // re-assign our file path, row index, and sidecar member variables
      }

      // Case 4: Else, we need to emit this batch!
```

mmmyr: Great advice! I'll do that. I think this will make the code super clean and easy-to-read.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
```diff
@@ -84,7 +87,10 @@ public class ActionsIterator implements CloseableIterator<ActionWrapper> {
    */
   private Optional<CloseableIterator<ActionWrapper>> actionsIter;
 
-  private final Optional<PaginationContext> paginationContextOpt;
+  private final Optional<PaginationContext> paginationContextOpt;;
+  private long numCheckpointFilesSkipped = 0;
+
```
scottsand-db: nit: delete white space; you can group all these member variables together if there's no comments in between them

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
```diff
@@ -112,13 +118,70 @@ public ActionsIterator(
...
+    String nextFilePath = nextLogFile.getFile().getPath();
+    if (nextFilePath.endsWith(".json")) {
+      return true; // never skip JSON files (to build tombstone hashsets)
+    }
+    // If nextFilePath is less than or equal to lastReadLogFilePath, it means nextLogFile
+    // comes before or is the same as the last read log file, so we should include it.
+    if (nextFilePath.compareTo(lastReadLogFilePathOpt.get()) <= 0) {
+      return true;
+    }
+    logger.info("Skip reading log file{}", nextFilePath);
```
scottsand-db: nit: add a space between `file{}`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
```diff
@@ -112,13 +118,70 @@ public ActionsIterator(
...
+    numCheckpointFilesSkipped++;
```
scottsand-db: is it ever possible that we read a `.crc` file here?

what if there is a future `00007.json.foo.bar` file in delta? is this code future proof?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
```diff
@@ -112,13 +118,70 @@ public ActionsIterator(
...
+      return true; // never skip v2 manifest checkpoint file
+    }
+    String nextFilePath = nextLogFile.getFile().getPath();
+    if (nextFilePath.endsWith(".json")) {
+      return true; // never skip JSON files (to build tombstone hashsets)
+    }
+    // If nextFilePath is less than or equal to lastReadLogFilePath, it means nextLogFile
+    // comes before or is the same as the last read log file, so we should include it.
+    if (nextFilePath.compareTo(lastReadLogFilePathOpt.get()) <= 0) {
+      return true;
```
scottsand-db, line 170: would be useful to log (a) `lastReadLogFilePat` and (b) `nextFilePath`, too

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
```diff
@@ -292,18 +355,28 @@ public ColumnarBatch extractSidecarsFromBatch(
...
 
+    logger.info("Extracting sidecar files from V2 manifest file");
     Path deltaLogPath = new Path(checkpointFileStatus.getPath()).getParent();
 
     // Sidecars will exist in schema. Extract sidecar files, then remove sidecar files from
     // batch output.
     List<DeltaLogFile> outputFiles = new ArrayList<>();
     int sidecarIndex = columnarBatch.getSchema().fieldNames().indexOf(LogReplay.SIDECAR_FIELD_NAME);
     ColumnVector sidecarVector = columnarBatch.getColumnVector(sidecarIndex);
+    int sidecarCnt = -1; // sidecar index start from 0
```
scottsand-db: would a better name be `sidecarIdx`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
```diff
@@ -292,18 +355,28 @@ public ColumnarBatch extractSidecarsFromBatch(
...
     for (int i = 0; i < columnarBatch.getSize(); i++) {
       SidecarFile sidecarFile = SidecarFile.fromColumnVector(sidecarVector, i);
       if (sidecarFile == null) {
         continue;
       }
+      sidecarCnt++;
+      if (paginationContextOpt.get().getLastReadSidecarFileIdx().isPresent()
```
scottsand-db: you are doing `paginationContextOpt.get()` but I don't think you have checked if the optional is present?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +157,173 @@ public boolean hasNext() {
...
+      final FilteredColumnarBatch batch = baseFilteredScanFilesIter.next();
+
+      validateBatch(batch);
+
+      final String batchFilePath = batch.getFilePath().get();
+      final long numRowsInBatch = batch.getData().getSize();
+
+      // Case 1: if the batch's from a fully consumed file (all data have been included in previous
```
scottsand-db: nit: it is more gramatically correct to say `if the batch is from`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +157,173 @@ public boolean hasNext() {
...
+      // pages), skip it. For example, if last page ends at 13.json, all batches from 14.json and
+      // 15.json will be skipped
+      if (isBatchFromFullyConsumedFile(
+          batchFilePath, tokenLastReadFilePathOpt, tokenLastReadSidecarFileIdxOpt)) {
+        // Only fully consumed JSON files and V2 manifest files won't be skipped in ActionsIterator.
```
scottsand-db: this would be a very useful place to log, eh?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +157,173 @@ public boolean hasNext() {
...
+        checkArgument(
+            batchFilePath.endsWith(".json") || FileNames.isV2CheckpointFile(batchFilePath));
+        continue;
+      }
+
+      // Case 2: if the batch is from the same last read file as recorded in the page token, decide
+      // if to skip it based on row index.
+      // For example, if last page ends at 13.json, we will skip batches in 13.json based on its row
```
scottsand-db: `based on its row index` --> what is the "it" here? could you clarify?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +157,173 @@ public boolean hasNext() {
...
+      // index.
+      else if (isBatchFromLastFileInToken(
+          batchFilePath, tokenLastReadFilePathOpt, tokenLastReadSidecarFileIdxOpt)) {
+        // Compare batch's row index to page token row
+        Optional<Long> tokenRowIndex = paginationContext.getLastReturnedRowIndex();
```
scottsand-db: can you help me to understand how we could get to this code and the `tokenRowIndex` is empty?

this makes me wonder if our PaginationContext or PageToken API is wrong or weird ... we should only ever have a valid paginationcontext/pagetoken and it should be really easy to use/consume in our code

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +157,173 @@ public boolean hasNext() {
...
+        checkArgument(tokenRowIndex.isPresent(), "Token row index is empty!");
+        // calculate the row index of the last row in current batch within the file
+        lastReturnedRowIndex += numRowsInBatch;
+        // Skip this batch if its last row index is smaller than or equal to the value in token.
+        if (lastReturnedRowIndex <= tokenRowIndex.get()) {
+          continue;
```
scottsand-db, line 241: this would be a great place to log

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
scottsand-db: `tokenLastReturnedRowIndex` would be a better variable name

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +157,173 @@ public boolean hasNext() {
...
+          lastReturnedRowIndex = -1;
+          logger.info("Reading new file: {}", lastReadLogFilePath);
+
+          // Sidecar index starts at -1 if none was seen in the previous page.
+          if (isSidecar(batchFilePath)) {
+            lastSidecarIndex++;
+          }
+        }
+        // calculate the row index of the last row in current batch within the file
+        lastReturnedRowIndex += numRowsInBatch;
```
scottsand-db: Case 3 says `if the batch is from a new file...` but this line of code is not within the `if (isBatchFromNewFile(batchFilePath)) {` scope?

and I'm not sure if that is intentional or not?

i.e. case 3 is not an `else if (...)`  it is an `else { if () { ...} and then another line. ofcode`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +157,173 @@ public boolean hasNext() {
...
+   *   <li>If the batch corresponds to a log file that appears earlier (in reverse lexicographic
+   *       order) than the file recorded in the token, it has already been fully processed and
+   *       should be skipped.
+   * </ul>
+   */
+  private boolean isBatchFromFullyConsumedFile(
+      String batchFilePath,
+      Optional<String> tokenFilePathOpt,
+      Optional<Long> tokenSidecarIndexOpt) {
+
```
scottsand-db: nit: in general can delete these empty lines at the start of methods

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +157,173 @@ public boolean hasNext() {
...
+  }
+
+  /**
+   * Returns {@code true} if the current batch comes from a fully consumed file, as determined by
+   * the page token.
+   *
+   * <p>Skip conditions:
+   *
+   * <ul>
+   *   <li>If a sidecar file was read in the previous page, all log files in the current page are
```
scottsand-db: sorry this confuses me -- you are comparing `sidecar` file to `log files`, can you give an example of this? or elabroate further about why this is the case?

scottsand-db: when you say "log files" do you mean "delta .json" files? if so can you please specify that

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +157,173 @@ public boolean hasNext() {
...
+          isSidecar(batchFilePath)
+              || lastReadLogFilePath == null
+              || batchFilePath.compareTo(lastReadLogFilePath) < 0);
+      return true;
+    }
+    return false;
+  }
+
+  private boolean isSidecar(String filePath) {
+    if (filePath.contains("/_delta_log/_sidecars/") && filePath.endsWith(".parquet")) {
```
scottsand-db: is there no utility in FileNames.java that helps here btw? if not, can you add it please? and add tests to FIleNamesSuite? (this can be refactor PR merged before this PR)

scottsand-db (changes requested):
Getter much better! Left some comments and questions

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
mmmyr: You are right, it's impossible. This is already ensured when building pagination context (page token is checked valid too).

mmmyr: I thought it's always good to have extra checkArgument() ? Maybe it's unnecessary.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
mmmyr: Let's say our log segment is: 10.json, 11.json, 12.json, 13.json. 
If last page we ends reading at 13.json,  isBatchFromNewFile only returns true when we reading the first batch of 12.json. When we're reading batch 2 of 12.json, isBatchFromNewFile() will return false, because the intent of this function is check if we should update lastReadLogFilePath and reset lastReturnedRowIndex to -1 if we're reading the first batch of a new file.

mmmyr: I think the function name and comments are confusing, I;ll add the example there : ).

mmmyr: Let's say number of rows in batch 1 and batch 2 in 12.json are both 5. When reading batch 1 in 12.json, lastReturnedRowIndex will first be reset to -1, and then become (-1+5) = 4. When reading batch 2 in 12.json, isBatchFromNewFile() will return false, so the new row index will be come (4+5) = 9. Hope this example makes sense : ) .

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
mmmyr: Yep. Because sidecar files always come at the end. Let's say from a pagination context we know the last read file is a sidecar file, it means we should skip all checkpoint files and json files before that. It might be good to add an example here.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
mmmyr: No function like "isSidecarFile(path)" in FileNames.java. I'll add it in separate PR.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
mmmyr: Kernel should not read a crc file in log segment. I'll add a checkArgument here.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
mmmyr: Yes, sorry about that.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
mmmyr: I update the name of function isBatchFromNewFIle() to isFirstBatchFromNewFile(), hope this clarifies the code.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
mmmyr: I'm working on testing the sidecar files now, will do the refactor after finishing testing the sidecar files (so that I can remove the UnsupportOperationException() here :) )

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
```diff
@@ -112,13 +116,73 @@ public ActionsIterator(
...
+            .map(DeltaLogFile::forFileStatus)
+            .filter(this::paginatedFilter)
+            .collect(Collectors.toList()));
     this.deltaReadSchema = deltaReadSchema;
     this.checkpointReadSchema = checkpointReadSchema;
     this.actionsIter = Optional.empty();
     this.schemaContainsAddOrRemoveFiles = LogReplay.containsAddOrRemoveFileActions(deltaReadSchema);
-    this.paginationContextOpt = paginationContextOpt;
+    if (paginationContextOpt.isPresent()) {
+      logger.info("Number of checkpoint files skipped is {}", numCheckpointFilesSkipped);
```
huan233usc: Why we need this log?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
```diff
@@ -112,13 +116,73 @@ public ActionsIterator(
...
+   * files are not included in the log segment list. Sidecar files are handled separately later,
+   * after the V2 manifest file has been read, specifically in the {@code extractSidecarFiles()}
+   * method.
+   *
+   * @param nextLogFile the log file to evaluate
+   * @return {@code true} to include the file; {@code false} to skip it
+   */
+  private boolean paginatedFilter(DeltaLogFile nextLogFile) {
+    String nextFilePath = nextLogFile.getFile().getPath();
+    checkArgument(nextFilePath.endsWith(".json") || nextFilePath.endsWith(".parquet"));
```
huan233usc: Can we just use the utils in kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/FileNames.java?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
```diff
@@ -112,13 +116,93 @@ public ActionsIterator(
...
+    logger.info("Pagination: nextFilePath is {}", nextFilePath);
+    switch (nextLogFile.getLogType()) {
+      case COMMIT:
+      case LOG_COMPACTION:
+      case CHECKPOINT_CLASSIC:
+      case V2_CHECKPOINT_MANIFEST:
+        return true;
+      case MULTIPART_CHECKPOINT:
+        if (isFullyConsumedFile(nextFilePath, lastReadLogFilePathOpt.get())) {
+          logger.info("Pagination: skip reading log file {}", nextFilePath);
```
scottsand-db: can we just say `skip reading multi-part checkpoint file {}`? i.e. why say "log file" when we know the exact type of file it is?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
```diff
@@ -112,13 +116,93 @@ public ActionsIterator(
...
+        throw new IllegalArgumentException("Unknown log file type!");
+    }
+  }
+
+  /**
+   * Returns true if the given file was already fully consumed in a previous page that ends at
+   * lastReadLogFilePath.
+   */
+  private boolean isFullyConsumedFile(String filePath, String lastReadLogFilePath) {
+    return filePath.compareTo(lastReadLogFilePath) > 0;
```
scottsand-db, line 205: Can you add an inline comment: why does "`>`" mean "before"? i.e. what are our assumptions and requirements in the code/sort order?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
```diff
@@ -292,18 +376,39 @@ public ColumnarBatch extractSidecarsFromBatch(
       FileStatus checkpointFileStatus, long checkpointVersion, ColumnarBatch columnarBatch) {
     checkArgument(columnarBatch.getSchema().fieldNames().contains(LogReplay.SIDECAR_FIELD_NAME));
 
+    logger.info("Extracting sidecar files from V2 manifest file");
```
scottsand-db: nit: we can delete this

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
```diff
@@ -292,18 +376,39 @@ public ColumnarBatch extractSidecarsFromBatch(
...
       SidecarFile sidecarFile = SidecarFile.fromColumnVector(sidecarVector, i);
       if (sidecarFile == null) {
         continue;
       }
+      sidecarIndexInV2Manifest++;
+      if (paginationContextOpt.isPresent()) {
+        logger.info(
+            "Pagination: next sidecar file is : index={}, path={}",
+            sidecarIndexInV2Manifest,
+            sidecarFile.getPath());
```
scottsand-db: IMO it is sufficient to log what we skip, and we don't need to log here always.
So, you can delete this logger.
Also, this would result in a double log when skip, which isn't great to look at.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -89,19 +116,29 @@ public PaginatedScanFilesIteratorImpl(
...
+    }
     // TODO: replace hash value of predicate and log segment
     Row pageTokenRow =
         new PageToken(
                 lastReadLogFilePath,
                 lastReturnedRowIndex,
-                Optional.empty() /* sidecar file index */,
+                (lastSidecarIndex == -1)
+                    ? Optional.empty()
+                    : Optional.of(lastSidecarIndex) /* sidecar file index */,
```
scottsand-db: nit: `Optional.of(lastSidecarIndex) /* sidecar file index */` --> the `lastSidecarIndex` already tells us what this is, we do don't need the `/* sidecar file index */`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +157,226 @@ public boolean hasNext() {
...
+      // than the last read file recorded in the page token.
+      //
+      // Example:
+      //   - Suppose the previous page ends at file 13.json
+      //   - Then, all files after 13.json (e.g., 14.json, 15.json, etc.)
+      //     have already been processed and are considered fully consumed.
+      //   - Any batches from these files should be skipped here.
+      if (isBatchFromFullyConsumedFile(
+          batchFilePath, tokenLastReadFilePathOpt, tokenLastReadSidecarFileIdxOpt)) {
+        // Only fully consumed JSON files and V2 manifest files won't be skipped in ActionsIterator.
```
scottsand-db: Would it be clearer just to say what _will_ be skipped? As opposed to saying what _wont_ be skipped?

scottsand-db: Btw we shouldn't just say "JSON" files, as that is not a formal Delta log type. There are delta files and minor log compaction files

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +157,226 @@ public boolean hasNext() {
...
+          batchFilePath, tokenLastReadFilePathOpt, tokenLastReadSidecarFileIdxOpt)) {
+        Optional<Long> tokenLastReturnedRowIndexOpt = paginationContext.getLastReturnedRowIndex();
+        // Calculate the row index of the last row in current batch within the file.
+        lastReturnedRowIndex += numRowsInBatch;
+        // Skip this batch if its last row index is smaller than or equal to last returned row index
+        // in token.
+        if (tokenLastReturnedRowIndexOpt.isPresent()
+            && lastReturnedRowIndex <= tokenLastReturnedRowIndexOpt.get()) {
+          logger.info(
+              "Pagination: skipping batch from a partially consumed file : {}", batchFilePath);
```
scottsand-db: would be great to log the indices here (i.e. these two indices are the critical pieces of information that inform whether or not we skip this batch)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +157,226 @@ public boolean hasNext() {
...
+        }
+      }
+
+      // Case 3: The batch is from a completely new file - one that has not been seen in previous
+      // pages.
+      // Do not skip it — this file is still fully unread.
+      // Example:
+      //   - Suppose the last page ended at file 13.json.
+      //   - In reverse lexicographic order, the next files to process will be 12.json, 11.json,
+      // etc.
```
scottsand-db: nit: indent this `etc.` so that it comes right underneath `In reverse`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +157,226 @@ public boolean hasNext() {
...
+        // in token.
+        if (tokenLastReturnedRowIndexOpt.isPresent()
+            && lastReturnedRowIndex <= tokenLastReturnedRowIndexOpt.get()) {
+          logger.info(
+              "Pagination: skipping batch from a partially consumed file : {}", batchFilePath);
+          continue;
+        }
+      }
+
+      // Case 3: The batch is from a completely new file - one that has not been seen in previous
```
scottsand-db: Can you clarify "new"? This is a "new" file that was not in the previous page, but it could be a file that we have already seen and emitted batches from in _this_ page, correct?

- page 1 was 20.json, 19.json, half of 18.json
- page 2 will skip the first half of 18.json, then start emitting
- page 2 now sees 17.json. we see the very first batch and emit it
- page 2 now sees the 2nd batch of 17.json. is 17.json considered a "completely new file"? see how "new" is a bit confusing? we need to give some more clarify/context here

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +157,226 @@ public boolean hasNext() {
...
+          "numSelectedAddFilesInBatch: {}, numRowsInBatch: {}",
+          numSelectedAddFilesInBatch,
+          numRowsInBatch);
+
+      // Found a valid batch, break out of the loop
+      break;
+    }
+  }
+
+  /** Validate current batch. */
```
scottsand-db: nit: superfluous comment

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +157,226 @@ public boolean hasNext() {
...
-        "pre-computed number of selected rows doesn't exist!");
-    if (!batch.getFilePath().get().equals(lastReadLogFilePath)) {
-      lastReadLogFilePath = batch.getFilePath().get();
-      lastReturnedRowIndex = -1;
-      logger.info("filePath {}", lastReadLogFilePath);
+        "Pre-computed number of selected rows doesn't exist!");
+  }
+
+  /**
+   * Returns {@code true} if the current batch comes from a fully consumed file, as determined by
```
scottsand-db: would `previously fully consumed file` be more clear?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +157,226 @@ public boolean hasNext() {
...
+   * the page token.
+   *
+   * <p>Logic:
+   *
+   * <ul>
+   *   <li>If a sidecar file was read in the previous page, all non-sidecar files (i.e., JSON and V2 manifest)
+   *  *       are considered consumed.
+   *   <li>Delta log files are ordered in reverse lexicographic order (i.e., higher file names appear earlier).
+   *  *       If the current batch’s log file name is greater than the last one recorded in the token,
+   *  *       it means this file appeared earlier in the segment and has already been processed.
```
scottsand-db: We can push these comments into the function. We don't need to read these comments here describing the logic, then look at the code, and then look at more comments again describing the logic

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +157,226 @@ public boolean hasNext() {
...
+  /**
+   * Returns true if the current batch is from the same file (either log or sidecar) that the
+   * previous page ended at, based on the file path and sidecar index recorded in the page token.
+   *
+   * <p>Logic:
+   *
+   * <ul>
+   *   <li>For regular log files: returns true if the file path matches the token’s last read path.
+   *   <li>For sidecar files: additionally checks the sidecar index presents and matches.
+   * </ul>
```
scottsand-db: Same here. Let's push the comments on the logic into the method

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -117,27 +125,92 @@ class PaginatedScanSuite extends AnyFunSuite with TestUtilsWithTableManagerAPIs
...
+
+      logger.info(s"New PageToken: lastReadLogFileName = $lastReadLogFilePath")
+      logger.info(s"New PageToken: lastReadRowIndex = $lastReturnedRowIndex")
+    }
+
+    (nextPageToken, returnedBatchesInPage)
   }
 
-  // TODO: test call hasNext() twice
+  private def runSingleTest(
```
scottsand-db: what does `Single` mean here? would `runSingleJsonFileTest` be a better name?

scottsand-db (changes requested):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +155,213 @@ public boolean hasNext() {
...
+              lastReturnedRowIndex);
+          continue;
+        }
+      }
+
+      // Case 3: This batch belongs to an "unseen file" — meaning a file whose content was
+      // not read at all in any previous page. In other words, this file is fully unconsumed:
+      // it was neither partially read nor fully read before.
+      //
+      // Batches from such files won't be skipped.
```
scottsand-db, line 248: what if the batch is entirely empty? e.g. the batch is entirely RemoveFiles? shouldn't we skip (not emit) that batch?

mmmyr: In normal scan, will this kind of batch be emitted?

scottsand-db: you don't need to implement/solve this in this PR, but if this case is possible/applicable then let's leave a TODO

mmmyr: But we build on top of normal scan, every batches returned from normal scan will all be returned in pagination.

scottsand-db: and also I think this should be done in the ScanImpl not in the PaginedScanImpl, so definitely not the responsibility of this PR

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +155,213 @@ public boolean hasNext() {
...
+   */
+  private boolean isBatchFromFullyConsumedFile(
+      String batchFilePath,
+      Optional<String> tokenLastReadLogFilePathOpt,
+      Optional<Long> tokenLastReadSidecarFileIdxOpt) {
+
+    if (tokenLastReadSidecarFileIdxOpt.isPresent()) {
+      // If a sidecar file was read in the previous page, all non-sidecar files (i.e., delta commit,
+      // log compaction
+      // and V2 checkpoint) are considered fully consumed.
```
scottsand-db: nit: weird new line in the middle of the comment

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
mmmyr: If we skip empty batch from normal scan in paginated iterator, it might be hard to test. because now we compare batches from normal scan to paginated scan.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +155,213 @@ public boolean hasNext() {
...
+      return !isSidecar(batchFilePath);
+    } else if (tokenLastReadLogFilePathOpt.isPresent()) {
+      //  Delta log files are ordered in reverse lexicographic order (i.e., higher file names appear
```
scottsand-db: nit: why two spaces before each starting letter in this comment?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
mmmyr: Yep we can check if ActiveAddFiles iterator do this check. If not, we can add that check : )

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -120,34 +155,213 @@ public boolean hasNext() {
...
+    return false;
+  }
+
+  /**
+   * Returns {@code true} if the current batch is the first one from a different file than the last
+   * seen, indicating the start of a new unseen file during pagination.
+   */
+  private boolean isFirstBatchFromUnseenFile(String batchFilePath) {
+    if (!batchFilePath.equals(lastReadLogFilePath)) {
+      // If batch isn't from a sidecar, it must appear after than lastReadLogFilePath in reverse
```
scottsand-db: nit: grammar `after than lastReadLogFilePath`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
mmmyr: Prob due to copy paste. Fixed.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -85,59 +85,149 @@ class PaginatedScanSuite extends AnyFunSuite with TestUtilsWithTableManagerAPIs
     buffer
   }
 
-  private def validatePageResults(
+  /** Test case to check if the result of a single page request is expected */
+  case class SinglePageRequestTestCase(
```
scottsand-db: This is only ever used for a first page, right? (That's my conclusion from looking at the test case code below)
Will it ever be used for a non-first page?
If not, can't we rename it to `FirstPageRequestTestContext` and instead of all of the `FirstPage` letters in the variable names, we can just say `expFileCnt, expBatchCnt, expLastReadLogFile, expLastReadRowIdx` ?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -85,59 +85,149 @@ class PaginatedScanSuite extends AnyFunSuite with TestUtilsWithTableManagerAPIs
...
+        pageSize = testCase.pageSize)
+      pageTokenOpt = newPageTokenOpt
+      allBatchesPaginationScan ++= newReturnedBatchesInPage
+    }
+
+    val normalScan =
+      getScanBuilder(tablePath = tablePath, tableVersionOpt = tableVersionOpt).build()
+
+    val iter = normalScan.getScanFiles(customEngine)
+    val allBatchesNormalScan = collectPaginatedBatches(iter)
```
scottsand-db: you are calling `collectPaginatedBatches` but the `iter` is from a normal scan? perhaps your helper method should just be called `collectBatches`?

btw, `CloseableIterator` has a method on it called `toinMemoryList` you could probably use that?

https://github.com/delta-io/delta/blob/8b7fd6d67d33c321f1751d44103b90e187285867/kernel/kernel-api/src/main/java/io/delta/kernel/utils/CloseableIterator.java#L260

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -237,62 +325,59 @@ class PaginatedScanSuite extends AnyFunSuite with TestUtilsWithTableManagerAPIs
...
+      expLastReadLogFileInFirstPage = JSON_FILE_0,
+      expLastReadRowIdxInFirstPage = 7)).foreach { testCase =>
     test(s"Multiple JSON files - page size ${testCase.pageSize}") {
-      val tablePath = getTestResourceFilePath("kernel-pagination-all-jsons")
-      val nextPageToken = runSinglePaginationTestCase(
-        testName = s"Multiple JSON files - page size ${testCase.pageSize}",
-        tablePath = tablePath,
-        tableVersionOpt = Optional.empty(),
-        testCase)
+      runCompletePaginationTest(
```
scottsand-db, line 378: so this looks like you are refactoring existing tests now? 

could we split off this prefactor into its own PR? that would help make this PR (>950 LOC) a bit smaller, wdyt?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
mmmyr: Created #4941

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -72,72 +66,168 @@ class PaginatedScanSuite extends AnyFunSuite with TestUtilsWithTableManagerAPIs
...
+  private def createPaginatedScan(
+      tablePath: String,
+      tableVersionOpt: Optional[Long],
+      pageSize: Long,
+      pageTokenOpt: Optional[Row] = Optional.empty()): PaginatedScan = {
+    getScanBuilder(tablePath, tableVersionOpt).buildPaginated(pageSize, pageTokenOpt)
   }
 
-  private def validatePageResults(
+//  private def collectBatchesFromScan(
```
scottsand-db: nit: old code, let's delete this?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -381,64 +464,354 @@ class PaginatedScanSuite extends AnyFunSuite with TestUtilsWithTableManagerAPIs
...
+      runCompletePaginationTest(
+        testCase = testCase,
+        tableVersionOpt = Optional.of(0L),
+        tablePath = getTestResourceFilePath("kernel-pagination-multi-part-checkpoints"))
+    }
+  }
+
+  // ===== Multi-part checkpoint files with JSON files test cases =====
+  /**
+   * Log Segment List:
```
scottsand-db:
```
   * Log Segment List:
   * 00000000000000000002.json (1 AddFile)
   * Batch 1 (A): 1 AddFile, 2 rows
   * 00000000000000000001.json (1 AddFile)
   * Batch 1 (B): 1 AddFile, 2 rows
   * ```

File: kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala [hunk 57]
```diff
@@ -381,64 +464,354 @@ class PaginatedScanSuite extends AnyFunSuite with TestUtilsWithTableManagerAPIs
...
+   * Batch 2: 2 AddFile, 2 row
+   * 00000000000000000000.checkpoint.0000000001.0000000003.parquet (5 AddFile)
+   * Batch 1: 3 AddFile, 5 row
+   * Batch 2: 2 AddFile, 2 row
+   */
+  Seq(
+    FirstPageRequestTestContext(
+      pageSize = 1,
+      expFileCnt = 1,
+      expBatchCnt = 1,
```
scottsand-db: expBatchCnt = 1, // Batches: A

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -381,64 +464,354 @@ class PaginatedScanSuite extends AnyFunSuite with TestUtilsWithTableManagerAPIs
...
+      expLastReadLogFile = JSON_FILE_2,
+      expLastReadRowIdx = 1),
+    FirstPageRequestTestContext(
+      pageSize = 2,
+      expFileCnt = 2,
+      expBatchCnt = 2,
```
scottsand-db: expBatchCnt = 2, // Batches: A, B

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -72,72 +66,168 @@ class PaginatedScanSuite extends AnyFunSuite with TestUtilsWithTableManagerAPIs
...
+//      val batch = paginatedIter.next()
+//      buffer += batch
+//    }
+//    buffer
+//  }
+
+  /** Test case to check if the result of a single page request is expected */
+  case class FirstPageRequestTestContext(
+      pageSize: Int,
+      expFileCnt: Int,
```
scottsand-db: `expScanFileCnt`

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -381,64 +464,354 @@ class PaginatedScanSuite extends AnyFunSuite with TestUtilsWithTableManagerAPIs
...
+  /**
+   * 2.checkpoint.uuid.json:
+   *  - Batch 1: 5 rows, 0 selected AddFiles
+   * sidecar 1:
+   *  - Batch 1: 1 rows, 1 selected AddFiles
+   * sidecar 2:
+   *  - Batch 1: 3 row, 3 selected AddFiles
+   */
+  val JSON_MANIFEST = "00000000000000000002.checkpoint.6374b053-df23-479b-b2cf-c9c550132b49.json"
+  val JSON_SIDECAR_1 =
```
scottsand-db: `JSON_MANIFEST_SIDECAR_1`

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -381,64 +464,354 @@ class PaginatedScanSuite extends AnyFunSuite with TestUtilsWithTableManagerAPIs
...
+   * - Batch 1: 5 rows, 0 selected AddFiles
+   * sidecar 1:
+   * - Batch 1: 3 rows, 3 selected AddFiles
+   * sidecar 2:
+   * - Batch 1: 1 row, 1 selected AddFiles
+   */
+
+  val PARQUET_SIDECAR_1 =
+    "00000000000000000002.checkpoint.0000000001.0000000002.055454d8-329c-4e0e-864d-7f867075af33.parquet"
+  val PARQUET_SIDECAR_2 =
```
scottsand-db: `PARQUET_MANIFEST_SIDECAR_2`

scottsand-db (changes requested):

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -72,72 +66,220 @@ class PaginatedScanSuite extends AnyFunSuite with TestUtilsWithTableManagerAPIs
...
+      val normalBatch = allBatchesNormalScan(i)
+      val paginatedBatch = allBatchesPaginationScan(i)
+      assert(normalBatch.getFilePath.equals(paginatedBatch.getFilePath))
+      assert(normalBatch.getData.getSize == paginatedBatch.getData.getSize)
+    }
   }
 
+  // ==== Test Paginated Iterator Behaviors ======
   // TODO: test call hasNext() twice
+  test("Call getCurrentPageToken() without calling any next() on an iterator") {
```
scottsand-db: `Call getCurrentPageToken() without calling any next() on an iterator` -- so this is what the test _does_ but not what the test is actually testing/checking.

A better test name would be: `Calling getCurrentPageToken() without calling next() throws Exception`

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -72,72 +66,220 @@ class PaginatedScanSuite extends AnyFunSuite with TestUtilsWithTableManagerAPIs
...
+        firstPaginatedIter.hasNext
+        firstPaginatedIter.getCurrentPageToken.get
+      }
+      assert(e.getMessage.contains("Can't call getCurrentPageToken()"))
+
+      firstPaginatedIter.close()
+    }
+  }
+
+  test("Call getCurrentPageToken() right after hasNext()") {
```
scottsand-db: a better test name would be `getCurrentPageToken() is impacted only by next() calls, not hasNext() calls`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -156,11 +260,137 @@ public FilteredColumnarBatch next() {
...
+   * Returns {@code true} if the current batch is the first one from a different file than the last
+   * seen, indicating the start of a new unseen file during pagination.
+   */
+  private boolean isFirstBatchFromUnseenFile(String batchFilePath) {
+    if (!batchFilePath.equals(lastReadLogFilePath)) {
+      // If batch isn't from a sidecar, it must appear after lastReadLogFilePath in reverse
+      // lexicographic order.
+      // If the batch's file path differs from {@code lastReadLogFilePath}, it's considered an
+      // unseen
+      // file.
```
scottsand-db: nit: this can be moved up to the above line

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -72,72 +66,220 @@ class PaginatedScanSuite extends AnyFunSuite with TestUtilsWithTableManagerAPIs
...
+  private def createPaginatedScan(
+      tablePath: String,
+      tableVersionOpt: Optional[Long],
+      pageSize: Long,
+      pageTokenOpt: Optional[Row] = Optional.empty()): PaginatedScan = {
+    getScanBuilder(tablePath, tableVersionOpt).buildPaginated(pageSize, pageTokenOpt)
   }
 
-  private def validatePageResults(
+  /** Test case to check if the result of a single page request is expected */
```
scottsand-db: nit: superfluous comment

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
```diff
@@ -112,13 +116,92 @@ public ActionsIterator(
...
+   *
+   * @param nextLogFile the log file to evaluate
+   * @return {@code true} to include the file; {@code false} to skip it
+   */
+  @VisibleForTesting
+  // TODO: verify numCheckpointFilesSkipped is correct in E2E test
+  // TODO: add unit test for this method
+  public boolean paginatedFilter(DeltaLogFile nextLogFile) {
+    // Pagination isn't enabled.
+    if (!paginationContextOpt.isPresent()) return true;
```
mmmyr, line 164: non null check: paginatino context

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ActionsIterator.java
```diff
@@ -297,13 +380,29 @@ public ColumnarBatch extractSidecarsFromBatch(
...
     for (int i = 0; i < columnarBatch.getSize(); i++) {
       SidecarFile sidecarFile = SidecarFile.fromColumnVector(sidecarVector, i);
       if (sidecarFile == null) {
         continue;
       }
+      sidecarIndexInV2Manifest++;
+      if (paginationContextOpt.isPresent()) {
+        if (paginationContextOpt.get().getLastReadSidecarFileIdx().isPresent()
+            && sidecarIndexInV2Manifest
+                < paginationContextOpt.get().getLastReadSidecarFileIdx().get()) {
```
mmmyr, line 399: doc: sidecar naming is not in order. so we must use index.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -57,11 +58,36 @@ public class PaginatedScanFilesIteratorImpl implements PaginatedScanFilesIterato
...
+   *
+   * <p>Initialization:
+   *
+   * <ul>
+   *   <li>If the pagination token includes a log file path, this value is initialized from it.
+   *   <li>If the pagination token does not include a log file path (i.e., the previous page did not
+   *       read any log file), this value is initialized to {@code null}.
+   * </ul>
    */
   private String lastReadLogFilePath = null;
```
mmmyr, line 73: ccv2? ratified commit.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -156,11 +260,134 @@ public FilteredColumnarBatch next() {
     if (!hasNext()) {
       throw new NoSuchElementException();
     }
+    hasLeastOneBatchConsumed = true;
     final FilteredColumnarBatch ret = currentBatch.get();
+    // Batch has been validated in prepareNext(), no need to re-validate it.
+    final long numSelectedAddFilesInBatch = ret.getPreComputedNumSelectedRows().get();
```
huan233usc, line 267: nit maybe checkState ret.getPreComputedNumSelectedRows() non empty, same below

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -156,11 +260,134 @@ public FilteredColumnarBatch next() {
...
+  private boolean isBatchFromLastFileInToken(
+      String batchFilePath,
+      Optional<String> tokenLastReadFilePathOpt,
+      Optional<Long> tokenLastReadSidecarFileIdxOpt) {
+    // Check if batch file path matches last read file path recorded in the page token (if
+    // present).
+    boolean isSameFilePath =
+        tokenLastReadFilePathOpt.isPresent()
+            && batchFilePath.equals(tokenLastReadFilePathOpt.get());
+    if (isSameFilePath) {
```
huan233usc: nit: for readablity you can early return
if(!isSameFilePath) {
return false
}
if (isSidecar(batchFilePath)) {
        checkArgument(
            tokenLastReadSidecarFileIdxOpt.isPresent()
                && lastSidecarIndex == tokenLastReadSidecarFileIdxOpt.get(),
            "Sidecar index mismatch for file: %s",
            batchFilePath);
      }
      return true;

huan233usc (approved):
LGTM, just some minor

PR #4905, opened 2025-07-09
[Kernel] [Pagination] Refactor + Check table path & table version when building Paginated Scan

mmmyr:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/PaginatedScanImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/PaginatedScanImpl.java
@@ -33,11 +33,22 @@ public class PaginatedScanImpl implements PaginatedScan {
   private final long pageSize;
   private final Optional<PageToken> pageTokenOpt;
   private final ScanImpl baseScan;
+  PaginationContext paginationContext;
 
   public PaginatedScanImpl(ScanImpl baseScan, Optional<Row> pageTokenRowOpt, long pageSize) {
     this.baseScan = baseScan;
     this.pageTokenOpt = pageTokenRowOpt.map(PageToken::fromRow);
     this.pageSize = pageSize;
+    this.paginationContext =
+        pageTokenOpt
+            .map(
+                token ->
+                    PaginationContext.forPageWithPageToken(
+                        pageSize, token, baseScan.getDataPath(), baseScan.getTableVersion()))
+            .orElseGet(
+                () ->
+                    PaginationContext.forFirstPage(
+                        pageSize, baseScan.getDataPath(), baseScan.getTableVersion()));
   }
 
   @Override
@@ -56,10 +67,6 @@ public class PaginatedScanImpl implements PaginatedScan {
   }
 
   public PaginatedScanFilesIterator getScanFiles(Engine engine, boolean includeStates) {
-    PaginationContext paginationContext =
-        pageTokenOpt
-            .map(token -> PaginationContext.forPageWithPageToken(pageSize, token))
-            .orElseGet(() -> PaginationContext.forFirstPage(pageSize));
     CloseableIterator<FilteredColumnarBatch> filteredScanFilesIter =
         baseScan.getScanFiles(engine, includeStates, Optional.of(paginationContext));
     return new PaginatedScanFilesIteratorImpl(filteredScanFilesIter, paginationContext);
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
@@ -19,6 +19,7 @@ import static io.delta.kernel.internal.DeltaErrors.wrapEngineException;
 import static io.delta.kernel.internal.skipping.StatsSchemaHelper.getStatsSchema;
 import static io.delta.kernel.internal.util.PartitionUtils.rewritePartitionPredicateOnCheckpointFileSchema;
 import static io.delta.kernel.internal.util.PartitionUtils.rewritePartitionPredicateOnScanFileSchema;
+import static io.delta.kernel.internal.util.Preconditions.checkArgument;
 import static java.util.function.Function.identity;
 import static java.util.stream.Collectors.toMap;
 
@@ -233,6 +234,15 @@ public class ScanImpl implements Scan {
   @Override
   public Optional<Predicate> getRemainingFilter() {
     return getDataFilters();
+  }
+
+  protected String getDataPath() {
+    return dataPath.toString();
+  }
+
+  protected long getTableVersion() {
+    checkArgument(snapshotReport.getVersion().isPresent(), "table  version isn't available!");
+    return snapshotReport.getVersion().get();
   }
 
   private Optional<Tuple2<Predicate, Predicate>> splitFilters(Optional<Predicate> filter) {
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
@@ -144,6 +144,14 @@ public class PageToken {
     return lastReadSidecarFileIdx;
   }
 
+  public String getTablePath() {
+    return tablePath;
+  }
+
+  public long getTableVersion() {
+    return tableVersion;
+  }
+
   @Override
   public boolean equals(Object obj) {
     if (this == obj) {
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginationContext.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginationContext.java
@@ -23,13 +23,19 @@ import java.util.Optional;
 /** {@code PaginationContext} carries pagination-related information. */
 public class PaginationContext {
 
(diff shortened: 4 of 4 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -235,6 +236,15 @@ public Optional<Predicate> getRemainingFilter() {
     return getDataFilters();
   }
 
+  protected String getDataPath() {
+    return dataPath.toString();
+  }
+
+  protected long getTableVersion() {
+    checkArgument(snapshotReport.getVersion().isPresent(), "table  version isn't available!");
```
scottsand-db: nit: double space `table  version`

scottsand-db: Can't we use a more reliable way to get the table version? Like, why even consider a solution that might be empty? We can just pass through this information from the scanBuilder

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PageToken.java
```diff
@@ -144,6 +144,16 @@ public Optional<Long> getLastReadSidecarFileIdx() {
...
 
+  public String getTablePath() {
+    return tablePath;
+  }
+
+  public long getTableVersion() {
+    return tableVersion;
+  }
+
+  public String getKernelVersion() {return kernelVersion;}
```
scottsand-db: formatting

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginationContext.java
```diff
@@ -38,12 +47,19 @@ public static PaginationContext forFirstPage(long pageSize) {
...
 
+  private final String tablePath;
+
+  private final long tableVersion;
+
   // TODO: add cached log replay hashsets related info
 
-  private PaginationContext(long pageSize, Optional<PageToken> pageToken) {
+  private PaginationContext(
+      long pageSize, Optional<PageToken> pageToken, String tablePath, long tableVersion) {
```
scottsand-db: nit: Optional params should come at the end

and please re-arrange the getters and the member variable declarations to match the same order

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginationContext.java
```diff
@@ -23,13 +25,20 @@
 /** {@code PaginationContext} carries pagination-related information. */
 public class PaginationContext {
 
-  public static PaginationContext forPageWithPageToken(long pageSize, PageToken pageToken) {
+  public static PaginationContext forPageWithPageToken(
+      long pageSize, PageToken pageToken, String tablePath, long tableVersion) {
     Objects.requireNonNull(pageToken, "page token is null");
-    return new PaginationContext(pageSize, Optional.of(pageToken));
+    Objects.requireNonNull(tablePath, "table Path is null");
```
scottsand-db: table `path` not `Path`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginationContext.java
```diff
@@ -23,13 +25,20 @@
...
+    checkArgument(tablePath.equals(pageToken.getTablePath()), "table path changes!");
```
scottsand-db: can we throw a clearer error message here?

i.e. the X in the pageToken does not match the X that was taken from the ScanBuilder/Scan context -- i.e. you probably passed in the incorrect pageToken?

We want these to be helpful, descriptive error messages that the pagination connector can use/see to reason about what went wrong

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
mmmyr:
```
public class ScanBuilderImpl implements ScanBuilder {

  private final Path dataPath;
  private final Metadata metadata;
  private final StructType snapshotSchema;
  private final LogReplay logReplay;
  private final SnapshotReport snapshotReport;

  private StructType readSchema;
  private Optional<Predicate> predicate;
```
ScanBuilder class doesn't have a member variable of "tableVersion"; it's saved in snapshortReport too, and SnapshotReport::getTableVersion() returns an empty value. 

But Snapshot class knows table version, we can pass down table version from snapshot to scanbuilder? 
```
public ScanBuilder getScanBuilder() {
    return new ScanBuilderImpl(
        dataPath, protocol, metadata, getSchema(), logReplay, getSnapshotReport());
  }
```
Like add a parameter "tableVersion" here.

scottsand-db: > But Snapshot class knows table version, we can pass down table version from snapshot to scanbuilder?

Yeah, let's do that 👍

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginationContext.java
```diff
@@ -17,21 +17,41 @@
...
 /** {@code PaginationContext} carries pagination-related information. */
 public class PaginationContext {
 
-  public static PaginationContext forPageWithPageToken(long pageSize, PageToken pageToken) {
+  public static PaginationContext forPageWithPageToken(
+      String tablePath, long tableVersion, long pageSize, PageToken pageToken) {
     Objects.requireNonNull(pageToken, "page token is null");
-    return new PaginationContext(pageSize, Optional.of(pageToken));
+    Objects.requireNonNull(tablePath, "table path is null");
+    checkArgument(
```
scottsand-db, line 31: Tests? and please update the PR description, thanks!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/PaginatedScanImpl.java
```diff
@@ -30,14 +30,27 @@
 
 /** Implementation of {@link PaginatedScan} */
 public class PaginatedScanImpl implements PaginatedScan {
+  private final ScanImpl baseScan;
+  PaginationContext paginationContext;
```
scottsand-db: is this purposefully non-final and using default package visibility? not private?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginationContext.java
mmmyr: Yep didn't add tests yet, will do : )

kernel/kernel-api/src/main/java/io/delta/kernel/internal/PaginatedScanImpl.java
mmmyr: I didn't push last commit when you comment : ). Already changed to private final, thanks!

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/PaginationContextSuite.scala
```diff
@@ -92,9 +108,54 @@ class PaginationContextSuite extends AnyFunSuite {
...
+        TEST_TABLE_PATH,
+        TEST_TABLE_VERSION,
+        negativePageSize,
+        validPageToken)
     }
     assert(e.getMessage === "Page size must be greater than zero!")
   }
 
+  test("should throw exception when the requested kernel version doesn't " +
+    "match the value in page token ") {
```
scottsand-db: nit: extra space at end of test name `n "`

scottsand-db (approved):
LGTM with 1 nit! Thanks!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginationContext.java
```diff
@@ -17,21 +17,41 @@
...
+        tablePath.equals(pageToken.getTablePath()),
+        "Invalid page token: table path does not match the requested table path.");
```
scottsand-db: can you also please print the values? e.g. it would be really helpful to see how the path, tableVersion, or kernelVersion differ

mmmyr: Added.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginationContext.java
```diff
@@ -17,21 +17,50 @@
...
 
-  public static PaginationContext forPageWithPageToken(long pageSize, PageToken pageToken) {
+  public static PaginationContext forPageWithPageToken(
+      String tablePath, long tableVersion, long pageSize, PageToken pageToken) {
     Objects.requireNonNull(pageToken, "page token is null");
-    return new PaginationContext(pageSize, Optional.of(pageToken));
+    Objects.requireNonNull(tablePath, "table path is null");
+    checkArgument(
+        tablePath.equals(pageToken.getTablePath()),
+        "Invalid page token: table path does not match the requested table path. "
```
scottsand-db: `table path does not match the requested table path` -- can you make it clearer what does `table path` mean and what does `requested table path` mean?

You can do this by adding the word `token`: `Invalid page token: token table path does not match the requested table path`

PR #4916, opened 2025-07-11
[Kernel] [CatalogManaged] Basic TransactionV2 and CommitContext implementation

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/4916/files) to review incremental changes.
- [**stack/kernel_catalog_managed_writes_2_simple**](https://github.com/delta-io/delta/pull/4916) [[Files changed](https://github.com/delta-io/delta/pull/4916/files)]
  - [stack/kernel_catalog_managed_writes_3_e2e_uc_prototype](https://github.com/delta-io/delta/pull/4921) [[Files changed](https://github.com/delta-io/delta/pull/4921/files/9b94774ba22e35303f32e94a1ff09e6d865cbf66..cb3e22772f249c23d4542a81c9d0502367d6fdd0)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This PR adds a sparse, lightweight, and basic implementation of TransactionV2 and CommitContext.

## How was this patch tested?

Very sparse CommitContext suite. More tests can be added later.

No.


kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/CommitContextImpl.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.stream.Collectors;
+
+public class CommitContextImpl implements CommitContext {
+
+  ////////////////////////////
+  // Static factory methods //
+  ////////////////////////////
+
+  /**
+   * Creates a {@link CommitContext} to be used for an initial commit only (e.g. not a retry, which
```
raveeram-db: Had a similar comment earlier, would be good to clarify `initial commit only` is for the initial commit attempt of a transaction to avoid confusion any potential confusion

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/CommitContextImpl.java
```diff
@@ -0,0 +1,154 @@
...
+    return toCloseableIterator(metadataActions.iterator());
+  }
+
+  private long getCommitAsVersion() {
+    return txnState.readTableOpt.map(readTable -> readTable.getVersion() + 1).orElse(0L);
+  }
+
+  private CommitInfo getCommitInfo() {
+    return new CommitInfo(
+        Optional.empty(), /* inCommitTimestampOpt */ // TODO: support ICT
```
raveeram-db, line 137: perhaps add issues for these. Seem like great first issues

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/CommitContextImpl.java
```diff
@@ -0,0 +1,154 @@
...
+        commitAttemptTimestampMs,
+        "Kernel-" + Meta.KERNEL_VERSION + "/" + txnState.engineInfo, /* engineInfo */
```
raveeram-db, line 139: Nice!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/transaction/TransactionV2State.java
```diff
@@ -0,0 +1,96 @@
...
+      String dataPath,
+      String logPath,
+      Optional<TransactionDataSource> readTableOpt,
+      Protocol protocol,
+      Metadata initialMetadata,
+      boolean isProtocolUpdate,
+      boolean isMetadataUpdate,
+      Clock clock,
+      Optional<SetTransaction> setTxnOpt) {
+    this.txnId = UUID.randomUUID().toString();
```
raveeram-db, line 110: should we allow engines to define/set a txnId as well and use this as a fallback?

raveeram-db (commented):
Overall flow looks great! Nice work!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/transaction/TransactionV2State.java
scottsand-db: Good question. That's never been a requirement so far -- we can add that when there is a need?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/CommitContextImpl.java
scottsand-db: Yup! Clarified and renamed the method to be even clearer. Thanks!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/transaction/TransactionV2Impl.java
```diff
@@ -0,0 +1,60 @@
...
+
+  @Override
+  public CommitContext getCommitContextForFirstCommitAttempt(
+      Engine engine, CloseableIterator<Row> dataActions) {
+    if (TableFeatures.isRowTrackingSupported(txnState.protocol)) {
+      throw new UnsupportedOperationException("Row Tracking not yet supported in TransactionV2");
+    }
+    if (IN_COMMIT_TIMESTAMPS_ENABLED.fromMetadata(txnState.initialMetadata)) {
+      throw new UnsupportedOperationException(
+          "In Commit Timestamps  not yet supported in TransactionV2");
```
scottsand-db: double space after `Timestamps`

kernel/kernel-api/src/main/java/io/delta/kernel/transaction/TransactionV2.java
```diff
@@ -57,6 +57,20 @@
...
+ * <p><b>Example Usages:</b>
+ *
+ * <p>Case 1: No action materialization, no retries
+ *
+ * <pre>{@code
+ * CloseableIterator<Row> allDataActionsIter = ... // Collect from executors
+ * CommitContext context = txn.getCommitContextForFirstCommitAttempt(engine, allDataActionsIter);
+ * resolvedTable.getCommitter.commit(
+ *   engine,
+ *   context.getFinalizedActions,
```
scottsand-db: `getFinalizedActions()`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/CommitContextImpl.java
```diff
@@ -0,0 +1,155 @@
...
+   * Creates a {@link CommitContext} to be used for the first commit attempt of a transaction only
+   * (e.g. not a retry, which requires conflict detection and potential action reconciliation and
+   * rebasing).
+   *
+   * @param finalizedDataActions the finalized data actions to be committed. That is, all updates to
+   *     the data actions (e.g. row tracking updates) have been applied. Note that these are the
+   *     data actions, not the metadata actions (like CommitInfo, Protocol, Metadata,
+   *     SetTransaction, DomainMetadata, etc.).
+   */
+  public static CommitContextImpl forFirstCommitAttempt(
```
nicklan, line 54: why `forFirstCommitAttempt`? Do we expect different arguments for subsequent attempts?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/CommitContextImpl.java
```diff
@@ -0,0 +1,155 @@
...
+    this.commitInfo = getCommitInfo();
+    this.iteratorConsumed = false;
+  }
+
+  /////////////////
+  // Public APIs //
+  /////////////////
+
+  @Override
+  public CloseableIterator<Row> getFinalizedActions() {
```
nicklan, line 89: Do we just blanket state that most kernel apis aren't thread safe? Should we be using an atomic testAndSet here?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/transaction/TransactionDataSource.java
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * As we transition from {@link io.delta.kernel.Transaction} (and {@link io.delta.kernel.Snapshot})
+ * to {@link io.delta.kernel.transaction.TransactionV2} (and {@link io.delta.kernel.ResolvedTable}),
+ * different parts of internal Kernel code will have to deal with both types of table "sources"
+ * (Snapshot, ResolvedTable).
+ *
+ * <p>This is a common interface for the minimal set of methods that both sources will have to
+ * expose.
+ */
+public interface TransactionDataSource {
```
nicklan: So does this _only_ exist so we can have _either_ a `Snapshot` _or_ a `ResolvedTable` back the txn? 

I'm kinda okay with that, but I do think we'll probably want `Snapshot` to go away at some point and just become `ResolvedTable`. That's not fully settled yet. 

Is it hard to have `Snapshot` implement `ResolvedTable`? Looking at the `ResolvedTabled` APIs, I think it should be quite simple. Then a txn could just have a `ResolvedTable` that backs it and we can get rid of this interface.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/transaction/TransactionV2State.java
```diff
@@ -0,0 +1,96 @@
...
+  // TODO public final Optional<List<Column>> clusteringColumnsOpt;
+
+  public TransactionV2State(
+      boolean isCreateOrReplace,
+      String engineInfo,
+      Operation operation,
+      String dataPath,
+      String logPath,
+      Optional<TransactionDataSource> readTableOpt,
+      Protocol protocol,
```
nicklan: is this the new or old protocol? why do we have `initialMetadata` and not `initialProtocol`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/CommitContextImpl.java
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  @Override
+  public CommitMetadata getCommitMetadata() {
+    return new CommitMetadata(
+        getCommitAsVersion(),
+        txnState.logPath,
+        commitInfo,
+        txnState.readTableOpt.map(TransactionDataSource::getProtocol),
+        txnState.readTableOpt.map(TransactionDataSource::getMetadata),
```
nicklan: why not `txnState.initialMetadata`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/transaction/TransactionV2State.java
```diff
@@ -0,0 +1,96 @@
...
+  // TODO: final boolean shouldUpdateClusteringDomainMetadata
+
+  // ===== Test infrastructure =====
+  public final Clock clock;
+
+  // ===== Table feature specific fields =====
+  public final Optional<SetTransaction> setTxnOpt;
+  // TODO public final Optional<List<Column>> clusteringColumnsOpt;
+
+  public TransactionV2State(
```
nicklan, line 89: this is again "not really public" right, because it's an `internal` class?

nicklan (commented):
left some comments. one thought:
What if we didn't have a `getFinalizedActions()` method with no args. What if, instead, we had:
```java
  // don't take the finalized actions in the factory method(s)
  public static CommitContextImpl forFirstCommitAttempt(Engine engine, TransactionV2State txnState)
```
and then had a:

```java
  public CloseableIterator<Row> getFinalizedActions(CloseableIterator<Row> finalizedDataActions)
```

method, that just could generate the metadata actions every time and `combine` it with the passed `finalizedDataActions`. Now it's up to the engine to pass me a valid iterator each time it calls that, and it can figure out however it wants to do that, and the `CommitContextImpl` doesn't have to worry about if the iterator has been consumed or not.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/CommitContextImpl.java
```diff
@@ -0,0 +1,155 @@
...
+  private CommitInfo getCommitInfo() {
+    return new CommitInfo(
+        Optional.empty(), /* inCommitTimestampOpt */ // TODO: support ICT
+        commitAttemptTimestampMs,
+        "Kernel-" + Meta.KERNEL_VERSION + "/" + txnState.engineInfo, /* engineInfo */
+        txnState.operation.getDescription(), /* description */
+        getOperationParameters(), /* operationParameters */
+        true, /* isBlindAppend */
+        txnState.txnId, /* txnId */
+        Collections.emptyMap() /* operationMetrics */);
```
huan233usc, line 144: We may also want some TODO here

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/CommitContextImpl.java
scottsand-db: Yup! There will be separate `TransactionV2` APIs for detecting + resolving conflicts and rebasing. That will need to take in additional information. It follows that we will need a different API, too, for constructing the CommtiContextImpl.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/CommitContextImpl.java
scottsand-db: Great question. I don't think we blanket state or declare it anywhere, though we do have several spots like this in our code base.

I'd be happy to use a testAndSet to make this thread safe 👍 and we should also double click on this and perhaps do an audit of our APIs, or at least figure out the best implementation/semantic etc for the Kernel Java project as a whole

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/CommitContextImpl.java
scottsand-db: Great question. I should add better documentation for `txnState.initialMetadata`.

Suppose we read a table at version 10 with no table properties (tihs is just an example). Let's call this metadata `A`.
Suppose in our transaction builder we add table property `foo=bar`. Let's call this metadata `B`.

readTableOpt (TransactionDataSource::getMetadata) <==> `A`

txnState.initialMetadata ==> `B`

And so, ultimately, we use the `readTableOpt` metadata because this parameter for `CommitMetadata` is the readMetadataOpt, not the `newMetadataOpt`

If we update the metadata each commit attempt, then we will be creating `B'`, `B''`, `B'''`, etc. We do this for example when we update the ICT enablement table properties.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/transaction/TransactionDataSource.java
scottsand-db: > So does this only exist so we can have either a Snapshot or a ResolvedTable back the txn?

Yes. We have lots of helper methods today (e.g. in TransactionImpl.java, in ConflictChecker.java, for Row Tracking, for Domain Metadata, etc.)

We will have to update all of those APIs to take in _something_ common to both ResolvedTable and Snapshot, since we have to continue to support the Snapshot public API for now.

I think `interface TransactionDataSource` is a good way to go about that 👍 

> Is it hard to have Snapshot implement ResolvedTable?

I don't think we want to change our public Snapshot API? It would have to be able to produce a Committer and a TransactionBuilder

kernel/kernel-api/src/main/java/io/delta/kernel/internal/transaction/TransactionV2State.java
scottsand-db: Right 👍

kernel/kernel-api/src/main/java/io/delta/kernel/internal/transaction/TransactionV2State.java
scottsand-db: Great question. I can add javadocs to clarify.

The protocol actual should not (as of Kernel code now) ever change from commit to commit. So it is both the `initialProtocol` but also the **only** Protocol that each commit attempt will see.

As discussed https://github.com/delta-io/delta/pull/4916/files#r2208812123, the metadata _can_ change from commit to commit.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/CommitContextImpl.java
scottsand-db: I'll add a comment like this to clarify:

```java
  /**
   * The protocol to use for this transaction, updated with all applicable input from the
   * transaction builders.
   *
   * <p>For example, if you have a table with protocol P1 at version 10, then create a ResolvedTable
   * at version 10, then create a TransactionBuilder, and enable a new table feature on that
   * builder, and finally create a TransactionV2, then:
   *
   * <ul>
   *   <li>the readTableOpt's protocol will be P1
   *   <li>this value will be that new Protocol P2 with that table feature enabled.
   * </ul>
   *
   * <p>As of this writing, it never changes during a transaction, e.g. from commit attempt to
   * commit attempt.
   */
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/transaction/TransactionV2State.java
scottsand-db: I'll update it to this:

```java
  /**
   * The protocol to use for this transaction, updated with all applicable input from the
   * transaction builder.
   *
   * <p>For example, if you have a table with protocol P1 at version 10, then create a ResolvedTable
   * at version 10, then create a TransactionBuilder, then enable a new table feature on that
   * builder, and finally create a TransactionV2, then:
   *
   * <ul>
   *   <li>the readTableOpt's protocol will be P1
   *   <li>this value will be that new Protocol P2 with that table feature enabled
   * </ul>
   *
   * <p>As of this writing, it never changes during a transaction, e.g. from commit attempt to
   * commit attempt.
   */
  public final Protocol protocol;

  /**
   * The initial metadata to use for this transaction, updated with all the applicable input from
   * the transaction builder.
   *
   * <p>It is the <b>initial</b> metadata because the metadata to use from commit attempt to commit
   * attempt may change, actually, for example in the case when we are enabling ICT and must set the
   * ICT enablement table properties, such as delta.inCommitTimestampEnablementVersion
   */
  public final Metadata initialMetadata;
```

scottsand-db: I can perhaps rename `protocol` to `finalProtocol` ?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/transaction/TransactionDataSource.java
nicklan: Mmm okay. The point is that we wouldn't have to update those APIs at all when we deprecated snapshot if they just took a `ResolvedTable`. Assumably nowhere that we'd use a snapshot like that would we actually try and get a committer or builder. So we could have the implementation for those just throw, but I agree it's a bit awkward.

I'll leave it up to your judgement here.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/transaction/TransactionV2State.java
nicklan: Mmm yeah, those are kinda confusing. Maybe `resultantProtocol` or even just `newProtocol`. I think I'd also call it `newMetadata` or something. "Initial" gives a strong sense that it's the metadata from the _start_ of the txn.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/CommitContextImpl.java
nicklan: Is this because we don't have intent based APIs, so if we're rebasing we may find existing properties that weren't there before and have to update them? I'm not clear why that's not also true for the protocol, or maybe we update the metadata for a different reason. I'm not familiar with the flow for ICT enablement.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/transaction/TransactionV2State.java
```diff
@@ -0,0 +1,96 @@
...
+  public final Operation operation;
+
+  // ===== File System paths =====
+  public final String dataPath;
+  public final String logPath;
+
+  // ===== Table state =====
+  public final Optional<TransactionDataSource> readTableOpt;
+  public final Protocol protocol;
+  public final Metadata initialMetadata;
```
allisonport-db: This was a little confusing to me at first; maybe we can rename or add more docs on this?

initialMetadata to me sounds like the metadata of the snapshot read. Maybe `currentMetadata`? Open to ideas or just clarifying here what it is

allisonport-db: I see there's already comment on this below :) Just +1 then good with the proposal there

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/CommitContextImpl.java
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  private CommitInfo getCommitInfo() {
+    return new CommitInfo(
+        Optional.empty(), /* inCommitTimestampOpt */ // TODO: support ICT
+        commitAttemptTimestampMs,
+        "Kernel-" + Meta.KERNEL_VERSION + "/" + txnState.engineInfo, /* engineInfo */
+        txnState.operation.getDescription(), /* description */
+        getOperationParameters(), /* operationParameters */
+        true, /* isBlindAppend */
```
allisonport-db, line 142: I think we want to not hard-code this but we will revisit this when we come back to our APIs anyways

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/CommitContextImpl.java
scottsand-db: > Is this because we don't have intent based APIs, so if we're rebasing we may find existing properties that weren't there before and have to update them?

Good question. No that's not the case / why. As far as I know, we only update the metadata between commit attempts because of In-Commit-Timestamp requirements.

We need to set delta table properties in the commit that enables ICT. These properties include the enablement timetamp and version.

So if we want to write commit 10, we will use an enablement version of 10.

If we conflict with 10.json, we will then try again and write 11.json. But now in this commit attempt we need to update the ICT enablement version property to be 11.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/CommitContextImpl.java
scottsand-db: Yeah I'm just hard-coding this for now (trying to move fast). As we allow non-blind-appends and have new and improvemed intent-based APIs, we will for sure update this 👍

kernel/kernel-api/src/main/java/io/delta/kernel/internal/transaction/TransactionDataSource.java
scottsand-db: Fair point, Nick. A minor annoyance/complexity here is that ResovledTableInternal has `String getLogPath` while SnapshotImpl has `Path getLogPath`.

I think we can move forwards with the ResolvedTableInternal, not the TransactionDataSource, and we will inline just create a SnapshotToResolvedTableAdapter that resolves the `String/Path` minor difference 👍

scottsand-db:
Hi Nick. Great question. We can do that. I think there's just one or two cons to that approach
- Note that Transaction V2 takes in the "dataActions" but not the "finalizedDataActions": `  CommitContext getCommitContextForFirstCommitAttempt(Engine engine, CloseableIterator<Row> dataActions);`
- Note that CommitContextImpl is instantiated with the `finalizedDataActions` -- these are the data actions **updated** with the RowTrackingId stuff
- So, to back up, there are
  - dataActions: original data actions from Engine, this is taken in by Kernel (input)
  - finalizedDataActions: data actions updated by Kernel, this is not exposed in any public API
  - finalizedActions: the metadata actions + finalizedDataActions, this is expose by Kernel (output)
- So, in your proposed approach, `getFinalizedActions` would not be called with the `finalizedDataActions` but rather the `dataActions` -- since the Engine is in charge of giving us the initial data actions, and Kernel is in charge of figuring out the finalized data actions with all the row tracking stuff applied
- So, this means that Kernel would have to re-apply the row tracking stuff to the inputted dataActions every time `getFinalizedActions(CloseableIterator<Row> dataActions)` is called. That's probably fine but also maybe not great?
- I guess I'm also wondering: why is the Engine calling `getFInalizedActions` multiple times? If you got the finalized actions, but did not materialize it, and then you want to get the finalized actions again, well that means that you materialized the original & un-finalized data actions? but why did you do that? if you can materialize anything, you should materialize the finalized actions?

The benefit to that approach is, as you say, that CommitContextImpl doesn't have to worry aout if the iterator has been consumed or not. But is that that bad?

scottsand-db:
@nicklan -- Great chat offline. Posting here for visibility. As we discussed, the only acceptable way that we can rebase for commit attempt N is if the finalized actions from commit attempt N-1 were materialized.

That is, it is not acceptable to try and perform rebase for commit attempr 4 if we are only using the original data actions.

That's because: each rebase attempt we have to read the delta log files of all conflicting commits. If we have to re-read these files every time, thats O(N^2) effort (file IO, too) and that's not acceptable.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/transaction/TransactionV2State.java
```diff
@@ -0,0 +1,118 @@
...
+  /**
+   * The metadata to use for the first commit attempt of this transaction, updated with all the
+   * applicable input from the transaction builder.
+   *
+   * <p>Note that the metadata might change from commit attempt to commit attempt. For example, when
+   * ICT is enabled, each commit attempt needs to be updated to write the Delta table property
+   * 'delta.inCommitTimestampEnablementVersion', which changes from commit attempt to commit
+   * attempt.
+   */
+  public final Metadata updatedMetadataForFirstCommitAttempt;
```
allisonport-db: Are these null if `isProtocolUpdate=false` or `isMetadataUpdate=false`? Or are they just the same as the read protocol/metadata?

allisonport-db (approved):
LGTM please clarify in the docs for the 1 open question

kernel/kernel-api/src/main/java/io/delta/kernel/internal/transaction/TransactionV2State.java
scottsand-db: Good question. Certainly never `null` -- we should use `Optional` if we ever want to express that intent.

I'll clarify in the docs.

scottsand-db: The alternative is we get rid of `isProtocolUpdate` and `isMetadataUpdate` and just make `updatedProtocol` and `updatedMetadataForFirstCommitAttempt ` Optional

And really this decision comes down to how the Builder/Factory creates this information.

https://github.com/delta-io/delta/blob/257025d09c38acfc819ec30eda9480cbd4755d0a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java#L362

allisonport-db: I think optional makes the most sense to me for now -- we can always change it in a future PR if needed.

allisonport-db: As long as we have access to the read snapshot metadata/protocol (which we do) then optional should be sufficient

scottsand-db: Updated to use optional!

PR #4920, opened 2025-07-13
[Kernel

mmmyr:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
@@ -105,7 +105,7 @@ public class PaginatedScanFilesIteratorImpl implements PaginatedScanFilesIterato
                 Meta.KERNEL_VERSION,
                 paginationContext.getTablePath() /* table path */,
                 paginationContext.getTableVersion() /* table version */,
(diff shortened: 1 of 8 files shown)


scottsand-db (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
```diff
@@ -105,7 +105,7 @@ public Optional<Row> getCurrentPageToken() {
                 Meta.KERNEL_VERSION,
                 paginationContext.getTablePath() /* table path */,
                 paginationContext.getTableVersion() /* table version */,
-                -1 /* predicate hash */,
+                paginationContext.getPredicateHash() /* predicate hash */,
```
huan233usc: nit: you could remove /* predicate hash */ as paginationContext.getPredicateHash() is clear enough.

huan233usc (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/PaginatedScanFilesIteratorImpl.java
mmmyr: Yep thanks for reminding!

PR #4923, opened 2025-07-14
[Protocol] Fix typo in name of a row tracking domain metadata field in protocol spec

lbhm:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [x] Other (Protocol spec)

## Description

The protocol spec has a typo in the row tracking metadata domain. [Kernel implements the high water mark as `rowIdHighWaterMark`](https://github.com/delta-io/delta/blob/master/kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/RowTrackingMetadataDomain.java#L60) but the spec defines it to be `highWaterMark`. Since the current implementation is already used in production (i.e., delta-spark has been using that name for years), I suggest to adapt that specification instead of opting for a breaking change in the code base.

Since this PR touches the spec anyways, I also fixed some non-critical typos along the way.

## How was this patch tested?

No since no code was changed.

This PR introduces user-facing changes to anyone who wants to implement the row tracking feature as part of a Delta connector.

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -1312,7 +1312,7 @@ Writers that do not implement VACUUM do not need to change anything and can safe
 For tables with Vacuum Protocol Check enabled, readers don’t need to understand or change anything new; they just need to acknowledge the feature exists.
 
 Making this feature a ReaderWriter feature (rather than solely a Writer feature) ensures that:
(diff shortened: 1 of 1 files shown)


johanl-db:
@allisonport-db can you merge this PR?

@ryan-johnson-databricks is OOO (not sure when he'll be back), it's prob. not worth waiting for him for this. If he has concerns, we can address them in a follow-up

felipepessoto:
@lbhm @johanl-db is Spark-delta using the incorrect name? Or when it was fixed?

lbhm:
@felipepessoto delta-spark has always used `rowIdHighWaterMark`, which means that the name is already used by all Delta tables (written from Spark) today. That's why we made sure to also reflect this in the protocol spec.

PR #4926, opened 2025-07-14
[Kernel] [Pagination] Check Log Segment Changes between Page Requests

mmmyr:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

Check if the log segment has changed between page reqeusts. 

## How was this patch tested?

PaginationContextSuite.scala.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
@@ -389,6 +389,11 @@ public class LogSegment {
         lastCommitTimestamp);
   }
 
+  @Override
+  public int hashCode() {
+    return 0;
+  }
+
   private String formatList(List<FileStatus> list) {
     if (list.isEmpty()) {
       return "";
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
@@ -384,10 +384,9 @@ public class LogSegment {
         lastCommitTimestamp);
   }
 
-  /** Hash code function to check if two log segments are same. */
   @Override
   public int hashCode() {
-    // TODO: change FileStatus::hashCode for staged commits?
+    // TODO: support staged commits #4927
     return Objects.hash(deltas, checkpoints, compactions);
   }
 
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
@@ -26,12 +26,7 @@ import io.delta.kernel.internal.lang.ListUtils;
 import io.delta.kernel.internal.util.FileNames;
 import io.delta.kernel.internal.util.Tuple2;
 import io.delta.kernel.utils.FileStatus;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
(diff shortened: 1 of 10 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
```diff
@@ -389,6 +384,12 @@ public String toString() {
         lastCommitTimestamp);
   }
 
+  @Override
+  public int hashCode() {
+    // TODO: support staged commits #4927
+    return Objects.hash(deltas, checkpoints, compactions);
```
huan233usc, line 390: You may not really need compactions? cc @nicklan qq if compacted log exists, will logSegment still contain full delta?

scottsand-db: @huan233usc it will contain them 👍

mmmyr: If log segment will contain full delta when compacted log exists, it's worth checking compactions right? (like what we do here).

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/PaginationContextSuite.scala
```diff
@@ -182,11 +196,26 @@ class PaginationContextSuite extends AnyFunSuite {
...
         TEST_TABLE_VERSION,
+        TEST_LOG_SEGMENT_HASH,
         TEST_WRONG_PREDICATE_HASH,
         pageSize,
         validPageToken)
     }
     assert(e.getMessage.contains("Invalid page token: token predicate"))
   }
 
+  test("should throw exception for when the requested log segment doesn't " +
```
scottsand-db, line 207: Can we do an e2e test for this?

We could create a delta table with e.g. v2manfiest file A;
Then delete that file and add a new v2manifest file B

And our pagination logic should catch this

mmmyr: Sure, I'll include that test in this PR.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
```diff
@@ -137,6 +141,60 @@ class PaginatedScanSuite extends AnyFunSuite with TestUtilsWithTableManagerAPIs
...
+      spark.range(200, 250, 1, 5).write.format("delta").mode("append").save(tablePath)
+
+      // Request first page
+      val firstPageSize = 2L
+      val firstPaginatedScan = createPaginatedScan(
+        tablePath = tablePath,
+        tableVersionOpt = Optional.empty(),
+        pageSize = firstPageSize)
+      val firstPaginatedIter = firstPaginatedScan.getScanFiles(customEngine)
+      while (firstPaginatedIter.hasNext) firstPaginatedIter.next() // exhaust iterator
```
scottsand-db: btw, any reason we need to do this? I thought we designed our APIs such that we can call `getCurrentPageToken` at any point during the iteration?

huan233usc (approved):

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/PaginatedScanSuite.scala
mmmyr: We need to call next() at least once but no need to exhaust the iterator. Changed "while" to "if" here.

PR #4936, opened 2025-07-15
[Kernel] [CatalogManaged] Add new builder .withCommitter API; Create DefaultCommitter skeleton

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/4936/files) to review incremental changes.
- [**stack/kernel_catalog_managed_writes_get_committer**](https://github.com/delta-io/delta/pull/4936) [[Files changed](https://github.com/delta-io/delta/pull/4936/files)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

https://github.com/delta-io/delta/pull/4814 introduced new TransactionV2, Committer, CommitContext, etc APIs.

This PR now adds the abilities for Engines to inject into Kernel the actual Committer.

This PR also includes a no-op Default committer implementation that validates that it is not interacting with catalog-managed tables. We do this because: ResolvedTable does not return an Optional committer, but rather just a committer. So we need to give it _something_ in the case that no committer was provided to the table builder.

## How was this patch tested?

New UTs.

New .withCommitter API

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/DefaultFileSystemManagedTableOnlyCommitter.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/DefaultFileSystemManagedTableOnlyCommitter.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright (2025) The Delta Lake Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.delta.kernel.internal.commit;
+
+import io.delta.kernel.commit.CommitFailedException;
+import io.delta.kernel.commit.CommitMetadata;
+import io.delta.kernel.commit.CommitResponse;
+import io.delta.kernel.commit.Committer;
+import io.delta.kernel.data.Row;
+import io.delta.kernel.engine.Engine;
+import io.delta.kernel.internal.DeltaErrorsInternal;
+import io.delta.kernel.internal.actions.Protocol;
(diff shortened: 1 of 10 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/DefaultFileSystemManagedTableOnlyCommitter.java
```diff
@@ -0,0 +1,50 @@
...
+import io.delta.kernel.commit.CommitResponse;
+import io.delta.kernel.commit.Committer;
+import io.delta.kernel.data.Row;
+import io.delta.kernel.engine.Engine;
+import io.delta.kernel.internal.DeltaErrorsInternal;
+import io.delta.kernel.internal.actions.Protocol;
+import io.delta.kernel.internal.tablefeatures.TableFeatures;
+import io.delta.kernel.utils.CloseableIterator;
+
+public class DefaultFileSystemManagedTableOnlyCommitter implements Committer {
```
scottsand-db, line 30: note: This will get more complex later on as we implement it properly

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/DefaultFileSystemManagedTableOnlyCommitter.java
```diff
@@ -0,0 +1,50 @@
...
+  @Override
+  public CommitResponse commit(
+      Engine engine, CloseableIterator<Row> finalizedActions, CommitMetadata commitMetadata)
+      throws CommitFailedException {
+    commitMetadata.getReadProtocolOpt().ifPresent(this::validateProtocol);
+    commitMetadata.getNewProtocolOpt().ifPresent(this::validateProtocol);
+    throw new UnsupportedOperationException("Default Committer not yet implemented");
+  }
+
+  private void validateProtocol(Protocol protocol) {
```
huan233usc, line 45: Is it intentionally included in this PR? it looks a bit wired that we do some check then throw unsupported error. (but ok with that given DefaultFileSystemManagedTableOnlyCommitter will be implemented shortly)

scottsand-db: Good question! Yup, intentionally included.

I think it's important for us to all agree from the get go that the default committer should not be able to write into catalog managed tables.

emkornfield (approved):

allisonport-db (approved):
LGTM

PR #4937, opened 2025-07-16
[Spark] Fix null pointer error of `partitionOverwriteMode` in `isDynamicPartitionOverwriteMode`

zhipengmao-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
It checks whether `mode`(`partitionOverwriteMode` option) is null before invoking `mode.equalsIgnoreCase` in `isDynamicPartitionOverwriteMode`. If `mode` is null, an invalid option exception will be thrown.

## How was this patch tested?
UT.

Yes. If `partitionOverwriteMode` is null, we used to throw


andreaschat-db (approved):
LGTM.

PR #4940, opened 2025-07-17
[Kernel] Update the GenerateIcebergCompatActionUtils methods to accept a pre-parsed physicalSchema param

raveeram-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

Today, all the action generators in this class parse the physicalSchema from its JSON representation in the TransactionState on every call. 
This PR updates the APIs to allow a pre-parsed physicalSchema to be plumbed through. This will be a helpful optimization for better batch performance when callers are creating batches of actions with the same physicalSchema.

## How was this patch tested?

Updated existing unit tests to use a mix of pre-parsed and Optional.empty() to cover both options. 

No

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/GenerateIcebergCompatActionUtils.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/GenerateIcebergCompatActionUtils.java
@@ -57,6 +57,9 @@ public final class GenerateIcebergCompatActionUtils {
    * @param partitionValues the partition values for the add
    * @param dataChange whether or not the add constitutes a dataChange (i.e. append vs. compaction)
    * @param tags key-value metadata to be attached to the add action
+   * @param physicalSchemaOpt An optional pre-parsed physical schema. Improves performance for batch
+   *     operations by avoiding repeated JSON parsing. Recommended when generating many actions with
+   *     the same schema.
    * @return add action row that can be included in the transaction
    * @throws UnsupportedOperationException if icebergWriterCompatV1 is not enabled
    * @throws UnsupportedOperationException if maxRetries != 0 in the transaction
@@ -129,7 +132,9 @@ public final class GenerateIcebergCompatActionUtils {
    * @param dataChange whether or not the add constitutes a dataChange (i.e. append vs. compaction)
    * @param tags key-value metadata to be attached to the add action
    * @param deletionVectorDescriptor optional deletion vector descriptor for the add action
-   * @param physicalSchemaOpt An optional pre-parsed physical schema of the table.
+   * @param physicalSchemaOpt An optional pre-parsed physical schema. Improves performance for batch
+   *     operations by avoiding repeated JSON parsing. Recommended when generating many actions with
+   *     the same schema.
    * @return add action row that can be included in the transaction
    * @throws UnsupportedOperationException if icebergWriterCompatV3 is not enabled
    * @throws UnsupportedOperationException if maxRetries != 0 in the transaction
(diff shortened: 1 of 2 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/GenerateIcebergCompatActionUtils.java
```diff
@@ -68,7 +68,8 @@ public static Row generateIcebergCompatWriterV1AddAction(
       DataFileStatus fileStatus,
       Map<String, Literal> partitionValues,
       boolean dataChange,
-      Map<String, String> tags) {
+      Map<String, String> tags,
+      Optional<StructType> physicalSchemaOpt) {
```
OussamaSaoudi, line 75: add @param

OussamaSaoudi (approved):
LGTM.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/GenerateIcebergCompatActionUtils.java
emkornfield: please also make sure to document performance considerations for this.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/GenerateIcebergCompatActionUtils.java
```diff
@@ -182,6 +193,7 @@ public static Row generateIcebergCompatWriterV3AddAction(
    * @param partitionValues the partition values for the remove
    * @param dataChange whether or not the remove constitutes a dataChange (i.e. delete vs.
    *     compaction)
+   * @param physicalSchemaOpt An optional pre-parsed physical schema of the table.
```
emkornfield: nit: add performance implications for tis.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/actions/GenerateIcebergCompatActionUtilsSuite.scala
```diff
@@ -71,15 +71,17 @@ class GenerateIcebergCompatActionUtilsSuite extends AnyFunSuite {
           txnStateRow,
           dataFileStatus,
           partitionValues,
-          dataChange)
+          dataChange,
+          Optional.empty())
```
emkornfield: nit: parameter comment.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/actions/GenerateIcebergCompatActionUtilsSuite.scala
```diff
@@ -71,15 +71,17 @@ class GenerateIcebergCompatActionUtilsSuite extends AnyFunSuite {
...
       }.getMessage.contains(expectedErrorMessageContains))
     assert(
       intercept[UnsupportedOperationException] {
         GenerateIcebergCompatActionUtils.generateIcebergCompatWriterV1RemoveAction(
           txnStateRow,
           dataFileStatus,
           partitionValues,
-          dataChange)
+          dataChange,
+          Optional.empty())
```
emkornfield: nit: parameter comment.

raveeram-db:
> LGTM, it seems like there is still a potential for misuse of the API but this at least makes it possible to avoid the performance penalty.

Agreed, it's a stop-gap solution and long-term we'd want to think about a safer API if possible.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/actions/GenerateIcebergCompatActionUtilsSuite.scala
raveeram-db: added

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/actions/GenerateIcebergCompatActionUtilsSuite.scala
raveeram-db: thanks, added.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/GenerateIcebergCompatActionUtils.java
raveeram-db: Good point, updated with performance considerations, thank you both.

PR #4924, opened 2025-07-14
[Connectors] Add golden table for row tracking tests

lbhm:
#### Which Delta project/connector is this regarding?

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [x] Other (Connetors/Golden tables)

## Description

This PR adds a golden table with row tracking enabled as a resource for testing the row tracking reader in Delta kernel (to be implemented soon).

To verify whether Delta kernel can correctly read a table with row tracking that was created by Delta Spark, I create a small dummy table and then run a `MERGE` op that updates a value and inserts a new value so that we have all combinations of materialized/not-materialized row IDs and row commit versions.

## How was this patch tested?

This PR adds new resources/reference solutions for future tests.

## Does this PR introduce _any_ user-facing changes?

No

PR #4943, opened 2025-07-17
[Kernel] Support reading materialized row tracking metadata columns

lbhm:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This PR adds initial support for reading row tracking metadata columns to kernel. Right now, this feature is limited to reading the materialized row ID and row commit version (i.e., values will be null if they were not materialized). Adding support for computing missing values on the fly will be added in a later PR.

The PR partially builds upon #4897 and #4924 and should be merged afterwards.

## How was this patch tested?

The PR adds several tests to the `RowTrackingSuite`, both based on a newly generated Delta table and a golden table (see #4924) that was generated with delta-spark.

This PR introduces (incomplete) new features to kernel. It does not modify any existing features.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
@@ -19,6 +19,7 @@ package io.delta.kernel;
 import io.delta.kernel.annotation.Evolving;
 import io.delta.kernel.data.*;
 import io.delta.kernel.engine.Engine;
+import io.delta.kernel.exceptions.InvalidTableException;
 import io.delta.kernel.expressions.Predicate;
 import io.delta.kernel.internal.InternalScanFileUtils;
 import io.delta.kernel.internal.actions.DeletionVectorDescriptor;
@@ -173,6 +174,16 @@ public interface Scan {
         initIfRequired();
         ColumnarBatch nextDataBatch = physicalDataIter.next();
 
+        // Add partition columns
+        // NOTE: Adding partition columns must be the first thing we do with the data batch because
+        // it is the last thing we remove from the physicalReadSchema.
+        nextDataBatch =
+            PartitionUtils.withPartitionColumns(
+                engine.getExpressionHandler(),
+                nextDataBatch,
+                InternalScanFileUtils.getPartitionValues(scanFile),
+                physicalReadSchema);
+
         DeletionVectorDescriptor dv =
             InternalScanFileUtils.getDeletionVectorDescriptorFromRow(scanFile);
 
@@ -185,8 +196,9 @@ public interface Scan {
           selectionVector = Optional.empty();
         } else {
           if (rowIndexOrdinal == -1) {
-            throw new IllegalArgumentException(
-                "Row index column is not " + "present in the data read from the Parquet file.");
+            throw new InvalidTableException(
+                tablePath,
+                "Row index column is not present in the data read from the Parquet file.");
           }
           if (!dv.equals(currDV)) {
             Tuple2<DeletionVectorDescriptor, RoaringBitmapArray> dvInfo =
@@ -198,16 +210,9 @@ public interface Scan {
           selectionVector = Optional.of(new SelectionColumnVector(currBitmap, rowIndexVector));
         }
         if (rowIndexOrdinal != -1) {
+          // TODO: Only remove rowIndex if it was not explicitly requested by the user
           nextDataBatch = nextDataBatch.withDeletedColumnAt(rowIndexOrdinal);
         }
-
-        // Add partition columns
-        nextDataBatch =
-            PartitionUtils.withPartitionColumns(
-                engine.getExpressionHandler(),
-                nextDataBatch,
-                InternalScanFileUtils.getPartitionValues(scanFile),
-                physicalReadSchema);
 
         // Change back to logical schema
         ColumnMappingMode columnMappingMode = ScanStateRow.getColumnMappingMode(scanState);
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
@@ -173,9 +173,9 @@ public interface Scan {
         initIfRequired();
         ColumnarBatch nextDataBatch = physicalDataIter.next();
 
-        // Add partition columns
-        // NOTE: Adding partition columns must be the first thing we do with the data batch because
-        // it is the last thing we remove from the physicalReadSchema when creating a ScanStateRow.
+        // Partition columns are the last thing we remove from the physicalReadSchema when creating
+        // a ScanStateRow. Therefore, we add them back first so that the data batch's schema matches
+        // the physicalReadSchema.
         nextDataBatch =
             PartitionUtils.withPartitionColumns(
                 engine.getExpressionHandler(),
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
@@ -176,7 +176,7 @@ public interface Scan {
 
         // Add partition columns
         // NOTE: Adding partition columns must be the first thing we do with the data batch because
-        // it is the last thing we remove from the physicalReadSchema.
+        // it is the last thing we remove from the physicalReadSchema when creating a ScanStateRow.
         nextDataBatch =
             PartitionUtils.withPartitionColumns(
                 engine.getExpressionHandler(),
@@ -196,8 +196,7 @@ public interface Scan {
           selectionVector = Optional.empty();
         } else {
           if (rowIndexOrdinal == -1) {
-            throw new InvalidTableException(
-                tablePath,
+            throw new IllegalArgumentException(
                 "Row index column is not present in the data read from the Parquet file.");
           }
           if (!dv.equals(currDV)) {
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
@@ -19,7 +19,6 @@ package io.delta.kernel;
 import io.delta.kernel.annotation.Evolving;
 import io.delta.kernel.data.*;
 import io.delta.kernel.engine.Engine;
-import io.delta.kernel.exceptions.InvalidTableException;
 import io.delta.kernel.expressions.Predicate;
 import io.delta.kernel.internal.InternalScanFileUtils;
 import io.delta.kernel.internal.actions.DeletionVectorDescriptor;
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaErrors.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaErrors.java
@@ -28,6 +28,7 @@ import io.delta.kernel.utils.DataFileStatus;
 import java.io.IOException;
 import java.sql.Timestamp;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Supplier;
@@ -442,6 +443,19 @@ public final class DeltaErrors {
 
   public static KernelException cannotToggleRowTrackingOnExistingTable() {
     return new KernelException("Row tracking support cannot be changed once the table is created.");
+  }
+
+  public static KernelException missingRowTrackingColumnRequested(String columnName) {
+    return new KernelException(
+        String.format(
+            "Row tracking is not enabled, but row tracking column '%s' was requested.",
+            columnName));
+  }
+
+  public static KernelException missingMetadataConfigField(String key, Map<String, String> config) {
+    return new KernelException(
+        String.format(
+            "Missing required field '%s' in the metadata configuration: %s", key, config));
   }
 
   public static KernelException cannotModifyAppendOnlyTable(String tablePath) {
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaErrors.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaErrors.java
@@ -28,7 +28,6 @@ import io.delta.kernel.utils.DataFileStatus;
 import java.io.IOException;
 import java.sql.Timestamp;
 import java.util.List;
-import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Supplier;
@@ -450,12 +449,6 @@ public final class DeltaErrors {
         String.format(
             "Row tracking is not enabled, but row tracking column '%s' was requested.",
             columnName));
-  }
-
-  public static KernelException missingMetadataConfigField(String key, Map<String, String> config) {
-    return new KernelException(
-        String.format(
-            "Missing required field '%s' in the metadata configuration: %s", key, config));
   }
 
   public static KernelException cannotModifyAppendOnlyTable(String tablePath) {
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
@@ -35,6 +35,7 @@ import io.delta.kernel.internal.metrics.ScanReportImpl;
 import io.delta.kernel.internal.metrics.Timer;
 import io.delta.kernel.internal.replay.LogReplay;
 import io.delta.kernel.internal.replay.PaginationContext;
+import io.delta.kernel.internal.rowtracking.MaterializedRowTrackingColumn;
 import io.delta.kernel.internal.skipping.DataSkippingPredicate;
 import io.delta.kernel.internal.skipping.DataSkippingUtils;
 import io.delta.kernel.internal.util.*;
@@ -200,10 +201,23 @@ public class ScanImpl implements Scan {
 
   @Override
   public Row getScanState(Engine engine) {
+    // Check for metadata columns in the read schema and convert them to physical columns
+    List<StructField> convertedFields = new ArrayList<>();
+    for (StructField field : readSchema.fields()) {
+      Optional<StructField> converted =
+          MaterializedRowTrackingColumn.convertToPhysicalColumn(field, metadata);
+      if (converted.isPresent()) {
+        convertedFields.add(converted.get());
+      } else {
+        convertedFields.add(field);
+      }
+    }
+    StructType convertedSchema = new StructType(convertedFields);
+
     // Physical equivalent of the logical read schema.
     StructType physicalReadSchema =
         ColumnMapping.convertToPhysicalSchema(
-            readSchema,
+            convertedSchema,
             snapshotSchema,
             ColumnMapping.getColumnMappingMode(metadata.getConfiguration()));
 
@@ -213,7 +227,7 @@ public class ScanImpl implements Scan {
     List<String> partitionColumns = VectorUtils.toJavaList(metadata.getPartitionColumns());
     StructType physicalDataReadSchema =
         PartitionUtils.physicalSchemaWithoutPartitionColumns(
-            readSchema, /* logical read schema */
+            convertedSchema, /* logical read schema */
             physicalReadSchema,
             new HashSet<>(partitionColumns));
 
@@ -224,7 +238,7 @@ public class ScanImpl implements Scan {
     return ScanStateRow.of(
         metadata,
         protocol,
-        readSchema.toJson(),
+        convertedSchema.toJson(),
         physicalReadSchema.toJson(),
         physicalDataReadSchema.toJson(),
         dataPath.toUri().toString());
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
@@ -201,20 +201,11 @@ public class ScanImpl implements Scan {
 
   @Override
   public Row getScanState(Engine engine) {
-    // Check for metadata columns in the read schema and convert them to physical columns
-    List<StructField> convertedFields = new ArrayList<>();
-    for (StructField field : readSchema.fields()) {
-      Optional<StructField> converted =
-          MaterializedRowTrackingColumn.convertToPhysicalColumn(field, metadata);
-      if (converted.isPresent()) {
-        convertedFields.add(converted.get());
-      } else {
-        convertedFields.add(field);
-      }
-    }
-    StructType convertedSchema = new StructType(convertedFields);
+    // Check for row tracking columns in the read schema and convert them to physical columns
+    StructType convertedSchema =
+        MaterializedRowTrackingColumn.convertToPhysicalSchema(readSchema, metadata);
 
-    // Physical equivalent of the logical read schema.
+    // Convert regular columns to physical columns if column mapping is enabled
     StructType physicalReadSchema =
         ColumnMapping.convertToPhysicalSchema(
             convertedSchema,
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
@@ -36,6 +36,7 @@ import io.delta.kernel.internal.metrics.Timer;
 import io.delta.kernel.internal.replay.LogReplay;
 import io.delta.kernel.internal.replay.PaginationContext;
 import io.delta.kernel.internal.rowtracking.MaterializedRowTrackingColumn;
+import io.delta.kernel.internal.rowtracking.RowTracking;
 import io.delta.kernel.internal.skipping.DataSkippingPredicate;
 import io.delta.kernel.internal.skipping.DataSkippingUtils;
 import io.delta.kernel.internal.util.*;
@@ -201,36 +202,22 @@ public class ScanImpl implements Scan {
 
   @Override
   public Row getScanState(Engine engine) {
-    // Check for row tracking columns in the read schema and convert them to physical columns
-    StructType convertedSchema =
-        MaterializedRowTrackingColumn.convertToPhysicalSchema(readSchema, metadata);
-
-    // Convert regular columns to physical columns if column mapping is enabled
-    StructType physicalReadSchema =
-        ColumnMapping.convertToPhysicalSchema(
-            convertedSchema,
-            snapshotSchema,
-            ColumnMapping.getColumnMappingMode(metadata.getConfiguration()));
+    StructType physicalSchema = convertToPhysicalSchema();
 
     // Compute the physical data read schema, basically the list of columns to read
     // from a Parquet data file. It should exclude partition columns and include
     // row_index metadata columns (in case DVs are present)
+    // NOTE: Removing partition columns is the last thing we do with the read schema
     List<String> partitionColumns = VectorUtils.toJavaList(metadata.getPartitionColumns());
     StructType physicalDataReadSchema =
         PartitionUtils.physicalSchemaWithoutPartitionColumns(
-            readSchema, /* logical read schema */
-            physicalReadSchema,
-            new HashSet<>(partitionColumns));
-
-    if (protocol.getReaderFeatures().contains("deletionVectors")) {
-      physicalDataReadSchema = physicalDataReadSchema.add(StructField.METADATA_ROW_INDEX_COLUMN);
-    }
+            readSchema /* logical read schema */, physicalSchema, new HashSet<>(partitionColumns));
 
     return ScanStateRow.of(
         metadata,
         protocol,
         readSchema.toJson(),
-        physicalReadSchema.toJson(),
+        physicalSchema.toJson(),
         physicalDataReadSchema.toJson(),
         dataPath.toUri().toString());
   }
@@ -238,6 +225,39 @@ public class ScanImpl implements Scan {
   @Override
   public Optional<Predicate> getRemainingFilter() {
     return getDataFilters();
+  }
+
+  private StructType convertToPhysicalSchema() {
+    StructType physicalSchema = new StructType();
+
+    ColumnMapping.ColumnMappingMode mode =
+        ColumnMapping.getColumnMappingMode(metadata.getConfiguration());
+    for (StructField logicalField : readSchema.fields()) {
+      if (logicalField.isMetadataColumn()) {
+        if (RowTracking.isRowTrackingColumn(logicalField)) {
+          physicalSchema =
+              MaterializedRowTrackingColumn.convertToPhysicalColumn(
+                  logicalField, readSchema, physicalSchema, metadata);
+        } else {
+          // As of now, metadata columns other than row tracking columns do not require
+          // any special handling, so we can just add them to the physical schema as is.
+          physicalSchema = physicalSchema.add(logicalField);
+        }
+      } else {
+        physicalSchema =
+            physicalSchema.add(
+                ColumnMapping.convertToPhysicalColumn(logicalField, snapshotSchema, mode));
+      }
+    }
+
+    if (protocol.getReaderFeatures().contains("deletionVectors")) {
+      if (physicalSchema.indexOf(StructField.METADATA_ROW_INDEX_COLUMN_NAME) == -1) {
+        // If the row index column is not already present, add it to the physical read schema
+        physicalSchema = physicalSchema.add(StructField.METADATA_ROW_INDEX_COLUMN);
+      }
+    }
+
+    return physicalSchema;
   }
 
   private Optional<Tuple2<Predicate, Predicate>> splitFilters(Optional<Predicate> filter) {
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
@@ -218,7 +218,7 @@ public class ScanImpl implements Scan {
     List<String> partitionColumns = VectorUtils.toJavaList(metadata.getPartitionColumns());
     StructType physicalDataReadSchema =
         PartitionUtils.physicalSchemaWithoutPartitionColumns(
-            convertedSchema, /* logical read schema */
+            readSchema, /* logical read schema */
             physicalReadSchema,
             new HashSet<>(partitionColumns));
 
@@ -229,7 +229,7 @@ public class ScanImpl implements Scan {
     return ScanStateRow.of(
         metadata,
         protocol,
-        convertedSchema.toJson(),
+        readSchema.toJson(),
         physicalReadSchema.toJson(),
         physicalDataReadSchema.toJson(),
         dataPath.toUri().toString());
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
@@ -202,7 +202,7 @@ public class ScanImpl implements Scan {
 
   @Override
   public Row getScanState(Engine engine) {
-    StructType physicalSchema = getPhysicalSchema();
+    StructType physicalSchema = createPhysicalSchema();
 
     // Compute the physical data read schema (i.e., the columns to read from a Parquet data file).
     // The only difference to the physical schema is that we exclude partition columns. ALl other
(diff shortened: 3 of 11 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/types/StructField.java
```diff
@@ -34,14 +34,26 @@ public class StructField {
...
 
   /**
    * The name of a row index metadata column. When present this column must be populated with row
    * index of each row when reading from parquet.
    */
   public static String METADATA_ROW_INDEX_COLUMN_NAME = "_metadata.row_index";
 
+  /**
+   * The name of a row ID metadata column. When present this column must be populated with the
+   * unique row ID of each column.
```
johanl-db: nit:

```suggestion
   * The name of the row ID metadata column. When present this column must be populated with the
   * unique ID of each row.
```

kernel/kernel-api/src/main/java/io/delta/kernel/types/StructField.java
```diff
@@ -50,6 +62,22 @@ public class StructField {
...
+          FieldMetadata.builder().putBoolean(IS_METADATA_COLUMN_KEY, true).build(),
+          Collections.emptyList());
+
+  public static StructField METADATA_ROW_COMMIT_VERSION_COLUMN =
+      new StructField(
+          METADATA_ROW_COMMIT_VERSION_COLUMN_NAME,
+          LongType.LONG,
+          false,
+          FieldMetadata.builder().putBoolean(IS_METADATA_COLUMN_KEY, true).build(),
+          Collections.emptyList());
```
johanl-db: I would define these fields in RowTracking.scala instead of here, along with the field names `METADATA_ROW_ID_COLUMN_NAME` and `METADATA_ROW_COMMIT_VERSION_COLUMN_NAME`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -145,6 +148,39 @@ public static Optional<Metadata> assignMaterializedColumnNamesIfNeeded(Metadata
...
+   * @param metadata The current metadata of the table.
+   * @return An Optional containing the physical column if conversion is applicable;
+   *     Optional.empty() otherwise.
+   */
+  public static Optional<StructField> convertToPhysicalColumn(
+      StructField column, Metadata metadata) {
+    if (column.equals(METADATA_ROW_ID_COLUMN)) {
+      return Optional.of(
+          new StructField(
+              metadata.getConfiguration().get(ROW_ID.getMaterializedColumnNameProperty()),
```
johanl-db: Throw a proper exception if the materialized column name isn't properly set. Define a helper for that.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -145,6 +148,39 @@ public static Optional<Metadata> assignMaterializedColumnNamesIfNeeded(Metadata
...
+              column.getDataType(),
+              column.isNullable()));
```
johanl-db: The materialized fields should always be nullable. E.p. the current metadata struct fields are defined as non-nullable (which isn't actually correct yet but will be once we populate default values), but we do expect nulls from the materialized column.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -145,6 +148,39 @@ public static Optional<Metadata> assignMaterializedColumnNamesIfNeeded(Metadata
...
+   * parquet reader is concerned, these columns are not metadata columns.
+   *
+   * @param column The logical column to convert.
+   * @param metadata The current metadata of the table.
+   * @return An Optional containing the physical column if conversion is applicable;
+   *     Optional.empty() otherwise.
+   */
+  public static Optional<StructField> convertToPhysicalColumn(
+      StructField column, Metadata metadata) {
+    if (column.equals(METADATA_ROW_ID_COLUMN)) {
```
johanl-db: We'll need to fail if a Row Tracking metadata column is requested but row tracking isn't enabled on the table. You'll need to check that the rowTracking table feature is supported in the table `Protocol`. Also add a test for this case

johanl-db: Match metadata column based on their name instead of the whole struct. In the current model, the name is the unique identifier. We don't wnat differences in e.g. nullability to cause metadata columns to be ignored.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -200,10 +201,23 @@ protected CloseableIterator<FilteredColumnarBatch> getScanFiles(
...
+    for (StructField field : readSchema.fields()) {
+      Optional<StructField> converted =
+          MaterializedRowTrackingColumn.convertToPhysicalColumn(field, metadata);
+      if (converted.isPresent()) {
+        convertedFields.add(converted.get());
+      } else {
+        convertedFields.add(field);
+      }
+    }
+    StructType convertedSchema = new StructType(convertedFields);
```
johanl-db: Move all of this logic inside `convertToPhysicalColumn`, that then directly returns a `StructType` with row tracking columns replaced.

We can pick between two options, either:
1. We consider this logic to be specific to row tracking metadata columns, and everything should be handled in convertToPhysicalColumn.
2. We consider this logic to be more generic for metadata columns in general, in that case `convertToPhysicalColumn` should be defined here in ScanImpl.java and be written more generically, e.g. by using a `switch` over supported metadata columns and failing if we don't recognize one.

I think 1. is fine for now, we don't really have evidence at this stage that we'll add metadata columns that map to a physical column beyond row tracking

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -200,10 +201,23 @@ protected CloseableIterator<FilteredColumnarBatch> getScanFiles(
...
+
```
johanl-db: Move this whole logic into `convertToPhysicalColumn`

There are two approaches available, either:
1. we consider this logic of mapping to physical columns to be specific to row tracking, in which case it should be fully contained in `convertToPhysicalColumn`
2. We consider it to be generic and applying to possibly other metadata columns and convertToPhysicalColumn should be moved to ScanImpl.java and be implemented in a more generic manner, e.g. using a switch over known metadata column.

I think 1. is fine, we don't expect at this point to be adding new metadata column that map to physical columns

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -213,7 +218,7 @@ public Row getScanState(Engine engine) {
     List<String> partitionColumns = VectorUtils.toJavaList(metadata.getPartitionColumns());
     StructType physicalDataReadSchema =
         PartitionUtils.physicalSchemaWithoutPartitionColumns(
-            readSchema, /* logical read schema */
+            convertedSchema, /* logical read schema */
```
johanl-db: This should still be `readSchema` afaict. This doesn't seem to matter much since row tracking columns can't be partition columns, but since `convertedSchema` is produced by `convertToPhysicalSchema`, it's not really a logical schema anymore. It' actually in a weird state, with a mix of physical and logical columns, meaning we probably don't want it to be used anywhere, esp. not escaping on line 232

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -145,6 +147,61 @@ public static Optional<Metadata> assignMaterializedColumnNamesIfNeeded(Metadata
...
+  public static StructType convertToPhysicalSchema(StructType logicalSchema, Metadata metadata) {
+    List<StructField> physicalFields = new ArrayList<>();
+    // Check if row tracking is enabled and error if row tracking columns are requested
+    boolean rowTrackingEnabled = TableConfig.ROW_TRACKING_ENABLED.fromMetadata(metadata);
+
+    for (StructField field : logicalSchema.fields()) {
+      if (!rowTrackingEnabled
+          && (field.getName().equals(METADATA_ROW_ID_COLUMN_NAME)
+              || field.getName().equals(METADATA_ROW_COMMIT_VERSION_COLUMN_NAME))) {
+        throw new InvalidTableException(
```
johanl-db: Define this exception in `DeltaErrors`. All user-facing errors should be defined there, and a query engine can request row tracking columns even though row tracking is disabled so this failure is user-facing

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -145,6 +147,61 @@ public static Optional<Metadata> assignMaterializedColumnNamesIfNeeded(Metadata
...
+            metadata.getId(),
+            String.format(
+                "Row tracking is not enabled, but row tracking column '%s' was requested.",
+                field.getName()));
+      }
+
+      if (field.getName().equals(METADATA_ROW_ID_COLUMN_NAME)) {
+        physicalFields.add(
+            new StructField(getPhysicalColumnName(ROW_ID, metadata), field.getDataType(), true));
```
johanl-db: nit: named arguments
```suggestion
            new StructField(getPhysicalColumnName(ROW_ID, metadata), field.getDataType(), true  /* nullable */));
```

below also

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/RowTrackingSuite.scala
```diff
@@ -376,6 +424,118 @@ class RowTrackingSuite extends DeltaTableWriteSuiteBase with ParquetSuiteBase {
...
+          readCols = Seq("id", "_metadata.row_id"),
+          metadataCols =
+            Seq(RowTracking.METADATA_ROW_ID_COLUMN, RowTracking.METADATA_ROW_COMMIT_VERSION_COLUMN),
+          engine = engine)
+      }
+      assert(e.getMessage.contains("Row tracking is not enabled, but row tracking column"))
+    }
+  }
+
+  test("Read row tracking columns from golden table") {
```
johanl-db: Update name, below also

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
lbhm: I agree with you on line 232, but for this line I am not so sure. Based on the variable name, I would expect `physicalDataReadSchema` to contain the entire physical schema that I (as a parquet reader) need to read, so I would expect that I contains the physical (not the logical) metadata columns.

lbhm: Nevermind - I looked more closely into `PartitionUtils` and agree with you. Will fix this. I also noticed an error in the documentation of the public method in `PartitionUtils` that I will fix along the way.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaErrors.java
```diff
@@ -444,6 +445,19 @@ public static KernelException cannotToggleRowTrackingOnExistingTable() {
...
+        String.format(
+            "Row tracking is not enabled, but row tracking column '%s' was requested.",
+            columnName));
+  }
+
+  public static KernelException missingMetadataConfigField(String key, Map<String, String> config) {
+    return new KernelException(
+        String.format(
+            "Missing required field '%s' in the metadata configuration: %s", key, config));
+  }
```
allisonport-db: This shouldn't be possible to hit unless a table has `rowTracking` enabled and is missing this field (corrupt metadata) right? If so -- this can just be an `InvalidTableException` and you can put it either inline in the code or add it to `DeltaErrorsInternal`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -200,10 +201,14 @@ protected CloseableIterator<FilteredColumnarBatch> getScanFiles(
 
   @Override
   public Row getScanState(Engine engine) {
-    // Physical equivalent of the logical read schema.
+    // Check for row tracking columns in the read schema and convert them to physical columns
+    StructType convertedSchema =
+        MaterializedRowTrackingColumn.convertToPhysicalSchema(readSchema, metadata);
+
+    // Convert regular columns to physical columns if column mapping is enabled
```
allisonport-db: Can you make sure to test with column mapping all modes enabled? I think you will encounter an error in this conversion since you don't add a column id in `MaterializedRowTrackingColumn.convertToPhysicalSchema(readSchema, metadata)`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/RowTracking.java
```diff
@@ -33,6 +36,32 @@
 
 /** A collection of helper methods for working with row tracking. */
 public class RowTracking {
+  /**
```
allisonport-db, line 40: Can you add a TODO here that we will be changing this API in the future please? just for future clarity in case anyone looks at this

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaErrors.java
lbhm: You're right. I've inlined this into an `InvalidTableException`.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/ColumnMapping.java
```diff
@@ -251,6 +251,15 @@ private static StructType convertToPhysicalSchema(
       StructType prunedSchema, StructType fullSchema, boolean includeFieldId) {
     StructType newSchema = new StructType();
     for (StructField prunedField : prunedSchema.fields()) {
+      if (fullSchema.indexOf(prunedField.getName()) == -1) {
```
huan233usc: Do we want to hard check the column is a valid metadata column checkState()? (both name and type matches the available metadata column)

lbhm: This is a bit tricky to do because with the current architecture, the row tracking columns have already been translated to their physical counterparts (which, at the physical level, are regular columns, so they are not tagged as metadata columns). `ColumnMapping.java` should not be concerned about the physical names of row tracking columns so I would prefer to handle this elsewhere.

I started drafting a refactor of `getScanState` that handles the translation of regular and metadata columns to physical columns in a more principled way. My plan is to discuss the idea with @johanl-db tomorrow and then bring it up for discussion here.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
lbhm: You are right! I added tests for column mapping modes and pushed sort of a hot fix that ensures that the tests pass. The fix is not ideal IMO, so I've been discussing with @johanl-db how to address this properly. I'll ping you once I have a draft ready.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/ColumnMapping.java
huan233usc: For context, I was worrying about a case that prunedSchema contain some random fields which are not parts of table schema.

I see there is one TODO for adding check at the ScanBuilder. Maybe you could update the TODO
https://github.com/delta-io/delta/blob/74988e2ff705f3483a066675ed1592b8340aa966/kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanBuilderImpl.java#L75

// TODO: validate the readSchema is a subset of the table schema ->
// TODO: validate the readSchema is a subset of the table schema or the extra fields are metadata column.

lbhm: You're right. I initially chose this kind of "hot fix" because the current main branch would also throw a NullPointerException at this point for unknown columns.

I'll look at your pointer tomorrow. We might want to address it once we agreed on one of the proposals for a metadata column API so that we can address regular and metadata columns holistically.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -200,32 +202,22 @@ protected CloseableIterator<FilteredColumnarBatch> getScanFiles(
 
   @Override
   public Row getScanState(Engine engine) {
-    // Physical equivalent of the logical read schema.
-    StructType physicalReadSchema =
-        ColumnMapping.convertToPhysicalSchema(
-            readSchema,
-            snapshotSchema,
-            ColumnMapping.getColumnMappingMode(metadata.getConfiguration()));
+    StructType physicalSchema = convertToPhysicalSchema();
```
johanl-db: Either name this `getPhysicalSchema()` or pass a schema to convert: `convertToPhysicalSchema(readSchema)`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -200,32 +202,22 @@ protected CloseableIterator<FilteredColumnarBatch> getScanFiles(
...
 
     // Compute the physical data read schema, basically the list of columns to read
     // from a Parquet data file. It should exclude partition columns and include
     // row_index metadata columns (in case DVs are present)
+    // NOTE: Removing partition columns is the last thing we do with the read schema
```
johanl-db: Update this comment, we don't remove the row_index field here anymore, it was handled before. This just removes the partition columns to get the actual schema to read from the Parquet data files

johanl-db: Also for the NOTE here and in Scan.java:
a better explanation may be that `physicalDataReadSchema` is the physical schema without partition columns. That motivates the order of operation, which otherwise seems a bit arbitrary.

Calling that out would also remove some of the confusion around the difference between `physicalReadSchema` and `physicalDataReadSchema`, now that it's only about partition columns and not also row index

kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
```diff
@@ -185,8 +196,9 @@ public FilteredColumnarBatch next() {
           selectionVector = Optional.empty();
         } else {
           if (rowIndexOrdinal == -1) {
-            throw new IllegalArgumentException(
-                "Row index column is not " + "present in the data read from the Parquet file.");
+            throw new InvalidTableException(
```
johanl-db: Keep the old exception, `InvalidTableException` doesn't seem very fitting since it's a problem with the parquet reader and/or the query engine/kernel, not an issue with the table itself

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -145,6 +147,68 @@ public static Optional<Metadata> assignMaterializedColumnNamesIfNeeded(Metadata
...
+   * <p>Note that we must not mark the physical columns as metadata columns because as far as the
+   * parquet reader is concerned, these columns are not metadata columns.
+   *
+   * @param logicalField The logical field to convert.
+   * @param logicalSchema The logical schema containing the field.
+   * @param physicalSchema The current physical schema to which the field will be added.
+   * @param metadata The current metadata of the table.
+   * @return A new StructType representing the updated physical schema with the added field.
+   */
+  public static StructType convertToPhysicalColumn(
```
johanl-db: You could split this in two, for `row_id` and `row_commit_version` respectively.
That way, you don't have to handle caller requesting a non-existing row tracking column

e.g. in ScanImpl::convertToPhysicalSchema:

```
if (RowTracking.isRowIDColumn(logicalField)) {
  physicalSchema.add(ROW_ID.getMaterializedColumn(metadata))
} else if (RowTracking.isRowCommitVersionColumn(logicalField)) {
  physicalSchema.add(ROW_COMMIT_VERSION.getMaterializedColumn(metadata))
}
```
You're exposing a bit the logic of Row tracking in ScanImpl, which you wanted to avoid, but I think it's ok in this case and much cleaner, otherwise you need to add extra logic to ensure that the caller does the right thing

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -145,6 +147,68 @@ public static Optional<Metadata> assignMaterializedColumnNamesIfNeeded(Metadata
...
+    }
+
+    if (logicalField.getName().equals(METADATA_ROW_ID_COLUMN_NAME)) {
+      // TODO: Use logicalSchema here to determine whether we also need to add row index
+      //  to the physical schema to compute non-materialized row IDs.
+      return physicalSchema.add(
+          new StructField(
+              getPhysicalColumnName(ROW_ID, metadata),
+              logicalField.getDataType(),
+              logicalField.isNullable()));
```
johanl-db: The physical field is always nullable.

I'd argue we can also hardcode the data type, as it must be Long. If the engine requesting something else, they visibly manually changed the logical field we define as `METADATA_ROW_ID_COLUMN`, I don't think we just use whatever the query engine decided the type should be

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -145,6 +147,68 @@ public static Optional<Metadata> assignMaterializedColumnNamesIfNeeded(Metadata
...
+    return Optional.ofNullable(
+            metadata.getConfiguration().get(column.getMaterializedColumnNameProperty()))
+        .orElseThrow(
+            () ->
+                new InvalidTableException(
+                    metadata.getId(),
+                    String.format(
+                        "Materialized column name `%s` is missing in the metadata configuration.",
+                        column.getMaterializedColumnNameProperty())));
+  }
```
johanl-db: Seems like this would be better defined in `MaterializedRowTrackingColumn` rather than a static method here

kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
lbhm: Reverted to the old exception

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
lbhm: As we just discussed, I would argue against this approach because the logical-to-physical conversion is different for Row ID and row commit version (Row ID might also have to add row index to the schema). Therefore, we cannot have one consistent implementation for all instances of `MaterializedRowTrackingColumn`.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
lbhm: I pushed my approach and did a first review pass with @johanl-db. The general idea is to have one central iteration over logical columns in `ScanImpl.java` (see `getPhysicalSchema()`), which then delegates the conversion logic to column mapping, row tracking, etc. depending on what type of column the current iteration looks at. For that, I included everything but the removal of partition columns since that step creates a separate variable `physicalDataReadSchema` that we include in the scan state.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/ColumnMapping.java
lbhm: I pushed the more holistic solution and also changed the todo that you mentioned. See [my response to Allison's comment](https://github.com/delta-io/delta/pull/4943#discussion_r2225239341) for a high-level description of the approach.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -235,6 +226,39 @@ public Optional<Predicate> getRemainingFilter() {
...
+          physicalSchema = physicalSchema.add(logicalField);
+        }
+      } else {
+        physicalSchema =
+            physicalSchema.add(
+                ColumnMapping.convertToPhysicalColumn(logicalField, snapshotSchema, mode));
+      }
+    }
+
+    if (protocol.getReaderFeatures().contains("deletionVectors")) {
```
huan233usc: Could we merge these two "if"?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -235,6 +226,39 @@ public Optional<Predicate> getRemainingFilter() {
     return getDataFilters();
   }
 
+  private StructType getPhysicalSchema() {
+    StructType physicalSchema = new StructType();
+
+    ColumnMapping.ColumnMappingMode mode =
+        ColumnMapping.getColumnMappingMode(metadata.getConfiguration());
+    for (StructField logicalField : readSchema.fields()) {
```
huan233usc, line 248: Personally I feel that the nested if-else is hard to read. 

I am thinking about if it worth the extract it to a smaller helper func, for example --

```
  for (StructField logicalField : readSchema.fields()) {
        physicalSchema = addPhysicalField(physicalSchema, logicalField, mode);
    }

private StructType addPhysicalField(
    StructType currentPhysicalSchema,
    StructField logicalField,
    ColumnMapping.ColumnMappingMode mode) {

    // early return for common case
    if (!logicalField.isMetadataColumn()) {
        // This is a regular data column
        return currentPhysicalSchema.add(
            ColumnMapping.convertToPhysicalColumn(logicalField, snapshotSchema, mode));
    }

    if (RowTracking.isRowTrackingColumn(logicalField)) {
        // This is a row tracking metadata column
        return MaterializedRowTrackingColumn.convertToPhysicalColumn(
            logicalField, readSchema, currentPhysicalSchema, metadata);
    }

   // other metadata column
    return currentPhysicalSchema.add(logicalField);
}
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -145,6 +148,64 @@ public static Optional<Metadata> assignMaterializedColumnNamesIfNeeded(Metadata
...
+      return physicalSchema.add(
+          new StructField(
+              ROW_ID.getPhysicalColumnName(metadata), LongType.LONG, true /* nullable */));
+    } else if (logicalField.getName().equals(METADATA_ROW_COMMIT_VERSION_COLUMN_NAME)) {
+      return physicalSchema.add(
+          new StructField(
+              ROW_COMMIT_VERSION.getPhysicalColumnName(metadata),
+              LongType.LONG,
+              true /* nullable */));
+    } else {
```
huan233usc: nit: we don't need this else {}, just 

if () {return abc}
throw new IllegalArgumentException

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/ColumnMapping.java
```diff
@@ -251,34 +278,44 @@ private static StructType convertToPhysicalSchema(
...
+   */
+  private static StructField convertToPhysicalColumn(
+      StructField logicalField, StructType fullSchema, boolean includeFieldId) {
+    StructField completeField = fullSchema.get(logicalField.getName());
+    DataType physicalType =
+        convertToPhysicalType(
+            logicalField.getDataType(), completeField.getDataType(), includeFieldId);
+    String physicalName = completeField.getMetadata().getString(COLUMN_MAPPING_PHYSICAL_NAME_KEY);
+
+    if (includeFieldId) {
```
huan233usc: same here
```
if(!includeFieldId) {
return new StructField(physicalName, physicalType, logicalField.isNullable());
}

remaining logic
```

huan233usc (approved):
Left some recommendation for readability, other than that lgtm.

kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
```diff
@@ -173,6 +173,16 @@ public FilteredColumnarBatch next() {
         initIfRequired();
         ColumnarBatch nextDataBatch = physicalDataIter.next();
 
+        // Add partition columns
+        // NOTE: Adding partition columns must be the first thing we do with the data batch because
+        // it is the last thing we remove from the physicalReadSchema when creating a ScanStateRow.
```
allisonport-db: Sorry I don't exactly follow this? Why is this / what necessitated this change? Was it wrong before?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -145,6 +148,64 @@ public static Optional<Metadata> assignMaterializedColumnNamesIfNeeded(Metadata
...
+   * <p>Note that we must not mark the physical columns as metadata columns because as far as the
+   * parquet reader is concerned, these columns are not metadata columns.
+   *
+   * @param logicalField The logical field to convert.
+   * @param logicalSchema The logical schema containing the field.
+   * @param physicalSchema The current physical schema to which the field will be added.
+   * @param metadata The current metadata of the table.
+   * @return A new StructType representing the updated physical schema with the added field.
+   */
+  public static StructType convertToPhysicalColumn(
```
allisonport-db: Is there a reason this can't return a StructField instead of a StructType? (and add the StructField to the schema outside this fx?) it sits strangely in the other function this way

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -235,6 +226,39 @@ public Optional<Predicate> getRemainingFilter() {
     return getDataFilters();
   }
 
+  private StructType getPhysicalSchema() {
```
allisonport-db: I think it would be useful to have docs here it's not obvious what this function does. I think it does the following essentially right?
1. convert logical fields to physical based on CM mode for non-metadata fields
2. convert metadata fields to physical fields if applicable (i.e. row tracking)
3. adds row_index if needed for DVs

allisonport-db: also nit - I think since this does real generation "get" would be better named "generate" or something

allisonport-db (commented):
Just a few minor comments then LGTM

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -200,32 +202,21 @@ protected CloseableIterator<FilteredColumnarBatch> getScanFiles(
...
-            snapshotSchema,
-            ColumnMapping.getColumnMappingMode(metadata.getConfiguration()));
-
-    // Compute the physical data read schema, basically the list of columns to read
-    // from a Parquet data file. It should exclude partition columns and include
-    // row_index metadata columns (in case DVs are present)
+    StructType physicalSchema = getPhysicalSchema();
+
+    // Compute the physical data read schema (i.e., the columns to read from a Parquet data file).
+    // The only difference to the physical schema is that we exclude partition columns. ALl other
```
qiyuandong-db: nit: ALl -> all

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -145,6 +148,64 @@ public static Optional<Metadata> assignMaterializedColumnNamesIfNeeded(Metadata
...
+              logicalField.getName()));
+    }
+  }
+
+  private String getPhysicalColumnName(Metadata metadata) {
+    return Optional.ofNullable(metadata.getConfiguration().get(getMaterializedColumnNameProperty()))
+        .orElseThrow(
+            () ->
+                new InvalidTableException(
+                    metadata.getId(),
```
qiyuandong-db: Shouldn't the first arg be the table path?
The signature seems to be 
`public InvalidTableException(String tablePath, String reason)`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
lbhm: The reason is that there is a 1:n mapping between logical and physical columns. Concretely, requesting the row ID column requires two physical columns (materialized row ID and row index). Since the row index might have already been requested elsewhere, we don't know in advance if we need to add the row index in this method call. In general, we need to return a `StructType` or some collection of `StructField`s though.

The 1:2 mapping is not part of this PR yet (see #4965), but I already wanted to put the necessary interface in place.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
lbhm: How about `createPhysicalSchema`? "Generate" implicates some sort of random generation to me but that might just be my subjective feeling.

kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
lbhm: I mainly added this comment to clarify the flow of schema modification. With the proposed refactor of `getScanState()`, we handle everything but the removal of partition columns in `createPhysicalSchema()`. That means that `physicalSchema` and `physicalDataReadSchema` are identical except for the missing partition columns.

Therefore, I figured that adding back partition columns first is the safest approach to ensure that we correctly "revert" the logical-to-physical conversion (i.e., create a logical batch of data from the information read from the physical files).

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
qiyuandong-db, line 248: nit: every `physicalSchema.add(...)` clones all fields. For N fields in the readSchema this for loop will be O(N^2). Maybe we can accumulate fields in an `ArrayList`, then create a single new `StructType` in the end.

qiyuandong-db (approved):
LGTM. Left some nit comments.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
lbhm: You're right. I thought that `metadata.getId()` would contain the table ID/path but that's not the case. I've changed the exception to an `IllegalArgumentException` since the table path is not directly available in the function's scope.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
allisonport-db: I prefer the change you did to List<StructField> thank you! Just for my understanding, why is this true? Is it just because we don't know if the rowId has been materialized or not per-file?
> requesting the row ID column requires two physical columns (materialized row ID and row index)

kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
allisonport-db: I see; I think maybe we could make this comment a little more clear? maybe add that adding this back converts our batch to have schema=physicalSchema which may or may not contain metadata columns?

allisonport-db (approved):
LGTM

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
lbhm: Row IDs are generally not materialized unless they are preserved during update/merge operations (kernel does not preserve IDs at the moment but the spec encourages it). To compute non-materialized row IDs, we need the row index.

PR #4953, opened 2025-07-21
[Kernel] Add `ADD` expressions

lbhm:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This PR adds support for `ADD` expressions as a specific kind of `ScalarExpressions`s. `ADD` expressions will be necessary to compute missing metadata column values on the fly. For example, if a row ID is not materialized, it is computed as `baseRowId + rowIndex`.

As part of this change, I changed the `longVector` utility in `VectorTestUtils` so that it accepts Java `long`s instead of Scala `Long`s. This is consistent with the other (existing) utility functions in `VectorTestUtils` and necessary for my changes since I must create vectors that include null values. As a consequence, this PR touches a couple of other test suites - although with trivial changes.

This PR is based on #4943 and should be merged after it.

## How was this patch tested?

I added unit tests in `DefaultExpressionEvaluatorSuite` to test the behavior of `ADD` expressions with the Kernel defaults.

This PR provides a new expression to users and requires query engines to implement it. The functionality of the expression is trivial and should be straightforward to implement for any engine.


kernel/kernel-api/src/test/scala/io/delta/kernel/internal/checkpoints/CheckpointerSuite.scala
```diff
@@ -248,7 +248,7 @@ object CheckpointerSuite extends VectorTestUtils {
     override def getSchema: StructType = CheckpointMetaData.READ_SCHEMA
 
     // empty vector for all columns
-    override def getColumnVector(ordinal: Int): ColumnVector = longVector()
+    override def getColumnVector(ordinal: Int): ColumnVector = longVector(Seq())
```
johanl-db: nit:

`Seq.empty` is more idiomatic

kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/internal/expressions/DefaultExpressionEvaluator.java
```diff
@@ -256,6 +257,21 @@ ExpressionTransformResult visitCoalesce(ScalarExpression coalesce) {
...
+          add.getChildren().stream().map(this::visit).collect(Collectors.toList());
+      if (children.size() != 2) {
+        throw unsupportedExpressionException(
+            add, "ADD requires exactly two arguments: left and right operands");
+      }
+
+      return new ExpressionTransformResult(
+          new ScalarExpression(
+              "ADD", Arrays.asList(children.get(0).expression, children.get(1).expression)),
+          children.get(0).outputType);
```
johanl-db, line 279: Similar to e.g. coalesce above, we should probably start by requiring both arguments to have the exact same type.
Otherwise, we need to define type coercion rules to infer the resulting type of ADD - e.g. int + decimal = ? - and that's a big undertaking

Ideally we'd also check that the type is one of the supported numeric type by the underlying implementation you provide, e.g. String + String shouldn't work

kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/internal/expressions/DefaultExpressionEvaluator.java
```diff
@@ -589,6 +605,56 @@ ColumnVector visitCoalesce(ScalarExpression coalesce) {
...
+          add.getChildren().stream().map(this::visit).collect(toList());
+      if (childResults.size() != 2) {
+        throw unsupportedExpressionException(
+            add, "ADD requires exactly two arguments: left and right operands");
+      }
+
+      return DefaultExpressionUtils.arithmeticVector(
+          childResults.get(0),
+          childResults.get(1),
+          new ArithmeticOperator() {
```
johanl-db, line 625: This implementation is good enough for now, add a comment though to mention that:
- it only supports adding operands of the same type
- it doesn't check for overflow and uses java semantics: integer values will wrap around when overflowing

kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/internal/expressions/DefaultExpressionEvaluator.java
```diff
@@ -589,6 +605,56 @@ ColumnVector visitCoalesce(ScalarExpression coalesce) {
...
+
+            @Override
+            public double apply(double a, double b) {
+              return a + b;
+            }
+
+            @Override
+            public BigDecimal apply(BigDecimal a, BigDecimal b) {
+              return a.add(b);
+            }
```
johanl-db: I would leave decimal out for now, we don't need it and we're setting a precedent for behavior that is most likely not what a sane query engine would expect. E.g. the result may not have the same precision and scale than the two operands, we may decide to increase the precision, or lose some of the scale from one of the operand in some cases.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/internal/expressions/DefaultExpressionEvaluatorSuite.scala
```diff
@@ -284,7 +284,45 @@ class DefaultExpressionEvaluatorSuite extends AnyFunSuite with ExpressionSuiteBa
...
+      new DefaultColumnarBatch(longCol1.getSize, longSchema, Array(longCol1, longCol2, longCol3))
+    val longCoalesceExpr = new ScalarExpression(
+      "COALESCE",
+      util.Arrays.asList(new Column("longCol1"), new Column("longCol2"), new Column("longCol3")))
+    val expLongOutput = longVector(Seq(1L, 2L, 3L, 4L))
+    val actLongOutput = evaluator(longSchema, longCoalesceExpr, LongType.LONG).eval(longBatch)
+    checkLongVectors(actLongOutput, expLongOutput)
+  }
 
+  test("evaluate expression: coalesce (string columns)") {
```
johanl-db, line 307: Can you add tests for other data types?
Ideally we cover everything but that'll be very verbose. I think it would be good to at least cover:
- timestamp (logical type is different from physical type)
- Structs (nested type)

johanl-db (approved):

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/internal/expressions/DefaultExpressionEvaluatorSuite.scala
lbhm: I added a test for timestamp columns. The necessary code changes to simulate and test nested types became very convoluted in my draft so I decided it was not worth the extra review/maintenance effort.
Conceptually, I would argue that map/array/struct types are safe to use with `COALESCE` since we really only care if a given value is `null` or not.

kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/internal/expressions/DefaultExpressionUtils.java
```diff
@@ -386,6 +386,87 @@ private ColumnVector getVector(int rowId) {
...
+   *
+   * <p>Only numeric data types are supported.
+   *
+   * @param left the left operand column vector
+   * @param right the right operand column vector
+   * @param operator the arithmetic operator to apply
+   * @return a new column vector representing the result of the arithmetic operation
+   */
+  static ColumnVector arithmeticVector(
+      ColumnVector left, ColumnVector right, ArithmeticOperator operator) {
```
allisonport-db, line 416: nit: `checkArgument(left.getDataType() = right.getDataType9))`?

allisonport-db: same thing for size maybe?

allisonport-db (approved):
1 minor comment LGTM

kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/internal/expressions/DefaultExpressionUtils.java
lbhm: Added both.

kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/internal/expressions/DefaultExpressionEvaluator.java
```diff
@@ -245,14 +245,37 @@ ExpressionTransformResult visitCoalesce(ScalarExpression coalesce) {
...
+
+    @Override
+    ExpressionTransformResult visitAdd(ScalarExpression add) {
+      List<ExpressionTransformResult> children =
+          add.getChildren().stream().map(this::visit).collect(Collectors.toList());
+      if (children.size() != 2) {
+        throw unsupportedExpressionException(
+            add, "ADD requires exactly two arguments: left and right operands");
+      }
+      if (children.get(0).outputType != children.get(1).outputType) {
```
qiyuandong-db: Looking at the rest of the code, we probably should use `outputType.equivalent` here.

PR #4956, opened 2025-07-22
[Spark] Reduce conflict checking in Row Tracking Backfill

tomvanbussel:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR improves the performance of the row tracking backfill on large tables. It does so by having each batch create its own transaction, instead of having them split off from a single transaction. This reduces the amount of versions that long running backfills have to conflict check against, and means that backfill batches no longer have to conflict check against previous batches of the same backfill. We do this by having each batch scan the delta log for files that have not been backfilled yet, instead of preparing this list of files ahead of time.

## How was this patch tested?

Existing tests to ensure that nothing breaks. Some small changes had to be made to the tests to account for there no longer being a main transaction, and to account for the removal of the `BackfillBatchIterator`.

No


andreaschat-db (approved):
LGTM.

PR #4959, opened 2025-07-22
[Kernel] Optimize path construction 

emkornfield:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This PR attempts to optimize path normalization, assuming that most paths are already normalized (don't contain more then. 1 front-slash in a row).  It first does a check to see if this condition holds, and if it doesn't it applies the regex replace.

Regex.replace is fairly heavy weight and by making this change we see a 40-50% reduction in Path construction time when the path is already normalized.  The additional overhead for when normalization is needed is within measurement noise on the computer I'm running in (best estimate is someplace between 1-5%).

Also remove an outdated TODO for removing path.

## How was this patch tested?

- Add unit tests (mostly AI generated)
- Added a JMH benchmark for cases when normalization is needed and not needed.

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/fs/Path.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/fs/Path.java
@@ -221,8 +221,7 @@ public class Path implements Comparable<Path>, Serializable, ObjectInputValidati
     // Validating this first before applying the regex saves ~40-50% of
     // Path construction time, with a potentially small overhead for
     // cases when paths need normalization.
-    boolean replace = containsRepeatedSlash(path);
-    if (replace) {
+    if (containsRepeatedSlash(path)) {
       // Remove duplicated slashes to ensure all equivalent Path's have
       // the same representation.
       path = SLASHES.matcher(path).replaceAll("/");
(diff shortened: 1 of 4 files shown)


huan233usc (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/fs/Path.java
```diff
@@ -240,6 +247,15 @@ private static String normalizePath(String scheme, String path) {
     return path;
   }
 
+  private static boolean containsRepeatedSlash(String path) {
+    for (int x = 0; x < path.length() - 1; x++) {
```
raveeram-db: would we better off using something like `return path.indexOf("//") != -1;` here (IIRC, the JIT treats indexOf as an intrinsic method and could be faster perhaps)

raveeram-db (commented):
Looks good, thanks for the PR. Just had a clarifying question

kernel/kernel-api/src/main/java/io/delta/kernel/internal/fs/Path.java
emkornfield: I didn't realize indexOf had an intrinsic, originally looking at the substring code it just seemed more complex which is why I went with this.  Benchmarking against `.contains` (which delegates to `.indexOf`) or `.indexOf` on my computer doesn't show definitive proof of being faster or slower, so i changed to use `.contains()` for simpler code.

raveeram-db (approved):

PR #4964, opened 2025-07-23
[Kernel] Refactor parameter names in `ScanStateRow.of()`

lbhm:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

The current parameter names in `ScanStateRow.of()` are kind of confusing and not aligned with the field names of `ScanStateRow.SCHEMA` (e.g., `readSchemaLogicalJson` maps to `logicalSchemaString`) . I propose to simplify and align them. There is only one method using `of()` (`ScanImpl.getScanState`) so this change should be safe.

## How was this patch tested?

`ScanStateRow.of()` is not directly used in any tests, so this change does not impact test behavior.

This is a user facing change for any user using named parameter access to `ScanStateRow.of()`. The order and functionality of the parameters remains unchanged.


lbhm:
Closed in favor of #5007

PR #4965, opened 2025-07-23
[Kernel] Enable coalesced reads of row tracking metadata columns

lbhm:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This PR builds upon #4943 and #4953 to enable fully functioning reads of row tracking metadata columns. Concretely, it adds functionality to compute non-materialized row IDs and row commit versions on the fly.

## How was this patch tested?

I modified the tests that I added in #4943 to now expect the fully correct result (instead of `null` for non-materialized values).

This PR turns the work started in #4943 into a usable new feature. Users/engines can now request row tracking metadata columns from Delta tables. At the moment, this PR still uses a preliminary API for requesting metadata columns until we agree on a stable metadata column API going forward.


lbhm:
Closing this in favor or #4994

PR #4975, opened 2025-07-25
[Spark] Add support for Row Tracking removal

andreaschat-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR adds support for row tracking removal. The main complexity is that row tracking cannot be disabled. The feature property, `delta.enableRowTracking`, only controls whether clients should expect all rows to contain row IDs. This means that concurrent transactions may continue generating row IDs while we clean up metadata.

Furthermore, the feature can be dropped and then re-enabled multiple times. Once the feature is dropped, there are no guarantees writers will preserve past metadata (either by copying action metadata or preserving them in the checkpoints). Therefore, we need to ensure that re-enablement is safe and it does not mix up (potentially incomplete) metadata from past enablements.

This can be aided with the introduction of an optional new configuration, `delta.suspendRowTracking`, that allows to suspend any row ID generation. This should help reduce the concurrent conflicting transactions. However, we cannot rely on this being respected by all clients. Thus, we need to design a solution that tolerates concurrent transactions.

We drop Row Tracking as follows:

a) Flip `delta.enableRowTracking` property. Readers/writers do not long expect all rows to be assigned with an ID.
b) Flip `delta.suspendRowTracking` property.
c) Unbackfill. This should only remove baseRowId and defaultRowCommitVersion from all actions. Note, in the meantime concurrent transactions can still produce row IDs. These should continue from the latest high watermark.
d) In a single commit, downgrade the protocol and remove all relevant properties (High watermark, materialized column names). Removing the materialized column names results in ignoring the columns in any future re-enablement of the feature.
e) Resolve any conflicting transactions by cleaning relevant metadata from add/remove actions. The conflict resolution should take place in both unbackfill commits as well as the drop feature commit.

## How was this patch tested?

Added new suite.

Yes. Now users can drop row tracking.

### DIFF
--- a/spark/src/main/resources/error/delta-error-classes.json
+++ b/spark/src/main/resources/error/delta-error-classes.json
@@ -2243,17 +2243,13 @@
   },
   "DELTA_ROW_TRACKING_BACKFILL_RUNNING_CONCURRENTLY_WITH_UNBACKFILL" : {
     "message" : [
-      "A Row Tracking enablement operation was detected running concurrently with a Row Tracking disablement.",
-      "Aborting the disablement operation.",
-      "Please retry the disablement operation if necessary when the enablement operation is complete."
+      "A Row Tracking enablement operation was detected running concurrently with a Row Tracking disablement. Aborting the disablement operation. Please retry the disablement operation if necessary when the enablement operation is complete."
     ],
     "sqlState" : "22KD0"
   },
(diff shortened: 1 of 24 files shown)


tomvanbussel (approved):
Nice! This will hopefully help a lot with increasing the adoption of row tracking, as this eliminates the risk of enabling it without being able to disable it.

PR #4981, opened 2025-07-28
[Spark] Run DeleteSQLSuite with both path-based and name-based access

ctring:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR makes `DeleteSQLSuite` an abstract suite and extends it to run with both path-based and name-based access

## How was this patch tested?

New and existing UTs

No


LukasRupprecht (approved):
LGTM!

PR #4982, opened 2025-07-28
Roll forward  [Spark] Split Large Test Suites (#4957), with increasing test shard

huan233usc:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Currently impression is splitting of test suite change how tests are sharding, and some test setup/turn down pollutes each other(deleting other tests path during cleanup

We increase test shard which seems to be working
## How was this patch tested?

workflow


larsk-db (approved):
LGTM

PR #4993, opened 2025-07-30
[Spark] Optimize MERGE DELETE to avoid scanning unnecessary target columns when CDC is disabled

mollyo-openai:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

When executing MERGE with DELETE actions, Delta’s write plan currently references all target payload columns even when CDC is off and we don’t need delete preimages. This happens in:

the matched DELETE branch, and

the target_is_null (source‑only) fallback used when no NOT MATCHED … INSERT applies.

Because those branches reference targetWriteCols, Spark cannot column‑prune the target payload; this inflates bytes scanned for delete‑only MERGEs (even with Deletion Vectors enabled).

With this PR, when CDC is disabled:
* In matched DELETE action vectors, replace targetWriteCols with typed NULLs
* In the target_is_null fallback vector, replace targetWriteCols with typed NULLs

## How was this patch tested?

tested against delete only merged and verified spark plans before and after, query similar to
```
MERGE INTO target_table t
                USING source_table i
                ON <condition>
                WHEN MATCHED THEN DELETE
```

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/commands/merge/MergeOutputGeneration.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/commands/merge/MergeOutputGeneration.scala
@@ -129,8 +129,12 @@ trait MergeOutputGeneration { self: MergeIntoCommandBase =>
     // unmatched source row when it should not be inserted. `target.output` will produce NULLs
     // which will get deleted eventually.
 
+    val deletedColsForUnmatchedTarget =
+      if (cdcEnabled) targetWriteCols
+      else targetWriteCols.map(e => Cast(Literal(null), e.dataType))
+
     val deleteSourceRowExprs =
-      (targetWriteCols ++
+      (deletedColsForUnmatchedTarget ++
         rowIdColumnExpressionOpt.map(_ => Literal(null)) ++
         rowCommitVersionColumnExpressionOpt.map(_ => Literal(null)) ++
         Seq(Literal(true))) ++
@@ -291,7 +295,15 @@ trait MergeOutputGeneration { self: MergeIntoCommandBase =>
             }
           }
           // Generate expressions to set the ROW_DROPPED_COL = true and mark as a DELETE
-          targetWriteCols ++
+          // Only read full target columns if CDC is enabled
+          val deletedDataExprs =
+            if (cdcEnabled) {
+              targetWriteCols
+            } else {
+              targetWriteCols.map(e => Cast(Literal(null), e.dataType))
+            }
+
+          deletedDataExprs ++
             rowIdColumnExpressionOpt ++
             rowCommitVersionColumnExpressionOpt ++
             Seq(incrCountExpr) ++


xupefei:
Thanks for the contribution! When you finish the update please change the PR status to "Ready for Review".

mollyo-openai:
thanks @xupefei! I have tested this in our fork for correctness and I see that all the tests are passing here. I am not sure how to best unit test this here, but any tips would be appreciated!

sunchao:
@xupefei please let us know what do you think about this PR. Does it look good to you overall (except tests)?

xupefei:
> @xupefei please let us know what do you think about this PR. Does it look good to you overall (except tests)?

Yes, the approach lgtm!

> thanks @xupefei! I have tested this in our fork for correctness and I see that all the tests are passing here. I am not sure how to best unit test this here, but any tips would be appreciated!

I would try to create a single DELETE query and inspect the query plan, down to the scan node and check which columns it will fetch -- once with CDC and once without.

xupefei:
Also, do we want to change the next `case` match: `DeltaMergeIntoNotMatchedBySourceDeleteClause`, to have the same behaviour?

spark/src/main/scala/org/apache/spark/sql/delta/commands/merge/MergeOutputGeneration.scala
```diff
@@ -291,7 +295,15 @@ trait MergeOutputGeneration { self: MergeIntoCommandBase =>
             }
           }
           // Generate expressions to set the ROW_DROPPED_COL = true and mark as a DELETE
-          targetWriteCols ++
+          // Only read full target columns if CDC is enabled
+          val deletedDataExprs =
```
johanl-db, line 299: +1 to @xupefei, I think we should also do the same for `DeltaMergeIntoNotMatchedBySourceDeleteClause`.

Also for consistency with https://github.com/delta-io/delta/pull/5004, can you try either raising an exception with `RaiseError` in both cases (as in that other PR), or return null in both?


I'm wondering if using `RaiseError` here can cause issues, in case the expression gets evaluated before the row is filtered out. I imagine not, since it's essentially the same as in the other PR

mollyo-openai: hi @xupefei @johanl-db , thanks I have added the DeltaMergeIntoNotMatchedBySourceDeleteClause.

RaiseError does cause issues here as the expression does end up being evaluated first (see failing tests on reverted commit https://github.com/delta-io/delta/pull/4993/commits/97fed8b6c4615bb73025b4490412eb69847bb7ba).

xupefei:
Hey @mollyo-openai, let me know if you plan to work on adding tests.

rqureshi-openai:
@xupefei just my 2c. It's totally possible someone someone modifies this code to implicitly trigger a table scan again but:
- white box tests that inspect the internal state are always brittle
- we have blackbox tests (this is a perf improvement, not creating net new functionality)

spark/src/main/scala/org/apache/spark/sql/delta/commands/merge/MergeOutputGeneration.scala
rqureshi-openai: +1 to molly's point.

@johanl-db the deleted rows get filtered out **afterwards** but the expression still gets evaluated. Rather than carry forward the target rows and then filter them out, we propagate nulls instead.

I wouldn't expect RaiseError to work here.

PR #4994, opened 2025-07-30
[Kernel] Enable fresh & materialized reads of row tracking metadata columns

lbhm:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This PR builds upon #4943 and #4953 to enable fully functioning reads of row tracking metadata columns. Concretely, it adds functionality to compute non-materialized row IDs and row commit versions on the fly.

## How was this patch tested?

I modified the tests that I added in #4943 to now expect the fully correct result (instead of `null` for non-materialized values).

This PR turns the work started in #4943 into a usable new feature. Users/engines can now request row tracking metadata columns from Delta tables. At the moment, this PR still uses a preliminary API for requesting metadata columns until we agree on a stable metadata column API going forward.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
@@ -21,6 +21,7 @@ import io.delta.kernel.data.*;
 import io.delta.kernel.engine.Engine;
 import io.delta.kernel.expressions.Predicate;
 import io.delta.kernel.internal.InternalScanFileUtils;
+import io.delta.kernel.internal.TableConfig;
 import io.delta.kernel.internal.actions.DeletionVectorDescriptor;
 import io.delta.kernel.internal.data.ScanStateRow;
 import io.delta.kernel.internal.data.SelectionColumnVector;
@@ -185,9 +186,12 @@ public interface Scan {
                 physicalReadSchema);
 
         // Transform physical row tracking columns to logical row tracking columns
-        nextDataBatch =
-            MaterializedRowTrackingColumn.transformPhysicalData(
-                nextDataBatch, scanFile, scanState, engine);
+        if (TableConfig.ROW_TRACKING_ENABLED.fromMetadata(
+            ScanStateRow.getConfiguration(scanState))) {
+          nextDataBatch =
+              MaterializedRowTrackingColumn.transformPhysicalData(
+                  nextDataBatch, scanFile, scanState, engine);
+        }
 
         // Get the selectionVector if DV is present
         DeletionVectorDescriptor dv =
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
@@ -219,8 +219,7 @@ public interface Scan {
 
         // If a column was only requested to compute other columns, we remove it
         for (StructField field : nextDataBatch.getSchema().fields()) {
-          if (field.getMetadata().contains(StructField.IS_INTERNAL_COLUMN_KEY)
-              && field.getMetadata().getBoolean(StructField.IS_INTERNAL_COLUMN_KEY)) {
+          if (field.isInternalColumn()) {
             int columnOrdinal = nextDataBatch.getSchema().indexOf(field.getName());
             if (columnOrdinal == -1) {
               // This should never happen since we only interact with a single schema
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
@@ -35,6 +35,7 @@ import io.delta.kernel.types.StructField;
 import io.delta.kernel.types.StructType;
 import io.delta.kernel.utils.CloseableIterator;
 import java.io.IOException;
+import java.util.Map;
 import java.util.Optional;
 
 /**
@@ -186,11 +187,11 @@ public interface Scan {
                 physicalReadSchema);
 
         // Transform physical row tracking columns to logical row tracking columns
-        if (TableConfig.ROW_TRACKING_ENABLED.fromMetadata(
-            ScanStateRow.getConfiguration(scanState))) {
+        Map<String, String> configuration = ScanStateRow.getConfiguration(scanState);
+        if (TableConfig.ROW_TRACKING_ENABLED.fromMetadata(configuration)) {
           nextDataBatch =
               MaterializedRowTrackingColumn.transformPhysicalData(
-                  nextDataBatch, scanFile, scanState, engine);
+                  nextDataBatch, scanFile, configuration, engine);
         }
 
         // Get the selectionVector if DV is present
@@ -218,12 +219,17 @@ public interface Scan {
 
         // If a column was only requested to compute other columns, we remove it
         for (StructField field : nextDataBatch.getSchema().fields()) {
-          if (field.getMetadata().contains(StructField.IS_INTERNAL_METADATA_COLUMN_KEY)
-              && field.getMetadata().getBoolean(StructField.IS_INTERNAL_METADATA_COLUMN_KEY)) {
+          if (field.getMetadata().contains(StructField.IS_INTERNAL_COLUMN_KEY)
+              && field.getMetadata().getBoolean(StructField.IS_INTERNAL_COLUMN_KEY)) {
             int columnOrdinal = nextDataBatch.getSchema().indexOf(field.getName());
-            if (columnOrdinal != -1) {
-              nextDataBatch = nextDataBatch.withDeletedColumnAt(columnOrdinal);
+            if (columnOrdinal == -1) {
+              // This should never happen since we only interact with a single schema
+              throw new IllegalArgumentException(
+                  String.format(
+                      "Column %s was requested internally but is not present in the data batch.",
+                      field.getName()));
             }
+            nextDataBatch = nextDataBatch.withDeletedColumnAt(columnOrdinal);
           }
         }
 
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
@@ -26,6 +26,7 @@ import io.delta.kernel.internal.data.ScanStateRow;
 import io.delta.kernel.internal.data.SelectionColumnVector;
 import io.delta.kernel.internal.deletionvectors.DeletionVectorUtils;
 import io.delta.kernel.internal.deletionvectors.RoaringBitmapArray;
+import io.delta.kernel.internal.rowtracking.MaterializedRowTrackingColumn;
 import io.delta.kernel.internal.util.ColumnMapping.ColumnMappingMode;
 import io.delta.kernel.internal.util.PartitionUtils;
 import io.delta.kernel.internal.util.Tuple2;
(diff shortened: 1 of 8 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -254,7 +256,15 @@ private StructType createPhysicalSchema() {
             .map(StructField::getName)
             .noneMatch(name -> name.equals(StructField.METADATA_ROW_INDEX_COLUMN_NAME))) {
       // If the row index column is not already present, add it to the physical read schema
-      physicalFields.add(StructField.METADATA_ROW_INDEX_COLUMN);
+      physicalFields.add(
```
johanl-db: Use `StructField.withNewMetadata` on the existing `StructField.METADATA_ROW_INDEX_COLUMN`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -17,14 +17,20 @@
...
+import io.delta.kernel.engine.ExpressionHandler;
 import io.delta.kernel.exceptions.InvalidTableException;
+import io.delta.kernel.expressions.*;
 import io.delta.kernel.internal.DeltaErrors;
+import io.delta.kernel.internal.InternalScanFileUtils;
 import io.delta.kernel.internal.TableConfig;
 import io.delta.kernel.internal.actions.Metadata;
 import io.delta.kernel.internal.util.ColumnMapping;
+import io.delta.kernel.types.*;
 import io.delta.kernel.types.LongType;
```
johanl-db: Remove `import io.delta.kernel.types.LongType;`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -173,17 +176,28 @@ public static List<StructField> convertToPhysicalColumn(
...
+      List<StructField> physicalFields = new ArrayList<>(2);
+      physicalFields.add(
           new StructField(
-              ROW_ID.getPhysicalColumnName(metadata), LongType.LONG, true /* nullable */));
+              ROW_ID.getPhysicalColumnName(metadata.getConfiguration()),
+              LongType.LONG,
+              true /* nullable */));
+      if (logicalSchema.indexOf(StructField.METADATA_ROW_INDEX_COLUMN_NAME) == -1) {
+        physicalFields.add(
+            new StructField(
```
johanl-db: See other comment, use `StructField.withNewMetadata`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -193,14 +207,119 @@ public static List<StructField> convertToPhysicalColumn(
             logicalField.getName()));
   }
 
-  private String getPhysicalColumnName(Metadata metadata) {
-    return Optional.ofNullable(metadata.getConfiguration().get(getMaterializedColumnNameProperty()))
+  /**
+   * Transforms physical data related to row tracking to the logical metadata requested by the user.
```
johanl-db: nit:

```suggestion
   * Computes row IDs and row commit versions based on their materialized values if present in the data returned by the Parquet reader, and using the base row ID and default row commit version from the AddFile otherwise.
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -193,14 +207,119 @@ public static List<StructField> convertToPhysicalColumn(
...
+              commitVersionColumnName,
+              exprHandler,
+              readSchema,
+              commitVersionOrdinal);
+    }
+
+    return dataBatch;
+  }
+
+  private static ColumnarBatch transformPhysicalRowId(
```
johanl-db, line 246: Add a small comment to mention with compute row IDs as COALESCE(materialize_row_id, base_row_id + row_index)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -193,14 +207,119 @@ public static List<StructField> convertToPhysicalColumn(
...
+      String rowIdColumnName,
+      ExpressionHandler exprHandler,
+      StructType readSchema,
+      int rowIdOrdinal) {
+    long baseRowId =
+        InternalScanFileUtils.getBaseRowId(scanFile)
+            .orElseThrow(
+                () ->
+                    DeltaErrors.rowTrackingMetadataMissingInFile(
+                        "Base row ID", InternalScanFileUtils.getFilePath(scanFile)));
```
johanl-db:
```suggestion
                        "baseRowId", InternalScanFileUtils.getFilePath(scanFile)));
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -193,14 +207,119 @@ public static List<StructField> convertToPhysicalColumn(
...
+      String commitVersionColumnName,
+      ExpressionHandler exprHandler,
+      StructType readSchema,
+      int commitVersionOrdinal) {
+    long defaultRowCommitVersion =
+        InternalScanFileUtils.getDefaultRowCommitVersion(scanFile)
+            .orElseThrow(
+                () ->
+                    DeltaErrors.rowTrackingMetadataMissingInFile(
+                        "Default row commit version", InternalScanFileUtils.getFilePath(scanFile)));
```
johanl-db:
```suggestion
                        "defaultRowCommitVersion", InternalScanFileUtils.getFilePath(scanFile)));
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -193,14 +207,119 @@ public static List<StructField> convertToPhysicalColumn(
...
+
+    // Remove the materialized row ID column and replace it with the coalesced vector
+    dataBatch =
+        dataBatch
+            .withDeletedColumnAt(rowIdOrdinal)
+            .withNewColumn(rowIdOrdinal, METADATA_ROW_ID_COLUMN, rowIdVector);
+    return dataBatch;
+  }
+
+  private static ColumnarBatch transformPhysicalCommitVersion(
```
johanl-db, line 284: Same, add a quick comment to explain row commit versions are computed as COALESCE(materialized_row_commit_version, default_row_commit_version)

johanl-db (commented):
The change looks good, mostly minor comments.

Small nit on the PR title: I believe 'coalesce reads' has a well defined meaning for memory access, might be a bit confusing for some people

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
lbhm: I created a `StructField.createInternalColumn()` factory that exactly replicates the object plus the additional `FieldMetadata` key. Using `withNewMetadata` would require me to iterate and copy the possibly existing metadata.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
lbhm: See above.

kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
```diff
@@ -183,17 +186,23 @@ public FilteredColumnarBatch next() {
...
                 physicalReadSchema);
 
-        DeletionVectorDescriptor dv =
-            InternalScanFileUtils.getDeletionVectorDescriptorFromRow(scanFile);
-
-        int rowIndexOrdinal =
-            nextDataBatch.getSchema().indexOf(StructField.METADATA_ROW_INDEX_COLUMN_NAME);
+        // Transform physical row tracking columns to logical row tracking columns
+        Map<String, String> configuration = ScanStateRow.getConfiguration(scanState);
+        if (TableConfig.ROW_TRACKING_ENABLED.fromMetadata(configuration)) {
```
allisonport-db, line 197: I'm not sure this is the right conditional? Shouldn't we never have a rowTracking column in the readSchema if it's not enabled?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -193,14 +198,126 @@ public static List<StructField> convertToPhysicalColumn(
...
+   * data returned by the Parquet reader, using the base row ID and default row commit version from
+   * the AddFile otherwise.
+   *
+   * @param dataBatch a batch of physical data read from the table.
+   * @param scanFile the {@link Row} representing the scan file metadata.
+   * @param configuration the configuration map containing table metadata.
+   * @param engine the {@link Engine} to use for expression evaluation.
+   * @return a new {@link ColumnarBatch} with logical row tracking columns
+   */
+  public static ColumnarBatch transformPhysicalData(
```
allisonport-db, line 213: cc @raveeram @vkorukanti this would be so much better with the expression fixup changes

allisonport-db (commented):
Just 1 minor comment

kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
lbhm: You're right. I added this conditional as a reading help (as in "don't even look at `MaterializedRowTrackingColumn.transformPhysicalData` if you're not concerned about row tracking") and as a micro optimization (i.e., we can skip the conditionals in `transformPhysicalData()` if row tracking is not enabled).

Would you prefer to have this removed?

allisonport-db: I was thinking maybe the conditional should be if the rowTracking columns are/aren't present in the read schema. Are you thinking this circumvents it? Maybe then just add something along the lines of
```
// If row tracking is enabled, check for row tracking columns in the read schema and transform as needed
```

lbhm: Checking if row tracking columns are enabled essentially is would the function we call here does so it would duplicate logic. 
I like your proposed comment to clarify things - I'll add it tomorrow.

kernel/kernel-api/src/main/java/io/delta/kernel/types/StructField.java
```diff
@@ -198,6 +209,19 @@ public StructField withDataType(DataType newType) {
...
+    FieldMetadata.Builder metadataBuilder =
+        FieldMetadata.builder()
+            .fromMetadata(field.metadata)
+            .putBoolean(IS_INTERNAL_COLUMN_KEY, true);
+    return new StructField(
+        field.getName(),
+        field.getDataType(),
+        field.isNullable(),
+        metadataBuilder.build(),
+        field.getTypeChanges());
```
johanl-db: nit
```suggestion
    FieldMetadata metadata =
        FieldMetadata.builder()
            .fromMetadata(field.metadata)
            .putBoolean(IS_INTERNAL_COLUMN_KEY, true)
            .build();
    return field.withNewMetadata(metadata);
```

kernel/kernel-api/src/main/java/io/delta/kernel/types/StructField.java
```diff
@@ -198,6 +209,19 @@ public StructField withDataType(DataType newType) {
     return new StructField(name, newType, nullable, metadata, typeChanges);
   }
 
+  public static StructField createInternalColumn(StructField field) {
```
johanl-db: Can we put this somewhere else so that it's not part of the public interface of `StructField`? We really don't want engines to create internal columns

johanl-db (approved):
Minor leftover comments, the change looks good to me

kernel/kernel-api/src/main/java/io/delta/kernel/types/StructField.java
lbhm: I moved the method to `ScanImpl`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -226,9 +227,23 @@ public Optional<Predicate> getRemainingFilter() {
     return getDataFilters();
   }
 
+  /** Helper method to create a copy of a column that is marked as an internal column. */
+  public static StructField createInternalColumn(StructField field) {
```
allisonport-db, line 231: I feel like this would be better in `SchemaUtils` but fine to do this in a follow up PR

PR #5003, opened 2025-07-31
[Kernel] [CatalogManaged] Removed ResolvedTable; Just use Snapshot

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/5003/files) to review incremental changes.
- [**stack/kernel_catalog_managed_merge_resolved_table_snapshot**](https://github.com/delta-io/delta/pull/5003) [[Files changed](https://github.com/delta-io/delta/pull/5003/files)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

I think we made a mistake in our implementation of the new CCv2 Kernel feature, with regard to the new entity "ResolvedTable".

I think we got carried away, or mis-directed, in this being something _new_, whereas in reality it's just a Snapshot with a few different capabilities (e.g. give me the committer; create a new transaction for me) and is created differently (from the builder).

This is some slightly new semantics/capabilities, but we can, and should, have just added these to Snapshot.

That's what this PR does -- remove ResolvedTable, add that functionality to Snapshot.

If we really want it called "ResolvedTable" -- then we can do a massive breaking API change later and rename things.

Existing UTs.


kernel/kernel-api/src/main/java/io/delta/kernel/Snapshot.java
```diff
@@ -76,4 +80,7 @@ public interface Snapshot {
    * @return an instance of {@link ScanBuilder}
    */
   ScanBuilder getScanBuilder();
+
+  /** @return a committer that owns and controls commits to this table */
+  Committer getCommitter();
```
allisonport-db, line 92: So this will always be present and will just be file-system committer for non-ccv2 tables?

kernel/kernel-api/src/main/java/io/delta/kernel/TableManager.java
```diff
@@ -17,29 +17,23 @@
...
-import io.delta.kernel.internal.table.ResolvedTableBuilderImpl;
+import io.delta.kernel.internal.table.SnapshotBuilderImpl;
 
-/** The entry point to load and create {@link ResolvedTable}s. */
+/** The entry point to load and create Delta tables. */
 @Experimental
 public interface TableManager {
-  // TODO static ResolvedTable forPathAtLatest(Engine engine, String path);
-  // TODO static ResolvedTable forPathAtVersion(Engine engine, String path, long version);
-  // TODO static ResolvedTable forPathAtTimestamp(Engine engine, String path, long timestamp);
```
allisonport-db, line 27: Do we still want these or is the `Table` interface the current stand-in for this?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/CommitContextImpl.java
allisonport-db, line 1: Are we getting rid of these classes as well? (future PR?)

kernel/kernel-api/src/main/java/io/delta/kernel/Snapshot.java
scottsand-db: Yup!

kernel/kernel-api/src/main/java/io/delta/kernel/TableManager.java
scottsand-db: The `TableManager.loadSnapshot` is how you build a Snapshot -- regardless of if it is for a path-based table or a catalog-maanged table. Deleting these comments is actually unrelated to this PR -- I just decided to clean it up now :)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/CommitContextImpl.java
scottsand-db: They won't be used, for a while. IMO we can keep them, and use them in the future when splitting up the transaction commit flow (give more power to the connector) becomes a priority.

allisonport-db: Maybe it would be helpful (in a future PR) to just add a note for all these unused classes that they aren't active code?

allisonport-db (approved):
LGTM

huan233usc (approved):

Issue #4992, opened 2025-07-30
[BUG] Running MERGE INTO query with a DELETE action causes full table scan


rqureshi-openai:
#### Which Delta project/connector is this regarding?
- [x] Spark

### Describe the problem
Running a MERGE INTO query that has a DELETE action causes the entire table to be scanned including all data/payload fields, rather than metadata fields and the required fields to resolve the merge condition and match conditions. This makes merging operations very slow and expensive since the entire target table needs to be read and shuffled.

#### Steps to reproduce

Run a `MERGE INTO` query with the following form:

```
MERGE INTO snapshot_table t
USING change_events c
ON t._id = c._id
WHEN MATCHED AND c.operation='DELETE' THEN DELETE
```

#### Observed results
We see the following table scan:

```
Scan snapshot_table
Output : [_id, **data**, __delta_internal_is_row_deleted, _tmp_metadata_row_index, file_path, file_name, file_size, file_block_start, file_block_length, file_modification_time]
```

#### Expected results
We expected the scan to look like this (notice: no data field):
```
Scan snapshot_table
Output : [_id, __delta_internal_is_row_deleted, _tmp_metadata_row_index, file_path, file_name, file_size, file_block_start, file_block_length, file_modification_time]
```

#### Further details

We drove an investigation and further fixed the issue in our fork.

In [MergeOutputGeneration.scala](https://github.com/delta-io/delta/blob/branch-3.2/spark/src/main/scala/org/apache/spark/sql/delta/commands/merge/MergeOutputGeneration.scala#L129) there is the following code snippet:

```
    val deleteSourceRowExprs =
      (targetWriteCols ++
        rowIdColumnExpressionOpt.map(_ => Literal(null)) ++
        rowCommitVersionColumnExpressionOpt.map(_ => Literal(null)) ++
        Seq(Literal(true))) ++
        (if (cdcEnabled) Seq(CDC_TYPE_NOT_CDC) else Seq())
```

Which we changed to the following:
```
   val deletedColsForUnmatchedTarget =
      if (cdcEnabled || !pruneTargetColsInDelete) targetWriteCols
      else targetWriteCols.map(e => Cast(Literal(null), e.dataType))
    val deleteSourceRowExprs =
      (targetWriteCols ++
      (deletedColsForUnmatchedTarget ++
        rowIdColumnExpressionOpt.map(_ => Literal(null)) ++
        rowCommitVersionColumnExpressionOpt.map(_ => Literal(null)) ++
[...]

PR #5004, opened 2025-08-01
[Spark] Avoid full table scan when MATCHED, NOT MATCHED predicates have conditionals 

rqureshi-openai:
Spark

## Description

Consider a MERGE INTO query as follows:

```
MERGE INTO target_table t
USING change_events c
on c.id = t.id
WHEN MATCHED and c.operation = 'Update' THEN UPDATE SET *
```

Originally, this would create projections like this which would reference the target table's columns, causing a full table scan:
```
If(update_condition) c.data else t.data
```

This PR changes the projection to be like this:
```
If(update_condition) c.data else RaiseError("unexpected row: this row should have been filtered out
```

The key question is: why can we do this? The basic answer is because we will never hit the else condition because in a previous step, we filtered to rows that will be modified and need to be rewritten. At this point, I will mention that we can optimize the projection by directly setting c.data instead of having the if-statement, but in the spirit of incremental, easy-to-read changes and for added safety, I decided to replace the fallbacks with RaiseErrors, since we want it to scream loudly that something is wrong. In a followup, we can implement the optimization.

Now let me try to explain more intuitively and with some code pointers. MERGE INTO is broken into two steps: rewriting modified rows, and writing delete vectors. Well let's think: why do we even need a fallback? If a row doesn't match any of the actions, we shouldn't even be processing it. And in fact, we don't, because there is an explicit filtering step in `ClassicMergeExecutor.scala` `def writeAllChanges` which is the method that constructs the overall plan/strategy for the rewriting phase. 

```

  /**
   * Write new files by reading the touched files and updating/inserting data using the source
   * query/table. This is implemented using a full-outer-join using the merge condition.
   *
   * Note that unlike the insert-only code paths with just one control column ROW_DROPPED_COL, this
   * method has a second control column CDC_TYPE_COL_NAME used for handling CDC when enabled.
   */
  protected def writeAllChanges(...) {
  ...
    val joinedDF =
      if (writeUnmodifiedRows) {
        joinedBaseDF
      } else {
        val filter = Or(generateFilterForModifiedRows(), generateFilterForNewRows())
        joinedBaseDF.filter(Column(filter))
      }
   ....
  }

  /**
   * Returns the expression that can be used for selecting the modified rows generated
   * by the merge operation. The expression is to designed to work irrespectively
   * of the join type used between the source and target tables.
   *
   * The expression consists of two parts, one for each of the action clause types that produce
   * row modifications: MATCHED, NOT MATCHED BY SOURCE. All actions of the same clause type form
   * a disjunctive clause. The result is then conjucted to an expression that filters the rows
   * of the particular action clause type. For example:
   *
   * MERGE INTO t
   * USING s
   * ON s.id = t.id
   * WHEN MATCHED AND id < 5 THEN ...
   * WHEN MATCHED AND id > 10 THEN ...
   * WHEN NOT MATCHED BY SOURCE AND id > 20 THEN ...
   *
   * Produces the following expression:
   *
   * ((as.id = t.id) AND (id < 5 OR id > 10))
   * OR
   * ((SOURCE TABLE IS NULL) AND (id > 20))
   */
  protected def generateFilterForModifiedRows(): Expression = {
    val matchedExpression = if (matchedClauses.nonEmpty) {
      And(Column(condition).expr, clauseDisjunction(matchedClauses))
    } else {
      Literal.FalseLiteral
    }

    val notMatchedBySourceExpression = if (notMatchedBySourceClauses.nonEmpty) {
      val combinedClauses = clauseDisjunction(notMatchedBySourceClauses)
      And(col(SOURCE_ROW_PRESENT_COL).isNull.expr, combinedClauses)
    } else {
      Literal.FalseLiteral
    }

    Or(matchedExpression, notMatchedBySourceExpression)
  }
```

We filter to rows that we *know* will land in one of the actions. With this invariant, we can safely remove the fallbacks and raise errors instead.

No.

### DIFF
--- a/scalastyle-config.xml
+++ b/scalastyle-config.xml
@@ -136,7 +136,7 @@ This file is divided into 3 sections:
   </check>
 
   <check customId="argcount" level="error" class="org.scalastyle.scalariform.ParameterNumberChecker" enabled="true">
-    <parameters><parameter name="maxParameters"><![CDATA[11]]></parameter></parameters>
+    <parameters><parameter name="maxParameters"><![CDATA[10]]></parameter></parameters>
   </check>
 
   <check level="error" class="org.scalastyle.scalariform.NoFinalizeChecker" enabled="true"></check>
--- a/scalastyle-config.xml
+++ b/scalastyle-config.xml
@@ -136,7 +136,7 @@ This file is divided into 3 sections:
   </check>
 
   <check customId="argcount" level="error" class="org.scalastyle.scalariform.ParameterNumberChecker" enabled="true">
-    <parameters><parameter name="maxParameters"><![CDATA[10]]></parameter></parameters>
+    <parameters><parameter name="maxParameters"><![CDATA[11]]></parameter></parameters>
   </check>
 
   <check level="error" class="org.scalastyle.scalariform.NoFinalizeChecker" enabled="true"></check>
--- a/spark/src/main/scala/org/apache/spark/sql/delta/commands/merge/MergeOutputGeneration.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/commands/merge/MergeOutputGeneration.scala
@@ -212,11 +212,9 @@ trait MergeOutputGeneration { self: MergeIntoCommandBase =>
         case "inner" => matchedExprs(i)
         case "leftOuter" =>
           // The source row is never null, so we only need to check if the target row is null.
-          assert(notMatchedBySourceExprs.isEmpty)
           If(ifTargetRowNull, notMatchedExprs(i), matchedExprs(i))
         case "rightOuter" =>
           // The target row is never null, so we only need to check if the source row is null.
-          assert(notMatchedExprs.isEmpty)
           If(ifSourceRowNull, notMatchedBySourceExprs(i), matchedExprs(i))
         case "fullOuter" =>
           CaseWhen(Seq(
--- a/spark/src/main/scala/org/apache/spark/sql/delta/commands/merge/MergeOutputGeneration.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/commands/merge/MergeOutputGeneration.scala
@@ -28,7 +28,6 @@ import org.apache.spark.sql.catalyst.analysis.UnresolvedAttribute
 import org.apache.spark.sql.catalyst.expressions._
 import org.apache.spark.sql.catalyst.plans.logical._
 import org.apache.spark.sql.functions._
-import org.apache.spark.sql.catalyst.expressions.RaiseError
 
 /**
  * Contains logic to transform the merge clauses into expressions that can be evaluated to obtain
--- a/spark/src/main/scala/org/apache/spark/sql/delta/commands/merge/MergeOutputGeneration.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/commands/merge/MergeOutputGeneration.scala
@@ -108,7 +108,6 @@ trait MergeOutputGeneration { self: MergeIntoCommandBase =>
       writeUnmodifiedRows: Boolean,
       clausesWithPrecompConditions: Seq[DeltaMergeIntoClause],
       cdcEnabled: Boolean,
-      joinType: String,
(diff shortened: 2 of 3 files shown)


rqureshi-openai:
Closes #5005

rqureshi-openai:
Fixes delta-io/delta#5005

spark/src/main/scala/org/apache/spark/sql/delta/commands/merge/MergeOutputGeneration.scala
```diff
@@ -29,6 +28,7 @@ import org.apache.spark.sql.catalyst.analysis.UnresolvedAttribute
 import org.apache.spark.sql.catalyst.expressions._
 import org.apache.spark.sql.catalyst.plans.logical._
 import org.apache.spark.sql.functions._
+import org.apache.spark.sql.catalyst.expressions.RaiseError
```
johanl-db: Remove, this is already included by import on line 28

scalastyle-config.xml
```diff
@@ -136,7 +136,7 @@ This file is divided into 3 sections:
   </check>
 
   <check customId="argcount" level="error" class="org.scalastyle.scalariform.ParameterNumberChecker" enabled="true">
-    <parameters><parameter name="maxParameters"><![CDATA[10]]></parameter></parameters>
+    <parameters><parameter name="maxParameters"><![CDATA[11]]></parameter></parameters>
```
johanl-db: Prefer locally disabling the style check if needed:
```
// scalastyle:off argcount

// scalastyle:on argcount
```

johanl-db (approved):
@rqureshi-openai The logic of this change is sound, we shouldn't need to copy rows over when DVs are enabled, so `noopCopyExprs` should never be evaluated.

I think we may actually be able to do something similar for DELETE clauses: we shouldn't have to read the target values in that case, even for rows that match, as long as we don't write CDF (which does write the deleted target row out).

This could be a follow up. In any case, I believe the current change works correctly with CDF

spark/src/main/scala/org/apache/spark/sql/delta/commands/merge/MergeOutputGeneration.scala
rqureshi-openai: Done!

rqureshi-openai:
> @rqureshi-openai The logic of this change is sound, we shouldn't need to copy rows over when DVs are enabled, so `noopCopyExprs` should never be evaluated.
> 
> I think we may actually be able to do something similar for DELETE clauses: we shouldn't have to read the target values in that case, even for rows that match, as long as we don't write CDF (which does write the deleted target row out).
> 
> This could be a follow up. In any case, I believe the current change works correctly with CDF

I believe this point is made and addressed in the other issue: https://github.com/delta-io/delta/issues/4992

scalastyle-config.xml
rqureshi-openai: It wasn't needed. Thanks!

johanl-db:
> > @rqureshi-openai The logic of this change is sound, we shouldn't need to copy rows over when DVs are enabled, so `noopCopyExprs` should never be evaluated.
> > I think we may actually be able to do something similar for DELETE clauses: we shouldn't have to read the target values in that case, even for rows that match, as long as we don't write CDF (which does write the deleted target row out).
> > This could be a follow up. In any case, I believe the current change works correctly with CDF
> 
> I believe this point is made and addressed in the other issue: #4992

Yes, that's it, I just took a look at the other PR, it looks correct

johanl-db:
@mollyo-openai

I'd like to make sure we align the change your making in https://github.com/delta-io/delta/pull/4993 with this PR before merging, could you take a look?

mollyo-openai:
> @mollyo-openai
> 
> I'd like to make sure we align the change your making in #4993 with this PR before merging, could you take a look?

hi @johanl-db , sorry for the delay, it has been a busy week for us 😄. Picking back up the other PR and will update with a comment when it is ready. thanks for reviewing!

raveeram-db:
Thanks for the reviews @johanl-db!

PR #5006, opened 2025-08-01
[Kernel] Remove `schemaString` from `Metadata` action constructor

lbhm:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

The current `Metadata` action constructor requires both a schema `StructType` and a string representation of the schema. This introduces a possible danger of data inconsistency.

This PR therefore removes `schemaString` from the `Metadata` constructor and adjusts all affected methods & tests.

## How was this patch tested?

Existing tests. This PR contains a refactoring without new functionality.

No - action implementations are part of `kernel.internal`.

### DIFF
--- a/connectors/flink/src/test/java/io/delta/flink/source/internal/KernelMetadataUtils.java
+++ b/connectors/flink/src/test/java/io/delta/flink/source/internal/KernelMetadataUtils.java
@@ -55,9 +55,8 @@ public class KernelMetadataUtils {
             Optional.ofNullable("name"),
             Optional.ofNullable("description"),
             new io.delta.kernel.internal.actions.Format("parquet", new HashMap<String, String>()),
-            "schemaString",
             schema,
-            new ArrayValue() { // paritionColumns
+            new ArrayValue() { // partitionColumns
                 @Override
                 public int getSize() {
                     return 2;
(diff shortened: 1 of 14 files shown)


johanl-db (approved):
LGTM

allisonport-db (approved):

lbhm:
@vkorukanti Good point about serialization cost. The three spots where this could be relevant IMO are:
- [`Metadata.toString()`](https://github.com/delta-io/delta/pull/5006/files#diff-f219f998f8d7f297268ddaa86ae5da83670beaa1f588663c0d5f6a8330dcafefR200) - probably not that critical?
- [`Metadata.toRow()`](https://github.com/delta-io/delta/pull/5006/files#diff-f219f998f8d7f297268ddaa86ae5da83670beaa1f588663c0d5f6a8330dcafefR296) - this seems to be relevant for committing and creating CRC infos. Am I correct to assume that this happens at most once per transaction?
- [`TransactionStateRow.of()`](https://github.com/delta-io/delta/pull/5006/files#diff-8179a7d4f88efff39639dc2fcc96db6ae43cc8b5d563da4635c598fe2f09be80R53) - this might be the most relevant spot? However, I see that we already serialize the physical schema with every call to this method. So if serialization cost is an issue here, I propose that we should keep both `logicalSchemaJson` and `physicalSchemaJSON` as a `Lazy<String>` inside `Metadata` to ensure that we serialize both schema at most once per `Metadata` instance.

Pinging @allisonport-db since you touched the `physicalSchemaString` in `TransactionStateRow` last.

My proposed options are:
- Keep the PR as is because we agree that serialization is not an issue here.
- Add both `physicalSchemaJson` and `logicalSchemaJson` as internal `Lazy<String>`s to `Metadata` and expose their `.get()` methods through the existing public getters of `Metadata`.

PR #5010, opened 2025-08-01
[Kernel] [CatalogManaged] Implement default commiter & refactor TransactionImpl to use it

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/5010/files) to review incremental changes.
- [**stack/kernel_catalog_managed_transaction_impl_invoke_committer**](https://github.com/delta-io/delta/pull/5010) [[Files changed](https://github.com/delta-io/delta/pull/5010/files)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

This PR implements the default committer and updates our transaction class to use it to perform the commit. Consequently, we have to change how our TransactionImpl decides when/if to retry the commit, based on the type of CommitFailedException thrown by the committer.

## How was this patch tested?

New unit and e2e tests.

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/exceptions/ConcurrentWriteException.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/exceptions/ConcurrentWriteException.java
@@ -24,13 +24,19 @@ import io.delta.kernel.annotation.Evolving;
  */
 @Evolving
 public class ConcurrentWriteException extends KernelException {
+  static final String msg =
+      "Transaction has encountered a conflict and can not be committed. "
+          + "Query needs to be re-executed using the latest version of the table.";
+
   public ConcurrentWriteException() {
-    super(
-        "Transaction has encountered a conflict and can not be committed. "
-            + "Query needs to be re-executed using the latest version of the table.");
+    super(msg);
   }
 
   public ConcurrentWriteException(String message) {
     super(message);
+  }
+
+  public ConcurrentWriteException(Throwable cause) {
+    super(msg, cause);
   }
 }
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/exceptions/ConcurrentWriteException.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/exceptions/ConcurrentWriteException.java
@@ -24,12 +24,12 @@ import io.delta.kernel.annotation.Evolving;
  */
 @Evolving
 public class ConcurrentWriteException extends KernelException {
-  static final String msg =
-      "Transaction has encountered a conflict and can not be committed. "
+  private static final String ERROR_MESSAGE =
+      "Transaction has encountered a conflict and cannot be committed. "
           + "Query needs to be re-executed using the latest version of the table.";
 
   public ConcurrentWriteException() {
-    super(msg);
+    super(ERROR_MESSAGE);
   }
 
   public ConcurrentWriteException(String message) {
@@ -37,6 +37,6 @@ public class ConcurrentWriteException extends KernelException {
   }
 
   public ConcurrentWriteException(Throwable cause) {
-    super(msg, cause);
+    super(ERROR_MESSAGE, cause);
   }
 }
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
@@ -24,6 +24,8 @@ import static io.delta.kernel.internal.util.Utils.toCloseableIterator;
 import static java.util.Collections.emptyMap;
 
 import io.delta.kernel.*;
+import io.delta.kernel.commit.CommitFailedException;
+import io.delta.kernel.commit.CommitMetadata;
 import io.delta.kernel.data.Row;
 import io.delta.kernel.engine.Engine;
 import io.delta.kernel.exceptions.ConcurrentWriteException;
@@ -34,6 +36,7 @@ import io.delta.kernel.internal.actions.*;
 import io.delta.kernel.internal.annotation.VisibleForTesting;
 import io.delta.kernel.internal.checksum.CRCInfo;
 import io.delta.kernel.internal.clustering.ClusteringUtils;
+import io.delta.kernel.internal.commit.DefaultFileSystemManagedTableOnlyCommitter;
 import io.delta.kernel.internal.compaction.LogCompactionWriter;
 import io.delta.kernel.internal.data.TransactionStateRow;
 import io.delta.kernel.internal.fs.Path;
@@ -51,7 +54,6 @@ import io.delta.kernel.internal.stats.FileSizeHistogram;
 import io.delta.kernel.internal.tablefeatures.TableFeatures;
 import io.delta.kernel.internal.util.*;
 import io.delta.kernel.internal.util.Clock;
-import io.delta.kernel.internal.util.FileNames;
 import io.delta.kernel.internal.util.InCommitTimestampUtils;
 import io.delta.kernel.internal.util.VectorUtils;
 import io.delta.kernel.metrics.TransactionMetricsResult;
@@ -61,7 +63,6 @@ import io.delta.kernel.utils.CloseableIterable;
 import io.delta.kernel.utils.CloseableIterator;
 import java.io.IOException;
 import java.io.UncheckedIOException;
-import java.nio.file.FileAlreadyExistsException;
 import java.util.*;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -311,27 +312,54 @@ public class TransactionImpl implements Transaction {
       }
 
       int numTries = 0;
-      while (numTries <= maxRetries) { // leq because the first is a try, not a retry
+      while (true) {
         logger.info("Committing transaction as version = {}.", commitAsVersion);
         try {
           transactionMetrics.commitAttemptsCounter.increment();
           return doCommit(
               engine, commitAsVersion, attemptCommitInfo, dataActions, transactionMetrics);
-        } catch (FileAlreadyExistsException fnfe) {
-          logger.info(
-              "Concurrent write detected when committing as version = {}.", commitAsVersion);
-          if (numTries < maxRetries) {
-            // only try and resolve conflicts if we're going to retry
+        } catch (CommitFailedException cfe) {
+          if (!cfe.isRetryable()) {
+            logger.error(
+                "Commit attempt for version {} failed with a non-retryable exception and will not "
+                    + "be retried. Error: {}",
+                commitAsVersion,
+                cfe.getMessage());
+            throw new RuntimeException(cfe);
+          } else if (numTries >= maxRetries) {
+            logger.error(
+                "Commit attempt for version {} failed with a retryable exception but will not be "
+                    + "retried because the maximum number of retries ({}) has been exhausted. "
+                    + "Error: {}",
+                commitAsVersion,
+                maxRetries,
+                cfe.getMessage());
+            throw new ConcurrentWriteException();
+          } else if (cfe.isConflict()) {
+            logger.warn(
+                "Commit attempt for version {} failed with a retryable exception due to a "
+                    + "physical conflict. {} Error: {}",
+                commitAsVersion,
+                cfe.getMessage());
+
             TransactionRebaseState rebaseState =
                 resolveConflicts(engine, commitAsVersion, attemptCommitInfo, numTries, dataActions);
             commitAsVersion = rebaseState.getLatestVersion() + 1;
             dataActions = rebaseState.getUpdatedDataActions();
             domainMetadataState.setComputedDomainMetadatas(rebaseState.getUpdatedDomainMetadatas());
             currentCrcInfo = rebaseState.getUpdatedCrcInfo();
-            // Action counters may be partially incremented from previous tries, reset the counters
-            // to 0 and drop fileSizeHistogram
+            // Action counters may be partially incremented from previous tries, reset the
+            // counters to 0 and drop fileSizeHistogram
             // TODO: reconcile fileSizeHistogram.
             transactionMetrics.resetActionMetricsForRetry();
+          } else {
+            // No conflict resolution needed - just retry with same version
+
+            logger.warn(
+                "Commit attempt for version {} failed with a retryable exception due to a "
+                    + "transient error. {} Skipping conflict resolution and trying again. Error: {}",
+                commitAsVersion,
+                cfe.getMessage());
           }
         }
         numTries++;
@@ -339,10 +367,6 @@ public class TransactionImpl implements Transaction {
     } finally {
       closed = true;
     }
-
-    // we have exhausted the number of retries, give up.
-    logger.info("Exhausted maximum retries ({}) for committing transaction.", maxRetries);
-    throw new ConcurrentWriteException();
   }
 
   private TransactionRebaseState resolveConflicts(
@@ -415,7 +439,7 @@ public class TransactionImpl implements Transaction {
       CommitInfo attemptCommitInfo,
       CloseableIterable<Row> dataActions,
       TransactionMetrics transactionMetrics)
-      throws FileAlreadyExistsException {
+      throws CommitFailedException {
     List<Row> metadataActions = new ArrayList<>();
     metadataActions.add(createCommitInfoSingleAction(attemptCommitInfo.toRow()));
     if (shouldUpdateMetadata) {
@@ -445,10 +469,25 @@ public class TransactionImpl implements Transaction {
       } else {
         completeFileActionIter = userStageDataIter;
       }
+
+      boolean isAppendOnlyTable = APPEND_ONLY_ENABLED.fromMetadata(metadata);
+
       // Create a new CloseableIterator that will return the metadata actions followed by the
       // data actions.
       CloseableIterator<Row> dataAndMetadataActions =
-          toCloseableIterator(metadataActions.iterator()).combine(completeFileActionIter);
+          toCloseableIterator(metadataActions.iterator())
+              .combine(completeFileActionIter)
+              .map(
+                  action -> {
+                    incrementMetricsForFileActionRow(transactionMetrics, action);
+                    if (!action.isNullAt(REMOVE_FILE_ORDINAL)) {
+                      RemoveFile removeFile = new RemoveFile(action.getStruct(REMOVE_FILE_ORDINAL));
+                      if (isAppendOnlyTable && removeFile.getDataChange()) {
+                        throw DeltaErrors.cannotModifyAppendOnlyTable(dataPath.toString());
+                      }
+                    }
+                    return action;
+                  });
 
       if (commitAsVersion == 0) {
         // New table, create a delta log directory
@@ -460,37 +499,27 @@ public class TransactionImpl implements Transaction {
         }
       }
 
-      boolean isAppendOnlyTable = APPEND_ONLY_ENABLED.fromMetadata(metadata);
+      final CommitMetadata commitMetadata =
+          new CommitMetadata(
+              commitAsVersion,
+              logPath.toString(),
+              attemptCommitInfo,
+              readSnapshot.getVersion() >= 0
+                  ? Optional.of(readSnapshot.getProtocol())
+                  : Optional.empty(),
+              readSnapshot.getVersion() >= 0
+                  ? Optional.of(readSnapshot.getMetadata())
+                  : Optional.empty(),
+              shouldUpdateProtocol ? Optional.of(protocol) : Optional.empty(),
+              shouldUpdateMetadata ? Optional.of(metadata) : Optional.empty());
 
-      // Write the staged data to a delta file
-      wrapEngineExceptionThrowsIO(
-          () -> {
-            engine
-                .getJsonHandler()
-                .writeJsonFileAtomically(
-                    FileNames.deltaFile(logPath, commitAsVersion),
-                    dataAndMetadataActions.map(
-                        action -> {
-                          incrementMetricsForFileActionRow(transactionMetrics, action);
-                          if (!action.isNullAt(REMOVE_FILE_ORDINAL)) {
-                            RemoveFile removeFile =
-                                new RemoveFile(action.getStruct(REMOVE_FILE_ORDINAL));
-                            if (isAppendOnlyTable && removeFile.getDataChange()) {
-                              throw DeltaErrors.cannotModifyAppendOnlyTable(dataPath.toString());
-                            }
-                          }
-                          return action;
-                        }),
-                    false /* overwrite */);
-            return null;
-          },
-          "Write file actions to JSON log file `%s`",
-          FileNames.deltaFile(logPath, commitAsVersion));
+      // May throw CommitFailedException
+      DefaultFileSystemManagedTableOnlyCommitter.INSTANCE.commit(
+          engine, dataAndMetadataActions, commitMetadata);
 
       return commitAsVersion;
-    } catch (FileAlreadyExistsException e) {
-      throw e;
     } catch (IOException ioe) {
+      // Error closing the CloseableIterator of actions or error creating the delta log directory
       throw new UncheckedIOException(ioe);
     }
   }
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
@@ -341,11 +341,11 @@ public class TransactionImpl implements Transaction {
                 commitAsVersion,
                 maxRetries,
                 cfe.getMessage());
-            throw new ConcurrentWriteException();
+            throw new RuntimeException(cfe);
           } else if (cfe.isConflict()) {
             logger.warn(
-                "Commit attempt for version {} failed with a retryable exception due to a "
-                    + "physical conflict. Error: {}",
+                "Commit attempt for version {} failed with a retryable exception due to a physical "
+                    + "conflict. Performing conflict resolution and trying again. Error: {}",
                 commitAsVersion,
                 cfe.getMessage());
 
@@ -355,12 +355,8 @@ public class TransactionImpl implements Transaction {
             dataActions = rebaseState.getUpdatedDataActions();
             domainMetadataState.setComputedDomainMetadatas(rebaseState.getUpdatedDomainMetadatas());
             currentCrcInfo = rebaseState.getUpdatedCrcInfo();
-            // Action counters may be partially incremented from previous tries, reset the counters
-            // to 0 and drop fileSizeHistogram
-            // TODO: reconcile fileSizeHistogram.
-            transactionMetrics.resetActionMetricsForRetry();
           } else {
-            // No conflict resolution needed - just retry with same version
+            // No conflict so no conflict resolution needed - just retry with same version
 
             logger.warn(
                 "Commit attempt for version {} failed with a retryable exception due to a "
@@ -368,6 +364,12 @@ public class TransactionImpl implements Transaction {
                 commitAsVersion,
                 cfe.getMessage());
           }
+          // We will be retrying the commit.
+          //
+          // Action counters may be partially incremented from previous tries, reset the counters
+          // to 0 and drop fileSizeHistogram
+          // TODO: reconcile fileSizeHistogram.
+          transactionMetrics.resetActionMetricsForRetry();
           numTries++;
         }
       }
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
@@ -512,12 +512,8 @@ public class TransactionImpl implements Transaction {
               commitAsVersion,
               logPath.toString(),
               attemptCommitInfo,
-              readSnapshot.getVersion() >= 0
-                  ? Optional.of(readSnapshot.getProtocol())
-                  : Optional.empty(),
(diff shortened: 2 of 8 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/DefaultFileSystemManagedTableOnlyCommitter.java
```diff
@@ -39,7 +46,40 @@ public CommitResponse commit(
...
+    try {
+      wrapEngineExceptionThrowsIO(
+          () -> {
+            engine
+                .getJsonHandler()
+                .writeJsonFileAtomically(jsonCommitFile, finalizedActions, false /* overwrite */);
+            return null;
+          },
+          String.format("Write file actions to JSON log file `%s`", jsonCommitFile));
+    } catch (FileAlreadyExistsException e) {
```
scottsand-db, line 62: @sumeet-db and @dhruvarya-db PTAL?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -275,38 +276,82 @@ private long commitWithRetry(
...
             dataActions = rebaseState.getUpdatedDataActions();
             domainMetadataState.setComputedDomainMetadatas(rebaseState.getUpdatedDomainMetadatas());
             currentCrcInfo = rebaseState.getUpdatedCrcInfo();
-            // Action counters may be partially incremented from previous tries, reset the counters
-            // to 0 and drop fileSizeHistogram
-            // TODO: reconcile fileSizeHistogram.
-            transactionMetrics.resetActionMetricsForRetry();
+          } else {
+            // Case 4: No conflict so no conflict resolution needed - just retry with same version.
+
```
scottsand-db: nit: delete whitespace

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -275,38 +276,82 @@ private long commitWithRetry(
...
+          if (!cfe.isRetryable()) {
+            // Case 1: Non-retryable exception. We must throw this.
+            logger.error(
+                "Commit attempt for version {} failed with a non-retryable exception and will not "
+                    + "be retried. Error: {}",
+                commitAsVersion,
+                cfe.getMessage());
+
+            throw new RuntimeException(cfe);
+          } else if (numTries >= maxRetries) {
```
scottsand-db: `>` not `>=`
e.g. if numTries is 1 and maxRetries is one, we should continue

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -275,38 +276,82 @@ private long commitWithRetry(
...
-        logger.info("Committing transaction as version = {}.", commitAsVersion);
+      while (true) {
+        // This loop exits upon either (a) commit success (return statement) or (b) commit failure,
+        // if the number of retries has been exhausted or the commit is not retryable.
+
+        logger.info(
+            "Attempting to commit transaction at version {}. Attempt {}/{}",
+            commitAsVersion,
+            numTries + 1,
+            maxRetries);
```
scottsand-db: maxRetries + 1

kernel/kernel-api/src/main/java/io/delta/kernel/exceptions/ConcurrentWriteException.java
```diff
@@ -24,13 +24,19 @@
  */
 @Evolving
 public class ConcurrentWriteException extends KernelException {
+  static final String msg =
+      "Transaction has encountered a conflict and can not be committed. "
```
emkornfield:
```suggestion
      "Transaction has encountered a conflict and cannot be committed. "
```
nit

kernel/kernel-api/src/main/java/io/delta/kernel/exceptions/ConcurrentWriteException.java
```diff
@@ -24,13 +24,19 @@
  */
 @Evolving
 public class ConcurrentWriteException extends KernelException {
+  static final String msg =
```
emkornfield: nit: 
```suggestion
  static final String ERROR_MESSAGE =
```
or at least all caps to indicate a constant?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -274,52 +277,95 @@ private long commitWithRetry(
...
               engine, commitAsVersion, attemptCommitInfo, dataActions, transactionMetrics);
-        } catch (FileAlreadyExistsException fnfe) {
-          logger.info(
-              "Concurrent write detected when committing as version = {}.", commitAsVersion);
-          if (numTries < maxRetries) {
-            // only try and resolve conflicts if we're going to retry
+        } catch (CommitFailedException cfe) {
+          if (!cfe.isRetryable()) {
+            // Case 1: Non-retryable exception. We must throw this.
+            logger.error(
```
emkornfield: It is sometimes considered an anti-pattern to log and rethrow.  Would it make sense to capture the information here into a more structured exception instead of rethrowing a RuntimException?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/DefaultFileSystemManagedTableOnlyCommitter.java
```diff
@@ -39,7 +46,40 @@ public CommitResponse commit(
...
+          true /* conflict */,
+          "Concurrent write detected for version " + commitMetadata.getVersion(),
+          e);
+    } catch (IOException e) {
+      throw new CommitFailedException(
+          true /* retryable */,
+          false /* conflict */,
+          "Failed to write commit file due to I/O error: " + e.getMessage(),
+          e);
+    } catch (Exception e) {
```
allisonport-db: This can be `RuntimeException` right? Are there any other checked exceptions possible here besides IOException?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/DefaultFileSystemManagedTableOnlyCommitter.java
```diff
@@ -39,7 +46,40 @@ public CommitResponse commit(
...
+      throw new CommitFailedException(
+          false /* retryable */,
+          false /* conflict */,
+          "Failed to write commit file: " + e.getMessage(),
+          e);
+    }
+
+    return new CommitResponse(ParsedLogData.forFileStatus(FileStatus.of(jsonCommitFile)));
```
allisonport-db, line 83: Check this return somewhere? I don't think it's in a test correct me if I'm wrong

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -274,52 +277,95 @@ private long commitWithRetry(
...
+            if (cfe.isConflict()) {
+              // Case 2A: We know why this commit failed, so we can throw a more specific exception.
+              throw new ConcurrentWriteException(cfe);
+            } else {
+              // Case 2B: We don't know why this commit failed, so we throw a generic exception.
+              throw new RuntimeException(cfe);
+            }
+          } else if (cfe.isConflict()) {
+            // Case 3: Retryable exception due to a conflict. We will resolve the conflict and retry
+            logger.warn(
```
emkornfield, line 344: this seems like a better case for logging, where a user would reasonably want to know about retries. which could cause issues.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -274,52 +277,95 @@ private long commitWithRetry(
...
           transactionMetrics.commitAttemptsCounter.increment();
           return doCommit(
               engine, commitAsVersion, attemptCommitInfo, dataActions, transactionMetrics);
-        } catch (FileAlreadyExistsException fnfe) {
-          logger.info(
-              "Concurrent write detected when committing as version = {}.", commitAsVersion);
-          if (numTries < maxRetries) {
-            // only try and resolve conflicts if we're going to retry
+        } catch (CommitFailedException cfe) {
+          if (!cfe.isRetryable()) {
```
emkornfield, line 330: wonder if it would make it easier to read if this was a helper function?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -274,52 +277,95 @@ private long commitWithRetry(
...
             domainMetadataState.setComputedDomainMetadatas(rebaseState.getUpdatedDomainMetadatas());
             currentCrcInfo = rebaseState.getUpdatedCrcInfo();
-            // Action counters may be partially incremented from previous tries, reset the counters
-            // to 0 and drop fileSizeHistogram
-            // TODO: reconcile fileSizeHistogram.
-            transactionMetrics.resetActionMetricsForRetry();
+          } else {
+            // Case 4: No conflict so no conflict resolution needed - just retry with same version.
+            logger.warn(
+                "Commit attempt for version {} failed with a retryable exception due to a "
```
emkornfield, line 359: Just to make sure I understand, the retry might encounter a conflict and make a retry?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -409,10 +455,25 @@ private long doCommit(
...
-          toCloseableIterator(metadataActions.iterator()).combine(completeFileActionIter);
+          toCloseableIterator(metadataActions.iterator())
+              .combine(completeFileActionIter)
+              .map(
+                  action -> {
+                    incrementMetricsForFileActionRow(transactionMetrics, action);
+                    if (!action.isNullAt(REMOVE_FILE_ORDINAL)) {
+                      RemoveFile removeFile = new RemoveFile(action.getStruct(REMOVE_FILE_ORDINAL));
+                      if (isAppendOnlyTable && removeFile.getDataChange()) {
+                        throw DeltaErrors.cannotModifyAppendOnlyTable(dataPath.toString());
```
emkornfield, line 492: not sure what this method throws, but it seems more accurate to say can't update or delete an append only table?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -424,37 +485,27 @@ private long doCommit(
...
-            return null;
-          },
-          "Write file actions to JSON log file `%s`",
-          FileNames.deltaFile(logPath, commitAsVersion));
+      final CommitMetadata commitMetadata =
+          new CommitMetadata(
+              commitAsVersion,
+              logPath.toString(),
+              attemptCommitInfo,
+              readSnapshot.getVersion() >= 0
```
emkornfield: We need this logic because before version zero Protocol would not be valid?  

Should these be pushed to methods on readSnapshot, instead of the ternary here?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -424,37 +485,27 @@ private long doCommit(
...
+                  ? Optional.of(readSnapshot.getProtocol())
+                  : Optional.empty(),
+              readSnapshot.getVersion() >= 0
+                  ? Optional.of(readSnapshot.getMetadata())
+                  : Optional.empty(),
+              shouldUpdateProtocol ? Optional.of(protocol) : Optional.empty(),
+              shouldUpdateMetadata ? Optional.of(metadata) : Optional.empty());
+
+      // May throw CommitFailedException
+      DefaultFileSystemManagedTableOnlyCommitter.INSTANCE.commit(
```
emkornfield, line 519: eventually this would be refactored to about the default filesystem?

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/commit/DefaultCommitterSuite.scala
```diff
@@ -73,4 +87,86 @@ class DefaultCommitterSuite extends AnyFunSuite
...
+      description = "FileAlreadyExistsException",
+      thrownException = new FileAlreadyExistsException("_delta_log/001.json"),
+      expectedRetryable = true,
+      expectedConflict = true,
+      expectedMessageFragment = "Concurrent write detected for version 1",
+      expectedCauseType = classOf[FileAlreadyExistsException]),
+    ExceptionTestCase(
+      description = "IOException",
+      thrownException = new IOException("Network timeout writing to _delta_log/001.json"),
+      expectedRetryable = true,
```
emkornfield, line 113: I'm not clear on what this is retryable.  We assume atomicity of operations here?  Does that generally hold?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/DefaultFileSystemManagedTableOnlyCommitter.java
scottsand-db: Sure!

For my own learning, it wasn't obvious to me why RuntimeException was better than Exception btw. Here's what claude taught me, in case this wasn't obvious to you either :D 

> Since the code already catches:
FileAlreadyExistsException (specific IOException subclass)
IOException (all other IOExceptions)
The only exceptions that could reach the catch (Exception e) block are unchecked exceptions - specifically RuntimeException and its subclasses (like NullPointerException, IllegalArgumentException, etc.).
So yes, this could be changed to:
However, there's a subtle consideration: catching Exception also catches Error subclasses (like OutOfMemoryError). Whether you want to catch and wrap Errors depends on your error handling philosophy. Many consider it bad practice to catch Errors as they typically indicate serious JVM-level problems that shouldn't be handled by application code.
Recommendation: Change it to catch (RuntimeException e) unless there's a specific requirement to handle Errors, which seems unlikely given this is a commit operation where you'd want errors to bubble up.

allisonport-db: I thought I was learning something but... catching exception does not catch errors in Java.. https://www.baeldung.com/java-error-catch

My reasoning for RuntimeException > Exception -- more specificity is better when the broader one is not possible

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
allisonport-db: This should look a lot better after rebasing on https://github.com/delta-io/delta/pull/5012

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWritesSuite.scala
```diff
@@ -1237,4 +1241,197 @@ class DeltaTableWritesSuite extends DeltaTableWriteSuiteBase with ParquetSuiteBa
...
+        override def getJsonHandler: JsonHandler = new TransientErrorJsonHandler()
+      }
+
+      val transientErrorEngine = new TransientErrorEngine()
+
+      val txn = createWriteTxnBuilder(table).withMaxRetries(maxRetries).build(transientErrorEngine)
+
+      val result = commitTransaction(txn, transientErrorEngine, emptyIterable())
+
+      assert(result.getVersion > 0)
```
allisonport-db: why > 0? and not == 1?

scottsand-db: this can be 1, you're right!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
scottsand-db: Hey Micah! Not sure I understand that question, sorry -- can you elaborate what the default filesystem is to you?

I will elaborate on this area of code more broadly though: eventually this will be refactored to just invoke the `committer` that is part of the readSnapshot 👍 it might be the default committer, it might. be a UC committer, etc.

emkornfield: sorry, yeah my comment wasn't clear.  Your response answered by question.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -274,52 +277,95 @@ private long commitWithRetry(
...
+        } catch (CommitFailedException cfe) {
+          if (!cfe.isRetryable()) {
+            // Case 1: Non-retryable exception. We must throw this.
+            logger.error(
+                "Commit attempt for version {} failed with a non-retryable exception and will not "
+                    + "be retried. Error: {}",
+                commitAsVersion,
+                cfe);
+
+            throw new RuntimeException(cfe);
```
scottsand-db: idea: NonRetryableCommitFailedException

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -274,52 +277,95 @@ private long commitWithRetry(
...
+            // retries. Will have to throw here, too.
+            logger.error(
+                "Commit attempt for version {} failed with a retryable exception but will not be "
+                    + "retried because the maximum number of retries ({}) has been exhausted. "
+                    + "Error: {}",
+                commitAsVersion,
+                maxRetries,
+                cfe);
+
+            if (cfe.isConflict()) {
```
scottsand-db: idea: MaxRetriesReachedConcurrentException
idea: MaxRetriesReachedException // e.g. io failure, not a concurrent write

preference: just throw MaxRetriedReachedException

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/DefaultFileSystemManagedTableOnlyCommitter.java
scottsand-db: idea: consider not catching this at all?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
allisonport-db: I'm good with MaxRetriedReachedException with a cause

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
allisonport-db: I feel like this one can just be a RuntimeException? But also if we change the commit method to not catch RuntimeExceptions -- we would never hit this in reality? which brings up the question is should we have a `isRetryable=false` case for CommitFailedException (not sure how that plays into the ccv2 design)

allisonport-db: Also good to address this in a separate PR

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
emkornfield: It seems more natural to propagate the last exception here (or I guess we should consider what we think callers would do on catching).

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
scottsand-db: Hi @emkornfield -- Great question. I'm aiming for a balance of best-design-and-implementation here, but also want to ensure velocity quite frankly. We have future plans to open up our transaction/commit APIs and to give the connector the control of catching these sorts of CommitFailedExceptions. We can improve a lot of things then.

Anyways, for now, to recap: connectors invoke `txn.commit`, `txn.commit` invoke the committer, and `txn.commit` will also catch exceptions and do the retries conflict detection/rebasing as needed.

So, I don't want to propagate a checked `CommitFailedException` `cfe` (this is what the committer throws) to the connector from this `txn.commit` API. Kernel should catch `CommitFailedException` and retry when applicable.

The two interesting cases are (a) when the exception is not retryable, and (b) when the exception is retryable but we have reached our `maxRetries` limit.

In case (a) -- I'm inclined to just wrap the `cfe` in a RuntimeException for now -- we don't expect engines to be able to recover from this.

In case (b) -- I proposed below to throw a MaxRetriesReachedException.

Does this sound like a decent plan for now? Happy to iterate and improve further in the future? But also want to move fast now to meet some deadlines.

scottsand-db: An alternative is an explicit `NonRetryableCommitFailedException`. We can iterate on that later perhaps?

scottsand-db: > It is sometimes considered an anti-pattern to log and rethrow. 

Implemented!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
scottsand-db: Yup! Any commit attempt might experience a conflict:

Imagine you are trying to do a blind append into a source table that has many concurrent appenders.

- read snapshot version: 10
- first commit attempt: we try to write 11.json
- conflict! there's actually already conflicts with 11.json to 15.json
- we detect any logical conflicts (none) and rebase
- second commit attempt: we try to write 16.json
- conflict! there's actually already conflicts with 16.json to 20.json
- we detect any logical conflicts (none) and rebase
- pattern continues ...

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
scottsand-db: agreed! will take care of that in another PR

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
emkornfield: Or actually probably more accurately we probably don't want to log here either but provide callbacks for handling retries.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
emkornfield: > Anyways, for now, to recap: connectors invoke txn.commit, txn.commit invoke the committer, and txn.commit will also catch exceptions and do the retries conflict detection/rebasing as needed.

Secondary question, was putting retries a into the Kernel vs Engine a conscious choice?  What are the trade-offs of leaving the loop to engine?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -310,52 +314,80 @@ private long commitWithRetry(
...
+            //         be able to recover from this.
+            throw new RuntimeException(
+                String.format(
+                    "Commit attempt for version %d failed with a non-retryable exception.",
+                    commitAsVersion),
+                cfe);
+          } else if (attempt >= maxCommitAttempts) {
+            // Case 2: Despite the error being retryable, we have exhausted the maximum number of
+            //         retries. We must throw here, too.
+            throw new MaxCommitRetriesReachedException(commitAsVersion, maxRetries, cfe);
```
emkornfield, line 341: if we go with this approach I think we need update user facing docs (i.e. on Transaction) as to expected exceptions that the connector would have to retry?

scottsand-db: Today, `txn.commit` is in charge of retries. We are actually _not_ expecting the Engine to retry. That's why we don't explicitly throw any checked exceptions to make the connector really think about this. Either the `txn.commit` succeeds (performing all the retries and rebasing as needed) or it doesn't.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/commit/DefaultCommitterSuite.scala
scottsand-db: If there's an intermittent or transient network exception, why can't you retry? the file `00N.json` doesn't exist yet?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
scottsand-db: Great question. I don't think at the time of the decision we knew about the benefits / the requirement of giving more control to engines for the commit. We now are aware of that requirement. I've done lots of design work to enable this. This is what the [CommitContext](https://github.com/delta-io/delta/blob/master/kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitContext.java) and [TransactionV2](https://github.com/delta-io/delta/blob/master/kernel/kernel-api/src/main/java/io/delta/kernel/transaction/TransactionV2.java) will enable.

However, for now, I cannot prioritize that development, in order to meet other release timelines. We will pick up that effort in the future.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
scottsand-db: As mentioned above -- we don't expect callers to do anything on catching. txn.commit is in charge of retries.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/commit/DefaultCommitterSuite.scala
emkornfield: How do you know if there is an intermittent or transient network that the file 00N.json doesn't exist?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
emkornfield: OK, this doesn't need to block this PR but it would be good to understand this choice in the context of how engines implementation isolation levels.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/commit/DefaultCommitterSuite.scala
scottsand-db: Fair point. It might exist, but it also might not exist yet. Isn't a re-try worthwhile though? We would expect a `FileAlreadyExistsException` if there is no transient error and the file does in fact already exist.

scottsand-db: This is the default behavior in https://github.com/delta-io/delta/blob/28548fde193a2f4d644b7ad6466dae2781968b42/storage/src/main/java/io/delta/storage/commit/uccommitcoordinator/UCTokenBasedRestClient.java#L230

emkornfield: > Fair point. It might exist, but it also might not exist yet. Isn't a re-try worthwhile though? We would expect a FileAlreadyExistsException if there is no transient error and the file does in fact already exist.

I might be misunderstanding control flow, but if the commit is blindly retried won't the `FileAlreadyExistsException` get converted into a [conflict](https://github.com/delta-io/delta/pull/5010/files#r2249025285)?  In this case would the rebase detect the successful commit (i.e. do we always have a transaction identifier to make it idempotent).

For the link to existing behavior, I'm not clear on the overall workflow, but it seems like it could be problematic also (at least it would be nice to guarantee idempotence, so if the commit actually succeeded but the service reported a failure it would still succeed.  (E.g. a 500 error or a timeout might of actually had the transaction be succesful, even if the the client doesn't acknowledge it).

scottsand-db: Ah, I see your point, Micah. Thanks for clarifying that and spotting this issue.

- we commit 007.json
- the json write succeeds, but we on the client are not made aware of it
- if we try to write 007.json again, we will see a FileAlreadyExistsException, and we will try to rebase and then write 008.json. This is bad/wrong.

In fact, the UC-Commit-Coordinator-Client (for delta-spark) handles this issue: https://github.com/delta-io/delta/blob/28548fde193a2f4d644b7ad6466dae2781968b42/storage/src/main/java/io/delta/storage/commit/uccommitcoordinator/UCCommitCoordinatorClient.java#L440

https://github.com/delta-io/delta/blob/28548fde193a2f4d644b7ad6466dae2781968b42/storage/src/main/java/io/delta/storage/commit/uccommitcoordinator/UCCommitCoordinatorClient.java#L708

I'll have to think on this 👍

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
emkornfield: Seems like this was covered above, and is orthogonal for now.

scottsand-db:
Abandoning this PR in favor of a simpler PR that adds the default committer only https://github.com/delta-io/delta/pull/5040

PR #5012, opened 2025-08-01
[WIP][KERNEL] Refactor `TransactionImpl` to make readSnapshot optional

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Currently readSnapshot is a required input to TransactionImpl. For new tables, we create a fake empty snapshot to fulfill this requirement with version = -1 and a fake empty logsegment. This propagates bad semantics to other functions within TransactionImpl because they rely on this concept of an "empty snapshot" instead of correctly coding to the concept of no existing snapshot for the case of a new table.

This also leaves questionable results for a lot of the methods on SnapshotImpl for this fake empty snapshot like...
- what should getTimestamp return? (currently -1)
- empty log segment
- getSchema/partitionColumns (currently we initialize this to be the ones created in the new table but this doesn't really make sense since the snapshot doesn't exist)
- getScanBuilder (?!)
- getSnapshotReport
- etc, etc, etc

This refactor lets us get rid of this "non-existent snapshot" throughout our code.

## How was this patch tested?

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
@@ -262,7 +262,7 @@ public class TransactionBuilderImpl implements TransactionBuilder {
           false, // isCreateOrReplace
           table.getDataPath(),
           table.getLogPath(),
-          latestSnapshot.get(),
+          latestSnapshot,
           engineInfo,
           operation,
           latestSnapshot.get().getProtocol(), // reuse latest protocol
@@ -327,14 +327,6 @@ public class TransactionBuilderImpl implements TransactionBuilder {
     Optional<Protocol> newProtocol = updatedProtocolAndMetadata._1;
     Optional<Metadata> newMetadata = updatedProtocolAndMetadata._2;
 
-    if (!latestSnapshot.isPresent()) {
-      // For now, we generate an empty snapshot (with version -1) for a new table. In the future,
-      // we should define an internal interface to expose just the information the transaction
-      // needs instead of the entire SnapshotImpl class. This should also let us avoid creating
-      // this fake empty initial snapshot.
-      latestSnapshot = Optional.of(getInitialEmptySnapshot(engine, baseMetadata, baseProtocol));
-    }
-
     // Block this for now - in a future PR we will enable this
     if (operation == Operation.REPLACE_TABLE) {
       if (newProtocol
@@ -351,7 +343,7 @@ public class TransactionBuilderImpl implements TransactionBuilder {
         isCreateOrReplace,
         table.getDataPath(),
         table.getLogPath(),
-        latestSnapshot.get(),
+        latestSnapshot,
         engineInfo,
         operation,
         newProtocol.orElse(baseProtocol),
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
@@ -21,6 +21,7 @@ import static io.delta.kernel.internal.actions.SingleAction.*;
 import static io.delta.kernel.internal.util.Preconditions.checkArgument;
 import static io.delta.kernel.internal.util.Preconditions.checkState;
 import static io.delta.kernel.internal.util.Utils.toCloseableIterator;
+import static java.util.Collections.emptyMap;
 
 import io.delta.kernel.*;
 import io.delta.kernel.data.Row;
@@ -79,10 +80,11 @@ public class TransactionImpl implements Transaction {
   private final boolean isCreateOrReplace;
   private final String engineInfo;
   private final Operation operation;
-  private final Path dataPath;
-  private final Path logPath;
+  public final Path dataPath;
+  // TODO make method
+  public final Path logPath;
   private final Protocol protocol;
-  private final SnapshotImpl readSnapshot;
+  private final Optional<SnapshotImpl> readSnapshot;
   private final Optional<SetTransaction> setTxnOpt;
   private final Optional<List<Column>> clusteringColumnsOpt;
   private final boolean shouldUpdateProtocol;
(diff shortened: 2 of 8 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -327,14 +327,6 @@ protected TransactionImpl buildTransactionInternal(
     Optional<Protocol> newProtocol = updatedProtocolAndMetadata._1;
     Optional<Metadata> newMetadata = updatedProtocolAndMetadata._2;
 
-    if (!latestSnapshot.isPresent()) {
-      // For now, we generate an empty snapshot (with version -1) for a new table. In the future,
-      // we should define an internal interface to expose just the information the transaction
-      // needs instead of the entire SnapshotImpl class. This should also let us avoid creating
-      // this fake empty initial snapshot.
-      latestSnapshot = Optional.of(getInitialEmptySnapshot(engine, baseMetadata, baseProtocol));
```
scottsand-db, line 336: can we delete the initial snapshot code now?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
```diff
@@ -368,7 +368,9 @@ private void handleTxn(ColumnVector txnVector) {
   }
 
   private List<FileStatus> getWinningCommitFiles(Engine engine) {
-    String firstWinningCommitFile = deltaFile(snapshot.getLogPath(), snapshot.getVersion() + 1);
+    // TODO doesn't current impl mean we always try to resolve conflicts from readSnapshot version
+    //  again? instead of iteratively on only the newest commit
+    String firstWinningCommitFile = deltaFile(transaction.getLogPath(), attemptVersion);
```
allisonport-db: Maybe just change this to `readSnapshot.map(SnapshotImpl::getVersion).orElse(-1) + 1)` for now (no change in behavior) and revisit this question separately?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
allisonport-db: oh yeah.. I meant to :)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
scottsand-db: good catch!
- attempt 1: we want to write `10.json`. We conflict and see that `10.json` to `15.json` exist. We must check against 10.json to 15.json
- attempt 2: we want to write `16.json`. We conflict and see that `16.json` to `20.json` exist. We must check against `16` to `20`. We should NOT check against 10 to 15, too

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/RowTracking.java
```diff
@@ -126,7 +126,10 @@ public void close() throws IOException {
       public CloseableIterator<Row> iterator() {
         // The row ID high watermark from the snapshot of the table that this transaction is reading
         // at. Any baseRowIds higher than this watermark are assigned by this transaction.
-        final long prevRowIdHighWatermark = readRowIdHighWaterMark(txnReadSnapshot);
+        final long prevRowIdHighWatermark =
+            txnReadSnapshot
+                .map(RowTracking::readRowIdHighWaterMark)
```
scottsand-db: Could we update `readRowIdHighWaterMark` to take in an `Optional<Snapshot>` instead? And then avoid the duplicate logic in these two invoker methods?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -82,7 +83,7 @@ public class TransactionImpl implements Transaction {
   private final Path dataPath;
   private final Path logPath;
   private final Protocol protocol;
-  private final SnapshotImpl readSnapshot;
+  private final Optional<SnapshotImpl> readSnapshot;
```
scottsand-db: nit: could we just pull the trigger and rename this to `readSnapshotOpt`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
```diff
@@ -368,7 +368,9 @@ private void handleTxn(ColumnVector txnVector) {
   }
 
   private List<FileStatus> getWinningCommitFiles(Engine engine) {
-    String firstWinningCommitFile = deltaFile(snapshot.getLogPath(), snapshot.getVersion() + 1);
+    // TODO this should be based on attemptVersion not readSnapshot version
```
scottsand-db: Can you please create an issue and link it here?
`// TODO: [delta-io/delta#1234] this should be based on attemptVersion not readSnapshot version`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
allisonport-db: yep good call

kernel/kernel-api/src/main/java/io/delta/kernel/internal/replay/ConflictChecker.java
allisonport-db: Done and added it to TODOs in https://github.com/delta-io/delta/issues/4573

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -201,17 +203,27 @@ public Protocol getProtocol() {
     return protocol;
   }
 
+  public Path getDataPath() {
```
scottsand-db, line 206: nit, not a blocker: would be great to move all of our `get` methods together to be a bit more organized, eh? could you move these 3 getters above? little things like this help a lot over time, thanks! (you can also push back, im ez)

scottsand-db (approved):
LGTM! Thx!

PR #5019, opened 2025-08-04
[KERNEL] Add clustering columns to `SnapshotImpl`

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Currently we have a method in ClusteringUtils that takes in a snapshot and returns the clustering columns. Since this requires parsing the domain metadata we try to avoid doing this more than once and instead pass around a `resolvedClusteringColumns` from txn building to txn implementation. This makes the fields in both much more confusing as we need to differentiate between newClusteringCols (set in the builder for a new or update table) vs existingClusteringCols for an existing table. Instead, we should lazily parse the clustering columns on SnapshotImpl and store it on that class.

## How was this patch tested?

Just a refactor existing tests should suffice.

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java
@@ -76,8 +76,11 @@ public class SnapshotImpl implements Snapshot {
     // io.delta.kernel.metrics.SnapshotMetricsResult#getLoadSnapshotTotalDurationNs}, are only
     // completed *after* the Snapshot has been constructed.
     this.lazySnapshotReport = new Lazy<>(() -> SnapshotReportImpl.forSuccess(snapshotContext));
-    this.lazyClusteringColumns = new Lazy<>(() -> ClusteringMetadataDomain.fromSnapshot(this)
-        .map(ClusteringMetadataDomain::getClusteringColumns));
+    this.lazyClusteringColumns =
+        new Lazy<>(
+            () ->
+                ClusteringMetadataDomain.fromSnapshot(this)
+                    .map(ClusteringMetadataDomain::getClusteringColumns));
   }
 
   /////////////////
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java
@@ -21,11 +21,13 @@ import static io.delta.kernel.internal.TableConfig.TOMBSTONE_RETENTION;
 import io.delta.kernel.ScanBuilder;
 import io.delta.kernel.Snapshot;
 import io.delta.kernel.engine.Engine;
+import io.delta.kernel.expressions.Column;
 import io.delta.kernel.internal.actions.CommitInfo;
 import io.delta.kernel.internal.actions.DomainMetadata;
 import io.delta.kernel.internal.actions.Metadata;
 import io.delta.kernel.internal.actions.Protocol;
 import io.delta.kernel.internal.checksum.CRCInfo;
+import io.delta.kernel.internal.clustering.ClusteringMetadataDomain;
 import io.delta.kernel.internal.fs.Path;
 import io.delta.kernel.internal.lang.Lazy;
 import io.delta.kernel.internal.metrics.SnapshotQueryContext;
@@ -51,6 +53,7 @@ public class SnapshotImpl implements Snapshot {
   private final LogSegment logSegment;
   private Optional<Long> inCommitTimestampOpt;
   private Lazy<SnapshotReport> lazySnapshotReport;
+  private Lazy<Optional<List<Column>>> lazyClusteringColumns;
 
   public SnapshotImpl(
       Path dataPath,
@@ -73,6 +76,8 @@ public class SnapshotImpl implements Snapshot {
     // io.delta.kernel.metrics.SnapshotMetricsResult#getLoadSnapshotTotalDurationNs}, are only
     // completed *after* the Snapshot has been constructed.
     this.lazySnapshotReport = new Lazy<>(() -> SnapshotReportImpl.forSuccess(snapshotContext));
+    this.lazyClusteringColumns = new Lazy<>(() -> ClusteringMetadataDomain.fromSnapshot(this)
+        .map(ClusteringMetadataDomain::getClusteringColumns));
   }
 
   /////////////////
@@ -153,6 +158,11 @@ public class SnapshotImpl implements Snapshot {
 
   public SnapshotReport getSnapshotReport() {
     return lazySnapshotReport.get();
+  }
+
+  // TODO: put this on the public Snapshot API? (need to convert to logical?)
+  public Optional<List<Column>> getClusteringColumns() {
+    return lazyClusteringColumns.get();
   }
 
   /**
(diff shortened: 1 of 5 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -79,12 +78,10 @@ public class TransactionBuilderImpl implements TransactionBuilder {
   // This represents logical column references before schema resolution is applied.
   // (e.g., case sensitivity, column mapping)
   private Optional<List<Column>> initialClusteringColumns = Optional.empty();
-
   // The resolved clustering columns that will be written into domain metadata. This reflects
   // case-preserved column names or physical column names if column mapping is enabled.
   // This would only be set after schema resolution and must align with the resolved schema.
-  private Optional<List<Column>> resolvedClusteringColumns = Optional.empty();
-  private boolean shouldUpdateClusteringDomainMetadata = false;
+  private Optional<List<Column>> newResolvedClusteringColumns = Optional.empty();
```
allisonport-db, line 78: `newResolvedClusteringColumns` + `initialClusteringColumns` is still not ideal but this should be further improved in my future main refactor PR. I also think this is still a lot better as the semantics are much clearer (newResolvedClusteringColumns only ever is initialClusteringColumns, not sometimes the new ones and sometimes the existing ones)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -201,6 +198,16 @@ public Protocol getProtocol() {
     return protocol;
   }
 
+  public Optional<List<Column>> getEffectiveClusteringColumns() {
+    if (isCreateOrReplace) {
+      return newClusteringColumnsOpt;
+    } else {
+      return newClusteringColumnsOpt.isPresent()
+          ? newClusteringColumnsOpt
+          : readSnapshot.getClusteringColumns();
```
allisonport-db: will need to be rebased on https://github.com/delta-io/delta/pull/5012

kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java
```diff
@@ -169,6 +177,11 @@ public SnapshotReport getSnapshotReport() {
     return lazySnapshotReport.get();
   }
 
+  /** @returns the physical clustering columns in this snapshot */
```
scottsand-db: are clustering columns always phsical? would `getPhysicalClusteringColumns` be a more precise and less error-prone API? just to help enforce/remind consumers of the semantic?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java
```diff
@@ -169,6 +177,11 @@ public SnapshotReport getSnapshotReport() {
...
+  public Optional<List<Column>> getClusteringColumns() {
```
scottsand-db: also: update javadoc to specify what:
- optional.empty means
- optional.of(empty list)
- optional.of(non-empty-list)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -84,9 +84,8 @@ public class TransactionImpl implements Transaction {
   private final Protocol protocol;
   private final SnapshotImpl readSnapshot;
   private final Optional<SetTransaction> setTxnOpt;
-  private final Optional<List<Column>> clusteringColumnsOpt;
+  private final Optional<List<Column>> newClusteringColumnsOpt;
```
scottsand-db, line 99: Can we again here add javadocs `/** */` to this variable indicating what the different states mean?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
scottsand-db: yup, a step in the right direction!

ideally, i think, `newResolvedClusteringColumns` shouldn't be a member variable of this class. e.g. it seems like `validateAndUpdateProtocolAndMetadata` just updates it, and then some other place in the code consumes it ... having this be a member variable is confusing; it should just be in the return type of `validateAndUpdateProtocolAndMetadata`

I'm sure you know all this :) i'm looking forward to your refactor :)

None-the-less, and sorry to be a stickler for documentation, but can we still add a bit more context here of how this variable is used? in the javadocs for this variable ...

Also, would a better name for `initialClusteringColumns` be `inputLogicalClusteringColumns`? e.g. this is the INPUT we are getting from the connector, and they are the LOGICAL names? I think that would be clearer than "initial" -- e.g. couldn't initial be interpreted as the clustering columns on the read snapshot?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/clustering/ClusteringUtils.java
```diff
@@ -37,13 +35,4 @@ public static DomainMetadata getClusteringDomainMetadata(List<Column> clustering
         ClusteringMetadataDomain.fromClusteringColumns(clusteringColumns);
     return clusteringMetadataDomain.toDomainMetadata();
   }
-
-  /**
-   * Extract ClusteringColumns from a given snapshot. Return None if the clustering domain metadata
-   * is missing.
-   */
-  public static Optional<List<Column>> getClusteringColumnsOptional(SnapshotImpl snapshot) {
```
scottsand-db, line 45: yup, love to see this deleted! This is weird C style struct code ... why would we invoke a method **on** a snapshot ... when we could just have all the snapshot APIs in one place and invoke a method **from** a snapshot. this makes discovering the capabilities and information we can extract from a snapshot much easier :) also this let's us cache the result

kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java
allisonport-db: yeah I think that's better

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
allisonport-db: Made a best effort to update the docs, but this var shouldn't be relevant after the future refactor. Also renamed the other variable to be more clear.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -203,6 +212,18 @@ public Protocol getProtocol() {
     return protocol;
   }
 
+  public Optional<List<Column>> getEffectiveClusteringColumns() {
```
scottsand-db, line 215: nit: is the below a bit cleaner? wdyt?

```java
public Optional<List<Column>> getEffectiveClusteringColumns() {
  if (isCreateOrReplace) {
    return newClusteringColumnsOpt;
  } else if (newClusteringColumnsOpt.isPresent()) {
    return newClusteringColumnsOpt;
  } else {
    // we know if !isCreateOrReplace then readSnapshotOpt must be present
    return readSnapshotOpt.flatMap(SnapshotImpl::getPhysicalClusteringColumns);
  }
}
```

scottsand-db (approved):
LGTM with 1 suggestion that is not a blocker!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
allisonport-db: Did something similar that I prefer a bit more

PR #5022, opened 2025-08-04
[SPARK] Block Time Travel beyond DeletedFileRetentionDuration

rajeshparangi:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
This change blocks time travel beyond deletedFileRetentionDuration. The affected commands are SELECT, RESTORE, CLONE and CDC which use the syntax AS OF.

## How was this patch tested?
Added new tests
Fixed existing tests to be compatible with the changes proposed.

Yes. Previously, user could time travel to any delta version beyond deletedFileRetentionDuration. With this change, time travel is restricted to be within the deletedFileRetentionDuration.

### DIFF
--- a/sharing/src/test/scala/io/delta/sharing/spark/DeltaSharingDataSourceDeltaSuite.scala
+++ b/sharing/src/test/scala/io/delta/sharing/spark/DeltaSharingDataSourceDeltaSuite.scala
@@ -16,6 +16,11 @@
 
 package io.delta.sharing.spark
 
+
+// scalastyle:off import.ordering.noEmptyLine
+
+import scala.concurrent.duration._
+
 import org.apache.spark.sql.delta.DeltaExcludedBySparkVersionTestMixinShims
 import org.apache.spark.sql.delta.sources.DeltaSQLConf
 import org.apache.spark.sql.delta.test.DeltaSQLCommandTest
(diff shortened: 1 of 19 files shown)


dhruvarya-db (approved):
LGTM

felipepessoto:
@rajeshparangi @dhruvarya-db @huan233usc could you help understand the motivation of this change?

not sure if I understand it correctly, does it mean it won’t be possible to time travel to versions older than 7 days by default?

rajeshparangi:
> @rajeshparangi @dhruvarya-db @huan233usc could you help understand the motivation of this change?
> 
> not sure if I understand it correctly, does it mean it won’t be possible to time travel to versions older than 7 days by default?

Yes, you are right. This is the motivation for that.

Today, customers can time-travel beyond Vacuum’s retention window if Vacuum hasn’t yet cleaned up the affected versions. This means it is possible to time-travel up to 30 days (default for delta.logRetentionDuration) even though the Vacuum retention window is 7 days (default for delta.deletedFileRetentionDuration). As mentioned earlier, this works only if the Vacuum hasn’t yet cleaned up those versions. If Vacuum has been run, the time-travel query may fail trying to access a file which may have already been deleted.

The proposal is to get rid of this uncanny dependency between Vacuum and time-travel and block all time-travel queries beyond Vacuum’s retention window. This ensures that if someone wants to preserve the ability to time-travel beyond Vacuum’s retention window, they need to appropriately set Vacuum retention window as well.

PR #5023, opened 2025-08-05
[Delta] Fix a couple of vacuum related tests to use the right clock

EstherBear:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This change fixes a couple of Vacuum related tests to use the right clock throughout the test logic.

## How was this patch tested?

Existing UTs.

No


xzhseh (approved):
LGTM, thanks for the fix!

PR #5027, opened 2025-08-05
[Kernel] Draft for a new Metadata column API (option 2)

lbhm:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This PR is a second draft for a new metadata column API based on offline discussions.

For option 1 see #4996. Instead of a new `DataType`, this PR introduces a `MetadataColumn` enum that is attached to a `StructField` via a special `FieldMetadata` key.

## How was this patch tested?

Existing tests.

Yes.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
@@ -32,6 +32,7 @@ import io.delta.kernel.internal.rowtracking.MaterializedRowTrackingColumn;
 import io.delta.kernel.internal.util.ColumnMapping.ColumnMappingMode;
 import io.delta.kernel.internal.util.PartitionUtils;
 import io.delta.kernel.internal.util.Tuple2;
+import io.delta.kernel.types.MetadataColumnType;
 import io.delta.kernel.types.StructField;
 import io.delta.kernel.types.StructType;
 import io.delta.kernel.utils.CloseableIterator;
@@ -194,10 +195,11 @@ public interface Scan {
         // If row tracking is enabled, check for physical row tracking columns in the data batch
         // and transform them to logical row tracking columns as needed
         Map<String, String> configuration = ScanStateRow.getConfiguration(scanState);
+        StructType logicalSchema = ScanStateRow.getLogicalSchema(scanState);
         if (TableConfig.ROW_TRACKING_ENABLED.fromMetadata(configuration)) {
           nextDataBatch =
               MaterializedRowTrackingColumn.transformPhysicalData(
-                  nextDataBatch, scanFile, configuration, engine);
+                  nextDataBatch, scanFile, logicalSchema, configuration, engine);
         }
 
         // Get the selectionVector if DV is present
@@ -207,8 +209,7 @@ public interface Scan {
         if (dv == null) {
           selectionVector = Optional.empty();
         } else {
-          int rowIndexOrdinal =
-              nextDataBatch.getSchema().indexOf(StructField.METADATA_ROW_INDEX_COLUMN_NAME);
+          int rowIndexOrdinal = nextDataBatch.getSchema().indexOf(MetadataColumnType.ROW_INDEX);
           if (rowIndexOrdinal == -1) {
             throw new IllegalArgumentException(
                 "Row index column is not present in the data read from the Parquet file.");
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
@@ -43,6 +43,7 @@ import io.delta.kernel.internal.util.*;
 import io.delta.kernel.metrics.ScanReport;
 import io.delta.kernel.metrics.SnapshotReport;
 import io.delta.kernel.types.FieldMetadata;
+import io.delta.kernel.types.MetadataColumnType;
 import io.delta.kernel.types.StructField;
 import io.delta.kernel.types.StructType;
 import io.delta.kernel.utils.CloseableIterator;
@@ -266,10 +267,13 @@ public class ScanImpl implements Scan {
 
     if (protocol.getReaderFeatures().contains("deletionVectors")
         && physicalFields.stream()
-            .map(StructField::getName)
-            .noneMatch(name -> name.equals(StructField.METADATA_ROW_INDEX_COLUMN_NAME))) {
+            .map(StructField::getMetadataColumnType)
+            .noneMatch(type -> type.equals(MetadataColumnType.ROW_INDEX))) {
       // If the row index column is not already present, add it to the physical read schema
-      physicalFields.add(createInternalColumn(StructField.METADATA_ROW_INDEX_COLUMN));
+      physicalFields.add(
+          createInternalColumn(
+              StructField.createMetadataColumn(
+                  StructField.DEFAULT_ROW_INDEX_COLUMN_NAME, MetadataColumnType.ROW_INDEX)));
     }
 
     return new StructType(physicalFields);
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/Metadata.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/Metadata.java
@@ -104,6 +104,8 @@ public class Metadata {
       ArrayValue partitionColumns,
       Optional<Long> createdTime,
       MapValue configurationMapValue) {
+    ensureNoMetadataColumns(schema);
+
     this.id = requireNonNull(id, "id is null");
     this.name = name;
     this.description = requireNonNull(description, "description is null");
@@ -347,5 +349,15 @@ public class Metadata {
       partitionColumnNames.add(partitionColName.toLowerCase(Locale.ROOT));
     }
     return Collections.unmodifiableSet(partitionColumnNames);
+  }
+
+  /** Helper method to ensure that a table schema never contains metadata columns. */
+  private static void ensureNoMetadataColumns(StructType schema) {
+    for (StructField field : schema.fields()) {
+      if (field.isMetadataColumn()) {
+        throw new IllegalArgumentException(
+            "Table schema cannot contain metadata columns: " + field.getName());
+      }
+    }
   }
 }
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
@@ -39,12 +39,12 @@ import java.util.stream.Stream;
 public final class MaterializedRowTrackingColumn {
 
   /** Static instance for the materialized row ID column. */
-  public static final MaterializedRowTrackingColumn ROW_ID =
+  public static final MaterializedRowTrackingColumn MATERIALIZED_ROW_ID =
       new MaterializedRowTrackingColumn(
           TableConfig.MATERIALIZED_ROW_ID_COLUMN_NAME, "_row-id-col-");
 
   /** Static instance for the materialized row commit version column. */
-  public static final MaterializedRowTrackingColumn ROW_COMMIT_VERSION =
+  public static final MaterializedRowTrackingColumn MATERIALIZED_ROW_COMMIT_VERSION =
       new MaterializedRowTrackingColumn(
           TableConfig.MATERIALIZED_ROW_COMMIT_VERSION_COLUMN_NAME, "_row-commit-version-col-");
 
@@ -80,7 +80,7 @@ public final class MaterializedRowTrackingColumn {
     Set<String> physicalColNames =
         schema.fields().stream().map(ColumnMapping::getPhysicalName).collect(Collectors.toSet());
 
-    Stream.of(ROW_ID, ROW_COMMIT_VERSION)
+    Stream.of(MATERIALIZED_ROW_ID, MATERIALIZED_ROW_COMMIT_VERSION)
         .map(column -> metadata.getConfiguration().get(column.getMaterializedColumnNameProperty()))
         .filter(Objects::nonNull)
         .forEach(
@@ -105,7 +105,7 @@ public final class MaterializedRowTrackingColumn {
     }
 
     // Check that both materialized column name configs are present when row tracking is enabled
-    Stream.of(ROW_ID, ROW_COMMIT_VERSION)
+    Stream.of(MATERIALIZED_ROW_ID, MATERIALIZED_ROW_COMMIT_VERSION)
         .forEach(
             column -> {
               if (!metadata
@@ -136,7 +136,7 @@ public final class MaterializedRowTrackingColumn {
 
     Map<String, String> configsToAdd = new HashMap<>();
 
-    Stream.of(ROW_ID, ROW_COMMIT_VERSION)
+    Stream.of(MATERIALIZED_ROW_ID, MATERIALIZED_ROW_COMMIT_VERSION)
         .filter(
             column ->
                 !metadata
@@ -175,21 +175,24 @@ public final class MaterializedRowTrackingColumn {
       throw DeltaErrors.missingRowTrackingColumnRequested(logicalField.getName());
     }
 
-    if (logicalField.getName().equals(METADATA_ROW_ID_COLUMN_NAME)) {
+    if (logicalField.getMetadataColumnType() == MetadataColumnType.ROW_ID) {
       List<StructField> physicalFields = new ArrayList<>(2);
       physicalFields.add(
           new StructField(
-              ROW_ID.getPhysicalColumnName(metadata.getConfiguration()),
+              MATERIALIZED_ROW_ID.getPhysicalColumnName(metadata.getConfiguration()),
               LongType.LONG,
               true /* nullable */));
-      if (logicalSchema.indexOf(StructField.METADATA_ROW_INDEX_COLUMN_NAME) == -1) {
-        physicalFields.add(ScanImpl.createInternalColumn(StructField.METADATA_ROW_INDEX_COLUMN));
+      if (!logicalSchema.contains(MetadataColumnType.ROW_INDEX)) {
+        physicalFields.add(
+            ScanImpl.createInternalColumn(
+                StructField.createMetadataColumn(
+                    StructField.DEFAULT_ROW_INDEX_COLUMN_NAME, MetadataColumnType.ROW_INDEX)));
       }
       return physicalFields;
-    } else if (logicalField.getName().equals(METADATA_ROW_COMMIT_VERSION_COLUMN_NAME)) {
+    } else if (logicalField.getMetadataColumnType() == MetadataColumnType.ROW_COMMIT_VERSION) {
       return Collections.singletonList(
           new StructField(
-              ROW_COMMIT_VERSION.getPhysicalColumnName(metadata.getConfiguration()),
+              MATERIALIZED_ROW_COMMIT_VERSION.getPhysicalColumnName(metadata.getConfiguration()),
               LongType.LONG,
               true /* nullable */));
     }
@@ -206,36 +209,48 @@ public final class MaterializedRowTrackingColumn {
    *
    * @param dataBatch a batch of physical data read from the table.
    * @param scanFile the {@link Row} representing the scan file metadata.
+   * @param logicalSchema the logical schema of the query.
    * @param configuration the configuration map containing table metadata.
    * @param engine the {@link Engine} to use for expression evaluation.
    * @return a new {@link ColumnarBatch} with logical row tracking columns
    */
   public static ColumnarBatch transformPhysicalData(
-      ColumnarBatch dataBatch, Row scanFile, Map<String, String> configuration, Engine engine) {
-    StructType readSchema = dataBatch.getSchema();
+      ColumnarBatch dataBatch,
+      Row scanFile,
+      StructType logicalSchema,
+      Map<String, String> configuration,
+      Engine engine) {
+    StructType physicalSchema = dataBatch.getSchema();
     ExpressionHandler exprHandler = engine.getExpressionHandler();
 
     // NOTE: We assume that each column is requested at most once in the read schema. This is
     // consistent with other parts of the codebase.
-    // TODO: Change the indexOf() check once we migrate to the new metadata API
-    String rowIdColumnName = ROW_ID.getPhysicalColumnName(configuration);
-    int rowIdOrdinal = readSchema.indexOf(rowIdColumnName);
+    String rowIdColumnName = MATERIALIZED_ROW_ID.getPhysicalColumnName(configuration);
+    int rowIdOrdinal = physicalSchema.indexOf(rowIdColumnName);
     if (rowIdOrdinal != -1) {
       dataBatch =
           transformPhysicalRowId(
-              dataBatch, scanFile, rowIdColumnName, exprHandler, readSchema, rowIdOrdinal);
+              dataBatch,
+              scanFile,
+              rowIdColumnName,
+              exprHandler,
+              logicalSchema,
+              physicalSchema,
+              rowIdOrdinal);
     }
 
-    String commitVersionColumnName = ROW_COMMIT_VERSION.getPhysicalColumnName(configuration);
-    int commitVersionOrdinal = readSchema.indexOf(commitVersionColumnName);
+    String rowCommitVersionColumnName =
+        MATERIALIZED_ROW_COMMIT_VERSION.getPhysicalColumnName(configuration);
+    int commitVersionOrdinal = physicalSchema.indexOf(rowCommitVersionColumnName);
     if (commitVersionOrdinal != -1) {
       dataBatch =
           transformPhysicalCommitVersion(
               dataBatch,
               scanFile,
-              commitVersionColumnName,
+              rowCommitVersionColumnName,
               exprHandler,
-              readSchema,
+              logicalSchema,
+              physicalSchema,
               commitVersionOrdinal);
     }
 
@@ -248,7 +263,8 @@ public final class MaterializedRowTrackingColumn {
       Row scanFile,
       String rowIdColumnName,
       ExpressionHandler exprHandler,
-      StructType readSchema,
+      StructType logicalSchema,
+      StructType physicalSchema,
       int rowIdOrdinal) {
     long baseRowId =
         InternalScanFileUtils.getBaseRowId(scanFile)
@@ -264,16 +280,23 @@ public final class MaterializedRowTrackingColumn {
                 new ScalarExpression(
                     "ADD",
                     Arrays.asList(
-                        new Column(StructField.METADATA_ROW_INDEX_COLUMN_NAME),
+                        new Column(
+                            dataBatch
+                                .getSchema()
+                                .at(dataBatch.getSchema().indexOf(MetadataColumnType.ROW_INDEX))
+                                .getName()),
                         Literal.ofLong(baseRowId)))));
     ColumnVector rowIdVector =
-        exprHandler.getEvaluator(readSchema, rowIdExpr, LongType.LONG).eval(dataBatch);
+        exprHandler.getEvaluator(physicalSchema, rowIdExpr, LongType.LONG).eval(dataBatch);
 
     // Remove the materialized row ID column and replace it with the coalesced vector
     dataBatch =
         dataBatch
             .withDeletedColumnAt(rowIdOrdinal)
-            .withNewColumn(rowIdOrdinal, METADATA_ROW_ID_COLUMN, rowIdVector);
+            .withNewColumn(
+                rowIdOrdinal,
+                logicalSchema.at(logicalSchema.indexOf(MetadataColumnType.ROW_ID)),
+                rowIdVector);
     return dataBatch;
   }
 
@@ -286,7 +309,8 @@ public final class MaterializedRowTrackingColumn {
       Row scanFile,
       String commitVersionColumnName,
       ExpressionHandler exprHandler,
-      StructType readSchema,
+      StructType logicalSchema,
+      StructType physicalSchema,
       int commitVersionOrdinal) {
     long defaultRowCommitVersion =
         InternalScanFileUtils.getDefaultRowCommitVersion(scanFile)
@@ -300,14 +324,16 @@ public final class MaterializedRowTrackingColumn {
             Arrays.asList(
                 new Column(commitVersionColumnName), Literal.ofLong(defaultRowCommitVersion)));
     ColumnVector commitVersionVector =
-        exprHandler.getEvaluator(readSchema, commitVersionExpr, LongType.LONG).eval(dataBatch);
+        exprHandler.getEvaluator(physicalSchema, commitVersionExpr, LongType.LONG).eval(dataBatch);
 
     // Remove the materialized row commit version column and replace it with the coalesced vector
     dataBatch =
         dataBatch
             .withDeletedColumnAt(commitVersionOrdinal)
             .withNewColumn(
-                commitVersionOrdinal, METADATA_ROW_COMMIT_VERSION_COLUMN, commitVersionVector);
+                commitVersionOrdinal,
+                logicalSchema.at(logicalSchema.indexOf(MetadataColumnType.ROW_COMMIT_VERSION)),
+                commitVersionVector);
     return dataBatch;
   }
 
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/types/DataTypeJsonSerDe.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/types/DataTypeJsonSerDe.java
@@ -363,7 +363,9 @@ public class DataTypeJsonSerDe {
       JsonNode value = entry.getValue();
       String key = entry.getKey();
 
-      if (value.isNull()) {
+      if (key.equals(StructField.METADATA_TYPE_KEY)) {
+        builder.putMetadataType(key, MetadataColumnType.fromString(value.textValue()));
+      } else if (value.isNull()) {
         builder.putNull(key);
       } else if (value.isIntegralNumber()) { // covers both int and long
         builder.putLong(key, value.longValue());
@@ -562,6 +564,8 @@ public class DataTypeJsonSerDe {
         gen.writeBoolean((Boolean) value);
       } else if (value instanceof String) {
         gen.writeString((String) value);
+      } else if (value instanceof MetadataColumnType) {
+        gen.writeString(value.toString());
       } else if (value instanceof FieldMetadata) {
         writeFieldMetadata(gen, (FieldMetadata) value);
       } else if (value instanceof Long[]) {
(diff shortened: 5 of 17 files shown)


kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/internal/parquet/ParquetFileReader.java
```diff
@@ -60,9 +60,7 @@ public ParquetFileReader(FileIO fileIO) {
   public CloseableIterator<ColumnarBatch> read(
       FileStatus fileStatus, StructType schema, Optional<Predicate> predicate) {
 
-    final boolean hasRowIndexCol =
-        schema.indexOf(StructField.METADATA_ROW_INDEX_COLUMN_NAME) >= 0
-            && schema.get(StructField.METADATA_ROW_INDEX_COLUMN_NAME).isMetadataColumn();
+    final boolean hasRowIndexCol = schema.contains(MetadataColumnType.ROW_INDEX);
```
vkorukanti: how do you distinguish between a column that has name `row_index` vs the metadata column `row_index`?

lbhm: This is an [overloaded method](https://github.com/delta-io/delta/pull/5027/files#diff-ace889e67ee9edc6fccc53da2be307d8faa8e8210380058ae655cd95e45300beR114-R116) that only accepts `MetadataColumnType` values. This way, we can safely distinguish between column name-based lookups and metadata column-based lookups.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/utils/MetadataColumnTestUtils.scala
```diff
@@ -0,0 +1,44 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.delta.kernel.defaults.utils
+
+import io.delta.kernel.types.{FieldMetadata, LongType, MetadataColumnType, StructField}
+
+trait MetadataColumnTestUtils {
+  val ROW_INDEX = new StructField(
```
emkornfield: Does this intentionally not  use `createMetadataColumn`

lbhm: No, that was an oversight. I'll fix this if we agree on this option.

lbhm: Done

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -39,12 +39,12 @@
 public final class MaterializedRowTrackingColumn {
 
   /** Static instance for the materialized row ID column. */
-  public static final MaterializedRowTrackingColumn ROW_ID =
+  public static final MaterializedRowTrackingColumn MATERIALIZED_ROW_ID =
```
lbhm, line 40: Renamed this to not create confusion with the `MetadataColumnType` instances.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/Metadata.java
```diff
@@ -104,6 +104,8 @@ public Metadata(
       ArrayValue partitionColumns,
       Optional<Long> createdTime,
       MapValue configurationMapValue) {
+    ensureNoMetadataColumns(schema);
```
lbhm, line 107: This blocks any write path attempt containing metadata columns.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -264,16 +281,23 @@ private static ColumnarBatch transformPhysicalRowId(
                 new ScalarExpression(
                     "ADD",
                     Arrays.asList(
-                        new Column(StructField.METADATA_ROW_INDEX_COLUMN_NAME),
+                        new Column(
+                            dataBatch
+                                .getSchema()
+                                .at(dataBatch.getSchema().indexOf(MetadataColumnType.ROW_INDEX))
+                                .getName()),
```
lbhm: We have to take this indirection because we don't know which name the engine decided to give to the row index metadata column.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/types/DataTypeJsonSerDe.java
```diff
@@ -363,7 +363,9 @@ private static FieldMetadata parseFieldMetadata(JsonNode json) {
       JsonNode value = entry.getValue();
       String key = entry.getKey();
 
-      if (value.isNull()) {
+      if (key.equals(StructField.METADATA_TYPE_KEY)) {
+        builder.putMetadataType(key, MetadataColumnType.fromString(value.textValue()));
+      } else if (value.isNull()) {
```
lbhm: Added this block since the value of the key is not a primitive (i.e., enum).

kernel/kernel-api/src/main/java/io/delta/kernel/types/StructField.java
```diff
@@ -140,14 +136,17 @@ public List<TypeChange> getTypeChanges() {
...
+    return metadata.contains(METADATA_TYPE_KEY);
   }
 
   public boolean isDataColumn() {
     return !isMetadataColumn();
   }
 
+  public MetadataColumnType getMetadataColumnType() {
+    return metadata.getMetadataType(METADATA_TYPE_KEY);
+  }
```
lbhm: I decided against returning an `Optional` here to stay consistent with other methods in this class, and since `isMetadataColumn()` allows you to verify that this won't be `null`.

kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java
```diff
@@ -182,10 +183,11 @@ public FilteredColumnarBatch next() {
         // Step 1: If row tracking is enabled, check for physical row tracking columns in the data
         // batch and transform them to logical row tracking columns as needed
         Map<String, String> configuration = ScanStateRow.getConfiguration(scanState);
+        StructType logicalSchema = ScanStateRow.getLogicalSchema(scanState);
```
vkorukanti: can we move these two calls to the iterator memeber variable section instead of each `next` call? There is a cost (deserializing schema etc.) in `ScanStateRow.getLogicalSchema`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/Metadata.java
```diff
@@ -348,4 +350,14 @@ private Set<String> loadPartitionColNames() {
     }
     return Collections.unmodifiableSet(partitionColumnNames);
   }
+
+  /** Helper method to ensure that a table schema never contains metadata columns. */
+  private static void ensureNoMetadataColumns(StructType schema) {
+    for (StructField field : schema.fields()) {
+      if (field.isMetadataColumn()) {
```
vkorukanti, line 357: do we need to check the nested fields?

kernel/kernel-api/src/main/java/io/delta/kernel/types/MetadataColumnType.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+package io.delta.kernel.types;
+
+/**
+ * Enumeration of metadata column types supported by Delta Kernel.
+ *
+ * <p>Metadata columns provide additional information about rows in a Delta table.
+ */
+public enum MetadataColumnType {
```
vkorukanti: super nit: The `Type` in the name is making it sound like a data type (this could just be me). Wondering if MetadataColumnKind or something else is better.

vkorukanti (approved):
LGTM, minor comments

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/Metadata.java
scovich: IMO metadata columns should always be top-level. 
If somebody wants to move them somewhere else later, they can apply an expression to do it?

Or do we envision use cases where that limitation would become annoying?

kernel/kernel-api/src/main/java/io/delta/kernel/types/MetadataColumnType.java
scovich: Devil's advocate: What would go wrong if we just called it `MetadataColumn`? 
e.g. `MetadataColumn.ROW_INDEX` seems clear enough, as does
```java
  public static StructField createMetadataColumn(String name, MetadataColumn metadataType) {
    switch (metadataType) {
      case ROW_INDEX:
```

kernel/kernel-api/src/main/java/io/delta/kernel/types/StructType.java
```diff
@@ -93,6 +102,22 @@ public int indexOf(String fieldName) {
     return fieldAndOrdinal != null ? fieldAndOrdinal._2 : -1;
   }
 
+  /** @return the index of the metadata column of the given type, or -1 if not found */
+  public int indexOf(MetadataColumnType type) {
+    // For now, we only allow each metadata column type to appear at most once in the schema.
```
scovich: ... and only at top level (not nested)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/Metadata.java
vkorukanti: Currently the StructType allows adding metadata column at any level. The check here is we don't write such metadata columns to table metadata. Here we are only checking the top-lvel columns. Asking whether we should check the nested structs are also free of metadata columns or is there a check elsewhere the prevents it from happening.

vkorukanti (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/types/StructType.java
```diff
@@ -40,6 +41,22 @@ public StructType() {
...
+    this.nameToFieldAndOrdinal = new HashMap<>();
+    for (int i = 0; i < fields.size(); i++) {
+      nameToFieldAndOrdinal.put(fields.get(i).getName(), new Tuple2<>(fields.get(i), i));
+    }
+  }
+
+  private StructType(List<StructField> fields, boolean validateColumns) {
+    if (validateColumns) {
+      validateNoNestedMetadataColumns(fields);
+    }
```
lbhm: This (almost) duplicate constructor is a workaround to prevent infinite recursion since I need a `StructType` constructor in `validateNoNestedMetadataColumns()`.

kernel/kernel-api/src/main/java/io/delta/kernel/types/StructType.java
```diff
@@ -40,6 +41,22 @@ public StructType() {
   }
 
   public StructType(List<StructField> fields) {
+    validateNoNestedMetadataColumns(fields);
```
lbhm: The validate method will repeatedly get called for every `add` call, but I don't see a way to prevent this as long as this constructor is public.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/Metadata.java
lbhm: I agree with @scovich and addressed this by adding a validation method to the `StructType` constructor that ensures that metadata columns are not nested.

kernel/kernel-api/src/main/java/io/delta/kernel/types/StructType.java
scovich: If the constructor always validates, why do we need to go deeper than one level?
* Base case -- The field to be added is a primitive type. No possibility of nested metadata.
* Recursive case -- The field to be added is a struct. The struct was already validated and so cannot contain nested metadata columns. We just need to validate that none of its primitive fields are metadata columns, since they're about to become nested columns.

lbhm: Good point. I'll simplify this.

kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/internal/parquet/ParquetFileReader.java
```diff
@@ -60,9 +60,7 @@ public ParquetFileReader(FileIO fileIO) {
   public CloseableIterator<ColumnarBatch> read(
       FileStatus fileStatus, StructType schema, Optional<Predicate> predicate) {
 
-    final boolean hasRowIndexCol =
-        schema.indexOf(StructField.METADATA_ROW_INDEX_COLUMN_NAME) >= 0
-            && schema.get(StructField.METADATA_ROW_INDEX_COLUMN_NAME).isMetadataColumn();
+    final boolean hasRowIndexCol = schema.contains(MetadataColumn.ROW_INDEX);
```
vkorukanti: should we throw error if we get any other metadata column type?

vkorukanti (approved):
One comment. If required can be addressed in a new PR.

kernel/kernel-api/src/main/java/io/delta/kernel/types/StructType.java
```diff
@@ -41,21 +40,7 @@ public StructType() {
...
-    for (int i = 0; i < fields.size(); i++) {
-      nameToFieldAndOrdinal.put(fields.get(i).getName(), new Tuple2<>(fields.get(i), i));
-    }
-  }
-
-  private StructType(List<StructField> fields, boolean validateColumns) {
-    if (validateColumns) {
-      validateNoNestedMetadataColumns(fields);
-    }
+    validateNoNestedMetadataColumns(fields, true /* top level */);
```
scovich: Given that the method has only one call site, I would recommend rearranging the logic a bit:
* Here, we iterate over the incoming struct fields
   * Filter out primitive columns, because they're allowed to be metadata
   * Pass the remaining items for validation
* Rename the method to `validateNoMetadataColumns`
   * It does what it says -- blows up if it sees any metadata columns
* For each input field, it checks:
   * If primitive -> blow up if metadata (top-level should have filtered those out)
   * if struct -> recurse into the fields
      * do _not_ filter out primitives
   * If array -> recurse into the array's element type
      * do _not_ filter out primitives
      * (or, update the array type's constructor to reject metadata columns as element)
   * if map -> recuse into the map's key and value types
      * do _not_ filter out primitives
      * (or, update the map type's constructor to reject metadata columns as key and/or value)

kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/internal/parquet/ParquetFileReader.java
lbhm: I added a check that looks for any other metadata column above this line.

kernel/kernel-api/src/main/java/io/delta/kernel/types/StructType.java
lbhm: Done, thank you for the suggestion!
I felt that recursing inside `validateNoMetadataColumns` was a less invasive change than touching the `MapType`/`ArrayType` constructor, so I went with the first option.

kernel/kernel-api/src/main/java/io/delta/kernel/types/StructField.java
```diff
@@ -140,14 +136,18 @@ public List<TypeChange> getTypeChanges() {
   }
 
   public boolean isMetadataColumn() {
-    return metadata.contains(IS_METADATA_COLUMN_KEY)
-        && (boolean) metadata.get(IS_METADATA_COLUMN_KEY);
+    return metadata != null && metadata.contains(METADATA_TYPE_KEY);
```
lbhm: `metadata` is `null` in some tests, so I need to guard against it.

kernel/kernel-api/src/main/java/io/delta/kernel/types/StructType.java
scovich: Fair about less-invasive... but isn't it pretty broken to use a metadata column as a map key? 
(in other words, do we anyway need that validation to prevent brokenness)?

lbhm: I agree with you that checks should be performed when constructing `MapType` and `ArrayType`. However, their precise semantics are a bit more involved, IMO. For example, do you only allow hashable types in map keys? Do you only allow strings to stay JSON-compliant?

IMO, making those decisions is outside the scope of this PR and should probably involve a design discussion among the kernel team. I have created #5143 to track this discussion.

scovich: Not sure I follow? Are you saying there is other, additional, verification that the map/array constructors miss? If so, I agree that's out of scope for this PR... but why does that prevent us from correctly validating the new behavior this PR introduces? 

Tho now that I think about it... maps and arrays don't have field metadata in the first place, so it would be hard to mark them as metadata columns?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
```diff
@@ -251,10 +252,13 @@ private StructType createPhysicalSchema() {
...
     if (protocol.getReaderFeatures().contains("deletionVectors")
         && physicalFields.stream()
-            .map(StructField::getName)
-            .noneMatch(name -> name.equals(StructField.METADATA_ROW_INDEX_COLUMN_NAME))) {
+            .map(StructField::getMetadataColumnType)
+            .noneMatch(MetadataColumn.ROW_INDEX::equals)) {
       // If the row index column is not already present, add it to the physical read schema
-      physicalFields.add(SchemaUtils.createInternalColumn(StructField.METADATA_ROW_INDEX_COLUMN));
+      physicalFields.add(
+          SchemaUtils.createInternalColumn(
```
allisonport-db: nit: can we rename createInternalColumn to `asInternalColumn` or something? found this sequence confusing without going and looking at all the method definitions

allisonport-db: Definitely do not want to address this in this PR, but this also makes me wonder if `IS_INTERNAL_COLUMN_KEY` should be stored on the StructField instead of just in the metadata cc @vkorukanti @raveeram-db @emkornfield

kernel/kernel-api/src/main/java/io/delta/kernel/types/StructField.java
```diff
@@ -209,6 +209,42 @@ public StructField withDataType(DataType newType) {
...
   }
 
+  /**
+   * Creates a metadata column of the given {@code colType} with the given {@code name}.
+   *
+   * @param name Name of the metadata column
+   * @param colType Type of the metadata column
+   * @return A StructField representing the metadata column
+   */
+  public static StructField createMetadataColumn(String name, MetadataColumn colType) {
```
allisonport-db: can you put this with the other static fields please?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/Metadata.java
```diff
@@ -348,4 +350,14 @@ private Set<String> loadPartitionColNames() {
     }
     return Collections.unmodifiableSet(partitionColumnNames);
   }
+
+  /** Helper method to ensure that a table schema never contains metadata columns. */
+  private static void ensureNoMetadataColumns(StructType schema) {
```
allisonport-db: nit - no reason to be static right?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -175,24 +173,28 @@ public static List<StructField> convertToPhysicalColumn(
...
+              MATERIALIZED_ROW_ID.getPhysicalColumnName(metadata.getConfiguration()),
               LongType.LONG,
               true /* nullable */));
-      if (logicalSchema.indexOf(StructField.METADATA_ROW_INDEX_COLUMN_NAME) == -1) {
-        physicalFields.add(SchemaUtils.createInternalColumn(StructField.METADATA_ROW_INDEX_COLUMN));
+      if (!logicalSchema.contains(MetadataColumn.ROW_INDEX)) {
+        physicalFields.add(
+            SchemaUtils.createInternalColumn(
+                StructField.createMetadataColumn(
+                    StructField.DEFAULT_ROW_INDEX_COLUMN_NAME, MetadataColumn.ROW_INDEX)));
```
allisonport-db: Maybe we can have a `DEFAULT_ROW_INDEX_STRUCT_FIELD` in MetadataColumn? And use that here + in `ScanImpl` and anywhere else we construct it using the default name?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -264,16 +279,23 @@ private static ColumnarBatch transformPhysicalRowId(
                 new ScalarExpression(
                     "ADD",
                     Arrays.asList(
-                        new Column(StructField.METADATA_ROW_INDEX_COLUMN_NAME),
+                        new Column(
+                            dataBatch
+                                .getSchema()
+                                .at(dataBatch.getSchema().indexOf(MetadataColumn.ROW_INDEX))
+                                .getName()),
```
allisonport-db: Can you store this in a variable `rowIndexMetadataColName` before? Will make this easier to parse

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
```diff
@@ -264,16 +279,23 @@ private static ColumnarBatch transformPhysicalRowId(
...
+        exprHandler.getEvaluator(physicalSchema, rowIdExpr, LongType.LONG).eval(dataBatch);
 
     // Remove the materialized row ID column and replace it with the coalesced vector
     dataBatch =
         dataBatch
             .withDeletedColumnAt(rowIdOrdinal)
-            .withNewColumn(rowIdOrdinal, METADATA_ROW_ID_COLUMN, rowIdVector);
+            .withNewColumn(
+                rowIdOrdinal,
+                logicalSchema.at(logicalSchema.indexOf(MetadataColumn.ROW_ID)),
```
allisonport-db: if this is the only place we use `logicalSchema` what do you think about instead having the method input be `rowIdLogicalStructField` or something?

allisonport-db: also maybe we can add a `StructType::get(MetadataColumn colType)` helper?

kernel/kernel-api/src/main/java/io/delta/kernel/types/MetadataColumn.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+package io.delta.kernel.types;
+
+/**
+ * Enumeration of metadata column types supported by Delta Kernel.
+ *
+ * <p>Metadata columns provide additional information about rows in a Delta table.
+ */
+public enum MetadataColumn {
```
allisonport-db: should this be `MetadataColumnType`? Before looking at this file I assumed this was some representation related to column/structField.

I think I remember a thread about this though, is there a reason why it's not that?

kernel/kernel-api/src/main/java/io/delta/kernel/types/MetadataColumn.java
```diff
@@ -0,0 +1,47 @@
...
+  ROW_INDEX("row_index"),
+  ROW_ID("row_id"),
+  ROW_COMMIT_VERSION("row_commit_version");
+
+  private final String textValue;
+
+  MetadataColumn(String textValue) {
```
allisonport-db: should this be private?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
allisonport-db: I think I was mis-understanding what was stored in MetadataColumn. Maybe this would belong in StructField instead?

kernel/kernel-api/src/main/java/io/delta/kernel/types/StructType.java
```diff
@@ -40,6 +40,11 @@ public StructType() {
   }
 
   public StructType(List<StructField> fields) {
+    validateNoMetadataColumns(
```
allisonport-db, line 44: can you rename this to `validateNoNestedMetadataColumns` and push this logic down? this is otherwise confusing I thought we were validating that there are no metadata columns at all

allisonport-db: also let's maybe add a second `validateNoDuplicateMetadataColumns` validation? `addMetadataColumn` doesn't stop you from adding duplicates via one of the other add methods

kernel/kernel-api/src/main/java/io/delta/kernel/types/StructType.java
```diff
@@ -163,4 +194,34 @@ public boolean equals(Object o) {
...
     return Objects.hash(nameToFieldAndOrdinal, fields, fieldNames);
   }
+
+  /**
+   * Validates that there are no metadata columns in a list of StructFields.
+   *
+   * @param fields The list of fields to validate
+   * @throws IllegalArgumentException if any nested metadata columns are found
+   */
+  private static void validateNoMetadataColumns(List<StructField> fields) {
```
allisonport-db: nit - no reason to be static for a private member helper I think

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/MetadataColumnSuite.scala
allisonport-db, line 1: move this to the `kernel-api` package please, this shouldn't use any engine utils

kernel/kernel-api/src/main/java/io/delta/kernel/types/StructType.java
allisonport-db: and add tests for these in `MetadataColumnSuite` (I think you're already testing most of this yay!)

kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/internal/parquet/ParquetFileReader.java
```diff
@@ -60,9 +60,13 @@ public ParquetFileReader(FileIO fileIO) {
...
 
-    final boolean hasRowIndexCol =
-        schema.indexOf(StructField.METADATA_ROW_INDEX_COLUMN_NAME) >= 0
-            && schema.get(StructField.METADATA_ROW_INDEX_COLUMN_NAME).isMetadataColumn();
+    if (Arrays.stream(MetadataColumn.values())
+        .filter(colType -> !colType.equals(MetadataColumn.ROW_INDEX))
+        .anyMatch(schema::contains)) {
+      throw new IllegalArgumentException(
+          "The parquet reader does not support metadata columns other than ROW_INDEX");
+    }
```
allisonport-db: I would slightly prefer this to iterate over the schema and check `isMetadataColumn` -> `metadataColumnType != ROW_INDEX`

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/utils/MetadataColumnTestUtils.scala
```diff
@@ -0,0 +1,27 @@
...
+
+import io.delta.kernel.types.{MetadataColumn, StructField}
+import io.delta.kernel.types.StructField.createMetadataColumn
+
+trait MetadataColumnTestUtils {
+  val ROW_INDEX: StructField =
+    createMetadataColumn("_metadata.row_index", MetadataColumn.ROW_INDEX)
+  val ROW_ID: StructField = createMetadataColumn("_metadata.row_id", MetadataColumn.ROW_ID)
+  val ROW_COMMIT_VERSION: StructField =
+    createMetadataColumn("_metadata.row_commit_version", MetadataColumn.ROW_COMMIT_VERSION)
```
allisonport-db: I think I suggested adding static default fields for these - if we go ahead with that maybe we can remove this all-together? seems like a useful util for both the src code and test code

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ScanImpl.java
lbhm: I renamed the method. Will leave this comment open in case others want to chime in on the future work on `IS_INTERNAL_COLUMN_KEY`.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/Metadata.java
lbhm: No, I changed it.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
lbhm: I added `DEFAULT_ROW_INDEX_COLUMN` to `StructField.java`.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/rowtracking/MaterializedRowTrackingColumn.java
lbhm: I refactored the input parameter to `logicalRowIdColumn` and `logicalCommitVersionColumn` 👍🏼

kernel/kernel-api/src/main/java/io/delta/kernel/types/MetadataColumn.java
lbhm: It used to be called that way, but @vkorukanti and @scovich argued against it. See https://github.com/delta-io/delta/pull/5027#discussion_r2304978757. I'm fine with either option, but wouldn't want to block the PR over this detail. If you have a strong opinion here, please discuss it directly with them.

kernel/kernel-api/src/main/java/io/delta/kernel/types/MetadataColumn.java
lbhm: I checked and IntelliJ tells me that `'private' is redundant for enum constructors`. So I guess it does not matter.

kernel/kernel-api/src/main/java/io/delta/kernel/types/StructType.java
lbhm: I changed the validation logic because @scovich specifically argued to redesign it this way. See https://github.com/delta-io/delta/pull/5027#discussion_r2308110446. Please discuss with them if you have a strong opinion here.

Regarding the no duplicate check: I also added a check to the regular `add` method that throws an error if you try to add a metadata column and that already exists.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/utils/MetadataColumnTestUtils.scala
lbhm: I added a static default field for row index because we use it in kernel. @johanl-db, in the past, argued against having static fields for the other metadata columns if they are not directly used by kernel. I'm somewhat indifferent here. IMO, this is not a big deal either way.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/MetadataColumnSuite.scala
lbhm: I moved this suite and `MetadataColumnTestUtils.scala` to `kernel-api`.

lbhm:
Thank you for the detailed review @allisonport-db! I addressed most of your comments and responded to the ones where others had a different opinion.

kernel/kernel-api/src/main/java/io/delta/kernel/types/MetadataColumn.java
scovich: Problem is `Type` is an overloaded/ambiguous term. Does it mean "kind" or "datatype"?

Calling it `MetadataColumnFoo` would be great, but only if "foo" were some other, unambiguous word that makes clear it's "kind" instead of "datatype" ? `MetadataColumnKind` and `MetadataColumnFlavor` both seemed unambiguous but a bit weird. If people like one of those -- or have a better idea -- I'm totally good with it.

raveeram-db (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/types/MetadataColumn.java
allisonport-db: I agree they're a bit weird, but I do think I prefer `MetadataColumnKind` or something of that sort to the current naming to avoid confusion. We could also do something like `MetadataColumnSpec`. I'm not too picky other than I don't like `MetadataColumn`.

Agree on not blocking the PR on this if we don't have an easy answer!

kernel/kernel-api/src/main/java/io/delta/kernel/types/StructType.java
allisonport-db: For the duplicate check - can we still just do so on the constructor? We only need to cover top level fields and this ensures we enforce this for all struct types, and reduces duplication across the various add methods.

lbhm: Good point. I removed the metadata duplicate check from the `add*()` methods and put it into the `StructType` constructor.

kernel/kernel-api/src/main/java/io/delta/kernel/types/MetadataColumn.java
scovich: IMO either of those names is better than `MetadataColumn[Type]`. Should we pick one and move on?

One question tho -- `MetadataColumnSpec` might be construed to imply that it somehow defines the column's data type and nullability as well. Which it doesn't seem to, but I'm also not clear on where/how the data type is decided/enforced in general? Do you just "have to know" these are all non-nullable LONG values?

scovich: Looks like it's hard-wired in the code of `createMetadataColumn`. Should we consider removing that duplication by tracking data type and nullability here? Or is that too difficult with java enums?

lbhm: I'll add `dataType` and `nullable` to the enum definition and call it `MetadataColumnSpec` so that we can move on.

lbhm: Done!

PR #5033, opened 2025-08-06
[KERNEL] Refactor TransactionBuilderImpl to have main transaction building logic in a separate class that can be shared by intent-based-builders

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Refactors the txn building logic to a separate class. This allows the current TransactionBuilder and future intent-based builders to share the same metadata update/validation code.

## How was this patch tested?

No.

------------------------------------------------------------------------------------------------------------------------
### Summary of main changes in this PR

1. Moves main metadata/protocol update + validation code to a separate class TransactionMetadataFactory
    - This class defines clear inputs and outputs for this code, with respect to individual operations.
        - Replace inputs: readSnapshot, schema, tableProperties, 1 of clusteringColumns or partitionColumns
        - Create inputs: schema, tableProperties, 1 of clusteringColumns or partitionColumns
        - Update inputs: readSnapshot, propertiesAdded, propertiesRemoved, newSchema, newClusteringColumns
        - Create and Replace outputs: new protocol & metadata, optionally resolved clustering cols if relevant
        - Replace outputs: new protocol & metadata _if they should be updated_,  optionally resolved clustering cols if relevant
    - TransactionBuilderImpl (and future intent-based builders) still must validate that they have the correct inputs for these entry points
2. Changes to the update/validation code
    - Move code that's specific to different operations out of shared update logic (i.e. unsetting table properties for update, getting the initial metadata for create, schema evolution for update, etc)
         - The true validation logic now takes new/updated metadata/protocol as an input.
    - Removes some code that executes in all cases but only makes sense for existing tables (i.e. `blockConfigChangeOnExistingTable ` doesn't make sense for a new table, but before we just called it with `isNewTable=true` where we allow this contrived change from "wasEnabled=false" -> "isEnabled=true" because we compare our original empty metadata with the one proposed for the new table).
    - Overall a lot of this code is still copied over and in future PRs we can iterate on this separate class to try to improve the logic and make it more intuitive

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
@@ -168,7 +168,7 @@ public class TransactionMetadataFactory {
       Optional<Map<String, String>> propertiesAdded,
       Optional<Set<String>> propertyKeysRemoved,
       Optional<StructType> newSchema,
-      Optional<List<Column>> userProvidedClusteringColumns) {
+      Optional<List<Column>> clusteringColumns) {
     if (propertiesAdded.isPresent() && propertyKeysRemoved.isPresent()) {
       Set<String> overlappingPropertyKeys =
           propertyKeysRemoved.get().stream()
@@ -209,7 +209,7 @@ public class TransactionMetadataFactory {
             newMetadata,
             Optional.empty(),
             false /* isCreateOrReplace */,
-            userProvidedClusteringColumns,
+            clusteringColumns,
             newSchema.isPresent() /* isSchemaEvolultion */)
         .finalOutput;
   }


kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
allisonport-db, line 1: A lot of this code is largely copied from TransactionBuilderImpl. I think there's more work to be done to improve the txn building iteration here and the expectations within the code -- but for now this factors out the inputs and outputs and allows us to improve this separately without changing the expectations for the input/output.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaReplaceTableSuite.scala
```diff
@@ -668,7 +671,11 @@ class DeltaReplaceTableSuite extends DeltaReplaceTableSuiteBase {
     }
   }
 
-  test("Column mapping maxFieldId is preserved during REPLACE TABLE") {
+  test("Column mapping maxFieldId is preserved during REPLACE TABLE " +
```
allisonport-db, line 674: This was actually a bug before -- we were accidentally not blocking going from (id -> none) when there were no property updates

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -256,67 +239,76 @@ protected TransactionImpl buildTransactionInternal(
...
+            && latestSnapshot
+                .get()
+                .getProtocol()
+                .supportsFeature(TableFeatures.CLUSTERING_W_FEATURE);
+    if (inputLogicalClusteringColumns.isPresent() && !clusteringEnabled) {
+      tablePropertiesWithFeatureEnablement.put(
+          TableFeatures.SET_TABLE_FEATURE_SUPPORTED_PREFIX
+              + TableFeatures.CLUSTERING_W_FEATURE.featureName(),
+          "supported");
+    }
```
allisonport-db: TODO: move this auto-enablement to inside of `TransactionMetadataFactory` -- the only one we should have to special case here is `withDomainMetadataSupported`

allisonport-db:
For context this is what `UpdateTransactionBuilderImpl` looks like using these changes https://github.com/allisonport-db/delta/blob/cf3e4fbc5ccf0d06707a83695fced8e33ec7ba95/kernel/kernel-api/src/main/java/io/delta/kernel/internal/UpdateTableTransactionBuilderImpl.java (I think super simple & short)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
```diff
@@ -0,0 +1,544 @@
...
+  private Optional<Protocol> newProtocol;
+  private Optional<List<Column>> resolvedNewClusteringColumns;
+
+  /* Fields that are fixed after validation and updates are finished */
+  private final Output finalOutput;
+
+  private TransactionMetadataFactory(
+      String tablePath,
+      Optional<SnapshotImpl> latestSnapshotOpt,
+      Optional<Metadata> initialNewMetadata,
```
allisonport-db, line 244: Clarify - this contains the _user provided_ updates, we may apply additional updates (auto enable cm mode for iceberg compat, etc)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -117,28 +117,31 @@ public TransactionImpl(
...
+    this.metadata = newMetadata.orElseGet(() -> readSnapshotOpt.get().getMetadata());
+    this.shouldUpdateMetadata = newMetadata.isPresent();
+
     this.isCreateOrReplace = isCreateOrReplace;
     this.dataPath = dataPath;
     this.logPath = logPath;
     this.readSnapshotOpt = readSnapshotOpt;
     this.engineInfo = engineInfo;
     this.operation = operation;
-    this.protocol = protocol;
```
scottsand-db, line 136: nit: can we leave the variable asignment in the same order? i.e. just move `this.protocol = newProtocol.orElseGet(() -> readSnapshotOpt.get().getProtocol());` to this line? that way the constructor param order better-matches the assignment order here?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
```diff
@@ -0,0 +1,544 @@
...
+/**
+ * Class for building the Protocol, Metadata, and ResolvedClusteringColumns to commit in a
+ * transaction. This means it validates and updates the protocol and metadata according to the
+ * inputs.
+ */
+public class TransactionMetadataFactory {
+  private static final Logger logger = LoggerFactory.getLogger(TransactionMetadataFactory.class);
+
+  /**
+   * Expectations with respect to the given operation: - Create table: both protocol and metadata
```
scottsand-db: nit: `<ul><li>` so that the `-` is not formatted like this

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
```diff
@@ -0,0 +1,544 @@
...
+   * will be present - Replace table: both protoocl and metadata will be present - Update table:
```
scottsand-db: `nit: protocol` typo

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
```diff
@@ -0,0 +1,544 @@
...
+    }
+  }
+
+  static Output buildCreateTableMetadata(
+      String tablePath,
+      StructType schema,
+      Map<String, String> tableProperties,
+      Optional<List<String>> partitionColumns,
+      Optional<List<Column>> clusteringColumns) {
+    checkArgument(!partitionColumns.isPresent() || !clusteringColumns.isPresent());
```
scottsand-db: is it worth having a comment accompanying this `checkArgument` that states

`// You cannot set both the partition and clustering columns when creating a new table.`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
```diff
@@ -0,0 +1,544 @@
...
+      this.resolvedNewClusteringColumns = resolvedNewClusteringColumns;
+    }
+  }
+
+  static Output buildCreateTableMetadata(
+      String tablePath,
+      StructType schema,
+      Map<String, String> tableProperties,
+      Optional<List<String>> partitionColumns,
+      Optional<List<Column>> clusteringColumns) {
```
scottsand-db, line 95: @btw, every time I see `Optional<List<Column>> clusteringColumns` it seems like I need to remind myself what Optional.of(empty list) means.

Maybe worth asking claude if a good solution to this problem is just created a class `ResolvedPhysicalClusteringColumns` that has that one single field inside of it, and _that_ class can have the comment explaining the semantics? instead of repeating the documentation (comment) everywhere?

This could be a followup PR?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
```diff
@@ -0,0 +1,544 @@
...
+    // In the case of Replace table there are a few delta-specific properties we want to preserve
+    Map<String, String> properties =
+        readSnapshot.getMetadata().getConfiguration().entrySet().stream()
+            .filter(
+                e ->
+                    ReplaceTableTransactionBuilderImpl.TABLE_PROPERTY_KEYS_TO_PRESERVE.contains(
+                        e.getKey()))
+            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+    properties.putAll(tableProperties);
+    Metadata metadata =
```
scottsand-db: nit: here, `metadata` is declared as a variable, but above in `buildCreateTableMetadata` it is declared inline in the call to `TransactionMetadataFactory`. can we pick one or the other?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
```diff
@@ -0,0 +1,544 @@
...
+  static Output buildReplaceTableMetadata(
+      String tablePath,
+      SnapshotImpl readSnapshot,
+      StructType schema,
+      Map<String, String> tableProperties,
+      Optional<List<String>> partitionColumns,
+      Optional<List<Column>> clusteringColumns) {
+    checkArgument(!partitionColumns.isPresent() || !clusteringColumns.isPresent());
+    // In the case of Replace table there are a few delta-specific properties we want to preserve
+    Map<String, String> properties =
```
scottsand-db: can we have some clearer names than `tableProperties` and `properties`

`tableProperties -> userInputProperties`? (not sure, I defer to you)

`properties -> userInputPropertiesUpdatedWithRequiredProperties` // something like that

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
```diff
@@ -0,0 +1,544 @@
...
+
+  static Output buildUpdateTableMetadata(
+      String tablePath,
+      SnapshotImpl readSnapshot,
+      Optional<Map<String, String>> propertiesAdded,
+      Optional<Set<String>> propertyKeysRemoved,
+      Optional<StructType> newSchema,
+      Optional<List<Column>> userProvidedClusteringColumns) {
+    if (propertiesAdded.isPresent() && propertyKeysRemoved.isPresent()) {
+      Set<String> invalidPropertyKeys =
```
scottsand-db: not even a nit, just an idea (personal preference) -- `overalappingPropertyKeys`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
```diff
@@ -0,0 +1,544 @@
...
+            tablePath,
+            Optional.of(readSnapshot),
+            newMetadata,
+            Optional.empty(),
+            false /* isCreateOrReplace */,
+            userProvidedClusteringColumns,
+            newSchema.isPresent() /* isSchemaEvolultion */)
+        .finalOutput;
+  }
+
```
scottsand-db, line 216: You know me :P my personal preference is some header blocks like this to help split up the file a bit

```java
  ////////////////////////////
  // Static factory methods //
  ////////////////////////////

  ///////////////////////////////
  // Instance Fields / Methods //
  ///////////////////////////////
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
```diff
@@ -0,0 +1,544 @@
...
+  private TransactionMetadataFactory(
+      String tablePath,
+      Optional<SnapshotImpl> latestSnapshotOpt,
+      Optional<Metadata> initialNewMetadata,
+      Optional<Protocol> initialNewProtocol,
+      boolean isCreateOrReplace,
+      Optional<List<Column>> userProvidedClusteringColumns,
+      boolean isSchemaEvolution) {
+    checkArgument(
```
scottsand-db, line 249: Idea for code clarity: could each of these `checkArguments` become super simple methods we call with very detailed names that throw detailed error messages? and then there's no need for a comment next to the `checkArgumetn` and the combinations of `!` and `||` and `&&` gets abstracted away to that new, well-named method? I know it adds some line overhead to the overall file but ... I think these super clean check arguments could improve readability. wdyt? This PR is already big ==> i'm fine making it a bit bigger, ha.

```java
validatePandMConsistency(...)
validateSchemaEvolutionForUpdateOnly(...)
validateSnapshotRequiredForUpdate(...)
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
```diff
@@ -0,0 +1,544 @@
...
+    checkArgument(!isSchemaEvolution || !isCreateOrReplace);
+    checkArgument(isCreateOrReplace || latestSnapshotOpt.isPresent());
+    this.tablePath = tablePath;
+    this.latestSnapshotOpt = latestSnapshotOpt;
+    this.isCreateOrReplace = isCreateOrReplace;
+    this.isSchemaEvolution = isSchemaEvolution;
+    this.newMetadata = initialNewMetadata;
+    this.newProtocol = initialNewProtocol;
+
+    performProtocolUpgrades();
```
scottsand-db: What if we moved these processing/logic methods into `def getFinalOutput` and kept the constructor to just the checkArguments and the assignment?

Then, when `getFinalOutput` runs, we KNOW for sure that every member variable has been assigned -- because the constructor has finished already!

whereas when logic runs in the constructor, there's a chance that some new develoepr added a logic method before some new instance variable has finished being assigned?

wdyt? not a blocker, just some food for thought. it's okay with me if you think this isn't necessary.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
```diff
@@ -0,0 +1,544 @@
...
+                newMetadata = Optional.of(getEffectiveMetadata().withNewSchema(structType)));
+      }
+    }
+
+    MaterializedRowTrackingColumn.throwIfColumnNamesConflictWithSchema(getEffectiveMetadata());
+  }
+
+  private static Metadata getInitialMetadata(
+      StructType schema, Map<String, String> tableProperties, List<String> partitionColumns) {
+    List<String> partitionColumnsCasePreserving =
```
scottsand-db: could this be declared inline? the function name is already clear enough 👍  wdyt?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
```diff
@@ -0,0 +1,544 @@
...
+    // Fact: either newMetadata is defined upon initiation or latestSnapshotOpt is present
+    return newMetadata.orElseGet(() -> latestSnapshotOpt.get().getMetadata());
+  }
+
+  private Protocol getEffectiveProtocol() {
+    // Fact: either newProtocol is defined upon initiation or latestSnapshotOpt is present
+    return newProtocol.orElseGet(() -> latestSnapshotOpt.get().getProtocol());
+  }
+
+  private void validateForUpdateTableUsingOldMetadata(Consumer<Metadata> validateFx) {
```
scottsand-db: nit: do we usually use `Fx` for "function" or would `Fn` or `Func` be cleaner?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
```diff
@@ -0,0 +1,544 @@
...
+  /* Fields that set by input */
+  private final String tablePath;
+  private final Optional<SnapshotImpl> latestSnapshotOpt;
+  private final boolean isCreateOrReplace;
+  private final boolean isSchemaEvolution;
+
+  /* Fields that are updated by helper methods when updating and validating the metadata */
+  private Optional<Metadata> newMetadata;
+  private Optional<Protocol> newProtocol;
+  private Optional<List<Column>> resolvedNewClusteringColumns;
```
scottsand-db: btw, what does "resovled" mean? This also brings me to the `ResolvedClusteringColumns` class that could document this and answer my question for me

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -256,67 +239,76 @@ protected TransactionImpl buildTransactionInternal(
...
-                            e.getKey()))
-                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
-        baseMetadata = baseMetadata.withMergedConfiguration(propertiesToPreserve);
-      }
-    } else {
-      // Otherwise, use the existing table metadata
-      baseMetadata = latestSnapshot.get().getMetadata();
+    // Instead of special casing enabling clustering or domain metadata, we should just add them
+    // to the table properties which we already handle.
+    Map<String, String> tablePropertiesWithFeatureEnablement =
```
scottsand-db: could this code here and the clustering code below be refactored to N simple methods? void or not idk, I defer to you

```java
x = addEnableDomainMetadataViaTablePropertyIfApplicable(..)
x = addEnableClusteringViaTablePropertyIfApplicable(..)
```

scottsand-db (changes requested):
Looks AWESOME! No push back to any major logic decisions, just tiny code style suggestions that will just build upon the already awesome clean up and clarity improvements that this PR adds.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
allisonport-db: yeah I just copy pasted this -- updated

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
allisonport-db: So I actually changed this, I think clustering should be handled by `TransactionMetadataFactory` so this is simpler now

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
allisonport-db: yeah sorry we use this wording throughout. I think I will change this to `physicalNewClusteringColumns`? I think this is the same thing (AFAIK but I also didn't coin this original terminology) and then we just reference "physical clustering columns" and "logical clustering columns"!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
allisonport-db: yeah I considered this as well but kinda liked enforcing that the building is only run once.

> whereas when logic runs in the constructor, there's a chance that some new develoepr added a logic method before some new instance variable has finished being assigned?

If it uses that variable we would presumably catch this though right?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
allisonport-db: I added error messages which I think helps, let me know if you think helper methods are better. The constructor is not too lengthy so personally fine with either.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
allisonport-db: Let's address this separately -- I'm not sure adding a new class is necessarily necessary but we can consider other options. Also maybe this just needs better documentation on the clustering protocol level? I think the idea of empty clustering columns is exactly that -- but maybe that's a little undefined w.r.t. what this means for the _Delta table_.

huan233usc: I remember @KaiqiJinWow once mentioned that empty clustering columns is also a clustered table. It is confusing, though.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
scottsand-db: > yeah I considered this as well but kinda liked enforcing that the building is only run once.

`getFinalOutput` could just cache the result? and return that if it is already pre-computed? not trying to push too too hard for my suggestion here, but it does seem like it has a few benefits without much cost? wdyt?

> If it uses that variable we would presumably catch this though right?

it would be a runtime exception, not a compile-time exception. so we would catch it if that new code was correctly tested, yeah, but it's not guaranteed

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
scottsand-db: It seems like we have a `type` that has different states (3) that we want re-used and well-documented. To me, we can just have that be a class. We can have the documentation in exactly one place, and only one place, and can have convience APIs like `.isClusteringSupported` etc.

Again, not suggesting in this PR :) But a good food for thought for a future refactor.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
scottsand-db: LGTM

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
```diff
@@ -0,0 +1,575 @@
...
+  ////////////////////////////
+
+  static Output buildCreateTableMetadata(
+      String tablePath,
+      StructType schema,
+      Map<String, String> tableProperties,
+      Optional<List<String>> partitionColumns,
+      Optional<List<Column>> clusteringColumns) {
+    // You cannot set both the partition and clustering columns
+    checkArgument(!partitionColumns.isPresent() || !clusteringColumns.isPresent());
```
scottsand-db: nit: like what you did below, move the information from this comment into the error message?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
```diff
@@ -0,0 +1,575 @@
...
+                    getInitialMetadata(
+                        schema, tableProperties, partitionColumns.orElse(emptyList()))),
+                Optional.of(getInitialProtocol()),
+                true /* isCreateOrReplace */,
+                clusteringColumns,
+                false /* isSchemaEvolultion */)
+            .finalOutput;
+    checkState(
+        output.newMetadata.isPresent() && output.newProtocol.isPresent(),
+        "Expected non-null metadata and protocol for create table");
```
scottsand-db, line 113: nit: non-empty; or, put another way; `metadata and protocol must be present for create table`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
```diff
@@ -0,0 +1,575 @@
...
+    return output;
+  }
+
+  static Output buildReplaceTableMetadata(
+      String tablePath,
+      SnapshotImpl readSnapshot,
+      StructType schema,
+      Map<String, String> userInputTableProperties,
+      Optional<List<String>> partitionColumns,
+      Optional<List<Column>> clusteringColumns) {
```
scottsand-db, line 123: should this be `userProvidedClusteringColumns` ?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
```diff
@@ -0,0 +1,575 @@
...
+    // You cannot set both the partition and clustering columns
+    checkArgument(!partitionColumns.isPresent() || !clusteringColumns.isPresent());
```
scottsand-db: ditto -- same comment as above

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
allisonport-db: I wasn't 100% sure on this because i think within the context that's relatively clear, and not sure if I should thus rename everything `userProvidedSchema` and `userProvidedPartitionColumns` etc

scottsand-db: Totally fair. I asked that because `buildUpdateTableMetadata` calls it `userProvidedClusteringColumns`

Could we just align on naming? -- Alignemnt ==> less cognitive load to read these three methods and understanding how they are similar/different

scottsand-db (approved):
LGTM! Great refactor and nice succinct tests 👍

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
allisonport-db: Fixed

PR #5040, opened 2025-08-07
done first pass

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/5040/files) to review incremental changes.
- [**stack/kernel_default_file_system_managed_committer**](https://github.com/delta-io/delta/pull/5040) [[Files changed](https://github.com/delta-io/delta/pull/5040/files)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/DefaultFileSystemManagedTableOnlyCommitter.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/DefaultFileSystemManagedTableOnlyCommitter.java
@@ -16,6 +16,8 @@
 
 package io.delta.kernel.internal.commit;
 
+import static io.delta.kernel.internal.DeltaErrors.wrapEngineExceptionThrowsIO;
+
 import io.delta.kernel.commit.CommitFailedException;
 import io.delta.kernel.commit.CommitMetadata;
 import io.delta.kernel.commit.CommitResponse;
@@ -24,10 +26,21 @@ import io.delta.kernel.data.Row;
 import io.delta.kernel.engine.Engine;
 import io.delta.kernel.internal.DeltaErrorsInternal;
 import io.delta.kernel.internal.actions.Protocol;
+import io.delta.kernel.internal.files.ParsedLogData;
 import io.delta.kernel.internal.tablefeatures.TableFeatures;
+import io.delta.kernel.internal.util.FileNames;
 import io.delta.kernel.utils.CloseableIterator;
+import io.delta.kernel.utils.FileStatus;
+import java.io.IOException;
+import java.nio.file.FileAlreadyExistsException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class DefaultFileSystemManagedTableOnlyCommitter implements Committer {
+
+  private static final Logger logger =
+      LoggerFactory.getLogger(DefaultFileSystemManagedTableOnlyCommitter.class);
+
   public static final DefaultFileSystemManagedTableOnlyCommitter INSTANCE =
       new DefaultFileSystemManagedTableOnlyCommitter();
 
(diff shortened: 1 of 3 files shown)


kernel/kernel-api/src/test/scala/io/delta/kernel/internal/commit/DefaultCommitterSuite.scala
```diff
@@ -33,6 +38,15 @@ class DefaultCommitterSuite extends AnyFunSuite
 
   private val protocol12 = new Protocol(1, 2)
 
+  private val basicFileSystemCommitMetadataNoPMChange = new CommitMetadata(
+    1L,
+    "/fake/_delta_log",
+    testCommitInfo,
+    Optional.of(protocol12),
+    Optional.of(basicPartitionedMetadata),
+    Optional.empty(),
```
huan233usc: nit /* newProtocolOpt */

huan233usc (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/DefaultFileSystemManagedTableOnlyCommitter.java
```diff
@@ -39,7 +52,37 @@ public CommitResponse commit(
...
+            return null;
+          },
+          String.format("Write file actions to JSON log file `%s`", jsonCommitFile));
+    } catch (FileAlreadyExistsException e) {
+      throw new CommitFailedException(
+          true /* retryable */,
+          true /* conflict */,
+          "Concurrent write detected for version " + commitMetadata.getVersion(),
+          e);
+    } catch (IOException e) {
```
allisonport-db, line 76: For the retrying: we are mainly undecided about how Kernel / in the future the Connector reacts to failed commit attempts right? There's no debate about what exceptions the committer should throw like here? i.e. is there any reason we should make this isRetryable=false ?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/DefaultFileSystemManagedTableOnlyCommitter.java
```diff
@@ -39,7 +52,37 @@ public CommitResponse commit(
...
+      throw new CommitFailedException(
+          true /* retryable */,
+          false /* conflict */,
+          "Failed to write commit file due to I/O error: " + e.getMessage(),
+          e);
+    }
```
allisonport-db, line 82: And I guess we decided to just propagate other RuntimeExceptions?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/DefaultFileSystemManagedTableOnlyCommitter.java
scottsand-db: Great question. If things are reasonably retryable, we should throw a retryable exception.

Worst case, we try, and the file *already does exist*. Then, really, the question/semantic is what do we do with a `retryable=true, isConflict=true` exception *after* we have already seen a `retryable=true, isConflict=false` exception.

Well, we need to see if that 2nd exception is actually just us seeing the successful write that, somehow (network hiccup) causes the 1st exception!

So Kernel txn.commit rety loop can jsut handle this. You can either throw (totally valid) or try and look at the `.json` file and see which case it is

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/DefaultFileSystemManagedTableOnlyCommitter.java
scottsand-db: Yup! That's what the UC-Commit-Coordinator-Client does today! (for example)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/commit/DefaultFileSystemManagedTableOnlyCommitter.java
scottsand-db: We will document this well inside of txn.commit

allisonport-db (approved):

PR #5042, opened 2025-08-07
[Kernel] Have the txn invoke the committer and handle CommitFailedException

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/5042/files) to review incremental changes.
- [**stack/kernel_txn_invoke_committer**](https://github.com/delta-io/delta/pull/5042) [[Files changed](https://github.com/delta-io/delta/pull/5042/files)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

This PR updates the kernel `txn.commit` retry loop to invoke the committer.

I also just update the `Transaction` API to have the committer -- the committer needs to be on _that_ API, not Snapshot, because in the future TransactionBuilder world, there is no Snapshot when you do a create!

## How was this patch tested?

New UTs.

Some new exceptions.


kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -76,14 +78,19 @@ public class TransactionImpl implements Transaction {
 
   private final UUID txnId = UUID.randomUUID();
 
-  /* If the transaction is defining a new table from scratch (i.e. create table, replace table) */
```
scottsand-db, line 79: I just change the variable order to match the constructor

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -313,39 +323,89 @@ private long commitWithRetry(
...
+                "Commit attempt {} for version {} failed with a retryable exception due to a "
+                    + "physical conflict. Performing conflict resolution and trying again. "
+                    + "Exception: {}",
+                attempt,
+                commitAsVersion,
+                cfe);
+
             TransactionRebaseState rebaseState =
-                resolveConflicts(engine, commitAsVersion, attemptCommitInfo, numTries, dataActions);
+                resolveConflicts(engine, commitAsVersion, attemptCommitInfo, attempt, dataActions);
```
scottsand-db, line 383: todo better logging inside of `resolveConflicts` too

kernel/kernel-api/src/main/java/io/delta/kernel/Transaction.java
```diff
@@ -83,6 +85,10 @@ public interface Transaction {
    */
   Row getTransactionState(Engine engine);
 
+  /** @return a committer that owns and controls commits to this table */
+  @Experimental
+  Committer getCommitter();
+
```
allisonport-db, line 91: Is this still configured however when building the Snapshot (and in the case of create when building the create txn)? Or should it always instead be during txn building (for existing tables as well)?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -119,6 +122,7 @@ public TransactionImpl(
       Operation operation,
       Optional<Protocol> newProtocol,
       Optional<Metadata> newMetadata,
+      Committer committer,
```
allisonport-db, line 126: cool once this is merged I can update https://github.com/delta-io/delta/pull/5044 to configure the committer for create & pass it down

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -313,39 +323,89 @@ private long commitWithRetry(
...
+            //         be able to recover from this.
+            throw new RuntimeException(
+                String.format(
+                    "Commit attempt %d for version %d failed with a non-retryable exception.",
+                    attempt, commitAsVersion),
+                cfe);
+          } else if (attempt >= getMaxCommitAttempts()) {
+            // Case 2: Despite the error being retryable, we have exhausted the maximum number of
+            //         retries. We must throw here, too.
+            throw new MaxCommitRetryLimitReachedException(commitAsVersion, maxRetries, cfe);
```
allisonport-db, line 349: Should we distinguish here between `isConflict` and not? In theory this could be a similar case where the commit was actually successful right?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/TransactionCommitLoopSuite.scala
```diff
@@ -0,0 +1,189 @@
...
+
+      assert(result.getVersion == 1)
+      assert(attemptCount == attemptNumberToSucceedAt)
+      assert(attemptedFilePaths.size == 1) // we should only be attempting to write 001.json
+      assert(result.getTransactionReport.getTransactionMetrics.getNumCommitAttempts ==
+        attemptNumberToSucceedAt)
+    }
+  }
+
+  // TODO: Transaction will fail on CFE(isRetryable=false, isConflict=true/false). The default
```
allisonport-db: Is it worth checking that we *don't* retry for misc. runtime exceptions? wdyt?

kernel/kernel-api/src/main/java/io/delta/kernel/Transaction.java
scottsand-db: It should be configured when building the snapshot and when bilding the createTxnBuilder 👍 

This "creation" is done from the catalogPlugin -- it has all this information. If we just get back a snapshot without a Committer, and later on the Engine wants to commit a transaction, it might not have access to that catalog plugin reference anymore.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
scottsand-db: By "distinguish" do you mean throw different top-level exception types?

Because the root cause of the exception, the `cfe`, is passed into this top-level exception, so that will be visible.

I'm inclined to do the minimal change here -- if there's a desire for that specifity in the future, we can add it 👍

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/TransactionCommitLoopSuite.scala
scottsand-db: sg!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaErrors.java
```diff
@@ -367,6 +368,21 @@ public static KernelException enablingClusteringOnPartitionedTableNotAllowed(
             tablePath, partitionColNames, clusteringCols));
   }
 
+  public static ConcurrentWriteException unknownIfSafeRetryConcurrentWriteException(
+      long commitVersion, int commitAttempt, CommitFailedException cfe) {
+    final String msg =
+        String.format(
+            "Commit attempt %d for table version %d failed due to a concurrent write conflict. "
+                + "However, a previous commit attempt, call it C, also failed without conflict, "
```
raveeram-db: any chance we can simplify this error message? And also add a bit more context on the impact if the Kernel were to retry and how the client can unblock this table (e.g. "..if we retry, we could end up with duplicate records. please manualy validate the commit history..") or something to that effect

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaErrors.java
```diff
@@ -367,6 +368,21 @@ public static KernelException enablingClusteringOnPartitionedTableNotAllowed(
             tablePath, partitionColNames, clusteringCols));
   }
 
+  public static ConcurrentWriteException unknownIfSafeRetryConcurrentWriteException(
```
raveeram-db: Nit: `unknownIfSafeToRetryConcurrentWriteException `

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -313,39 +323,89 @@ private long commitWithRetry(
...
-            // TODO: reconcile fileSizeHistogram.
-            transactionMetrics.resetActionMetricsForRetry();
           }
         }
-        numTries++;
+        // We will be retrying the commit.
+        //
+        // Action counters may be partially incremented from previous tries, reset the counters
+        // to 0 and drop fileSizeHistogram
+        // TODO: reconcile fileSizeHistogram.
```
raveeram-db: Could we please create issues for these TODOs

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaErrors.java
scottsand-db: Will do!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
scottsand-db: xx @huan233usc

scottsand-db: @huan233usc can you fill this out with more details? Thanks! https://github.com/delta-io/delta/issues/5047

kernel/kernel-api/src/main/java/io/delta/kernel/exceptions/CommitStateUnknownException.java
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+
+package io.delta.kernel.exceptions;
+
+import io.delta.kernel.annotation.Evolving;
+import io.delta.kernel.commit.CommitFailedException;
+
+@Evolving
+public class CommitStateUnknownException extends ConcurrentWriteException {
```
emkornfield: Would people already be catching ConcurrentWriteException?  If they are it seems like they could easily mishandle this exception?

emkornfield: nit: Java doc

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -313,52 +324,101 @@ private long commitWithRetry(
...
-        } catch (FileAlreadyExistsException fnfe) {
-          logger.info(
-              "Concurrent write detected when committing as version = {}.", commitAsVersion);
-          if (numTries < maxRetries) {
-            // only try and resolve conflicts if we're going to retry
+        } catch (CommitFailedException cfe) {
+          if (!cfe.isRetryable()) {
+            // Case 1: Non-retryable exception. We must throw this. We don't expect connectors to
+            //         be able to recover from this.
+            throw new RuntimeException(
```
emkornfield: KernelException? Or comment why it isn't.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -313,52 +324,101 @@ private long commitWithRetry(
...
+            // Case 3: No conflict => No conflict resolution needed. Just retry with same version.
+            logger.warn(
+                "Commit attempt {} for table version {} failed with a retryable exception and "
+                    + "without conflict. Skipping conflict resolution and trying again. "
+                    + "Exception: {}",
+                attempt,
+                commitAsVersion,
+                cfe);
+            seenRetryableNonConflictException = true;
+          } else if (seenRetryableNonConflictException) {
```
emkornfield, line 358: add an assert that this is a conflict?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -313,52 +324,101 @@ private long commitWithRetry(
...
           transactionMetrics.commitAttemptsCounter.increment();
           return doCommit(
               engine, commitAsVersion, attemptCommitInfo, dataActions, transactionMetrics);
-        } catch (FileAlreadyExistsException fnfe) {
-          logger.info(
-              "Concurrent write detected when committing as version = {}.", commitAsVersion);
-          if (numTries < maxRetries) {
-            // only try and resolve conflicts if we're going to retry
+        } catch (CommitFailedException cfe) {
+          if (!cfe.isRetryable()) {
```
emkornfield, line 341: for case 1 and 2 it might make the code easier to follow if you have to helper guard methods:
```
throwIfNotRetryable(cfe);
throwIfRetriesExceeded(attempt, getMaxCommitAttempts();
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -313,52 +324,101 @@ private long commitWithRetry(
...
+            // - Commit Attempt #2: FileAlreadyExistsException => CFE(retryable=true, conflict=true)
+            //         Should we retry this commit? If it's case (A), then we should not, as we are
+            //         just conflicting with our previous commit attempt. If it's case (B), then we
+            //         should retry, since we are conflicting with some *other* writer's commit. In
+            //         the future we can add detection capabilities between these two cases (e.g.
+            //         check if the CommitInfo action is present and has a txnId, else compare the
+            //         other contents of the delta files).
+            throw new CommitStateUnknownException(commitAsVersion, attempt, cfe);
+          } else {
+            // Case 5: There is a conflict, and we have not previously seen a retryable and
```
emkornfield, line 378: add an assert that this is a conflict.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
emkornfield, line 358: moving the two conflict cases to a helper method might also make the flow easier to understand here.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/commit/DefaultCommitterSuite.scala
```diff
@@ -60,10 +61,11 @@ class DefaultCommitterSuite extends AnyFunSuite
       val schema = new StructType().add("col1", IntegerType.INTEGER)
       val metadata = testMetadata(schema, Seq[String]())
       val committer = TableManager.loadSnapshot(dataPath.toString)
+        .asInstanceOf[SnapshotBuilderImpl]
         .withProtocolAndMetadata(readProtocol, metadata)
         .atVersion(1)
         .build(emptyMockEngine)
-        .getCommitter
+        .getCommitter // Currently, the only `getCommitter` public API is on Transaction
```
emkornfield: nit: This comment seems like a partial thought to me?  What else should be expected here?

kernel/kernel-api/src/main/java/io/delta/kernel/exceptions/CommitStateUnknownException.java
scottsand-db: Good question. `txn.commit` doesn't throw any checked exceptions -- we aren't expecting engines to be catching these exceptions and doing anything particularly smart or helpful with them.

`txn.commit` means: I'm an engine. I'm asking kernel to go and do the commit and retry until it succeeds, or max retry limit reached, or some failure occurs. If a failure occurs, welp, that's it, txn failed, throw.

In the future, we will give the engine the complete control of the commit loop. Then, the Engine will have to catch `CommitFailedException` and handle it accordingly.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
scottsand-db: cc @allisonport-db any input here?

I _think_ it should be `RuntimeException` because the error originated from the Committer, not from Kernel.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/commit/DefaultCommitterSuite.scala
scottsand-db: I will elaborate!

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/TransactionCommitLoopSuite.scala
```diff
@@ -0,0 +1,223 @@
...
+      val initialTxn = createWriteTxnBuilder(table).withSchema(engine, testSchema).build(engine)
+      commitTransaction(initialTxn, engine, emptyIterable()) // 000.json
+
+      val kernelTxn = createWriteTxnBuilder(table).withMaxRetries(5).build(engine)
+
+      // Create 001.json. This will make the engine throw a FileAlreadyExistsException when trying
+      // to write 001.json. The default committer will turn this into a
+      // CFE(isRetryable=true, isConflict=true).
+
+      spark.range(0, 10)
```
emkornfield: not sure if we have policies on this, but it seems like it would be lighter wait to avoid spark in tests when possible (it seems like a heavy wait dependency).  Could do the commit with just kernel.

emkornfield: maybe abstract to a helper method `addConflictingTransaction`

emkornfield (changes requested):
Biggest concern is modelling of the CommitStateUnknown exception, and whether it should extend the current exception.

kernel/kernel-api/src/main/java/io/delta/kernel/exceptions/CommitStateUnknownException.java
emkornfield: Are we not already throwing CommitFailedExceptions?  I believe @allisonport-db but somehow connectors must be handling failed exceptions today at some level?  Are we saying every time a connector receives a CommitFailedException that entire app exits and human is reviewing the state of the table?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
scottsand-db: btw -- is this just defensive programming? because we only reach here if the `else if (!cfe.isConflict()) {` evaluates to false

emkornfield: Yes, it serves two purposes.  Defensive programming, and concise documentation.  I have a small brain-cache and needed to reread the code once or twice to convince myself of the state.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
scottsand-db: Hey @emkornfield -- Personally, I do like the structure of the `if, if else, else` statements here -- in your code above those `throwIf` would then contain the if statement -- which messes with the handling of the if, if else, etc. here

But I do agree having tiny helper methods in each `if` block would be nice.

- case 1: i can refactor to `DeltaErrors.throw____` <--- only change really that I like
- case 2: already a 1 liner
- case 3: log + set a variable -- I'm inclined to keep this?
- case 4: comment + throw -- I'm inclined to keep this?
- case 5: set a bunch of member variables --  I'm inclined to keep this?

wdyt?

emkornfield: If you and @allisonport-db are happy with current structure no need to change on my account. Note that the catch block is ~50 lines of code/comments (I know mostly comments which is a good thing) but generally it seems like some way of shortening what a reader needs to see at this level of abstraction would be my preference.

kernel/kernel-api/src/main/java/io/delta/kernel/exceptions/CommitStateUnknownException.java
scottsand-db: > Are we not already throwing CommitFailedExceptions? 

We are not. That's a checked exception thrown by the `Committer`. `txn.commit` does not re-throw it as a checked exception

scottsand-db: > but somehow connectors must be handling failed exceptions today at some level?

Not that I know of. Today, Kernel does the commit. And then Connectors either (a) learn it was successful, or (b) learn that it failed and throw.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
scottsand-db: Implemented some feedback and simplified!

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/commit/DefaultCommitterSuite.scala
scottsand-db: I can just delete this tbd 👍

kernel/kernel-api/src/main/java/io/delta/kernel/exceptions/CommitStateUnknownException.java
emkornfield: OK, thinking about this some more I don't think it makes sense for this to extend ConcurrentWriteException, even if all the above holds about visibility.

Reasoning:
1.  We don't actually know whether a concurrent write caused this issue or not (i.e. its not clear this is a "is a" relationship from a class hierarchy)
2. If at some point connectors do start handling these exceptions they probably want to handle ConcurrentWriteException separate from TransactionStateUnknown.

In the long term when connectors have full control over the loop this probably won't make a difference.  Is there an argument for making this extend the ConcurrentWriteException?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -341,26 +341,18 @@ private long commitWithRetry(
...
-                "Commit attempt {} for table version {} failed with a retryable exception and "
-                    + "without conflict. Skipping conflict resolution and trying again. "
-                    + "Exception: {}",
-                attempt,
-                commitAsVersion,
-                cfe);
+            printLogForRetryableNonConflict(attempt, commitAsVersion, cfe);
             seenRetryableNonConflictException = true;
           } else if (seenRetryableNonConflictException) {
+            assert cfe.isRetryable() && cfe.isConflict();
```
emkornfield: we might want to make this a argument check so it always runs (not sure what we do in other places for assertion like logic, as long as we are being consistent it is fine).

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
emkornfield: Thanks, it still seems a little bit long, so want to keep this thread open for another reviewer to chime in, if they have thoughts.

kernel/kernel-api/src/main/java/io/delta/kernel/exceptions/CommitStateUnknownException.java
scottsand-db: Fair points!

> Is there an argument for making this extend the ConcurrentWriteException?

We are trying to commit N.json, and we learn that N.json already exists. We don't know if _we_ wrote it, or if some other writer wrote it.

Is that your argument? i.e. because we are not sure, this should not be a ConcurrentWriteException?

I'm easy. I can make this not extend it 👍

emkornfield: > I'm easy. I can make this not extend it 👍

Thanks, I think I feel better if we don't extend.

emkornfield (approved):
Changes look reasonable as long as second reviewer is happy and we don't extend the ConcurrentWriteException.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
allisonport-db: Yeah I think this is kind of a gray area -- I'm not exactly sure what it should be but I think propagating the exception thrown from the committer seems valid to me

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
allisonport-db: I think the length is fine for me personally since I find it readable. This is a tiny nit but for me when reading it I think the many branches of if, if else, if else where the conditions each depend on different things are a little more confusing for me to parse.

Like
```
if (!isRetryable) {
   do X
    // Okay now i have to remember all branches below this must be retryable
} else if (too many attempts) {
   do Y
} else (isNotConflict) {
   do Z
    // Okay now i have to remember all branches below this must be conflict = true
} else (seenRetryableNonConflictException) {

} else {
}
```

where as it would be easier to parse if we kept if+else at levels that the conditions pair with each other or maybe just returned early (which I think is more similar to what @emkornfield was suggesting, sort of exit early for those first 2 cases)?

Maybe don't need to be separate functions for me personally, but I think taking them out of the long branch statement would make it easier to parse. Again this is super minor for me but I remember thinking it was a little confusing to track when I first saw it.

```
if (!isRetryable) {
   throw error
}
 if (too many attempts) {
   throw error
}

// Now just the retry/resolve logic together in 1 if/else
```

kernel/kernel-api/src/main/java/io/delta/kernel/exceptions/CommitStateUnknownException.java
allisonport-db: I know this is resolved but +1 not extending it makes me feel better about it as well -- since it's not _necessarily 100% of the time with certainty_ a concurrent write exception

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
```diff
@@ -213,6 +215,11 @@ protected TransactionImpl buildTransactionInternal(
         snapshot -> validateWriteToExistingTable(engine, snapshot, isCreateOrReplace));
     validateTransactionInputs(engine, isCreateOrReplace);
 
+    final Committer committer =
+        latestSnapshot
+            .map(SnapshotImpl::getCommitter)
```
huan233usc, line 220: Will it be problematic if a catalog owned cc table falling back to FileSystemManagedTableOnlyCommitter?

huan233usc (approved):
lgtm, just 1 q

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionBuilderImpl.java
scottsand-db: Yes -- engines must inject the right committer

PR #5043, opened 2025-08-08
[Spark] Fix CommitInfo Deserialization and retain legacy deserialization for legacy callsites

dhruvarya-db:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR completely fixes `CommitInfo`.operationParameters deserialization. A new method `getLegacyPostDeserializationOperationParameters` has been added to retain the legacy behaviour for old callsites. This PR, by itself, should be a no-op in terms of behaviour changes as the legacy callsites will use `getLegacyPostDeserializationOperationParameters` for now.

The newly introduced `JsonMapDeserializer` effectively performs an inverse of `JsonMapSerializer`. The original problem is explained in detail below:

The in-memory representation of operation params of any Delta Operation can be a combination of json encoded strings, simple strings, or primitives single-encoded as strings. The values can be any of:
- `"123"`
- `"true"`
- `"1.0"`
- `"\"true\""`
- `"\"1.0\""`
- or more complex json encoded strings

Due to how `JsonMapSerializer` strips one level of encoding for these values during serialization, these can end up being written out in this form:

- `"123"` → `123`
- `"true"` → `true`
- `"1.0"` → `1.0`
- `"\"true\""` → `"true"`
- `"\"1.0\""` → `"1.0"`

Since operationParameters is a `Map[String, String]`, during the deserialization phase, the deserializer intelligently converts primitive types from above to simple strings:

- `"123"` → `123` → `"123"`
- `"true"` → `true` → `"true"`
- `"1.0"` → `1.0` → `"1.0"`
- `"\"true\""` → `"true"` → `"true"`
- `"\"1.0\""` → `"1.0"` → `"1.0"`

Since we stripped one level of encoding during serialization, we need to add it back to get closer to the original in-memory representation:

- `"123"` → `123` → `"123"` → `"\"123\""`
- `"true"` → `true` → `"true"` → `"\"true\""`
- `"1.0"` → `1.0` → `"1.0"` → `"\"1.0\""`
- `"\"true\""` → `"true"` → `"true"` → `"\"true\""`
- `"\"1.0\""` → `"1.0"` → `"1.0"` → `"\"1.0\""`

Note how values that were single-encoded as strings originally are now double-encoded as strings (e.g., `"true"` → `true` → `"true"` → `"\"true\""`). This is because the deserializer converted the primitive values to strings as well as retained simple strings as strings. In this process, we lost some information about the original values.

Solution:

To fix this, we first deserialize the values as a `java.util.HashMap[String, Any]`:

- `"true"` → `true` → `true` (type: Boolean)
- `"\"true\""` → `"true"` → `"true"` (type: String)

And then JsonEncode them once to get the original in-memory representation:

- `"true"` → `true` → `true` → `"true"`
- `"\"true\""` → `"true"` → `"true"` → `"\"true\""`

## How was this patch tested?

Adds a new suite that tests both the new deserialization as well as the legacy behaviour. Most DeltaOperations are covered by this suite.

No

### DIFF
--- a/spark/src/test/scala/org/apache/spark/sql/delta/ActionSerializerSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/ActionSerializerSuite.scala
@@ -236,7 +236,7 @@ class ActionSerializerSuite extends QueryTest with SharedSparkSession with Delta
 
     assert(operationParameters === expectedOperationParameters)
 
-    val expectedPrettyOperationParameters = Map(
+    val expectedLegacyOperationParameters = Map(
       "catalogTable" -> "t1",
       "numFiles" -> "23",
       "partitionedBy" -> "[\"a\",false]",
@@ -253,14 +253,14 @@ class ActionSerializerSuite extends QueryTest with SharedSparkSession with Delta
     assert(roundTrippedCommitInfo1.operationParameters === expectedOperationParameters)
     assert(
       roundTrippedCommitInfo1.getLegacyPostDeserializationOperationParameters ===
-        expectedPrettyOperationParameters)
+        expectedLegacyOperationParameters)
 
     val roundTrippedCommitInfo2 =
       JsonUtils.fromJson[CommitInfo](JsonUtils.toJson(roundTrippedCommitInfo1))
     assert(roundTrippedCommitInfo2.operationParameters === expectedOperationParameters)
     assert(
       roundTrippedCommitInfo2.getLegacyPostDeserializationOperationParameters ===
-        expectedPrettyOperationParameters)
+        expectedLegacyOperationParameters)
   }
 
   test("round trip of operation parameters: non-primitive types in operation parameters") {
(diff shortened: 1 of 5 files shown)


sumeet-db:
LGTM! Thank you for fixing this!

1. Does `java.util.HashMap[String, Any]` also handle nulls properly? I presume yes, since we have test coverage for it.
2. Can we update the remaining callsites (if we can that is) in followup PRs?

spark/src/test/scala/org/apache/spark/sql/delta/ActionSerializerSuite.scala
```diff
@@ -213,6 +214,131 @@ class ActionSerializerSuite extends QueryTest with SharedSparkSession with Delta
...
+      "numFiles" -> "23",
+      "partitionedBy" -> "\"[\\\"a\\\",false]\"",
+      "sourceFormat" -> "\"parquet\"",
+      "collectStats" -> "false",
+      "k1" -> "null",
+      "" -> "null")
+
+    assert(operationParameters === expectedOperationParameters)
+
+    val expectedPrettyOperationParameters = Map(
```
sumeet-db: Nit: expectedLegacyOperationParameters?

sumeet-db (approved):
Nice fix!

dhruvarya-db:
> 1. Does java.util.HashMap[String, Any] also handle nulls properly? I presume yes, since we have test coverage for it.

Yes, they are. I copied tests from your PR https://github.com/delta-io/delta/pull/4311/files to validate this.

> Can we update the remaining callsites (if we can that is) in followup PRs?

Sounds good to me. I agree that doing that in follow up PRs would be better to not increase the scope of this change

PR #5044, opened 2025-08-08
[KERNEL] Intent-based builders for both UPDATE TABLE and CREATE TABLE

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Adds intent based builders for create and update.

## How was this patch tested?

Tests coming soon...

Yes new builder APIs.


kernel/kernel-api/src/main/java/io/delta/kernel/Snapshot.java
```diff
@@ -98,4 +99,10 @@ public interface Snapshot {
 
   /** @return a committer that owns and controls commits to this table */
   Committer getCommitter();
+
+  /**
+   * @return a {@link UpdateTableTransactionBuilder} to build an update table transaction
+   * @since 3.4.0
```
scottsand-db, line 101: nit: since 4.1.0

kernel/kernel-api/src/main/java/io/delta/kernel/TableManager.java
```diff
@@ -43,5 +46,17 @@ static SnapshotBuilder loadSnapshot(String path) {
...
 
-  // TODO: static CreateTableTransactionBuilder buildCreateTableTransaction(...)
+  /**
+   * Creates a {@link CreateTableTransactionBuilder} to build a create table transaction.
+   *
+   * @param path the file system path for the delta table being created
+   * @param engineInfo information about the engine that is making the updates.
+   * @param schema the schema for the delta table being created
+   * @return create table builder instance to build the transaction
+   * @since 3.4.0
```
scottsand-db, line 56: ditto

kernel/kernel-api/src/main/java/io/delta/kernel/TableManager.java
```diff
@@ -43,5 +46,17 @@ static SnapshotBuilder loadSnapshot(String path) {
...
+   */
+  static CreateTableTransactionBuilder buildCreateTableTransaction(
+      String path, String engineInfo, StructType schema) {
```
scottsand-db: nit: can this order be `path, schema, engineInfo` -- plus that is what aligns with the constructor below

kernel/kernel-api/src/main/java/io/delta/kernel/internal/CreateTableTransactionBuilderImpl.java
```diff
@@ -0,0 +1,137 @@
...
+            tableProperties.orElse(Collections.emptyMap()),
+            partitionColumns,
+            clusteringColumns);
+
+    Path dataPath = new Path(resolvedPath);
+    Path logPath = new Path(dataPath, "_delta_log");
+    return new TransactionImpl(
+        true, // isCreateOrReplace
+        dataPath,
+        logPath,
```
scottsand-db: nit: could we simplify our TransactionImpl constructor by 1 param and just pass in the `dataPath`? and `txnImpl` creates its own `logPath`?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/CreateTableTransactionBuilderImpl.java
```diff
@@ -0,0 +1,137 @@
...
+        txnMetadata.newProtocol,
+        txnMetadata.newMetadata,
+        Optional.empty(), // no setTransaction for create table
+        txnMetadata.physicalNewClusteringColumns,
+        maxRetries,
+        0, // logCompactionInterval - using default for create table
+        System::currentTimeMillis);
+  }
+
+  private String resolveTablePath(Engine engine, String tablePath) {
```
scottsand-db: Duplicate code to what is done in the SnapshotFactory -- can we move this to a common util?

https://github.com/delta-io/delta/blob/3c1273b5e66db61df76c4b4db7af411714a764be/kernel/kernel-api/src/main/java/io/delta/kernel/internal/table/SnapshotFactory.java#L115

kernel/kernel-api/src/main/java/io/delta/kernel/internal/CreateTableTransactionBuilderImpl.java
```diff
@@ -0,0 +1,137 @@
...
+                spec ->
+                    spec.getPartitionColumns().stream()
+                        .map(col -> col.getNames()[0])
+                        .collect(Collectors.toList()));
+    Optional<List<Column>> clusteringColumns =
+        dataLayoutSpec
+            .filter(DataLayoutSpec::hasClustering)
+            .map(DataLayoutSpec::getClusteringColumns);
+
+    // Build the transaction metadata using TransactionMetadataFactory
```
scottsand-db: tiny tiny tiny nit: IMO, the code is clear enough! No need for a comment. Just my personal preference though 👍

kernel/kernel-api/src/main/java/io/delta/kernel/internal/CreateTableTransactionBuilderImpl.java
```diff
@@ -0,0 +1,137 @@
...
+    TransactionMetadataFactory.Output txnMetadata =
+        TransactionMetadataFactory.buildCreateTableMetadata(
+            resolvedPath,
+            schema,
+            tableProperties.orElse(Collections.emptyMap()),
+            partitionColumns,
```
scottsand-db, line 97: Remind me: is it worth considering having the factory take in the DataLayoutSpec instead? would other builders do similar logic to the above?

or, for example, new partition columns isn't possible on UPDATE, so that should *not* be in the common factory API?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/UpdateTableTransactionBuilderImpl.java
```diff
@@ -0,0 +1,211 @@
...
+import io.delta.kernel.expressions.Column;
+import io.delta.kernel.internal.actions.SetTransaction;
+import io.delta.kernel.internal.tablefeatures.TableFeatures;
+import io.delta.kernel.transaction.UpdateTableTransactionBuilder;
+import io.delta.kernel.types.StructType;
+import java.util.*;
+
+public class UpdateTableTransactionBuilderImpl implements UpdateTableTransactionBuilder {
+
+  /** Timestamp when this builder was created, used for populating any {@link SetTransaction} set */
```
scottsand-db: `{@link SetTransaction} set */` --> instead of `set` should it just be `setTxnId` or something like that? it's not a `Set` collection?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/UpdateTableTransactionBuilderImpl.java
```diff
@@ -0,0 +1,211 @@
...
+  private final long currentTimeMillis = System.currentTimeMillis();
```
scottsand-db: idea: `current` tells me, yeah, this is the "current" time, but would we rather name it do what it really means? like `txnStartTime?` or `builderStartTime`? maybe that's a silly idea -- wdyt? totally fine with keeping it the current naming.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/UpdateTableTransactionBuilderImpl.java
```diff
@@ -0,0 +1,211 @@
...
+        txnMetadata.newMetadata,
+        setTxnOpt,
+        txnMetadata.physicalNewClusteringColumns,
+        maxRetries,
+        logCompactionInterval,
+        // TODO: support configuring clock if needed
+        System::currentTimeMillis);
+  }
+
+  private void validateTablePropertiesAddedRemovedOverlap() {
```
scottsand-db: `validateTableProperties*No*AddedRemovedOverlap`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/UpdateTableTransactionBuilderImpl.java
```diff
@@ -0,0 +1,211 @@
...
+   */
+  private int maxRetries = 200;
+
+  /** Number of commits between producing a log compaction file. */
+  private int logCompactionInterval = 0;
+
+  public UpdateTableTransactionBuilderImpl(
+      SnapshotImpl snapshot, String engineInfo, Operation operation) {
+    if (operation == Operation.CREATE_TABLE) {
+      throw new TableAlreadyExistsException(snapshot.getPath(), "Operation = CREATE_TABLE");
```
scottsand-db: This seems like an IllegalArgumentException?
you can have a common private helper method that prints out `Operation X is not compatible with <className>`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/UpdateTableTransactionBuilderImpl.java
```diff
@@ -0,0 +1,211 @@
...
+  @Override
+  public UpdateTableTransactionBuilder withLogCompactionInterval(int logCompactionInterval) {
+    checkArgument(logCompactionInterval >= 0, "logCompactionInterval must be >= 0");
+    this.logCompactionInterval = logCompactionInterval;
+    return this;
+  }
+
+  @Override
+  public Transaction build(Engine engine) {
+
```
scottsand-db: nit: whitespace

kernel/kernel-api/src/main/java/io/delta/kernel/internal/UpdateTableTransactionBuilderImpl.java
```diff
@@ -0,0 +1,211 @@
...
+  private Optional<Map<String, String>> tablePropertiesAddedOpt = Optional.empty();
+  private Optional<Set<String>> tablePropertiesRemovedOpt = Optional.empty();
+  private Optional<StructType> updatedSchemaOpt = Optional.empty();
+  private Optional<List<Column>> inputLogicalClusteringColumnsOpt = Optional.empty();
+  /**
+   * Number of retries for concurrent write exceptions to resolve conflicts and retry commit. In
+   * Delta-Spark, for historical reasons the number of retries is really high (10m). We are starting
+   * with a lower number by default for now. If this is not sufficient we can update it.
+   */
+  private int maxRetries = 200;
```
scottsand-db: `200` seems like a magic number that is re-used across a few classes -- this should be an internal static final default value, eh?

scottsand-db: you could also just pass in an `Optional.empty` into `txn` and then `txn` itself gets to decide what the default is

not trying to add too much more work to your plate, we could do that later too, im ez

kernel/kernel-api/src/main/java/io/delta/kernel/internal/UpdateTableTransactionBuilderImpl.java
```diff
@@ -0,0 +1,211 @@
...
+          }
+        });
+
+    boolean needsMetadataOrProtocolUpdate =
+        updatedSchemaOpt.isPresent() // schema evolution
+            || tablePropertiesAddedOpt.isPresent() // table properties updated
+            || tablePropertiesRemovedOpt.isPresent() // table properties unset
+            || inputLogicalClusteringColumnsOpt.isPresent(); // update clustering columns
+
+    if (!needsMetadataOrProtocolUpdate) {
```
scottsand-db: remind me: did this break a test?

kernel/kernel-api/src/main/java/io/delta/kernel/transaction/CreateTableTransactionBuilder.java
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Builder for creating a {@link Transaction} to create a new Delta table.
+ *
+ * @since 3.4.0
+ */
+@Evolving
+public interface CreateTableTransactionBuilder {
+
+  /**
+   * Set table properties (configuration) for the new Delta table. Note, user-properties (those
```
scottsand-db: `configuration` is more an internal (poorly named) term, eh? we can delete? `table properties` is very very clear 👍 

nit: `Note:` is usually on a new line `<p>`

kernel/kernel-api/src/main/java/io/delta/kernel/transaction/DataLayoutSpec.java
```diff
@@ -0,0 +1,153 @@
...
+
+  private DataLayoutSpec(List<Column> partitionColumns, List<Column> clusteringColumns) {
+    if (partitionColumns != null && clusteringColumns != null) {
+      throw new IllegalArgumentException("Cannot specify both partition and clustering columns");
+    }
+    if (partitionColumns != null && partitionColumns.isEmpty()) {
+      throw new IllegalArgumentException("Partition columns cannot be empty");
+    }
+    this.partitionColumns = partitionColumns;
+    this.clusteringColumns = Collections.unmodifiableList(clusteringColumns);
```
scottsand-db: can we use Optional instead of `null`? this will throw if `clusteringColumns` is null --- i'd rather just requireNonNull and go from there

kernel/kernel-api/src/main/java/io/delta/kernel/transaction/DataLayoutSpec.java
```diff
@@ -0,0 +1,153 @@
...
+import java.util.List;
+
+/**
+ * Specification for the data layout of a Delta table, including partitioning and clustering
+ * configurations.
+ *
+ * <p>This class supports different layout strategies:
+ *
+ * <ul>
+ *   <li><b>Unpartitioned:</b> No special data layout
```
scottsand-db: is a `clustered` table unpartitioned?

to me it seems like the three cases are:
- no data layout spec: this is not partitioned AND not clustered
- partitioned
- clustered

kernel/kernel-api/src/main/java/io/delta/kernel/transaction/DataLayoutSpec.java
```diff
@@ -0,0 +1,153 @@
...
+    return partitionColumns != null && !partitionColumns.isEmpty();
+  }
+
+  /**
+   * Returns true if this layout specification includes clustering.
+   *
+   * <p>Clustering can be enabled even with empty clustering columns, which indicates that
+   * clustering is enabled on the table but no specific columns are defined yet.
+   */
+  public boolean hasClustering() {
```
scottsand-db: do we ever need to ask/differentiate: clustering is SUPPORTED but no clustering columns ? (empty list) vs clustering is supported WITH clustering columns?

kernel/kernel-api/src/main/java/io/delta/kernel/transaction/UpdateTableTransactionBuilder.java
```diff
@@ -0,0 +1,160 @@
...
+  UpdateTableTransactionBuilder withTablePropertiesRemoved(Set<String> propertyKeys);
+
+  /**
+   * Update the clustering columns for the table and enable clustering if it is not already enabled.
+   * Note: clustering cannot be enabled for a partitioned table.
+   *
+   * @param clusteringColumns The columns to cluster by. Cannot be null.
+   * @throws IllegalArgumentException if the table is partitioned
+   */
+  // TODO: should this be a DataLayoutSpec instead?
```
scottsand-db: no 👍

scottsand-db (changes requested):
Looks great! Builders are SUPER simple, as they should be. Took an early look. Minor comments, thanks!

kernel/kernel-api/src/main/java/io/delta/kernel/Snapshot.java
allisonport-db: I think this will release next with 3.4.0. I wish there was a better way to make sure to update this before next release

allisonport-db: Oh but then it's not in 4.0.0 :/

kernel/kernel-api/src/main/java/io/delta/kernel/internal/CreateTableTransactionBuilderImpl.java
allisonport-db: I think this might be shared with ReplaceTableBuilder -- but notably not with Update. I honestly think there might be a fair amount of overlap between replace/create in the underlying implementation. I'm of the opinion to leave this as is for now, and address this in the individual PR for Replace where we can see what is common code and how's the best way to factor it out. It's easier to make those decisions once we have both builders being implemented.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/UpdateTableTransactionBuilderImpl.java
allisonport-db: I think I meant "set" as in a verb haha. I can clarify I can see the confusions.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/CreateTableTransactionBuilderImpl.java
scottsand-db: SGTM 👍

kernel/kernel-api/src/main/java/io/delta/kernel/internal/UpdateTableTransactionBuilderImpl.java
allisonport-db: No I think that's valid I'll update it. currentTime is not as clear

kernel/kernel-api/src/main/java/io/delta/kernel/internal/UpdateTableTransactionBuilderImpl.java
allisonport-db: yeah, let me remove this for now. We can address this separately in all of the builders at once (metadata validation when no metadata updates)

kernel/kernel-api/src/main/java/io/delta/kernel/transaction/DataLayoutSpec.java
allisonport-db: I am open to opinions on this, but my preference is that `getPartitionColumns` and `getClusteringColumns` do NOT return optional, and should not be called unless they are defined. Thus switching this to optional would just be internal and would just change how we store and access them upon return for these functions. So if we are leaving the method signatures as is, I would say `null` is fine. If you disagree with leaving the method signatures then I think that's a different point -- in which maybe?

allisonport-db: Oops I didn't see what you were referring to! I meant to remove that... instead switched to wrapping them as unmodifiable upon return for that very reason

kernel/kernel-api/src/main/java/io/delta/kernel/transaction/DataLayoutSpec.java
allisonport-db: yeah fair point, I feel like we are not consistent with terminology here. I am happy to change the docs here, thoughts on what the static factory method should be called though? I don't really like `noDataLayoutSpec`, `emptySpec`? idk

kernel/kernel-api/src/main/java/io/delta/kernel/transaction/DataLayoutSpec.java
allisonport-db: I think possibly, but then you would just use both `hasClustering()` and `getClusteringColumns()` -- no need to over-optimize this now I think

kernel/kernel-api/src/main/java/io/delta/kernel/internal/UpdateTableTransactionBuilderImpl.java
allisonport-db: yeah I prefer #2 -- let me do this separately

kernel/kernel-api/src/main/java/io/delta/kernel/TableManager.java
```diff
@@ -43,5 +46,17 @@ static SnapshotBuilder loadSnapshot(String path) {
     return new SnapshotBuilderImpl(path);
   }
 
-  // TODO: static CreateTableTransactionBuilder buildCreateTableTransaction(...)
+  /**
+   * Creates a {@link CreateTableTransactionBuilder} to build a create table transaction.
+   *
+   * @param path the file system path for the delta table being created
+   * @param engineInfo information about the engine that is making the updates.
```
scottsand-db: nit: `update` (singular) wdyt?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/CreateTableTransactionBuilderImpl.java
```diff
@@ -0,0 +1,121 @@
...
+    return this;
+  }
+
+  @Override
+  public Transaction build(Engine engine) {
+    requireNonNull(engine, "engine cannot be null");
+    String resolvedPath = resolvePath(engine, unresolvedPath);
+
+    // Extract partition and clustering columns from the data layout spec
+    Optional<List<String>> partitionColumns =
```
scottsand-db, line 83: Will any other code ever want to ask "What are the partition columns" given a DeltaLayoutSpec?

Could this be a utility on the DeltaLayoutSpec class?

TBH -- even if no one else needs to ask that, it just seems cleaner, eh?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/CreateTableTransactionBuilderImpl.java
```diff
@@ -0,0 +1,121 @@
...
+        dataPath,
+        Optional.empty(), // no existing snapshot for create table
+        engineInfo,
+        Operation.CREATE_TABLE,
+        txnMetadata.newProtocol,
+        txnMetadata.newMetadata,
+        Optional.empty(), // no setTransaction for create table
+        txnMetadata.physicalNewClusteringColumns,
+        maxRetries,
+        0, // logCompactionInterval - using default for create table
```
scottsand-db: can this be a private static final int, just like with `maxRetries` ?

scottsand-db: and btw should there be a `withLogCompactionInterval` API? or are we moving to a config?

kernel/kernel-api/src/main/java/io/delta/kernel/transaction/UpdateTableTransactionBuilder.java
```diff
@@ -0,0 +1,159 @@
...
+import io.delta.kernel.exceptions.*;
+import io.delta.kernel.expressions.Column;
+import io.delta.kernel.types.StructType;
+import io.delta.kernel.utils.CloseableIterable;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Builder for creating a {@link Transaction} to update an existing Delta table.
```
scottsand-db: instead of seeing the word `creating` here (which could be a tiny bit confusing, since this is an UPDATE txn builder not a CREATE txn builder` --> could we just say `Builds a {@link Transaction} to update an existing ..`

kernel/kernel-api/src/main/java/io/delta/kernel/transaction/DataLayoutSpec.java
```diff
@@ -0,0 +1,153 @@
...
+    return clusteringColumns != null;
+  }
+
+  /**
+   * Returns true if this is an unpartitioned layout.
+   *
+   * <p>An unpartitioned layout has neither partitioning nor clustering enabled.
+   */
+  public boolean isUnpartitioned() {
```
scottsand-db: `isNoDataLayoutSpec` or something more like that? e.g. above you say `<li><b>No data layout spec:</b> No special data layout`

scottsand-db: e.g. `hasPartitioning` = false doesn't actually mean that `isUnpartitioned=true` which is a tad confusing
but if this was `isNoDataLayoutSpec` that makes it clearer it applies to both partitioning and clustering

scottsand-db (commented):
Looking great. Just tiny nits / comments, nothing big! Looking forward to seeing the tests.

IMO the `DataLayoutSpec` could be a separate PR? With tiny tests that valdiate its APIs?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/CreateTableTransactionBuilderImpl.java
allisonport-db: added in https://github.com/delta-io/delta/pull/5056

kernel/kernel-api/src/main/java/io/delta/kernel/transaction/DataLayoutSpec.java
allisonport-db: updated this in https://github.com/delta-io/delta/pull/5056

kernel/kernel-api/src/main/java/io/delta/kernel/internal/CreateTableTransactionBuilderImpl.java
allisonport-db: I omitted it from this builder because it's not relevant (should only be relevant for updates I think?) -- we wouldn't compact on any interval for 000.json

allisonport-db: "using default" was maybe the wrong comment here

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/TransactionBuildersV2Suite.scala
```diff
@@ -0,0 +1,237 @@
...
+    } else {
+      var txnBuilder = TableManager.loadSnapshot(tablePath)
+        .asInstanceOf[SnapshotBuilderImpl].build(engine)
+        .buildUpdateTableTransaction(testEngineInfo, Operation.WRITE)
+      /*
+      TODO: in a separate PR clean up usage of createTxn such that schema != null ONLY when it
+        should either be updated for schema evolution or for a new table
+      if (schema != null) {
+        txnBuilder = txnBuilder.withUpdatedSchema(schema)
+      }
```
allisonport-db: Basically, a lot of tests provide the schema even for existing tables when there is no intention of schema evolution. Those tests fail with this change, because the tables don't support schema evolution due to column mapping not being enabled. For now, no tests that use this do schema evolution anyways, so this is not a blocker.

Here is a PR where I am trying to make those test changes https://github.com/delta-io/delta/pull/5057 (broke a lot of other tests so need to iterate on it)

allisonport-db: Example failures can be seen here https://github.com/delta-io/delta/actions/runs/16922189876/job/47950671018

kernel/kernel-api/src/main/java/io/delta/kernel/internal/CreateTableTransactionBuilderImpl.java
```diff
@@ -0,0 +1,143 @@
...
+            .map("supported"::equals)
+            .orElse(false);
+    if (isCatalogManaged) {
+      // For catalog managed tables we assume the catalog has ensured the table loc is not already
+      // a Delta table; return early
+      return;
+    }
+    // Otherwise, try loading the latest snapshot to ensure the table does not exist
+    try {
+      Snapshot snapshot = Table.forPath(engine, tablePath).getLatestSnapshot(engine);
```
scottsand-db: Use the future-proof API! `TableManager.loadSnapshot(...).build(...)`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -72,11 +69,23 @@
...
   private static final Logger logger = LoggerFactory.getLogger(TransactionImpl.class);
 
   public static final int DEFAULT_READ_VERSION = 1;
   public static final int DEFAULT_WRITE_VERSION = 2;
+  /**
+   * Default retries for concurrent write exceptions to resolve conflicts and retry commit. In
+   * Delta-Spark, for historical reasons the number of retries is really high (10m). We are starting
+   * with a lower number by default for now. If this is not sufficient we can update it.
+   */
+  private static final int defaultMaxRetries = 200;
```
scottsand-db: `DEFAULT_MAX_RETRIES`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/UpdateTableTransactionBuilderImpl.java
```diff
@@ -0,0 +1,183 @@
...
+          tablePropertiesRemovedOpt.get().stream()
+              .filter(tablePropertiesAddedOpt.get()::containsKey)
+              .collect(toSet());
+      if (!invalidPropertyKeys.isEmpty()) {
+        throw DeltaErrors.overlappingTablePropertiesSetAndUnset(invalidPropertyKeys);
+      }
+    }
+  }
+
+  private void validateOperationParam(Operation operation) {
```
scottsand-db: nit: `validateIsUpdateOperation` perhaps?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/TablePropertiesSuiteBase.scala
```diff
@@ -223,3 +233,5 @@ class TablePropertiesSuite extends DeltaTableWriteSuiteBase {
     assert(keys.forall(!metadata.getConfiguration.containsKey(_)))
   }
 }
+
+class TablePropertiesTransactionBuilderV1Suite extends TablePropertiesSuiteBase {}
```
scottsand-db: 1. can we move this to the top of the class? would be much more visible, above the abstract trait

2. can the `class TablePropertiesTransactionBuilderV2Suite extends TablePropertiesSuiteBase with UseTransactionBuilderV2` be put here too? then all of the children of this abstract test suite are here -- that is simple and clear

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/TransactionBuildersV2Suite.scala
```diff
@@ -0,0 +1,238 @@
...
+import io.delta.kernel.expressions.Column
+import io.delta.kernel.internal.table.SnapshotBuilderImpl
+import io.delta.kernel.internal.tablefeatures.TableFeatures
+import io.delta.kernel.internal.util.Clock
+import io.delta.kernel.transaction.DataLayoutSpec
+import io.delta.kernel.types.StructType
+import io.delta.kernel.utils.CloseableIterable.emptyIterable
+import io.delta.kernel.utils.CloseableIterator
+
+trait UseTransactionBuilderV2 extends DeltaTableWriteSuiteBase {
```
scottsand-db: I love this abstraction direction, thanks!

Can we do something a bit more like

https://github.com/delta-io/delta/blob/8c6fa6dd451625f5c8b66d1d83414a9e1198fda1/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableReadsSuite.scala#L69 and

https://github.com/delta-io/delta/blob/8c6fa6dd451625f5c8b66d1d83414a9e1198fda1/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/utils/TestUtils.scala#L72

That is:
- DeltaTableWriteSuiteBase -- KEEP this name to avoid breaking any existing codes! (but, tbh, it should probably be `DeltaTableWriteSuiteUtils` :( )
- in that file we create `AbstractDeltaTableWriteSuiteBase` (all of the methods, but createTxn is abstract)
- we have: `LegacyAPIsDeltaTableWriteSuite`
- we have `NewAPIsDeltaTableWriteSuite`
- etc etc

This could be another prefactor PR? (excluding the NewAPIsDeltaTableWriteSuite) -- and in that prefactor PR we could also make all of the `createTxn` changes, too? And maybe even the `DeltaTableWriteSuiteUtils` rename :D

scottsand-db:
```scala
trait DeltaTableWriteSuiteBase extends AbstractDeltaTableWriteSuiteBase {
  override def createTxn(...)
}

/**
 * DO NOT MODIFY this trait -- this is just syntactic sugar to clearly indicate we are extending the
 * "default" TestUtils which happens to use the legacy Kernel APIs
 */
trait DeltaTableWriteSuiteBaseWithLegacyAPIs extends DeltaTableWriteSuiteBase

trait DeltaTableWriteSuiteBaseWithLegacyAPIsWithNewAPIs extends DeltaTableWriteSuiteBase {
  override def createTxn(...)
}

trait AbstractDeltaTableWriteSuiteBase extends .... {

  def createTxn(...): ...

  // other methods
}
```

scottsand-db (approved):

PR #5058, opened 2025-08-13
[KERNEL][REFACTOR] Refactor some test utils + suites in preparation for https://github.com/delta-io/delta/pull/5044

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Refactor our write test utilities. This is in preparation for [https://github.com/delta-io/delta/pull/5044](https://github.com/delta-io/delta/pull/5058#top)
- Refactors DeltaWritesSuiteBase to a WriteUtils class that does not extend AnyFunSuite
- Separates it into a class hierarchy with abstract spec and concrete implementations so we can abstract specific write utilities to use different code paths (e.g. v1 vs v2 transaction builders, in the future commit methods, ccv2 code path, etc).

## How was this patch tested?

Test refactor.

No.

### DIFF
--- a/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/WriteUtils.scala
+++ b/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/WriteUtils.scala
@@ -56,7 +56,9 @@ import org.scalatest.funsuite.AnyFunSuite
 
 trait TransactionBuilderSupport {
 
-  def createWriteTxnBuilder(table: Table): TransactionBuilder
+  def createWriteTxnBuilder(
+      table: Table,
+      operation: Operation = Operation.WRITE): TransactionBuilder
 
   // scalastyle:off argcount
   def createTxn(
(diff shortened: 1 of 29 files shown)


kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/WriteUtils.scala
```diff
@@ -54,10 +54,101 @@ import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.Path
 import org.scalatest.funsuite.AnyFunSuite
 
+trait TransactionBuilderSupport {
```
scottsand-db: Love this!

Only note is: (a) `WriteUtils` should be moved into the `utils` package/folder, and (b) the `TransactionBuilderSupport` should be moved to a new file in that same package/folder, too.

scottsand-db (changes requested):
Looks great! Asked for 1 file re-org, and then I'll stamp. Thanks!

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/WriteUtils.scala
allisonport-db: Done

PR #5052, opened 2025-08-11
Finish first pass/sketch; sending for early review

scottsand-db:
## 🥞 Stacked PR
Use this [link](#) to review incremental changes.

---------

#### Which Delta project/connector is this regarding?

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

PR #5060, opened 2025-08-13
[Kernel] [CatalogManaged] Add helpful APIs to CommitMetadata, useful for Committers

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/5060/files) to review incremental changes.
- [**stack/kernel_commit_metadata_types**](https://github.com/delta-io/delta/pull/5060) [[Files changed](https://github.com/delta-io/delta/pull/5060/files)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This PR adds some helpful APIs to CommitMetadata, which will be used by https://github.com/delta-io/delta/pull/5052 (see their usage there; I will rebase that 5052 PR after this PR is merged).

## How was this patch tested?

New UTs.

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
@@ -87,7 +87,7 @@ public class CommitMetadata {
         readMetadataOpt.isPresent() || newMetadataOpt.isPresent(),
         "At least one of readMetadataOpt or newMetadataOpt must be present");
 
-    checkICTPresentIfCatalogManaged();
+    checkInCommitTimestampPresentIfCatalogManaged();
   }
 
   /** The version of the Delta table this commit is targeting. */
@@ -143,7 +143,7 @@ public class CommitMetadata {
    * the protocol that was read at the beginning of the commit.
    */
   public Protocol getEffectiveProtocol() {
-    return newProtocolOpt.orElseGet(() -> readProtocolOpt.get());
+    return newProtocolOpt.orElseGet(readProtocolOpt::get);
   }
 
   /**
@@ -152,7 +152,7 @@ public class CommitMetadata {
    * metadata that was read at the beginning of the commit.
    */
   public Metadata getEffectiveMetadata() {
-    return newMetadataOpt.orElseGet(() -> readMetadataOpt.get());
+    return newMetadataOpt.orElseGet(readMetadataOpt::get);
   }
 
   /**
@@ -181,7 +181,7 @@ public class CommitMetadata {
     }
   }
 
-  private void checkICTPresentIfCatalogManaged() {
+  private void checkInCommitTimestampPresentIfCatalogManaged() {
     if (TableFeatures.isCatalogManagedSupported(getEffectiveProtocol())) {
       checkArgument(
           commitInfo.getInCommitTimestamp().isPresent(),
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
@@ -23,6 +23,7 @@ import io.delta.kernel.annotation.Experimental;
 import io.delta.kernel.internal.actions.CommitInfo;
 import io.delta.kernel.internal.actions.Metadata;
 import io.delta.kernel.internal.actions.Protocol;
+import io.delta.kernel.internal.tablefeatures.TableFeatures;
 import java.util.Optional;
 
 /**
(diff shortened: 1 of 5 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
```diff
@@ -104,4 +136,56 @@ public Optional<Protocol> getNewProtocolOpt() {
...
+  public Metadata getEffectiveMetadata() {
+    return newMetadataOpt.orElseGet(() -> readMetadataOpt.get());
+  }
+
+  /**
+   * Determines the type of commit based on whether this is a table creation and the catalog-managed
+   * status of the table before and after the commit.
+   */
+  public CommitType getCommitType() {
+    final boolean isCreate = !readProtocolOpt.isPresent() && !readMetadataOpt.isPresent();
```
huan233usc, line 163: curious about the replace, are we going to treat it as update from the angle of commiter?

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
```diff
@@ -56,6 +76,18 @@ public CommitMetadata(
     this.readMetadataOpt = requireNonNull(readMetadataOpt, "readMetadataOpt is null");
     this.newProtocolOpt = requireNonNull(newProtocolOpt, "newProtocolOpt is null");
     this.newMetadataOpt = requireNonNull(newMetadataOpt, "newMetadataOpt is null");
+
+    checkArgument(
```
emkornfield, line 80: nit: I wonder if these checks mean we mean we should really have a concept that is the tuple of protocol and metadata.  Then have Option<NewType>.  Might be overly complex modelling if this is the only place this pattern occurs.

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
```diff
@@ -104,4 +136,56 @@ public Optional<Protocol> getNewProtocolOpt() {
...
     return newMetadataOpt;
   }
+
+  /**
+   * Returns the effective {@link Protocol} that will be in place after this commit. If a new
+   * protocol is being written as part of this commit, returns the new protocol. Otherwise, returns
+   * the protocol that was read at the beginning of the commit.
+   */
+  public Protocol getEffectiveProtocol() {
+    return newProtocolOpt.orElseGet(() -> readProtocolOpt.get());
```
emkornfield: nit: can the orElseGet argument be shortened to readProtocolOpt::get?

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
```diff
@@ -104,4 +136,56 @@ public Optional<Protocol> getNewProtocolOpt() {
...
+  }
+
+  /**
+   * Returns the effective {@link Metadata} that will be in place after this commit. If new metadata
+   * is being written as part of this commit, returns the new metadata. Otherwise, returns the
+   * metadata that was read at the beginning of the commit.
+   */
+  public Metadata getEffectiveMetadata() {
+    return newMetadataOpt.orElseGet(() -> readMetadataOpt.get());
```
emkornfield: same nit on shortening

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
```diff
@@ -104,4 +136,56 @@ public Optional<Protocol> getNewProtocolOpt() {
...
+    } else if (readVersionCatalogManaged && !writeVersionCatalogManaged) {
+      return CommitType.CATALOG_DOWNGRADE_TO_FILESYSTEM;
+    } else if (!readVersionCatalogManaged && writeVersionCatalogManaged) {
+      return CommitType.FILESYSTEM_UPGRADE_TO_CATALOG;
+    } else {
+      return CommitType.FILESYSTEM_WRITE;
+    }
+  }
+
+  private void checkICTPresentIfCatalogManaged() {
```
emkornfield: nit: maybe spell out ICT I'm not sure how well this is know.

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
scottsand-db: Great question! From the perspective of this method, for now, we should just treat REPLACE as a WRITE IMO.

e.g. replacing a filesystem or catalog table is just a new delta commit, and then the engine needs to handle what to do upon that successful commit

happy to change this in the future

scottsand-db: happy to also treat this as a separate case (how would we detect that?)

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
scottsand-db: i think so 👍

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
scottsand-db: This is a good idea. We could create a `ResolvedMetadata` internal class that contains: read version, table path, log path, protocol, metadata.

Are you okay if I treat this as a P1 and try and pick it up later?

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
huan233usc: I guess we could add new case in the future when needed(and when we have more signal and justification), especially iiuc this enum are mainly used internally by kernel.

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
emkornfield: Yeah, don't block on this.

PR #5062, opened 2025-08-13
[Kernel] UCCatalogManagedCommitter skeleton -- basic input validation only

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/5062/files) to review incremental changes.
- [**stack/kernel_uc_catalog_managed_committer_1b**](https://github.com/delta-io/delta/pull/5062) [[Files changed](https://github.com/delta-io/delta/pull/5062/files)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

This PR adds the new `UCCatalogManagedCommitter`. Lightweight implementaiton only, only input validation. Will add the rest of the implementation in future PRs.

## How was this patch tested?

New UTs.

No.

### DIFF
--- a/unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
+++ b/unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
@@ -31,6 +31,10 @@ import org.apache.hadoop.fs.Path;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * An implementation of {@link Committer} that handles commits to Delta tables managed by Unity
+ * Catalog. That is, these Delta tables must have the catalogManaged table feature supported.
+ */
 public class UCCatalogManagedCommitter implements Committer {
   private static final Logger logger = LoggerFactory.getLogger(UCCatalogManagedCommitter.class);
 
@@ -38,6 +42,13 @@ public class UCCatalogManagedCommitter implements Committer {
   private final String ucTableId;
   private final Path tablePath;
 
+  /**
+   * Creates a new UCCatalogManagedCommitter for the specified Unity Catalog-managed Delta table.
+   *
+   * @param ucClient the Unity Catalog client to use for commit operations
+   * @param ucTableId the unique Unity Catalog table identifier
+   * @param tablePath the path to the Delta table in the underlying storage system
+   */
   public UCCatalogManagedCommitter(UCClient ucClient, String ucTableId, String tablePath) {
     this.ucClient = requireNonNull(ucClient, "ucClient is null");
     this.ucTableId = requireNonNull(ucTableId, "ucTableId is null");
@@ -51,7 +62,7 @@ public class UCCatalogManagedCommitter implements Committer {
     requireNonNull(engine, "engine is null");
     requireNonNull(finalizedActions, "finalizedActions is null");
     requireNonNull(commitMetadata, "commitMetadata is null");
-    validateCommitMetadataLogPath(commitMetadata);
+    validateLogPathBelongsToThisUcTable(commitMetadata);
 
     final CommitMetadata.CommitType commitType = commitMetadata.getCommitType();
 
@@ -83,7 +94,7 @@ public class UCCatalogManagedCommitter implements Committer {
     return path.toUri().normalize().toString();
   }
 
-  private void validateCommitMetadataLogPath(CommitMetadata cm) {
+  private void validateLogPathBelongsToThisUcTable(CommitMetadata cm) {
     final String expectedDeltaLogPathNormalized = normalize(new Path(tablePath, "_delta_log"));
     final String providedDeltaLogPathNormalized = normalize(new Path(cm.getDeltaLogDirPath()));
     checkArgument(
--- a/unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
+++ b/unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
@@ -21,6 +21,8 @@ import java.util.Optional
 import scala.collection.JavaConverters._
 
 import io.delta.kernel.commit.CommitMetadata
+import io.delta.kernel.commit.CommitMetadata.CommitType
+import io.delta.kernel.internal.actions.{Metadata, Protocol}
 import io.delta.kernel.internal.tablefeatures.TableFeatures
 import io.delta.kernel.test.VectorTestUtils
 
@@ -30,6 +32,29 @@ class UCCatalogManagedCommitterSuite
     extends AnyFunSuite
     with UCCatalogManagedTestUtils
     with VectorTestUtils {
+
+  private def createCommitMetadata(
+      version: Long,
+      logPath: String = baseTestLogPath,
+      readProtocolOpt: Optional[Protocol] = Optional.empty(),
+      readMetadataOpt: Optional[Metadata] = Optional.empty(),
+      newProtocolOpt: Optional[Protocol] = Optional.empty(),
+      newMetadataOpt: Optional[Metadata] = Optional.empty()): CommitMetadata = new CommitMetadata(
+    version,
+    logPath,
+    testCommitInfo(),
+    readProtocolOpt,
+    readMetadataOpt,
+    newProtocolOpt,
+    newMetadataOpt)
+
+  private def catalogManagedWriteCommitMetadata(
+      version: Long,
+      logPath: String = baseTestLogPath): CommitMetadata = createCommitMetadata(
+    version = version,
+    logPath = logPath,
+    readProtocolOpt = Optional.of(protocolWithCatalogManagedSupport),
+    readMetadataOpt = Optional.of(basicPartitionedMetadata))
 
   test("constructor throws on null inputs") {
     val ucClient = new InMemoryUCClient("ucMetastoreId")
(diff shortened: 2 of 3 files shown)


unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -0,0 +1,95 @@
...
+import io.delta.kernel.commit.Committer;
+import io.delta.kernel.data.Row;
+import io.delta.kernel.engine.Engine;
+import io.delta.kernel.utils.CloseableIterator;
+import io.delta.storage.commit.uccommitcoordinator.UCClient;
+import org.apache.hadoop.fs.Path;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class UCCatalogManagedCommitter implements Committer {
```
emkornfield, line 38: nit: docs?

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -0,0 +1,95 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(UCCatalogManagedCommitter.class);
+
+  private final UCClient ucClient;
+  private final String ucTableId;
+  private final Path tablePath;
+
+  public UCCatalogManagedCommitter(UCClient ucClient, String ucTableId, String tablePath) {
```
emkornfield, line 52: nit: docs

unity/src/test/scala/io/delta/unity/UCCatalogManagedTestUtils.scala
```diff
@@ -19,17 +19,43 @@ package io.delta.unity
...
 
 import org.apache.hadoop.fs.{FileStatus => HadoopFileStatus, Path}
 
-trait UCCatalogManagedTestUtils extends TestUtils {
+trait UCCatalogManagedTestUtils extends TestUtils with ActionUtils {
   val fakeURI = new URI("s3://bucket/table")
+  val baseTestTablePath = "/path/to/table"
+  val baseTestLogPath = "/path/to/table/_delta_log"
+
+  def createCommitMetadata(
```
emkornfield: Is this going to be used in multiple tests?

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -0,0 +1,95 @@
...
+      // TODO: support this
+      throw new UnsupportedOperationException("Protocol change is not yet implemented");
+    }
+    if (commitMetadata.getNewMetadataOpt().isPresent()) {
+      // TODO: support this
+      throw new UnsupportedOperationException("Metadata change is not yet implemented");
+    }
+
+    // TODO: support this
+    throw new UnsupportedOperationException("Commit logic is not yet implemented");
```
allisonport-db, line 86: Are these each tied to a specific piece of implementation logic? (as the reason for this structure)

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -0,0 +1,95 @@
...
+
+  ////////////////////
+  // Helper methods //
+  ////////////////////
+
+  private String normalize(Path path) {
+    return path.toUri().normalize().toString();
+  }
+
+  private void validateCommitMetadataLogPath(CommitMetadata cm) {
```
allisonport-db: super simple docs? not obvious what this does without reading the code for me (or maybe just slightly more descriptive method name)

unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
```diff
@@ -0,0 +1,165 @@
...
+      committer.commit(defaultEngine, null, catalogManagedWriteCommitMetadata())
+    }
+
+    // Null commitMetadata
+    assertThrows[NullPointerException] {
+      committer.commit(defaultEngine, emptyActionsIterator, null)
+    }
+  }
+
+  test("commit throws if CommitMetadata is for a different table") {
```
allisonport-db, line 93: check that this does normalize the path?

unity/src/test/scala/io/delta/unity/UCCatalogManagedTestUtils.scala
```diff
@@ -19,17 +19,43 @@ package io.delta.unity
...
+      version: Long = 1L,
```
allisonport-db: why is the default version 1?

allisonport-db: Does it make sense to have 2 helpers, 1 new table (version = 0, readProtocolOpt/readMetadataOpt are empty, newProtcol/newMetadata are required) and 1 for existing table (version = 1, readProtocolOpt/readMetadataOpt are required, new are optional)?

Not sure if that makes sense with how it's going to be used in tests.

allisonport-db (commented):
Few minor qs/comments

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
scottsand-db: I'd say: yes, each of these `Unsupported...` are coupled to some distinct functionality.

I'm just trying to roll out and implement this class iteratively, and we can support one piece of logic without the others. e.g. we can implement the normal commit logic, without supporting the protocol/metadata changes (which requires mapping kernel protocol to the delta storage abstract protocol, for example)

unity/src/test/scala/io/delta/unity/UCCatalogManagedTestUtils.scala
scottsand-db: Great question/observation. 

I'm trying to have minimal test overhead / util bloat. I think for the case of commit, callers should just be sure to pass in a version. Alternatively, we could just make the version required and have everyone pass it in. I'd prefer the former option for now? i.e. what is currently present? wdyt?

unity/src/test/scala/io/delta/unity/UCCatalogManagedTestUtils.scala
scottsand-db: I'd think so? It's only used by one right now -- I'd be happy to move it out of here; and to move it to a common util when applicable 👍

unity/src/test/scala/io/delta/unity/UCCatalogManagedTestUtils.scala
allisonport-db: I think version should just be required then. I feel like defaults make sense when there's a reason it's the default, version 1 seems like it's a random version to me (i.e. why not 0 or 2).

Or maybe committing version 1 is the common-case and I'm just unaware?

scottsand-db: I'll make it required 👍

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
scottsand-db: i renamed the method to make it self-describing 👍

unity/src/test/scala/io/delta/unity/UCCatalogManagedTestUtils.scala
emkornfield: If you are pretty sure, fine to leave as is.

unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
scottsand-db: yeah this is interesting.

(a) I think our UCCatalogManagedClient should be resolving the `tablePath` input using the engine
(b) all I want to do in the committer is make sure the input delta log path belongs to that tablePath. think I can just compare the paths directly? normalizing to URI seems also fine to me ... but I'm not expecting any of the input to be weird/bad...

allisonport-db: then I think it's fine to compare directly (forces the client to resolve using the engine in theory)

scottsand-db: But part of me wonders: if the catalog told us that X is the table path, should the engine then override and "resolve" it (and maybe change it slightly?) hmmm

scottsand-db: cc @scovich curious if you have any thoughts?

scottsand-db: My opinion: The UCCatalogManagedCommitter should resolve the input path ... that's how everything inside of Kernel operates.

But normalization is probabler safer? Like i don't think we need to test edge cases for it ... we are not expecting any weird paths at this point in the commit? we just want a safe way to compare two Paths ... and URI.normalize is a good way to do that

scottsand-db: cc @vkorukanti or @huan233usc if you have any thoughts?

scottsand-db: Alas -- my updated opinoin is:

Let's just keep things simple. The engine/catalog should give the UCCatalogManagedClient the right tablePath. We should just use that. We can iterate and improve on this later

allisonport-db (approved):

emkornfield (approved):

PR #5074, opened 2025-08-14
[Kernel] UCCatalogManagedCommitter: finish the lightweight WRITE (not CREATE) implementation

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/5074/files) to review incremental changes.
- [**stack/kernel_uc_catalog_managed_committer_2b**](https://github.com/delta-io/delta/pull/5074) [[Files changed](https://github.com/delta-io/delta/pull/5074/files)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

This PR fleshes out the remaining WRITE (not CREATE) UCCatalogManagedCommitter implementation. e.g. write the staged commit file, and then send that info to UC.

## How was this patch tested?

Simple unit tests. E2E mock write tests will come later.

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
@@ -24,6 +24,7 @@ import io.delta.kernel.internal.actions.CommitInfo;
 import io.delta.kernel.internal.actions.Metadata;
 import io.delta.kernel.internal.actions.Protocol;
 import io.delta.kernel.internal.tablefeatures.TableFeatures;
+import io.delta.kernel.internal.util.FileNames;
 import io.delta.kernel.internal.util.Tuple2;
 import java.util.Optional;
 
@@ -175,6 +176,32 @@ public class CommitMetadata {
     } else {
       return CommitType.FILESYSTEM_WRITE;
     }
+  }
+
+  /**
+   * Returns the corresponding published Delta log file path for this commit, which is in the form
+   * of {@code <table_path>/_delta_log/0000000000000000000<version>.json}.
+   *
+   * <p>Usages:
+   *
+   * <ul>
+   *   <li>Filesystem-managed committers must write to this file path.
+   *   <li>Catalog-managed committers must backfill to this file path, if/when they so choose.
+   * </ul>
+   */
+  public String getPublishedDeltaFilePath() {
+    return FileNames.deltaFile(logPath, version);
+  }
+
+  /**
+   * Returns a new staged commit file path with a unique UUID for this commit. Each invocation
+   * returns a new, unique value, in the form of {@code
+   * <table_path>/_delta_log/_staged_commits/0000000000000000000<version>.<uuid>.json}
+   *
+   * <p>Catalog-managed committers may use this path to write new staged commits.
+   */
+  public String generateNewStagedCommitFilePath() {
+    return FileNames.stagedCommitFile(logPath, version);
   }
 
   private void checkReadStateAbsentIfAndOnlyIfVersion0() {
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
@@ -186,7 +186,7 @@ public class CommitMetadata {
    *
    * <ul>
    *   <li>Filesystem-managed committers must write to this file path.
-   *   <li>Catalog-managed committers must backfill to this file path, if/when they so choose.
+   *   <li>Catalog-managed committers must publish to this file path, if/when they so choose.
    * </ul>
    */
   public String getPublishedDeltaFilePath() {
--- a/unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
+++ b/unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
@@ -17,6 +17,8 @@
 package io.delta.unity;
 
 import static io.delta.kernel.internal.util.Preconditions.checkArgument;
+import static io.delta.kernel.internal.util.Preconditions.checkState;
+import static io.delta.unity.utils.OperationTimer.timeCheckedOperation;
 import static java.util.Objects.requireNonNull;
 
 import io.delta.kernel.commit.CommitFailedException;
@@ -25,8 +27,15 @@ import io.delta.kernel.commit.CommitResponse;
 import io.delta.kernel.commit.Committer;
 import io.delta.kernel.data.Row;
 import io.delta.kernel.engine.Engine;
+import io.delta.kernel.internal.annotation.VisibleForTesting;
+import io.delta.kernel.internal.files.ParsedLogData;
 import io.delta.kernel.utils.CloseableIterator;
+import io.delta.kernel.utils.FileStatus;
+import io.delta.storage.commit.Commit;
 import io.delta.storage.commit.uccommitcoordinator.UCClient;
+import io.delta.storage.commit.uccommitcoordinator.UCCommitCoordinatorException;
+import java.io.IOException;
+import java.util.Optional;
 import org.apache.hadoop.fs.Path;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -71,8 +80,6 @@ public class UCCatalogManagedCommitter implements Committer {
       throw new UnsupportedOperationException("Unsupported commit type: " + commitType);
     }
 
-    // TODO: lastKnownBackfilledVersion? Take that in as a hint.
-
     if (commitMetadata.getNewProtocolOpt().isPresent()) {
       // TODO: support this
       throw new UnsupportedOperationException("Protocol change is not yet implemented");
@@ -82,8 +89,12 @@ public class UCCatalogManagedCommitter implements Committer {
       throw new UnsupportedOperationException("Metadata change is not yet implemented");
     }
 
-    // TODO: support this
-    throw new UnsupportedOperationException("Commit logic is not yet implemented");
+    final FileStatus kernelStagedCommitFileStatus =
+        writeStagedCommitFile(engine, finalizedActions, commitMetadata);
+
+    commitToUC(commitMetadata, kernelStagedCommitFileStatus);
+
+    return new CommitResponse(ParsedLogData.forFileStatus(kernelStagedCommitFileStatus));
   }
 
   ////////////////////
@@ -102,5 +113,113 @@ public class UCCatalogManagedCommitter implements Committer {
         "Delta log path '%s' does not match expected '%s'",
         expectedDeltaLogPathNormalized,
         providedDeltaLogPathNormalized);
+  }
+
+  private FileStatus writeStagedCommitFile(
+      Engine engine, CloseableIterator<Row> finalizedActions, CommitMetadata commitMetadata)
+      throws CommitFailedException {
+    checkArgument(
+        commitMetadata.getVersion() > 0, "Can only write staged commit files for versions > 0");
+    final String stagedCommitFilePath = commitMetadata.generateNewStagedCommitFilePath();
+
+    return timeCheckedOperation(
+        logger,
+        "Write staged commit file: " + stagedCommitFilePath,
+        ucTableId,
+        () -> {
+          try {
+            // Do not use Put-If-Absent for staged commit files since we assume that UUID-based
+            // commit files are globally unique, and so we will never have concurrent writers
+            // attempting to write the same commit file.
+
+            engine
+                .getJsonHandler()
+                .writeJsonFileAtomically(
+                    stagedCommitFilePath, finalizedActions, true /* overwrite */);
+
+            // TODO: [delta-io/delta#5021] Use FileSystemClient::getFileStatus API instead
+            return FileStatus.of(stagedCommitFilePath);
+          } catch (IOException ex) {
+            throw new CommitFailedException(
+                true /* retryable */,
+                false /* conflict */,
+                "Failed to write staged commit file due to: " + ex.getMessage(),
+                ex);
+          }
+        });
+  }
+
+  private void commitToUC(CommitMetadata commitMetadata, FileStatus kernelStagedCommitFileStatus)
+      throws CommitFailedException {
+    timeCheckedOperation(
+        logger,
+        "Commit staged commit file to UC: " + kernelStagedCommitFileStatus.getPath(),
+        ucTableId,
+        () -> {
+          // commitToUc is only for normal catalog WRITES, not for CREATE, or UPGRADE, or DOWNGRADE,
+          // or anything filesystem related.
+          checkState(
+              commitMetadata.getCommitType() == CommitMetadata.CommitType.CATALOG_WRITE,
+              "Only supported commit type is CATALOG_WRITE, but got: %s");
+
+          try {
+            ucClient.commit(
+                ucTableId,
+                tablePath.toUri(),
+                Optional.of(getUcCommitPayload(commitMetadata, kernelStagedCommitFileStatus)),
+                Optional.empty() /* lastKnownBackfilledVersion */, // TODO: take this in as a hint
+                false /* isDisown */,
+                Optional.empty() /* newMetadata */, // TODO: support sending newProtocol
+                Optional.empty()); // TODO: support sending newProtocol
+            return null;
+          } catch (io.delta.storage.commit.CommitFailedException cfe) {
+            throw storageCFEtoKernelCFE(cfe);
+          } catch (IOException ex) {
+            throw new CommitFailedException(
+                true /* retryable */, false /* conflict */, ex.getMessage(), ex);
+          } catch (UCCommitCoordinatorException ucce) {
+            // For now, this catches all UC exceptions such as:
+            // - CommitLimitReachedException -> TODO: backfill in this case
+            // - InvalidTargetTableException
+            // - UpgradeNotAllowedException
+            // We can add specific catch statements for these exceptions if needed in the future.
+            throw new CommitFailedException(
+                false /* retryable */, false /* conflict */, ucce.getMessage(), ucce);
+          }
+        });
+  }
+
+  private Commit getUcCommitPayload(
+      CommitMetadata commitMetadata, FileStatus kernelStagedCommitFileStatus) {
+    return new Commit(
+        commitMetadata.getVersion(),
+        kernelFileStatusToHadoopFileStatus(kernelStagedCommitFileStatus),
+        // commitMetadata validates that the ICT is present if writing to a catalogManaged table
+        commitMetadata.getCommitInfo().getInCommitTimestamp().get());
+  }
+
+  @VisibleForTesting
+  public static org.apache.hadoop.fs.FileStatus kernelFileStatusToHadoopFileStatus(
+      io.delta.kernel.utils.FileStatus kernelFileStatus) {
+    return new org.apache.hadoop.fs.FileStatus(
+        kernelFileStatus.getSize() /* length */,
+        false /* isDirectory */,
+        1 /* blockReplication */,
+        128 * 1024 * 1024 /* blockSize (128MB) */,
+        kernelFileStatus.getModificationTime() /* modificationTime */,
+        kernelFileStatus.getModificationTime() /* accessTime */,
+        org.apache.hadoop.fs.permission.FsPermission.getFileDefault() /* permission */,
+        "unknown" /* owner */,
+        "unknown" /* group */,
+        new org.apache.hadoop.fs.Path(kernelFileStatus.getPath()) /* path */);
+  }
+
+  private static CommitFailedException storageCFEtoKernelCFE(
+      io.delta.storage.commit.CommitFailedException storageCFE) {
+    return new CommitFailedException(
+        storageCFE.getRetryable(),
+        storageCFE.getConflict(),
+        storageCFE.getMessage(),
+        storageCFE.getCause());
   }
 }
--- a/unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
+++ b/unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
@@ -132,6 +132,8 @@ public class UCCatalogManagedCommitter implements Committer {
             // commit files are globally unique, and so we will never have concurrent writers
             // attempting to write the same commit file.
 
+            // Note: the engine is responsible for closing the actions iterator once it has been
+            //       fully consumed.
             engine
                 .getJsonHandler()
                 .writeJsonFileAtomically(
@@ -169,8 +171,8 @@ public class UCCatalogManagedCommitter implements Committer {
                 Optional.of(getUcCommitPayload(commitMetadata, kernelStagedCommitFileStatus)),
                 Optional.empty() /* lastKnownBackfilledVersion */, // TODO: take this in as a hint
                 false /* isDisown */,
-                Optional.empty() /* newMetadata */, // TODO: support sending newProtocol
-                Optional.empty()); // TODO: support sending newProtocol
+                Optional.empty() /* newMetadata */, // TODO: support sending newMetadata
+                Optional.empty()); /* newProtocol */ // TODO: support sending newProtocol
             return null;
           } catch (io.delta.storage.commit.CommitFailedException cfe) {
             throw storageCFEtoKernelCFE(cfe);
@@ -179,7 +181,7 @@ public class UCCatalogManagedCommitter implements Committer {
                 true /* retryable */, false /* conflict */, ex.getMessage(), ex);
           } catch (UCCommitCoordinatorException ucce) {
             // For now, this catches all UC exceptions such as:
-            // - CommitLimitReachedException -> TODO: backfill in this case
+            // - CommitLimitReachedException -> TODO: publish in this case
             // - InvalidTargetTableException
             // - UpgradeNotAllowedException
             // We can add specific catch statements for these exceptions if needed in the future.
--- a/unity/src/main/java/io/delta/unity/utils/OperationTimer.java
+++ b/unity/src/main/java/io/delta/unity/utils/OperationTimer.java
@@ -23,9 +23,16 @@ import org.slf4j.Logger;
 public class OperationTimer {
   private OperationTimer() {}
 
-  /** Times an operation and logs the duration. */
-  public static <T> T timeOperation(
-      Logger logger, String operationName, String ucTableId, Supplier<T> operation) {
+  @FunctionalInterface
+  public interface ThrowingSupplier<T, E extends Exception> {
+    T get() throws E;
+  }
+
+  /** Times an operation that throws a checked exception of type E and logs the duration. */
+  @SuppressWarnings("unchecked")
+  public static <T, E extends Exception> T timeCheckedOperation(
+      Logger logger, String operationName, String ucTableId, ThrowingSupplier<T, E> operation)
+      throws E {
     final long startTime = System.nanoTime();
     try {
       final T result = operation.get();
@@ -36,8 +43,14 @@ public class OperationTimer {
       final long durationMs = nanoToMs(System.nanoTime() - startTime);
       logger.warn(
           "[{}] {} failed after {} ms: {}", ucTableId, operationName, durationMs, e.getMessage());
-      throw e;
+      throw (E) e; // Safe cast since operation can only throw E
     }
+  }
+
+  /** Times an operation and logs the duration. */
+  public static <T> T timeUncheckedOperation(
+      Logger logger, String operationName, String ucTableId, Supplier<T> operation) {
+    return timeCheckedOperation(logger, operationName, ucTableId, operation::get);
   }
 
   private static long nanoToMs(long nanoTime) {
--- a/unity/src/test/scala/io/delta/unity/UCCatalogManagedClientSuite.scala
+++ b/unity/src/test/scala/io/delta/unity/UCCatalogManagedClientSuite.scala
@@ -37,7 +37,11 @@ import org.scalatest.funsuite.AnyFunSuite
 /** Unit tests for [[UCCatalogManagedClient]]. */
 class UCCatalogManagedClientSuite extends AnyFunSuite with UCCatalogManagedTestUtils {
 
-  private def createUCCatalogManagedClientForTableWithNoRatifiedCommits(
+  /**
+   * When a new UC table is created, it will have Delta version 0 but the max ratified verison in
+   * UC is -1. This is a special edge case.
+   */
+  private def createUCCatalogManagedClientForTableWithMaxRatifiedVersionNegativeOne(
       ucTableId: String = "ucTableId"): UCCatalogManagedClient = {
     val ucClient = new InMemoryUCClient("ucMetastoreId")
     val tableData = new TableData(-1, ArrayBuffer[Commit]())
@@ -136,7 +140,8 @@ class UCCatalogManagedClientSuite extends AnyFunSuite with UCCatalogManagedTestU
     case (versionToLoad, description) =>
       test(s"table version 0 is loaded when UC maxRatifiedVersion is -1 -- $description") {
         val tablePath = getTestResourceFilePath("catalog-owned-preview")
-        val ucCatalogManagedClient = createUCCatalogManagedClientForTableWithNoRatifiedCommits()
+        val ucCatalogManagedClient =
+          createUCCatalogManagedClientForTableWithMaxRatifiedVersionNegativeOne()
         val snapshot = ucCatalogManagedClient
           .loadSnapshot(defaultEngine, "ucTableId", tablePath, versionToLoad)
 
@@ -198,7 +203,8 @@ class UCCatalogManagedClientSuite extends AnyFunSuite with UCCatalogManagedTestU
 
   test("creates snapshot with UCCatalogManagedCommitter") {
     val tablePath = getTestResourceFilePath("catalog-owned-preview")
-    val ucCatalogManagedClient = createUCCatalogManagedClientForTableWithNoRatifiedCommits()
+    val ucCatalogManagedClient =
+      createUCCatalogManagedClientForTableWithMaxRatifiedVersionNegativeOne()
     val snapshot = ucCatalogManagedClient
       .loadSnapshot(defaultEngine, "ucTableId", tablePath, Optional.of(0L))
       .asInstanceOf[SnapshotImpl]
--- a/unity/src/test/scala/io/delta/unity/UCCatalogManagedClientSuite.scala
+++ b/unity/src/test/scala/io/delta/unity/UCCatalogManagedClientSuite.scala
@@ -37,6 +37,7 @@ import org.scalatest.funsuite.AnyFunSuite
 /** Unit tests for [[UCCatalogManagedClient]]. */
 class UCCatalogManagedClientSuite extends AnyFunSuite with UCCatalogManagedTestUtils {
 
+  // TODO: [delta-io/delta#5118] If UC changes CREATE semantics, update logic here.
   /**
    * When a new UC table is created, it will have Delta version 0 but the max ratified verison in
    * UC is -1. This is a special edge case.
--- a/unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
+++ b/unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
@@ -16,7 +16,7 @@
 
 package io.delta.unity
 
-import java.io.{File, IOException}
+import java.io.IOException
 import java.nio.file.Files
 import java.util.{Optional, UUID}
 
@@ -90,6 +90,19 @@ class UCCatalogManagedCommitterSuite
       new KernelTuple2[Protocol, Metadata](
         protocolWithCatalogManagedSupport,
         basicPartitionedMetadata)))
+
+  private def getSingleElementRowIter(elem: String): CloseableIterator[Row] = {
+    import io.delta.kernel.defaults.integration.DataBuilderUtils
+    import io.delta.kernel.types.{StringType, StructField, StructType}
+
+    val schema = new StructType().add(new StructField("testColumn", StringType.STRING, true))
+    val simpleRow = DataBuilderUtils.row(schema, elem)
+    singletonCloseableIterator(simpleRow)
+  }
+
+  // ============================================================
+  // ===================== Misc. Unit Tests =====================
+  // ============================================================
 
   test("constructor throws on null inputs") {
     val ucClient = new InMemoryUCClient("ucMetastoreId")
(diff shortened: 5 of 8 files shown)


unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
```diff
@@ -185,15 +190,66 @@ class UCCatalogManagedCommitterSuite
...
-    val committer = new UCCatalogManagedCommitter(ucClient, "ucTableId", baseTestTablePath)
-    val commitMetadata = catalogManagedWriteCommitMetadata(version = 1)
+  test("commit writes staged commit file and invokes UC client commit API") {
+    withTempDir { tempDir =>
+      // ===== GIVEN =====
+      val tablePath = tempDir.getAbsolutePath
+      val logPath = s"$tablePath/_delta_log"
+
+      // TODO do this in kernel
+      defaultEngine.getFileSystemClient.mkdirs(s"$logPath/_staged_commits")
```
scottsand-db: Will make a prefactor PR for this

unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
```diff
@@ -185,15 +190,66 @@ class UCCatalogManagedCommitterSuite
...
   }
 
+  private def getSingleElementRowIter(elem: String): CloseableIterator[Row] = {
+    import io.delta.kernel.defaults.integration.DataBuilderUtils
+    import io.delta.kernel.types.{StringType, StructField, StructType}
+
+    val schema = new StructType().add(new StructField("testColumn", StringType.STRING, true))
+    val simpleRow = DataBuilderUtils.row(schema, elem)
+    singletonCloseableIterator(simpleRow)
+  }
```
scottsand-db, line 254: will do error handling tests in a subsequent PR (fast follow)

unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
```diff
@@ -185,15 +195,217 @@ class UCCatalogManagedCommitterSuite
...
+      // ===== GIVEN =====
+      val tablePath = tempDir.getAbsolutePath
+      val logPath = s"$tablePath/_delta_log"
 
-    val exMsg = intercept[UnsupportedOperationException] {
-      committer.commit(defaultEngine, emptyActionsIterator, commitMetadata)
-    }.getMessage
-    assert(exMsg.contains("Commit logic is not yet implemented"))
+      // TODO do this in kernel
+      defaultEngine.getFileSystemClient.mkdirs(s"$logPath/_staged_commits")
```
scottsand-db: TODO merge https://github.com/delta-io/delta/pull/5080 first

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitFailedException.java
```diff
@@ -32,6 +32,11 @@
 @Experimental
 public class CommitFailedException extends Exception {
 
+  public static final boolean RETRYABLE = true;
```
allisonport-db: Is this an attempt at named parameters in java? what about static factory methods instead?

allisonport-db: Or honestly I don't think this is much better than just booleans since the order is still the important part? you could easily do `new CommitFailedException(CONFLICT, NOT_RETRYABLE)`

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
```diff
@@ -181,6 +182,14 @@ public CommitType getCommitType() {
     }
   }
 
+  public String getPublishedDeltaFilePath() {
+    return FileNames.deltaFile(logPath, version);
+  }
+
+  public String getNewStagedCommitFilePath() {
```
allisonport-db: Is it possible that this will be called more than once? Do we want it to generate a new path each time? wondering if this should instead be stored -per commit metadata instance

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -103,4 +117,103 @@ private void validateLogPathBelongsToThisUcTable(CommitMetadata cm) {
...
+                ucTableId,
+                tablePath.toUri(),
+                Optional.of(getUCCommitPayload(commitMetadata, kernelStagedCommitFileStatus)),
+                Optional.empty() /* lastKnownBackfilledVersion */,
+                false /* isDisown */,
+                Optional.empty() /* newMetadata */, // TODO: support sending newProtocol
+                Optional.empty()); // TODO: support sending newProtocol
+            return null;
+          } catch (io.delta.storage.commit.CommitFailedException cfe) {
+            throw storageCFEtoKernelCFE(cfe);
```
allisonport-db, line 178: is this the error we'd hit if it is a conflict?

allisonport-db (commented):
Few comments, will take a look at the test suite tomorrow

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitFailedException.java
emkornfield: I think either static-factory or making them distinct enums would probably be a good idea.

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -103,4 +117,103 @@ private void validateLogPathBelongsToThisUcTable(CommitMetadata cm) {
...
+
+    return timeCheckedOperation(
+        logger,
+        "Write staged commit file: " + stagedCommitFilePath,
+        ucTableId,
+        () -> {
+          try {
+            // Do not use Put-If-Absent for staged commit files since we assume that UUID-based
+            // commit files are globally unique, and so we will never have concurrent writers
+            // attempting to write the same commit file.
```
emkornfield, line 133: Sorry if this is a duplicate comment, but could you expand on the harm doing overwrite = false does here?  Is it problematic for retries?  More expensive?

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -103,4 +117,103 @@ private void validateLogPathBelongsToThisUcTable(CommitMetadata cm) {
...
+          checkState(
+              commitMetadata.getCommitType() == CommitMetadata.CommitType.CATALOG_WRITE,
+              "Only supported commit type is CATALOG_WRITE, but got: %s");
+
+          try {
+            ucClient.commit(
+                ucTableId,
+                tablePath.toUri(),
+                Optional.of(getUCCommitPayload(commitMetadata, kernelStagedCommitFileStatus)),
+                Optional.empty() /* lastKnownBackfilledVersion */,
```
emkornfield: should there be a TODO here as well?  If not perhaps expand comment to explain why this is missing.

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -103,4 +117,103 @@ private void validateLogPathBelongsToThisUcTable(CommitMetadata cm) {
...
+            throw storageCFEtoKernelCFE(cfe);
+          } catch (IOException ex) {
+            throw new CommitFailedException(RETRYABLE, NOT_CONFLICT, ex.getMessage(), ex);
+          } catch (UCCommitCoordinatorException ucce) {
+            throw new CommitFailedException(NOT_RETRYABLE, NOT_CONFLICT, ucce.getMessage(), ucce);
+          }
+        });
+  }
+
+  private Commit getUCCommitPayload(
```
emkornfield: aside its names like this for why I like strict camel casing (getUcCommitPayload) but I think this aligns with most of the code.

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -103,4 +117,103 @@ private void validateLogPathBelongsToThisUcTable(CommitMetadata cm) {
...
+      io.delta.kernel.utils.FileStatus kernelFileStatus) {
+    return new org.apache.hadoop.fs.FileStatus(
+        kernelFileStatus.getSize() /* length */,
+        false /* isDirectory */,
+        1 /* blockReplication */,
+        128 * 1024 * 1024 /* blockSize (128MB) */,
+        kernelFileStatus.getModificationTime() /* modificationTime */,
+        kernelFileStatus.getModificationTime() /* accessTime */,
+        org.apache.hadoop.fs.permission.FsPermission.getFileDefault() /* permission */,
+        "user" /* owner */,
```
emkornfield: are "user" and "group" actually needed here can they be empty instead or a string that makes it clear they are not used?

unity/src/main/java/io/delta/unity/utils/OperationTimer.java
```diff
@@ -36,10 +43,16 @@ public static <T> T timeOperation(
       final long durationMs = nanoToMs(System.nanoTime() - startTime);
       logger.warn(
           "[{}] {} failed after {} ms: {}", ucTableId, operationName, durationMs, e.getMessage());
-      throw e;
+      throw (E) e; // Safe cast since operation can only throw E
```
emkornfield, line 46: does catching E on line 42 not work?

unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
```diff
@@ -185,15 +195,217 @@ class UCCatalogManagedCommitterSuite
...
+
+      // Verify the staged commit file actually exists on disk
+      val file = new java.io.File(stagedCommitFilePath)
+      assert(file.exists())
+      assert(file.isFile())
+      assert(file.length() > 0)
+
+      // Read the file content and verify our test value was written
+      val fileContent = scala.io.Source.fromFile(file).getLines().mkString("\n")
+      assert(fileContent.contains(testValue))
```
emkornfield, line 310: this makes the length check above redundant?

unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
```diff
@@ -185,15 +195,217 @@ class UCCatalogManagedCommitterSuite
...
+
+      // Verify the file is in the correct location
+      assert(stagedCommitFilePath.contains("_staged_commits"))
```
emkornfield: should this be more specific (is it written under a directory?)

unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
```diff
@@ -185,15 +195,217 @@ class UCCatalogManagedCommitterSuite
...
+      assert(file.getName.contains("00000000000000000001"))
+      assert(stagedCommitFilePath.startsWith(tablePath))
```
emkornfield: it might be clearer to assert this first, then assert what the remaining substring looks like (if the spec requires something specific here).

unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
```diff
@@ -185,15 +195,217 @@ class UCCatalogManagedCommitterSuite
...
+    val hadoopFileStatus =
+      UCCatalogManagedCommitter.kernelFileStatusToHadoopFileStatus(kernelFileStatus)
+
+    // ===== THEN =====
+    assert(hadoopFileStatus.getPath.toString == "/path/to/file.json")
+    assert(hadoopFileStatus.getLen == 1024L)
+    assert(hadoopFileStatus.getModificationTime == 1234567890L)
+    assert(hadoopFileStatus.getAccessTime == 1234567890L) // same as modification time
+    assert(!hadoopFileStatus.isDirectory)
+    assert(hadoopFileStatus.getReplication == 1)
```
emkornfield: from here and below, its not clear to me if these values are required or are just being passed through for construction?  I assume UC doesn't actually manage tables on HDFS?

unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
```diff
@@ -185,15 +195,217 @@ class UCCatalogManagedCommitterSuite
...
+      })
+
+      val ucClient = new InMemoryUCClient("ucMetastoreId")
+      val tableData = new TableData(maxRatifiedVersion = 1, commits = ArrayBuffer.empty[Commit])
+      ucClient.createTableIfNotExistsOrThrow("ucTableId", tableData)
+      val committer = new UCCatalogManagedCommitter(ucClient, "ucTableId", tablePath)
+      val commitMetadata = catalogManagedWriteCommitMetadata(2, logPath = logPath)
+
+      // ===== WHEN =====
+      val ex = intercept[CommitFailedException] {
```
emkornfield, line 347: is there a plan for more e2e tests using InMemoryUCCatalog and Kernel?

emkornfield (commented):
Mostly just questions.  I think testing strategy is likely the most important one.

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitFailedException.java
scottsand-db: Makes sense team. Will consider doing enums in another PR.

Btw, I agree someone could do `new CommitFailedException(CONFLICT, NOT_RETRYABLE)`, but someone could also do `new CommitFailedException(true /*conflict*/, false /*retryable*/)`.

All this adds is the semantic sugar of `RETRYABLE` instead of `true /* retryable */`, which is an improvement IMO.

Will make them enums nonetheless 👍

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
scottsand-db: Great question. I view this as a utility method that committers can use as their discretion. I can add javadocs that makes it clearer (hopefully the `New` in the method name is already clear enough). But IMO just returning a new UUID each time is fine

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
scottsand-db: Yup, it's up above but I should move it here. Will do! Thanks!

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
scottsand-db: Yup!

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
scottsand-db: I can do that, I'm easy!

unity/src/main/java/io/delta/unity/utils/OperationTimer.java
scottsand-db: `Cannot catch type parameters` :/

scottsand-db: That's what the compiler says ^^. Claude elaborates:

> This is because of type erasure in Java generics. At runtime, the generic type parameter E is erased and becomes just Exception (its upper bound). The Java compiler doesn't allow you to catch generic type parameters directly in catch blocks.

unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
scottsand-db: Will make this clearer!

unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
scottsand-db: Yup! Will come in subsequent PRs

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
scottsand-db: cc @sumeet-db, @LukasRupprecht, @dhruvarya-db 

Just copying what the original UC-CCC does: https://github.com/delta-io/delta/blob/3b75a3c3ea7a44d49cc7c3f8c2b89f5af76ce5c6/storage/src/main/java/io/delta/storage/commit/CoordinatedCommitsUtils.java#L130

I'll let the folks above elaborate/respond?

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
scottsand-db: Will do!

unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
scottsand-db: Yup! Used a regex!

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
allisonport-db: yep this is fine, I was more wondering if there was a scenario where a committer would want to call it more than once and have it be the same pinned value (i.e. each commit metadata = 1 staged commit file path). if the answer = no, sgtm!

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -103,4 +114,105 @@ private void validateLogPathBelongsToThisUcTable(CommitMetadata cm) {
...
+                false /* isDisown */,
+                Optional.empty() /* newMetadata */, // TODO: support sending newProtocol
+                Optional.empty()); // TODO: support sending newProtocol
+            return null;
+          } catch (io.delta.storage.commit.CommitFailedException cfe) {
+            throw storageCFEtoKernelCFE(cfe);
+          } catch (IOException ex) {
+            throw new CommitFailedException(
+                true /* retryable */, false /* conflict */, ex.getMessage(), ex);
+          } catch (UCCommitCoordinatorException ucce) {
```
allisonport-db, line 182: Can you add a tiny comment what this might consist of and why it's not retryable? Not clear to me

allisonport-db (approved):
LGTM assuming someone with more UC context looks at the UC-specific code!

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -103,4 +114,105 @@ private void validateLogPathBelongsToThisUcTable(CommitMetadata cm) {
...
+            // attempting to write the same commit file.
+
+            engine
+                .getJsonHandler()
+                .writeJsonFileAtomically(
+                    stagedCommitFilePath, finalizedActions, true /* overwrite */);
+
+            // TODO: [delta-io/delta#5021] Use FileSystemClient::getFileStatus API instead
+            return FileStatus.of(stagedCommitFilePath);
+          } catch (IOException ex) {
```
scottsand-db, line 144: @xzhseh and @sumeet-db -- is it worth catching FileAlreadyExistsException here and throwing a non-retryable error? because we should never see a UUID file that already exists

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -103,4 +114,105 @@ private void validateLogPathBelongsToThisUcTable(CommitMetadata cm) {
         expectedDeltaLogPathNormalized,
         providedDeltaLogPathNormalized);
   }
+
+  private FileStatus writeStagedCommitFile(
+      Engine engine, CloseableIterator<Row> finalizedActions, CommitMetadata commitMetadata)
+      throws CommitFailedException {
+    final String stagedCommitFilePath = commitMetadata.generateNewStagedCommitFilePath();
```
scottsand-db, line 123: checkArgument not version 0

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -103,4 +114,105 @@ private void validateLogPathBelongsToThisUcTable(CommitMetadata cm) {
...
+    return new org.apache.hadoop.fs.FileStatus(
+        kernelFileStatus.getSize() /* length */,
+        false /* isDirectory */,
+        1 /* blockReplication */,
+        128 * 1024 * 1024 /* blockSize (128MB) */,
+        kernelFileStatus.getModificationTime() /* modificationTime */,
+        kernelFileStatus.getModificationTime() /* accessTime */,
+        org.apache.hadoop.fs.permission.FsPermission.getFileDefault() /* permission */,
+        "unknown" /* owner */,
+        "unknown" /* group */,
```
huan233usc, line 215: I wonder if we need to fix these values in the future, maybe add a tracking to see if we need to add permission related metadata in kernel's file status?

scottsand-db: Great callout! In terms of "tracking", I think that we should just create the issue (or, even better, the PR) if/when this ever becomes a problem.

unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
```diff
@@ -57,6 +91,19 @@ class UCCatalogManagedCommitterSuite
         protocolWithCatalogManagedSupport,
         basicPartitionedMetadata)))
 
+  private def getSingleElementRowIter(elem: String): CloseableIterator[Row] = {
```
scottsand-db, line 94: Just moved up from below

huan233usc (approved):
Still have questions on why put if absent cannot be used. But we resolve it as follow up.

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -103,4 +114,112 @@ private void validateLogPathBelongsToThisUcTable(CommitMetadata cm) {
         expectedDeltaLogPathNormalized,
         providedDeltaLogPathNormalized);
   }
+
+  private FileStatus writeStagedCommitFile(
+      Engine engine, CloseableIterator<Row> finalizedActions, CommitMetadata commitMetadata)
```
sumeet-db, line 119: When is `finalizedActions` closed?

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -103,4 +114,112 @@ private void validateLogPathBelongsToThisUcTable(CommitMetadata cm) {
...
+        // commitMetadata validates that the ICT is present if writing to a catalogManaged table
+        commitMetadata.getCommitInfo().getInCommitTimestamp().get());
+  }
+
+  @VisibleForTesting
+  public static org.apache.hadoop.fs.FileStatus kernelFileStatusToHadoopFileStatus(
+      io.delta.kernel.utils.FileStatus kernelFileStatus) {
+    return new org.apache.hadoop.fs.FileStatus(
+        kernelFileStatus.getSize() /* length */,
+        false /* isDirectory */,
```
sumeet-db, line 208: How did we come up with these constants?

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -103,4 +114,112 @@ private void validateLogPathBelongsToThisUcTable(CommitMetadata cm) {
...
+                .writeJsonFileAtomically(
+                    stagedCommitFilePath, finalizedActions, true /* overwrite */);
+
+            // TODO: [delta-io/delta#5021] Use FileSystemClient::getFileStatus API instead
+            return FileStatus.of(stagedCommitFilePath);
+          } catch (IOException ex) {
+            throw new CommitFailedException(
+                true /* retryable */,
+                false /* conflict */,
+                "Failed to write staged commit file due to: " + ex.getMessage(),
```
sumeet-db, line 148: Add an error log with the version and/or file name here?

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
sumeet-db: Technically, we should have `overwrite = false` here since if the file already exists, it indicates a UUID clash or some other issue. However, it comes with a negligible-to-small overhead of `putIfAbsent`. 

Will let @LukasRupprecht add more details here.

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
sumeet-db: To me, it feels like handling and testing for a uuid-clash scenario would be an overkill here.

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -103,4 +114,112 @@ private void validateLogPathBelongsToThisUcTable(CommitMetadata cm) {
...
+                false /* isDisown */,
+                Optional.empty() /* newMetadata */, // TODO: support sending newProtocol
+                Optional.empty()); // TODO: support sending newProtocol
+            return null;
+          } catch (io.delta.storage.commit.CommitFailedException cfe) {
+            throw storageCFEtoKernelCFE(cfe);
+          } catch (IOException ex) {
+            throw new CommitFailedException(
+                true /* retryable */, false /* conflict */, ex.getMessage(), ex);
+          } catch (UCCommitCoordinatorException ucce) {
```
sumeet-db, line 182: What about other errors like permissions issue / network failures? How are they handled?

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
scottsand-db: Great question. The engine `writeJsonFileAtomically` consumer is responsible for closing it 👍

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
scottsand-db: Claude 😓 I agree I don't think this is best, and e.g. asked for input on the original draft [PR](https://github.com/delta-io/delta/pull/5052/files#r2270570433) too

I don't think UC uses any of these fields? I suggest we merge this and can followup and increase the information required to be stored in the io.delta.kernel.FileStatus if needed

scottsand-db: cc @vkorukanti btw

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
scottsand-db: I'm just copying what the existing UCCommitCoordinatorClient does.

For example, the UCTokenBasedRestClint will handle various HTTP errors and may throw a CommitFailedException -> in which case, we just forward that along here.

General network failures should be IOException => we handle that above

UpgradeNotAllowedException and InvalidTargetTableException are both CFE(false, false) ==> so that is covered here too

CommitLimitReachedException => there's a TODO comment to backfill in the future, for now we just fail

Any other specific concerns?

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
scottsand-db: @emkornfield has suggested in the past that logging and then throwing is an anti-pattern.
@emkornfield any input here? I'd be happy to log .. e.g. what if the consumer (catcher) doesn't log or doesn't propagate the error?

emkornfield: So the issue with logging and rethrowing is that something up the stack will likely log as well creating log noise.  If there are details we want to capture, they would ideally go in the message, or as a structured part of the exception being rethrown.

If the catcher of the exception doesn't log OR propagate the error then I think that seems like them not doing a good job.  I'm not sure how much we want to account for calling code not having great practices?  In this case it is kernel that is intended to be the catcher? (and I think it already logs)?

emkornfield: So the issue with logging and rethrowing is that something up the stack will likely log as well creating log noise.  If there are details we want to capture, they would ideally go in the message, or as a structured part of the exception being rethrown.

If the catcher of the exception doesn't log OR propagate the error then I think that seems like them not doing a good job.  I'm not sure how much we want to account for calling code not having great practices?  In this case it is kernel that is intended to be the catcher? (and I think it already logs)?

scottsand-db: Kernel is the caller (for now). Kernel will log 👍 

https://github.com/delta-io/delta/blob/master/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java#L362C8-L363C85

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
LukasRupprecht: +1 to what @sumeet-db said. UUID clashes are so unlikely that we want to keep it simple and just do a regular PUT to save any overhead that comes from putIfAbsent.

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
```diff
@@ -177,6 +178,32 @@ public CommitType getCommitType() {
...
 
+  /**
+   * Returns the corresponding published Delta log file path for this commit, which is in the form
+   * of {@code <table_path>/_delta_log/0000000000000000000<version>.json}.
+   *
+   * <p>Usages:
+   *
+   * <ul>
+   *   <li>Filesystem-managed committers must write to this file path.
+   *   <li>Catalog-managed committers must backfill to this file path, if/when they so choose.
```
xzhseh: Backfill, or publish - should we align the terminology with the RFC?
```suggestion
   *   <li>Catalog-managed committers must publish to this file path, if/when they so choose.
```

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
xzhseh: > The engine writeJsonFileAtomically consumer is responsible for closing it 👍

If so, let's add a comment to explicitly document this.

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
xzhseh: +1 to what @sumeet-db said! We could add a comment explaining why we are _not_ catching `FileAlreadyExistsException` (e.g., as it's an extreme edge case and likely caused by UUID-clash issue), though.

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -103,4 +114,112 @@ private void validateLogPathBelongsToThisUcTable(CommitMetadata cm) {
...
+  }
+
+  private void commitToUC(CommitMetadata commitMetadata, FileStatus kernelStagedCommitFileStatus)
+      throws CommitFailedException {
+    timeCheckedOperation(
+        logger,
+        "Commit staged commit file to UC: " + kernelStagedCommitFileStatus.getPath(),
+        ucTableId,
+        () -> {
+          // commitToUc is only for normal catalog WRITES, not for CREATE, or UPGRADE, or DOWNGRADE,
```
xzhseh, line 161: So what's the expected paths for CREATE / UPGRADE / DOWNGRADE if we would need to interact with (or notify) UC regarding those commits?

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -103,4 +114,112 @@ private void validateLogPathBelongsToThisUcTable(CommitMetadata cm) {
...
+
+          try {
+            ucClient.commit(
+                ucTableId,
+                tablePath.toUri(),
+                Optional.of(getUcCommitPayload(commitMetadata, kernelStagedCommitFileStatus)),
+                Optional.empty() /* lastKnownBackfilledVersion */, // TODO: take this in as a hint
+                false /* isDisown */,
+                Optional.empty() /* newMetadata */, // TODO: support sending newProtocol
+                Optional.empty()); // TODO: support sending newProtocol
```
xzhseh: Nit:
```suggestion
                Optional.empty()); /* newProtocol */ // TODO: support sending newProtocol
```

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -103,4 +114,112 @@ private void validateLogPathBelongsToThisUcTable(CommitMetadata cm) {
...
+            return null;
+          } catch (io.delta.storage.commit.CommitFailedException cfe) {
+            throw storageCFEtoKernelCFE(cfe);
+          } catch (IOException ex) {
+            throw new CommitFailedException(
+                true /* retryable */, false /* conflict */, ex.getMessage(), ex);
+          } catch (UCCommitCoordinatorException ucce) {
+            // For now, this catches all UC exceptions such as:
+            // - CommitLimitReachedException -> TODO: backfill in this case
```
xzhseh: Nit:
```suggestion
            // - CommitLimitReachedException -> TODO: publish in this case
```

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
xzhseh: > I don't think UC uses any of these fields?

AFAIK UC is agnostic to `FileStatus` at all, so it's okay to put them in this way for now. Also can we add a TODO here for subsequent follow-up / tracking?

unity/src/test/scala/io/delta/unity/UCCatalogManagedClientSuite.scala
```diff
@@ -37,7 +37,11 @@ import org.scalatest.funsuite.AnyFunSuite
 /** Unit tests for [[UCCatalogManagedClient]]. */
 class UCCatalogManagedClientSuite extends AnyFunSuite with UCCatalogManagedTestUtils {
 
-  private def createUCCatalogManagedClientForTableWithNoRatifiedCommits(
+  /**
+   * When a new UC table is created, it will have Delta version 0 but the max ratified verison in
+   * UC is -1. This is a special edge case.
```
xzhseh, line 43: Add a TODO here for future refactor in case we support UPGRADE / DOWNGRADE on UC side - so that the max ratified version would be up-to-date.

unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
```diff
@@ -205,15 +281,153 @@ class UCCatalogManagedCommitterSuite
...
   }
 
-  test("the remaining commit logic is not yet implemented") {
-    val ucClient = new InMemoryUCClient("ucMetastoreId")
-    val committer = new UCCatalogManagedCommitter(ucClient, "ucTableId", baseTestTablePath)
-    val commitMetadata = catalogManagedWriteCommitMetadata(version = 1)
+  test("CATALOG_WRITE: writes staged commit file and invokes UC client commit API") {
+    withTempTableAndLogPathAndStagedCommitFolderCreated { case (tablePath, logPath) =>
+      // ===== GIVEN =====
+      // Setup UC client with initial table with maxRatifiedVersion = 0, numCommits = 1
```
xzhseh: To be consistent with the current semantics, let's change the `maxRatifiedVersion` to `-1` for initial table that has just been created.

unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
```diff
@@ -205,15 +281,153 @@ class UCCatalogManagedCommitterSuite
...
+      assert(ex.getMessage.contains("Failed to write staged commit file due to: Network error"))
+    }
+  }
+
+  test("CATALOG_WRITE: i.d.s.c.CommitFailedException during UC commit => kernel CFE") {
+    withTempTableAndLogPathAndStagedCommitFolderCreated { case (tablePath, logPath) =>
+      // ===== GIVEN =====
+      val ucClient = new InMemoryUCClient("ucMetastoreId") {
+        override def forceThrowInCommitMethod(): Unit =
+          throw new io.delta.storage.commit.CommitFailedException(
```
xzhseh, line 362: Instead of mocking the `CommitFailedException` that will be thrown when there is any conflict - how about creating a actual concurrent test case that has two concurrent writers committing to the same version?

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
scottsand-db: Great question! My understanding is:
(a) for CREATE --> that is a filesystem-only 000.json write, right? not going through UC's commit API? that's what delta-spark and the UC-Commit-Coordinator-Client do.

(b) UPGRADE and DOWNGRADE can be added incrementally later :) not worrying about it in this PR.

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
scottsand-db: I don't love adding TODOs for casing that we _might_, or might not, want to support/fix in the future. I think we will know/remember in the futrue that if the FileStatus information in Kernel UC commits becomes important, that this is the area we have to update.

unity/src/test/scala/io/delta/unity/UCCatalogManagedClientSuite.scala
scottsand-db: Hey @xzhseh -- sorry, can you please elabroate on how a server-side UPGRADE/DOWNGRADE will impact the expected max ratified version of -1 for the CREATE case?

unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
scottsand-db: Yup! Will implement more integration-like tests in subsequent PRs! But for now I want to mock and have individual, specific, simple unit tests 👍

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
xzhseh: No strong opinion here, either works!

unity/src/test/scala/io/delta/unity/UCCatalogManagedClientSuite.scala
xzhseh: IMO the CREATE case is just a specialized version of UPGRADE. E.g., We are doing `FS_TO_CC` commit to "_upgrade_" a file-system based delta table to catalog-based for the 0th commit.

We could notify UC synchronously (and immediately) at the 0th commit as part of the server side UPGRADE/DOWNGRADE logic - so the max ratified version would then be 0 rather than -1.

cc @LukasRupprecht for visibility.

unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
xzhseh: Sounds good!

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
xzhseh: CREATE would eventually go through UC commit API - but I agree we can keep the current implementation and refactor afterwards if needed.

xzhseh (approved):
LGTM!

PR #5080, opened 2025-08-14
done

scottsand-db:
## 🥞 Stacked PR

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)


kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWritesSuite.scala
```diff
@@ -66,6 +66,18 @@ abstract class AbstractDeltaTableWritesSuite extends AnyFunSuite with AbstractWr
   // Create table tests
   ///////////////////////////////////////////////////////////////////////////
 
+  test("create table: _delta_log and _staged_commits directories are created") {
```
zachschuermann: could we still test by just passing in a specific commit type manually? (to test the _delta_log existing, no _staged_commits)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -980,4 +973,30 @@ private void printLogForRetryableWithConflictException(
         commitAsVersion,
         cfe);
   }
+
+  private void createDeltaLogAndStagedCommitsDirectoriesIfNeeded(
```
allisonport-db: What will happen if we have a catalog table without the _staged_commits directory?

Should we check for that and either (1) create the directory anyways or (2) throw invalid table exception?

allisonport-db:
I guess to piggy-back on @zachschuermann should it be conditional on whether catalogManaged is enabled? or `CATALOG_CREATE` possibly?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWritesSuite.scala
scottsand-db: Will add that test 👍

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
scottsand-db: I don't think we should check that, nor throw. For all we know, someone is using a filesystem-based catalog that uses S3 put-if-absent as it's commit broker. They never write into the `_staged_commits` directory. Perhaps the catalog deleted it?

I'm just trying to really quickly copy what delta-spark is doing https://github.com/delta-io/delta/blob/3b75a3c3ea7a44d49cc7c3f8c2b89f5af76ce5c6/spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala#L2052-L2058

https://github.com/delta-io/delta/blob/48203d8fd720d26f049d0649cc8c53196e8c9e5b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala#L528

IMO we can add further complexity / checks later

scottsand-db:
@allisonport-db & @zachschuermann -- I totally understand the argument to only create the `_staged_commits` directory if this is a catalogManaged table. For now, I'm trying to just move fast and copy the delta-spark logic: 

https://github.com/delta-io/delta/blob/3b75a3c3ea7a44d49cc7c3f8c2b89f5af76ce5c6/spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala#L2052-L2058

https://github.com/delta-io/delta/blob/48203d8fd720d26f049d0649cc8c53196e8c9e5b/spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala#L528

For now, I'd rather err on the side of alignment with delta-spark.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
allisonport-db: Hm, maybe I'm lost, but should this maybe be done by the committer instead then? Maybe it doesn't fit well anywhere there?

scottsand-db: Kernel already has the `engine` and can make the directory. No need to put this on the Committer.

allisonport-db (approved):
LGTM

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DirectoryCreationSuite.scala
```diff
@@ -0,0 +1,179 @@
...
+import org.scalatest.funsuite.AnyFunSuite
+
+class DirectoryCreationSuite extends AnyFunSuite with WriteUtils with ActionUtils {
+
+  ///////////////
+  // E2E Tests //
+  ///////////////
+
+  // TODO: test _staged_commits directory when catalogManaged writes are supported
+  test("create table with v2Checkpoints supported: _delta_log and _sidecars directories are" +
```
scottsand-db: https://github.com/delta-io/delta/pull/5088

PR #5083, opened 2025-08-14
[Kernel] Refactor: Make P & M in CommitMetadata constructor coupled

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/5083/files) to review incremental changes.
- [**stack/kernel_commit_metadata_p_and_m_refactor**](https://github.com/delta-io/delta/pull/5083) [[Files changed](https://github.com/delta-io/delta/pull/5083/files)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Minor clean up of logic in the CommitMetadata class. Wish I had done this sooner. Oop.

Also had some better version <=> readState (P & M) checks.

## How was this patch tested?

Update existing UTs and add a few new UTs.

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
@@ -23,7 +23,6 @@ import io.delta.kernel.annotation.Experimental;
 import io.delta.kernel.internal.actions.CommitInfo;
 import io.delta.kernel.internal.actions.Metadata;
 import io.delta.kernel.internal.actions.Protocol;
-import io.delta.kernel.internal.annotation.VisibleForTesting;
 import io.delta.kernel.internal.tablefeatures.TableFeatures;
 import io.delta.kernel.internal.util.Tuple2;
 import java.util.Optional;
@@ -140,7 +139,7 @@ public class CommitMetadata {
    * the protocol that was read at the beginning of the commit.
    */
   public Protocol getEffectiveProtocol() {
-    return newProtocolOpt.orElseGet(() -> readPandMOpt.get()._1);
+    return newProtocolOpt.orElseGet(() -> getReadProtocolOpt().get());
   }
 
   /**
@@ -149,7 +148,7 @@ public class CommitMetadata {
    * metadata that was read at the beginning of the commit.
    */
   public Metadata getEffectiveMetadata() {
-    return newMetadataOpt.orElseGet(() -> readPandMOpt.get()._2);
+    return newMetadataOpt.orElseGet(() -> getReadMetadataOpt().get());
   }
 
   /**
@@ -191,10 +190,5 @@ public class CommitMetadata {
           commitInfo.getInCommitTimestamp().isPresent(),
           "InCommitTimestamp must be present for commits to catalogManaged tables");
     }
-  }
-
-  @VisibleForTesting
-  public Optional<Tuple2<Protocol, Metadata>> getReadProtocolAndMetadata() {
-    return readPandMOpt;
   }
 }


kernel/kernel-api/src/test/scala/io/delta/kernel/internal/commit/CommitMetadataSuite.scala
```diff
@@ -32,88 +33,74 @@ class CommitMetadataSuite extends AnyFunSuite
...
         logPath,
         testCommitInfo(),
-        Optional.of(protocol12),
-        Optional.of(basicPartitionedMetadata),
-        Optional.empty(),
-        Optional.empty())
+        Optional.empty(), // readPandMOpt
+        Optional.empty(), // newProtocolOpt
+        Optional.empty()
+      ) // newMetadataOpt
```
scottsand-db, line 48: fix

allisonport-db (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
```diff
@@ -55,38 +57,33 @@ public enum CommitType {
...
   private final Optional<Protocol> newProtocolOpt;
   private final Optional<Metadata> newMetadataOpt;
 
   public CommitMetadata(
       long version,
       String logPath,
       CommitInfo commitInfo,
-      Optional<Protocol> readProtocolOpt,
-      Optional<Metadata> readMetadataOpt,
+      Optional<Tuple2<Protocol, Metadata>> readPandMOpt,
```
emkornfield, line 67: Is Typle2 the best option here, I thought on a prior PR you mentioned something about ResolvedSnapshot or something similar?

emkornfield: If not, it still might be more readable to define a simple java option to contain these to elements.

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
```diff
@@ -143,7 +140,7 @@ public Optional<Metadata> getNewMetadataOpt() {
    * the protocol that was read at the beginning of the commit.
    */
   public Protocol getEffectiveProtocol() {
-    return newProtocolOpt.orElseGet(readProtocolOpt::get);
+    return newProtocolOpt.orElseGet(() -> readPandMOpt.get()._1);
```
emkornfield: `getReadProtocol.get()`?

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
scottsand-db: > Is Typle2 the best option here, I thought on a prior PR you mentioned something about ResolvedSnapshot or something similar?

I had considered a `ResolvedMetadata`. Even a `Snapshot` ... but ... for the former, that doesn't exist yet, and I'm not sure if this change here necessitates it yet .. and I don't think we should take a in a full snapshot, that would make testing a bit harder.

Basically .. I'm fine with this current change. It's a step in the right direction. And when a use case/scenario presents itself that really crisply guides us to change this (e.g. create a `ResolvedMetadata` class, I'm open to refactoring this then.
 
> If not, it still might be more readable to define a simple java option to contain these to elements.

By this do you mean a new `class`?

emkornfield: sorry yes, new class.

kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
```diff
@@ -55,38 +57,33 @@ public enum CommitType {
   private final long version;
   private final String logPath;
   private final CommitInfo commitInfo;
-  private final Optional<Protocol> readProtocolOpt;
-  private final Optional<Metadata> readMetadataOpt;
+  private final Optional<Tuple2<Protocol, Metadata>> readPandMOpt;
   private final Optional<Protocol> newProtocolOpt;
   private final Optional<Metadata> newMetadataOpt;
 
   public CommitMetadata(
```
raveeram-db, line 63: Not related to this PR, but would be given this is a public API, woudl be good to document this Ctor's signature and expectations. Especially the expected semantics around `checkReadStateAbsentIfAndOnlyIfVersion0`

PR #5084, opened 2025-08-15
[Spark] CDC Reader Optimization

sezruby:

- [X] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Optimize CDC Reader path by using logical plan directly, instead of RDD scan implementation at runtime. 
This approach helps apply Spark's optimization rules for CDC Scan, improves performance with existing Spark/Parquet optimizations. It also makes Spark query plan statistics visible in Spark UI.

The feature can be controlled via 2 configs:
```
  val DELTA_CDF_BATCH_STATIC_READER =
    buildConf("changeDataFeed.batch.staticReader")
      .doc(
        s"""If enabled, Delta uses an explicit query plan to read CDC changes instead of RDD impl.
           |This enables query plan optimizations via Spark Plan Optimizer when applicable.
           |The original CDF reader implementation uses Spark PrunedFilteredScan API,
           |which performs a slower RDD-based scan and does not show jobs in the Spark UI.
           |This optimization is applied to CDF reads with start and end timestamps/versions
           |specified.
           |If changeDataFeed.batch.staticReader.startOnly.enabled is set to true, it will
           |also apply the optimization to the CDF reads with start version only
           |and the end version is decided at Spark Query Plan Analyzer, not the execution time.
           """.stripMargin)
      .internal()
      .booleanConf
      .createWithDefault(true)

  val DELTA_CDF_BATCH_STATIC_READER_START_ONLY =
    buildConf("changeDataFeed.batch.staticReader.startOnly.enabled")
      .doc(
        s"""If enabled, changeDataFeed.batch.staticReader optimization also applies
           |CDF read queries with start version only, non streaming scenario.
           |The main caveat with the optimization is that, if the end version is not specified,
           |it might not use the latest version if the table is updated after analysis.
           """.stripMargin)
      .internal()
      .booleanConf
      .createWithDefault(true)
```

`changeDataFeed.batch.staticReader.startOnly.enabled = true` might cause functionality change as it decides the end version at query analysis time. If users relies on the execution time end version, it would be breaking change.

Example for better understanding of the limitation:
1)	val cdcDf = spark.sql("select * from table_changes('table1', 0)") // no end version is specified, startVersion = 0, table version = 10
2)	println(cdcDf.schema) or cdcDf.queryExecution.optimizedPlan => df.queryExecution.analyzedPlan built
3)	newDataDf.write.mode(“append”).saveAsTable(“table1”) // new commit made to table1, table version 11
4)	cdcDf.collect() // this CDC read doesn’t include table version 11, new DataDf in 3)
5)	newDataDf2.write.mode(“append”).saveAsTable(“table1”) // new commit made to table1, table version 12
6)	cdcDf.collect() // this CDC read doesn’t include table version 11-12, new DataDf2 in 5)

## How was this patch tested?

Unit test coverage / quick manual perf testing:

Without fix:
Screenshot 2025-08-14 162826
With fix:
Screenshot 2025-08-14 162916

20% gain, local 14 core cluster. The performance gain would be bigger when using a native engine and bigger clusters.

Test script:
```
from pyspark.sql.functions import col, lit
spark.range(100000000).withColumn("value", col("id") + 4).write.format("delta").mode("overwrite").save(tablePath)
spark.sql("delete from delta.`" + tablePath + "` where id = 2")
spark.range(100000000).withColumn("value", col("id") + 4).write.format("delta").mode("overwrite").save(tablePath)
spark.conf.set("spark.databricks.delta.changeDataFeed.batch.staticReader", "true")
spark.sql("select sum(id), sum(value) from tabLe_chaNges_By_pAth('" + tablePath + "', 0, 2) group by _change_type").collect()
```

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/commands/cdc/CDCReader.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/commands/cdc/CDCReader.scala
@@ -237,7 +237,16 @@ object CDCReader extends CDCReaderImpl
         snapshotWithSchemaMode.snapshot
     }
 
-    override val schema: StructType = cdcReadSchema(snapshotForBatchSchema.metadata.schema)
+    override val schema: StructType = {
+      cdcReadSchema(
+        DeltaTableUtils.removeInternalDeltaMetadata(
+          sqlContext.sparkSession,
+          DeltaTableUtils.removeInternalWriterMetadata(
+            sqlContext.sparkSession, snapshotForBatchSchema.metadata.schema
+          )
+        )
+      )
+    }
 
     def getScanDf: DataFrame = {
       startingVersion match {
@@ -834,7 +843,14 @@ trait CDCReaderImpl extends DeltaLogging {
       readSchemaSnapshot, isStreaming, spark)
     dfs.append(deletedAndAddedRows: _*)
 
-    val readSchema = cdcReadSchema(readSchemaSnapshot.metadata.schema)
+    val readSchema = cdcReadSchema(
+      DeltaTableUtils.removeInternalDeltaMetadata(
+        spark,
+        DeltaTableUtils.removeInternalWriterMetadata(
+          spark, readSchemaSnapshot.metadata.schema
+        )
+      )
+    )
     // build an empty DS. This DS retains the table schema and the isStreaming property
     // NOTE: We need to manually set the stats to 0 otherwise we will use default stats of INT_MAX,
     // which causes lots of optimizations to be applied wrong.
--- a/spark/src/test/scala/org/apache/spark/sql/delta/DeltaTimeTravelSuite.scala
+++ b/spark/src/test/scala/org/apache/spark/sql/delta/DeltaTimeTravelSuite.scala
@@ -967,21 +967,23 @@ class DeltaTimeTravelSuite extends QueryTest
 
   // 2) SELECT ... CHANGES AS OF
   test("Block CDC beyond deletedFileRetention") {
-    withTestTable { (tbl, t4, t6) =>
-      Seq(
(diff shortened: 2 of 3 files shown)


sharing/src/test/scala/io/delta/sharing/spark/DeltaSharingDataSourceCMSuite.scala
```diff
@@ -96,46 +96,31 @@ class DeltaSharingDataSourceCMSuite
       deltaTableName: String,
       sharedTablePath: String,
       startingVersion: Int): Unit = {
-    val schema = spark.read
```
sezruby, line 99: Moved definitions as it throws the expected exception early with the change.

spark/src/main/scala/org/apache/spark/sql/delta/commands/cdc/CDCReader.scala
```diff
@@ -193,6 +193,73 @@ object CDCReader extends CDCReaderImpl
     }
   }
 
+  /**
+   * A special BaseRelation wrapper for CDF reads for optimization at query planning stage.
+   */
+  case class DeltaCDFRelationAtAnalysis(
```
sezruby, line 199: copy of DeltaCDFRelation , without `with CatalystScan` / `override def buildScan/unhandledFilters`. 
new getScanDf for query plan generation

sezruby:
Disabling the config by default would be totally okay.
With a native engine, it shows more than 2x gain for the scan part. Apparently RDD execution at query time doesn't fully get benefit from query optimizer.

8 core * 8 executors, with native engine:

Without fix:
image

With fix:
image

PR #5088, opened 2025-08-15
done first pass

scottsand-db:
## 🥞 Stacked PR

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
```diff
@@ -330,6 +337,54 @@ private void performProtocolUpgrades(boolean clusteringRequired) {
...
+    // catalogManaged is being enabled in this transaction. catalogManaged dependsOn
+    // inCommitTimestamp. The inCommitTimestamp feature should have been auto-supported in the
+    // protocol by now.
+    checkState(
+        getEffectiveProtocol().supportsFeature(TableFeatures.IN_COMMIT_TIMESTAMP_W_FEATURE),
+        "inCommitTimestamp feature expected to have already been supported");
+
+    // Case 2: Txn is explicitly disabling ICT. Throw.
+    final String txnIctEnabledValue =
+        originalUserInputProperties.get(TableConfig.IN_COMMIT_TIMESTAMPS_ENABLED.getKey());
```
emkornfield, line 382: I thought there were helpers to extract this given the map, instead of checking the map explicitly?  Maybe I was imagining it because I couldn't find it on a quick search?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
```diff
@@ -330,6 +337,54 @@ private void performProtocolUpgrades(boolean clusteringRequired) {
     }
   }
 
+  /** STEP 1.5: Handle catalogManaged enablement. Updates the METADATA if applicable. */
+  private void handleCatalogManagedEnablement() {
+    final boolean readVersionSupportsCatalogManaged =
```
emkornfield, line 359: It seems this handles create, but does it also need to handle replace logic, or there is nothing different needed to consider for replace?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
```diff
@@ -246,6 +250,7 @@ private TransactionMetadataFactory(
       Optional<SnapshotImpl> latestSnapshotOpt,
       Optional<Metadata> initialNewMetadata,
       Optional<Protocol> initialNewProtocol,
+      Map<String, String> originalUserInputProperties,
```
allisonport-db, line 270: Can't we just use the table properties in `initialNewMetadata`?

allisonport-db: If there is no newMetadata -- no need to do this check anyways since there was no table properties input

scottsand-db: Great question.

We need access to the userInputProperties specified right now.

- if the table is at version 10, and ICT is already explicitly disbaled at this vesion, then if you enable catalogManaged in this txn, we will enable ICT for you
- if the table is at version 10, and then you explicitily enable catalogManaged AND disable ICT in this txn, then we throw

See how we need to be able to differentiate between: ICT is already disabled on the table, and ICT is explicitly being disabled by the table properties set in _this_ transaction?

allisonport-db: Can you take a look at how we do this for iceberg compat? We can definitely do it without this input using the table properties either from newMetadata and existingMetadata.

newMetadata.tableProperties essentially should == userInputProperties

scottsand-db: `newMetadata.tableProperties essentially should == userInputProperties`

I don't think that's true. Look at the UPDATE code, for example:

```java
    if (!newProperties.isEmpty()) {
      newMetadata = Optional.of(readSnapshot.getMetadata().withMergedConfiguration(newProperties));
    }
```

`newMetadata.tableProperties` is equal to the existing table metadata with all new user properties merged in

allisonport-db: I see

emkornfield: probably worth providing a short synopsis of the reasoning here as a comment.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
scottsand-db: > I thought there were helpers to extract this given the map, instead of checking the map explicitly? 

Yup! There is. In fact, we use it below!

```java
final boolean isIctAlreadyEnabled =
        TableConfig.IN_COMMIT_TIMESTAMPS_ENABLED.fromMetadata(getEffectiveMetadata());
```

However, this utility combines two important cases: the field does not exist (default is false) and the field is set to false.

But what we need to check is if it is explicitly set to false (which is different than it not being set).

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
scottsand-db: Great question. I don't think we need to do anything special for REPLACE. We just need to look at (a) what are the user input properties, and (b) what are the current latest new metadata table properties, and validate/update/throw from there.

scottsand-db: Err. Actually. REPLACE and enabling catalogManaged should be blocked. I can add that in a future PR.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
emkornfield: maybe we should add this helper to the Config classes?  OK to do as a follow-up.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
emkornfield: Is it hard to do here, so we don't forget about it?

scottsand-db: I'll add it to the project plan 👍 I won't forget about it 🫡 Im trying to keep these PRs small

PR #5100, opened 2025-08-18
[Kernel] Block unsupported catalogManaged enablement / ICT disablement on catalogManaged table scenarios

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/5100/files) to review incremental changes.
- [**stack/kernel_block_replace_while_enabling_catalogManaged**](https://github.com/delta-io/delta/pull/5100) [[Files changed](https://github.com/delta-io/delta/pull/5100/files)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Block setting catalogManaged=supported during REPLACE operation. Also block a few other invalid operations, like disabling ICT on a catalogManaged table, whether it is during an UPDATE or a REPLACE.

## How was this patch tested?

New UT.

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
@@ -419,10 +419,17 @@ public class TransactionMetadataFactory {
     final boolean txnExplicitlyDisablesICT =
         txnIctEnabledValue != null && txnIctEnabledValue.equalsIgnoreCase("false");
 
+    // Case 1: Txn is not explicitly disabling ICT. Exit.
     if (!txnExplicitlyDisablesICT) {
       return;
     }
 
+    // Case 2: Txn is explicitly disabling ICT on a catalogManaged table. Throw.
+    if (getEffectiveProtocol().supportsFeature(TableFeatures.CATALOG_MANAGED_R_W_FEATURE_PREVIEW)) {
+      throw new KernelException("Cannot disable inCommitTimestamp on a catalogManaged table");
+    }
+
+    // Case 3 (normal case): Txn is explicitly disabling ICT. Remove the ICT enablement properties.
     final Set<String> ictKeysToRemove =
         new HashSet<>(
             Arrays.asList(
--- a/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/catalogManaged/CatalogManagedEnablementSuite.scala
+++ b/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/catalogManaged/CatalogManagedEnablementSuite.scala
@@ -31,7 +31,7 @@ import org.scalatest.funsuite.AnyFunSuite
 
 class CatalogManagedEnablementSuite extends AnyFunSuite with TestUtils {
 
-  case class CatalogManagedEnablementTestCase(
+  case class CatalogManagedTestCase(
       testName: String,
       operationType: String, // "CREATE", "UPDATE", or "REPLACE"
       initialTableProperties: Map[String, String] = Map.empty,
@@ -42,82 +42,101 @@ class CatalogManagedEnablementSuite extends AnyFunSuite with TestUtils {
       expectedCatalogManagedSupported: Boolean = false)
 
   val catalogManagedTestCases = Seq(
-    CatalogManagedEnablementTestCase(
-      testName = "CREATE: catalogManaged enablement flag => enables catalogManaged and ICT",
+    // ===== CREATE cases =====
+    CatalogManagedTestCase(
+      testName = "CREATE: set catalogManaged=supported => enables catalogManaged and ICT",
       operationType = "CREATE",
       transactionProperties = Map("delta.feature.catalogOwned-preview" -> "supported"),
       expectedSuccess = true,
       expectedIctEnabled = true,
       expectedCatalogManagedSupported = true),
-    CatalogManagedEnablementTestCase(
-      testName = "UPDATE: catalogManaged enablement flag => enables catalogManaged and ICT",
+    CatalogManagedTestCase(
+      testName = "ILLEGAL CREATE: set catalogManaged=supported and explicitly disable ICT => THROW",
+      operationType = "CREATE",
+      transactionProperties = Map(
+        "delta.feature.catalogOwned-preview" -> "supported",
+        "delta.enableInCommitTimestamps" -> "false"),
+      expectedSuccess = false,
+      expectedExceptionMessage =
+        Some("Cannot disable inCommitTimestamp when enabling catalogManaged")),
+
+    // ===== UPDATE cases =====
+    CatalogManagedTestCase(
+      testName = "UPDATE: set catalogManaged=supported => enables catalogManaged and ICT",
       operationType = "UPDATE",
       initialTableProperties = Map.empty, // Start with basic table
       transactionProperties = Map("delta.feature.catalogOwned-preview" -> "supported"),
       expectedSuccess = true,
       expectedIctEnabled = true,
       expectedCatalogManagedSupported = true),
-    CatalogManagedEnablementTestCase(
-      testName = "UPDATE: catalogManaged enablement flag => enables ICT if previously disabled",
+    CatalogManagedTestCase(
+      testName = "UPDATE: set catalogManaged=supported => enables ICT if previously disabled",
       operationType = "UPDATE",
       initialTableProperties = Map("delta.enableInCommitTimestamps" -> "false"),
       transactionProperties = Map("delta.feature.catalogOwned-preview" -> "supported"),
       expectedSuccess = true,
       expectedIctEnabled = true,
       expectedCatalogManagedSupported = true),
-    CatalogManagedEnablementTestCase(
-      testName =
-        "ILLEGAL CREATE: catalogManaged enablement flag but ICT explicitly disabled too => THROW",
-      operationType = "CREATE",
-      transactionProperties = Map(
-        "delta.feature.catalogOwned-preview" -> "supported",
-        "delta.enableInCommitTimestamps" -> "false"),
-      expectedSuccess = false,
-      expectedExceptionMessage =
-        Some("Cannot disable inCommitTimestamp when enabling catalogManaged")),
-    CatalogManagedEnablementTestCase(
-      testName =
-        "ILLEGAL UPDATE: catalogManaged enablement flag but ICT explicitly disabled too => THROW",
-      operationType = "UPDATE",
-      initialTableProperties = Map.empty,
-      transactionProperties = Map(
-        "delta.feature.catalogOwned-preview" -> "supported",
-        "delta.enableInCommitTimestamps" -> "false"),
-      expectedSuccess = false,
-      expectedExceptionMessage =
-        Some("Cannot disable inCommitTimestamp when enabling catalogManaged")),
-    CatalogManagedEnablementTestCase(
-      testName = "UPDATE: catalogManaged enablement flag => ICT already enabled",
+    CatalogManagedTestCase(
+      testName = "UPDATE: set catalogManaged=supported and ICT already enabled => Okay",
       operationType = "UPDATE",
       initialTableProperties = Map("delta.enableInCommitTimestamps" -> "true"),
       transactionProperties = Map("delta.feature.catalogOwned-preview" -> "supported"),
       expectedSuccess = true,
       expectedIctEnabled = true,
       expectedCatalogManagedSupported = true),
-    CatalogManagedEnablementTestCase(
-      testName = "No-op: catalogManaged not being enabled should not affect ICT",
+    CatalogManagedTestCase(
+      testName = "ILLEGAL UPDATE: set catalogManaged=supported and disable ICT => THROW",
+      operationType = "UPDATE",
+      initialTableProperties = Map.empty,
+      transactionProperties = Map(
+        "delta.feature.catalogOwned-preview" -> "supported",
+        "delta.enableInCommitTimestamps" -> "false"),
+      expectedSuccess = false,
+      expectedExceptionMessage =
+        Some("Cannot disable inCommitTimestamp when enabling catalogManaged")),
+    CatalogManagedTestCase(
+      testName = "ILLEGAL UPDATE: catalogManaged already supported, then disable ICT => THROW",
+      operationType = "UPDATE",
+      initialTableProperties = Map("delta.feature.catalogOwned-preview" -> "supported"),
+      transactionProperties = Map("delta.enableInCommitTimestamps" -> "false"),
+      expectedSuccess = false,
+      expectedExceptionMessage =
+        Some("Cannot disable inCommitTimestamp on a catalogManaged table")),
+    CatalogManagedTestCase(
+      testName = "NO-OP UPDATE: catalogManaged not being enabled should not affect ICT",
       operationType = "UPDATE",
       initialTableProperties = Map.empty,
       transactionProperties = Map(),
       expectedSuccess = true,
       expectedIctEnabled = false,
       expectedCatalogManagedSupported = false),
-    CatalogManagedEnablementTestCase(
-      testName = "ILLEGAL REPLACE: catalogManaged enablement flag => THROW",
+
+    // ===== REPLACE cases =====
+    CatalogManagedTestCase(
+      testName = "REPLACE: normal replace should succeed on a catalogManaged table",
+      operationType = "REPLACE",
+      initialTableProperties = Map("delta.feature.catalogOwned-preview" -> "supported"),
+      transactionProperties = Map(),
+      expectedSuccess = true,
+      expectedIctEnabled = true,
+      expectedCatalogManagedSupported = true),
+    CatalogManagedTestCase(
+      testName = "ILLEGAL REPLACE: set catalogManaged=supported => THROW",
       operationType = "REPLACE",
       initialTableProperties = Map.empty,
       transactionProperties = Map("delta.feature.catalogOwned-preview" -> "supported"),
       expectedSuccess = false,
       expectedExceptionMessage =
         Some("Cannot enable the catalogManaged feature during a REPLACE command.")),
-    CatalogManagedEnablementTestCase(
-      testName = "REPLACE: should succeed on a catalogManaged table",
+    CatalogManagedTestCase(
+      testName = "ILLEGAL REPLACE: catalogManaged already supported, then disable ICT => THROW",
       operationType = "REPLACE",
       initialTableProperties = Map("delta.feature.catalogOwned-preview" -> "supported"),
-      transactionProperties = Map(),
-      expectedSuccess = true,
-      expectedIctEnabled = true,
-      expectedCatalogManagedSupported = true))
+      transactionProperties = Map("delta.enableInCommitTimestamps" -> "false"),
+      expectedSuccess = false,
+      expectedExceptionMessage =
+        Some("Cannot disable inCommitTimestamp on a catalogManaged table")))
 
   catalogManagedTestCases.foreach { testCase =>
     test(testCase.testName) {
@@ -199,7 +218,7 @@ class CatalogManagedEnablementSuite extends AnyFunSuite with TestUtils {
         } else {
           // Transaction building should fail
           val exception = intercept[Exception] {
-            txnBuilder.build(defaultEngine).commit(defaultEngine, emptyIterable[Row])
+            txnBuilder.build(defaultEngine)
           }
 
           testCase.expectedExceptionMessage.foreach { expectedMsg =>


kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
```diff
@@ -658,4 +659,13 @@ private static void validateSchemaAndPartColsCreateOrReplace(
     SchemaUtils.validateSchema(schema, isColumnMappingModeEnabled(mappingMode));
     SchemaUtils.validatePartitionColumns(schema, partitionColumns);
   }
+
+  private static void validateNotEnablingCatalogManagedOnReplace(
```
allisonport-db, line 699: If the table is already a catalog managed table is that okay? Should it be preserved (remain catalog managed)? (it won't be currently fyi)

scottsand-db: ~~Ah, it should be maintained 👍 Will include that here, too.~~

scottsand-db: So, REPLACE will retain certain table properties, `TABLE_PROPERTY_KEYS_TO_PRESERVE`.

But this is a protocol value with no matching table property.

And replace just keeps the protocol from the previous snapshot.

So this should be fine?

scottsand-db: Updated so that we always retain ICT enablement flag. Synced with @xzhseh that this is what we should be doing.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/tablefeatures/TableFeatures.java
```diff
@@ -607,6 +607,17 @@ public static Optional<Tuple2<Protocol, Set<TableFeature>>> autoUpgradeProtocolB
...
 
+  /**
+   * Checks if a table feature is being manually supported through user property {@code
+   * delta.feature.<featureName>=supported}.
+   */
+  public static boolean isPropertiesManuallySupportingTableFeature(
+      Map<String, String> userProperties, TableFeature tableFeature) {
+    final String featurePropKey = SET_TABLE_FEATURE_SUPPORTED_PREFIX + tableFeature.featureName();
+    final String propertyValue = userProperties.get(featurePropKey); // will be null if not found
+    return "supported".equals(propertyValue);
```
xzhseh: Should we consider making `"supported"` a `public/protected static final` somewhere?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ReplaceTableTransactionBuilderV2Impl.java
```diff
@@ -27,12 +27,23 @@
...
 
+  /**
+   * Generally for replace table we want to reset all table state, however there are a few
+   * delta-specific properties that we should preserve
+   */
+  static final Set<String> TABLE_PROPERTY_KEYS_TO_PRESERVE =
+      new HashSet<String>() {
+        {
+          add(TableConfig.COLUMN_MAPPING_MAX_COLUMN_ID.getKey());
+          add(TableConfig.IN_COMMIT_TIMESTAMPS_ENABLED.getKey());
```
xzhseh: Let's add a comment here explaining why ICT needs to be preserved during an active REPLACE commands. E.g., We would want to keep the ICT enablement flag / version / timestamp to properly track the table history, etc. cc @dhruvarya-db.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ReplaceTableTransactionBuilderV2Impl.java
```diff
@@ -27,12 +27,23 @@
...
+          // TODO are there any other table properties we should preserve?
```
xzhseh: I'm wondering do we have any available docs that track the interactions between different table features <-> DDL commands? If not, should we start consolidating one to align on the intended semantics?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
```diff
@@ -658,4 +659,13 @@ private static void validateSchemaAndPartColsCreateOrReplace(
...
+      Map<String, String> tableProperties) {
+    if (TableFeatures.isPropertiesManuallySupportingTableFeature(
+        tableProperties, TableFeatures.CATALOG_MANAGED_R_W_FEATURE_PREVIEW)) {
+      throw new UnsupportedOperationException(
+          "Cannot enable the catalogManaged feature during a REPLACE operation.");
```
xzhseh: Nit:
```suggestion
          "Cannot enable the catalogManaged feature during a REPLACE command.");
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
xzhseh, line 699: Can we be more specific for the input parameter here - IIUC the `tableProperties` should be explicit user overrides rather than all the existing table properties. Also let's add a comment explaining why we wouldn't throw the exception for an existing CC table to this helper.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/catalogManaged/CatalogManagedEnablementSuite.scala
```diff
@@ -170,26 +197,22 @@ class CatalogManagedEnablementSuite extends AnyFunSuite with TestUtils {
             .build(defaultEngine)
             .asInstanceOf[SnapshotImpl]
 
+          val protocol = snapshot.getProtocol
+
           // Check if catalogManaged feature is supported
-          val catalogManagedSupported = snapshot.getProtocol
+          val catalogManagedSupported = protocol
             .supportsFeature(TableFeatures.CATALOG_MANAGED_R_W_FEATURE_PREVIEW)
-          assert(
```
xzhseh, line 161: Out of curiosity, why are we removing the debug assertion messages here (and below)?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/catalogManaged/CatalogManagedEnablementSuite.scala
```diff
@@ -157,6 +173,17 @@ class CatalogManagedEnablementSuite extends AnyFunSuite with TestUtils {
               .build(defaultEngine)
               .buildUpdateTableTransaction("engineInfo", Operation.MANUAL_UPDATE)
               .withTablePropertiesAdded(testCase.transactionProperties.asJava)
+
+          case "REPLACE" =>
+            val replaceSchema = schema.add("col2", IntegerType.INTEGER)
+
+            TableManager
```
xzhseh, line 180: QQ: Does the `buildReplaceTableTransaction` cover both `REPLACE` and `RTAS` scenarios? Also, can we add independent command-level tests (and how hard would that be)?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ReplaceTableTransactionBuilderV2Impl.java
scottsand-db: I'd defer to the ICT/metadata experts there for what that comment should be :) 

@dhruvarya-db could you comment in this thread what you'd like it to be?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/catalogManaged/CatalogManagedEnablementSuite.scala
scottsand-db: IMO they add verbosity to the test suite -- the asserts are usually quite clear and self-explanatory. I'd rather aim for readability and succinctness

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/catalogManaged/CatalogManagedEnablementSuite.scala
scottsand-db: > Does the buildReplaceTableTransaction cover both REPLACE and RTAS scenarios? 

I defer to @allisonport-db 

> Also, can we add independent command-level tests (and how hard would that be)?

@xzhseh -- Can you please elabroate a bit on what you mean?

xzhseh: > @xzhseh -- Can you please elabroate a bit on what you mean?

E.g., Do we have similar test that can cover the below scenarios in kernel - so we have some E2E test coverage as well.
```sql
REPLACE TABLE test_table (col1 INTEGER, col2 INTEGER) ...

REPLACE TABLE test_table (col1 INTEGER, col2 INTEGER) AS SELECT ...
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ReplaceTableTransactionBuilderV2Impl.java
dhruvarya-db: We should also preserve `IN_COMMIT_TIMESTAMP_ENABLEMENT_VERSION` and `IN_COMMIT_TIMESTAMP_ENABLEMENT_TIMESTAMP`. 
1. By default, we should retain all three properties.
2. If the client explicitly disables `IN_COMMIT_TIMESTAMPS_ENABLED`, we should drop all three properties though.

Retaining `IN_COMMIT_TIMESTAMP_ENABLEMENT_VERSION` and `IN_COMMIT_TIMESTAMP_ENABLEMENT_TIMESTAMP` is important because otherwise the client will not know when ICT was enabled. This means could either result in a query failure or incorrect results being served during time travel depending on the implementation.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/catalogManaged/CatalogManagedEnablementSuite.scala
allisonport-db: yes you can just add data using the normal path via Transaction for RTAS

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ReplaceTableTransactionBuilderV2Impl.java
allisonport-db: Do we implement this here in this PR? (Retaining the `IN_COMMIT_TIMESTAMP_ENABLEMENT_VERSION` and `IN_COMMIT_TIMESTAMP_ENABLEMENT_TIMESTAMP`). I think not but maybe I'm missing something.

scottsand-db: > Do we implement this here in this PR? 

Not yet! Was waiting for all reviewers to take a look before I pushed changes.

allisonport-db: @dhruvarya-db comment SGTM

kernel/kernel-api/src/main/java/io/delta/kernel/internal/tablefeatures/TableFeatures.java
scottsand-db: Done!

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/catalogManaged/CatalogManagedEnablementSuite.scala
scottsand-db: I'm inclined to just keep this as is for now ... more robust RTAS testing in general can be done in the future, but IMO not a P0 here

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ReplaceTableTransactionBuilderV2Impl.java
scottsand-db: Hey @xzhseh -- can you give some examples of such interactions? Here is a concrete example of two table features (ccv2, ICT) and REPLACE. Any others? Agreed we should track this.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ReplaceTableTransactionBuilderV2Impl.java
```diff
@@ -27,12 +27,27 @@
...
-import java.util.Optional;
+import java.util.*;
 
 public class ReplaceTableTransactionBuilderV2Impl implements ReplaceTableTransactionBuilder {
 
+  /**
+   * Generally for replace table we want to reset all table state, however there are a few
+   * delta-specific properties that we should preserve
+   */
+  static final Set<String> TABLE_PROPERTY_KEYS_TO_PRESERVE =
```
allisonport-db: @dhruvarya-db Following up on your prior comment, if ICT is disabled explicitly, we should drop all ICT properties right?

I think we might need to add some custom logic to implement this?

allisonport-db: I think we need to test this change in logic as well in `DeltaReplaceTableSuite` as well. Maybe we should make this fix in a separate PR? ICT property preservation is independent of this catalogManged logic

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/catalogManaged/CatalogManagedEnablementSuite.scala
```diff
@@ -99,13 +95,29 @@ class CatalogManagedEnablementSuite extends AnyFunSuite with TestUtils {
...
     CatalogManagedEnablementTestCase(
-      testName = "No-op: catalogOwned not being enabled should not affect ICT",
+      testName = "No-op: catalogManaged not being enabled should not affect ICT",
       operationType = "UPDATE",
       initialTableProperties = Map.empty,
       transactionProperties = Map(),
       expectedSuccess = true,
       expectedIctEnabled = false,
-      expectedCatalogManagedSupported = false))
+      expectedCatalogManagedSupported = false),
```
allisonport-db, line 113: Should there be a failure case for replace where ICT is explicitly disabled ==> illegal config since ICT is required for catalogManaged

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ReplaceTableTransactionBuilderV2Impl.java
scottsand-db: > if ICT is disabled explicitly, we should drop all ICT properties right?

My takeaway: these are the properties that we should preserve, unless that is explicitly changed in the txn. Any new overrides should correctly override.

So if you do a normal REPLACE, then ICT props are preserverd.

If you do a REPLACE and then explicitly disable ICT, then ICT should be explicitly disabled.

And I believe that's what's implemented here

I kinda am inclined to ship this PR (maybe a few more tests in CatalogManagedEnablementSuite) as we have a  users that want this feature soon, and perhaps more tests in DeltaReplaceTableSuite owned by @dhruvarya-db in another PR?

But if there's big pushback, I can move all this ICT stuff to another PR.

allisonport-db: @dhruvarya-db Can you clarify? Should the `IN_COMMIT_TIMESTAMP_ENABLEMENT_VERSION` and `IN_COMMIT_TIMESTAMP_ENABLEMENT_TIMESTAMP` props be dropped if ICT is explicitly disabled?

@scottsand-db Just to clarify, I am concerned about the non-user-facing ICT properties here (^)

allisonport-db: I am fine with following up in another PR the removal of those non-user facing properties + the tests for that in `DeltaReplaceTableSuite`. Could we add basic tests in `DeltaReplaceTableSuite` for the behavior change in this PR though?

scottsand-db: > PR the removal of those non-user facing properties

@allisonport-db -- can you clarify what this section of the code needs to do for the non-user-facing properties? My understanding is: we need to keep them in the delta metadata configuration. And what is/is-not user-facing is just a matter of filtering properties when someone runs a `SHOW TBLPROPERTIES` command.

allisonport-db: My understanding from @dhruvarya-db was that if the user explicitly disables ICT in the table properties, we should also remove the non-user-facing properties from the config since ICT is no longer enabled. But not sure I interpreted this correctly.

By user-facing, I mean set by the user vs set internally, not so much about what should be displayed/returned to the user.

I think maybe we are just not aligned on "we need to keep them in the delta metadata configuration" so happy to just let @dhruvarya-db clarify.

dhruvarya-db: There three state relevant state transitions:
1. ICT disabled -> ICT enabled: We should add both the enablement tracking properties
2. ICT enabled -> ICT enabled (REPLACE, any DML, etc): We should retain the enablement tracking properties
3. ICT enabled -> ICT disabled: it is highly recommended that the two enablement tracking properties be removed. Even if we forget to remove them, it should not cause any issues.

scottsand-db: It seems like this PR has fundamentally turned into an ICT PR/discussion. I'll make a new prefactor PR explicitly for ICT and we can discuss there

scottsand-db: Hi @dhruvarya-db -- Regarding point 1 in the code, I see the test case:

https://github.com/delta-io/delta/blob/c8c830424ec2874b9201e1e4a98b5ac18a8264b6/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/InCommitTimestampSuite.scala#L245

So this is an exception to point 1? or is CREATE jsut a special case?

Also -- can you help me understand why we do this?

```
  test("Enablement tracking properties should not be added if ICT is enabled on commit 0") {
      setTablePropAndVerify(
        engine = engine,
        tablePath = tablePath,
        key = IN_COMMIT_TIMESTAMPS_ENABLED,
        value = "true",
        expectedValue = true)

      val ver0Snapshot =
        Table.forPath(engine, tablePath).getLatestSnapshot(engine).asInstanceOf[SnapshotImpl]
      assertHasNoMetadataProp(ver0Snapshot, IN_COMMIT_TIMESTAMP_ENABLEMENT_TIMESTAMP)
      assertHasNoMetadataProp(ver0Snapshot, IN_COMMIT_TIMESTAMP_ENABLEMENT_VERSION)
    }
  }
```

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/catalogManaged/CatalogManagedEnablementSuite.scala
xzhseh: +1

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ReplaceTableTransactionBuilderV2Impl.java
xzhseh: > Hey @xzhseh -- can you give some examples of such interactions? Here is a concrete example of two table features (ccv2, ICT) and REPLACE. Any others? Agreed we should track this.

E.g., CCv2's specific semantics with CLONE / REPLACE / CREATE, etc.
- For CLONE we won't preserve CCv2 unless explicitly specified.
- For REPLACE we retain the existing CCv2 / non-CCv2 properties.
- Also there are other cases when handling dependent features such as ICT...

And for ICT and other table features they have their own specific treatments w.r.t. these DDL commands that should be explicitly documented / tracked.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/catalogManaged/CatalogManagedEnablementSuite.scala
scottsand-db: Oop still need to add. Will do.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ReplaceTableTransactionBuilderV2Impl.java
scottsand-db: Created https://github.com/delta-io/delta/pull/5156

scottsand-db:
Pausing this. Want to merge https://github.com/delta-io/delta/pull/5156 first.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ReplaceTableTransactionBuilderV2Impl.java
scottsand-db: I see. Thanks for elaborating! I think each component of the code should track this, at the very least. For example in this PR, for ICT + REPLACE, we document this here: https://github.com/delta-io/delta/pull/5156/files#diff-d3436360ac15655035f35b01d8e7c2bc27db976db99b04b1709ed90d64be372aR54-R61

We should at the very least do that. Also, if you happen to have a list of these such semantics in your head, perhaps we should work to document them in some README.md somewhere?

cc @nicklan and @raveeram-db for some context

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/catalogManaged/CatalogManagedEnablementSuite.scala
```diff
@@ -46,42 +42,51 @@ class CatalogManagedEnablementSuite extends AnyFunSuite with TestUtils {
...
-      testName = "UPDATE: catalogManaged enablement flag => enables ICT if previously disabled",
+    CatalogManagedTestCase(
+      testName = "UPDATE: set catalogManaged=supported => enables ICT if previously disabled",
       operationType = "UPDATE",
       initialTableProperties = Map("delta.enableInCommitTimestamps" -> "false"),
       transactionProperties = Map("delta.feature.catalogOwned-preview" -> "supported"),
       expectedSuccess = true,
       expectedIctEnabled = true,
       expectedCatalogManagedSupported = true),
-    CatalogManagedEnablementTestCase(
```
scottsand-db, line 72: Just moved this up. Also, I clean up some of the test case names to be more succinct and clearer

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/catalogManaged/CatalogManagedEnablementSuite.scala
```diff
@@ -90,22 +95,48 @@ class CatalogManagedEnablementSuite extends AnyFunSuite with TestUtils {
...
       operationType = "UPDATE",
       initialTableProperties = Map.empty,
       transactionProperties = Map(),
       expectedSuccess = true,
       expectedIctEnabled = false,
-      expectedCatalogManagedSupported = false))
+      expectedCatalogManagedSupported = false),
+
+    // ===== REPLACE cases =====
+    CatalogManagedTestCase(
```
scottsand-db, line 116: New test cases

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
```diff
@@ -418,10 +419,17 @@ private void handleInCommitTimestampDisablement() {
     final boolean txnExplicitlyDisablesICT =
         txnIctEnabledValue != null && txnIctEnabledValue.equalsIgnoreCase("false");
 
+    // Case 1: Txn is not explicitly disabling ICT. Exit.
     if (!txnExplicitlyDisablesICT) {
       return;
     }
 
+    // Case 2: Txn is explicitly disabling ICT on a catalogManaged table. Throw.
```
xzhseh, line 427: For kernel - is it true that we don't currently have a generic interface that automatically prevents dependent features being turned off? The current implementation LGTM but we may want to generalize it a bit more..

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/catalogManaged/CatalogManagedEnablementSuite.scala
```diff
@@ -19,25 +19,21 @@ package io.delta.kernel.defaults.catalogManaged
 import scala.collection.JavaConverters._
```
xzhseh, line 19: **_Not a blocker for this PR_**: Since we are also including DDL commands <> CCv2 semantics, I feel like we need a separate `CatalogManagedPropertySuite` to test all the interactions. Also we need to check for `ucTableId` as we currently store it in delta log.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/catalogManaged/CatalogManagedEnablementSuite.scala
```diff
@@ -90,22 +95,48 @@ class CatalogManagedEnablementSuite extends AnyFunSuite with TestUtils {
       expectedSuccess = false,
       expectedExceptionMessage =
         Some("Cannot disable inCommitTimestamp when enabling catalogManaged")),
-    CatalogManagedEnablementTestCase(
-      testName = "UPDATE: catalogManaged enablement flag => ICT already enabled",
+    CatalogManagedTestCase(
```
xzhseh, line 98: Can we also test REPLACE + ICT + CCv2 interactions?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
scottsand-db: We don't have that yet, yes.

We are able to say which table features require other table features -- https://github.com/delta-io/delta/blob/a0826e32207f8f22e5f8c2a53a90ee757c7c0251/kernel/kernel-api/src/main/java/io/delta/kernel/internal/tablefeatures/TableFeatures.java#L92

But this has to do with the **protocol**, which is different from the enablement flag.

It would be nice to be able to additionally specify that not only does this protocol feature X require another protocol Y to be supported, but Y must also be enabled, and then you cannot disable Y, too.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/catalogManaged/CatalogManagedEnablementSuite.scala
scottsand-db: +1 on having a CatalogManagedPropertySuite! Will add that to the tracker.

> Also we need to check for ucTableId as we currently store it in delta log.

Great question / callout! That is UC specific -- and is completely outside the scope of Delta Kernel! That would be done in the `delta-unity` JAR (in this repo) -- but do remember that Kernel never cares what type of catalogManaged table/catalog (UC, glue, etc.) it is reading/talking to.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/catalogManaged/CatalogManagedEnablementSuite.scala
scottsand-db: I have `ILLEGAL REPLACE: catalogManaged already supported, then disable ICT => THROW"` below. is there another combination you would like me to test?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/catalogManaged/CatalogManagedEnablementSuite.scala
scottsand-db: For example, we should add some checks + tests in delta-unity that prevent us from removing the uc table ID property perhaps?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/ReplaceTableTransactionBuilderV2Impl.java
xzhseh: > Also, if you happen to have a list of these such semantics in your head, perhaps we should work to document them in some README.md somewhere?

IMO there would be two ways to do it:
- Maintain a delta kernel table features <> DDL commands track list somewhere as the source of truth for all table feature implementations in kernel.
- Or, in a more general (and breaking) way, update the delta protocol to include the specific DDL-related semantics of each table feature in its corresponding section.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionMetadataFactory.java
xzhseh: > But this has to do with the protocol, which is different from the enablement flag.

Yeah unless we also introduce the metadata enablement flag <> table feature mappings.. E.g., we know table feature X is automatically supported/disabled when the X-enablement-flag is turned on/off.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/catalogManaged/CatalogManagedEnablementSuite.scala
xzhseh: > That is UC specific -- and is completely outside the scope of Delta Kernel!

Ah yes, let's add those UTs in `delta-unity` then.

> For example, we should add some checks + tests in delta-unity that prevent us from removing the uc table ID property perhaps?

Exactly. We should throw exception(s) whenever someone tries to (explicitly) modify `ucTableId` via any DDL commands.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/catalogManaged/CatalogManagedEnablementSuite.scala
xzhseh: > I have ILLEGAL REPLACE: catalogManaged already supported, then disable ICT => THROW" below. is there another combination you would like me to test?

Ah missed this one. Yeah it seems sufficient to me, thanks!

Issue #5001, opened 2025-07-31
[Feature Request]Implement Dsv2 in kernel


huan233usc:
## Overview

Please see https://github.com/delta-io/delta/issues/4718

Also see https://github.com/delta-io/delta/issues/4985 for kernel improvement for Dsv2

## Project Tracking

Merged = ✅ 
Needs Review = 👀
Waiting for merge = ☑   
Abandoned = 🛑  

### [COMPLETE ✅ ] Milestone 0.1: Setup repo

| Description | PR | Status | Created | Merged |
|--------|--------|--------|--------|--------|
| setup separate module | https://github.com/delta-io/delta/commit/f2b79c1d4b66c2aea6279a49c778b5dcc7e89fa6 | ✅ | 07/10/25 | 07/30/25 |

### [In progress ] Milestone 0.2: Basic read

| Description | PR | Status | Created | Merged |
|--------|--------|--------|--------|--------|
| Schema translation | https://github.com/delta-io/delta/pull/4997 | ✅ | 07/30/25 | 08/05/25 |
| Table scaffolding | https://github.com/delta-io/delta/pull/5024 |  ✅ | 08/05/25 | |
| Test catalog binding Table implementation | https://github.com/delta-io/delta/pull/5035 |  ✅ |  | 08/16/25|
| Support Read Mix-in interface | https://github.com/delta-io/delta/pull/5053 | ✅   |  | 08/16/25|
| Scan Builder | https://github.com/delta-io/delta/pull/5053 |  ✅  |  | 08/16/25|
| Batch Interface |  |  |  | |
| Partition Reader |  |  |  | |
| Path based table in test catalog | https://github.com/delta-io/delta/pull/5106 | 👀 |  | |
| E2E integration for basic case|  |  |  | |

PR #5102, opened 2025-08-19
[Kernel][Dsv2]Implement Batch interface for Batch Scan

huan233usc:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This PR implements Batch interface for Batch Scan in DSV2. Note that the plan partition could be called multiple times in the lifecycle of query and Scan.getScanFile could only called once, we implement a wrapper to cache the iterator result and make it be able to accessed multiple times.

see https://github.com/delta-io/delta/issues/5001 for overall tracking

## How was this patch tested?

Unit

No


spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/scan/KernelSparkScanContext.java
```diff
@@ -0,0 +1,112 @@
...
+          List<FilteredColumnarBatch> batchList = new ArrayList<>();
+          try (CloseableIterator<FilteredColumnarBatch> scanFiles =
+              kernelScan.getScanFiles(engine)) {
+            while (scanFiles.hasNext()) {
+              FilteredColumnarBatch batch = scanFiles.next();
+              batchList.add(batch);
+            }
+          } catch (IOException e) {
+            LOG.warn("Failed to read from kernel scan", e);
+            throw new UncheckedIOException("Failed to read added files from kernel scan", e);
```
huan233usc, line 102: Follow iceberg to handle ioe during table expansion

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/scan/KernelSparkScanBuilder.java
```diff
@@ -30,9 +31,11 @@ public class KernelSparkScanBuilder implements org.apache.spark.sql.connector.re
 
   private final ScanBuilder kernelScanBuilder;
   private final StructType sparkReadSchema;
+  private final Engine engine;
 
-  public KernelSparkScanBuilder(SnapshotImpl snapshot) {
+  public KernelSparkScanBuilder(SnapshotImpl snapshot, Engine engine) {
```
scottsand-db: nit: in every Kernel API I can think of, the `Engine` param comes first. I like to think of the Engine as being one of the most important params. Just a tiny style that I like.

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/scan/KernelSparkScanBuilder.java
```diff
@@ -30,9 +31,11 @@ public class KernelSparkScanBuilder implements org.apache.spark.sql.connector.re
 
   private final ScanBuilder kernelScanBuilder;
   private final StructType sparkReadSchema;
+  private final Engine engine;
```
scottsand-db: btw -- no other way to "inject" the engine, eh? We typically try to avoid saving a reference to it. cc @vkorukanti

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/scan/KernelSparkScan.java
```diff
@@ -26,18 +29,22 @@
  */
 public class KernelSparkScan implements org.apache.spark.sql.connector.read.Scan {
```
scottsand-db, line 31: btw -- is this created on the driver or executor? in general btw I would really like it if our classes really made it clear where they are ceated (Driver, Executor) and when/if they need to be serialized sent back and forth

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/scan/KernelSparkScanContext.java
```diff
@@ -0,0 +1,112 @@
...
+ * Shared context for Kernel-based Spark scans that manages scan state and partition planning.
+ * Caches FilteredColumnarBatch data and converts them to Spark InputPartitions on demand.
+ */
+public class KernelSparkScanContext {
+
+  private static final Logger LOG = LoggerFactory.getLogger(KernelSparkScanContext.class);
+
+  private final Scan kernelScan;
+  private final Engine engine;
+  private final String serializedScanState;
```
scottsand-db: I wonder: Can we abstract away the serialization format?

could this be a class `KernelSerializedState` -- and perhaps there's a runtime json implementation of it, but it can easily be switched out later to be bytes/something else?

I'd like to avoid hard-coding to String and I wonder if we can do ourselves a favor and make updating this later a bit easier

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/scan/KernelSparkScanContext.java
```diff
@@ -0,0 +1,112 @@
...
+  private final AtomicReference<Optional<List<FilteredColumnarBatch>>> cachedBatches;
```
scottsand-db: cachedBatches of _what_?

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/scan/KernelSparkScanContext.java
```diff
@@ -0,0 +1,112 @@
...
+   * metadata.
+   * 
+   * <p>Results are cached on first invocation using thread-safe lazy initialization. Subsequent
+   * calls return the partitions calculated based on the same cached batches..
+   * 
+   * @return array of InputPartitions for Spark to process
+   * @throws UncheckedIOException if kernel scan fails or partition creation fails
+   */
+  public InputPartition[] planPartitions() {
+    Optional<List<FilteredColumnarBatch>> batches = cachedBatches.get();
```
scottsand-db: could do we use some tiny methods?

```java
public InputPartition[] planPartitions() {
  final List<FilteredColumnarBatch> batches = loadScanBatches()
  return convertBatchesToInputPartitions(..);
}

private List<FilteredColumnarBatch> loadScanBatches() {
  ...
}
```

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/scan/KernelSparkScanContext.java
```diff
@@ -0,0 +1,112 @@
...
+            throw new UncheckedIOException("Failed to read added files from kernel scan", e);
+          }
+          batches = Optional.of(batchList);
+          cachedBatches.set(batches);
+        }
+      }
+    }
+
+    // Create InputPartitions from cached batches
+    // Implementation note: not caching partitions for the future extension of runtime filtering.
```
scottsand-db: Can we make this note just a bit clearer and expressive? e.g. just elaborate a bit more?

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/scan/batch/KernelSparkBatchScan.java
```diff
@@ -0,0 +1,44 @@
...
+import static java.util.Objects.requireNonNull;
+
+import io.delta.spark.dsv2.scan.KernelSparkScanContext;
+import java.io.Serializable;
+import org.apache.spark.sql.connector.read.Batch;
+import org.apache.spark.sql.connector.read.InputPartition;
+import org.apache.spark.sql.connector.read.PartitionReaderFactory;
+
+/** Spark Batch implementation backed by Delta Kernel Scan. */
+public class KernelSparkBatchScan implements Batch, Serializable {
```
scottsand-db: Sorry - why this class and also `KernelSparkScanContext`?

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/scan/batch/KernelSparkInputPartition.java
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Spark InputPartition implementation that holds serialized Delta Kernel scan information. Contains
+ * both scan state and add files.
+ */
+public final class KernelSparkInputPartition implements InputPartition, Serializable {
+
+  private final String serializedScanState;
+  // TODO: [delta-io/delta#5109] implement the logic to group files in to partition based on file
+  // size.
+  // Json representation of one add file in kernel
```
scottsand-db: nit: for TODOs, let's alsign the `size.` to be under the `[` --> this indentation will make both lines show up properly in intellij for example

```java
// TODO: <first line>
//       next line...
```
nit: method docs and variable docs should use `/** */` -- for example, one might assume that `// Json representation of one add file in kernel` is part of the TODO comment.

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/scan/batch/KernelSparkInputPartition.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.io.Serializable;
+import java.util.Objects;
+import org.apache.spark.sql.connector.read.InputPartition;
+
+/**
+ * Spark InputPartition implementation that holds serialized Delta Kernel scan information. Contains
+ * both scan state and add files.
+ */
+public final class KernelSparkInputPartition implements InputPartition, Serializable {
```
scottsand-db, line 30: would love to have it documented that this is serialized between Driver and Executor. I know it implements `Serializable` but I do think an explicit comment could help new engineers onboard

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/scan/KernelSparkScanBuilder.java
huan233usc: I changed to passing hadoop configuration, and only create engine if needed

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/scan/KernelSparkScan.java
huan233usc: Driver, added documentation and added for batch and streaming(for future extension) table scan.

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/scan/KernelSparkScanContext.java
huan233usc: Added a wrapper class

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/scan/batch/KernelSparkBatchScan.java
huan233usc: (Serializable is added accidentally -- we don't need it)

Mainly for future use as KernelSparkScanContext need to be mutable (like pushing runtime filter)

https://github.com/apache/spark/blob/master/sql/catalyst/src/main/java/org/apache/spark/sql/connector/read/SupportsRuntimeV2Filtering.java


KernelSparkBatchScan  scan1 = Scan.toBatch // file after data skipping using static predicate set by user
Scan.filter() // mutate KernelSparkScanContext by adding new filter
KernelSparkBatchScan  scan2 = Scan.toBatch // file after data skipping using dynamic predicate produce by planner

scottsand-db: Makes sense. Can we add some of this info/context/usage details to the classdocs? And explicitly call out this class' relationship to KernelSparkScanContext?

i.e. probably just duplicate a short 1-2 sentences for both classes javadocs

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/scan/KernelSparkScanContext.java
```diff
@@ -0,0 +1,124 @@
...
+import org.apache.spark.sql.connector.read.InputPartition;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Shared context for Kernel-based Spark scans that manages scan state and partition planning.
+ *
+ * <p>This class is created on the Driver and handles the conversion of Delta Kernel scan files into
+ * Spark InputPartitions. It caches scan file batches to support multiple calls to planPartitions()
+ * and ensures consistent partition planning.
```
scottsand-db, line 44: I commented this elsewhere but commenting here for visibility: let's explicitly include what is the relationship between this class and KernelSparkScan?

scottsand-db: Or perhaps in relation to KernelSparkBatchScan i think? I defer to you

spark-kernel-dsv2/src/test/java/io/delta/spark/dsv2/scan/KernelSparkScanContextTest.java
```diff
@@ -0,0 +1,138 @@
...
+      KernelSparkInputPartition p2 = (KernelSparkInputPartition) partitions2[i];
+      assertEquals(p1.getSerializedScanState(), p2.getSerializedScanState());
+      assertEquals(p1.getSerializedScanFileRow(), p2.getSerializedScanFileRow());
+    }
+  }
+
+  //////////////////////
+  // Private helpers //
+  /////////////////////
+  private void createTestTable(String path, String tableName) {
```
scottsand-db: perhaps `createTestTableWIthData`

spark-kernel-dsv2/src/test/java/io/delta/spark/dsv2/scan/batch/KernelSparkBatchScanTest.java
```diff
@@ -0,0 +1,116 @@
...
+    UnsupportedOperationException exception =
+        assertThrows(UnsupportedOperationException.class, batchScan::createReaderFactory);
+
+    assertEquals("reader factory is not implemented", exception.getMessage());
+  }
+
+  //////////////////////
+  // Private helpers //
+  /////////////////////
+  private void createTestTable(String path, String tableName) {
```
scottsand-db: question: is it worth moving this to a common test utils class? we don't have to; it's only one method that is duplicated here and in the test suite above. i don't want to end up with just muddied, complex, tangled test utils that are hard to navigate. perhaps this test table is just unique to this class, and so putting it somewhere common isn't needed? i defer to you

allisonport-db (approved):
LGTM

PR #5106, opened 2025-08-19
[Kernel][Dsv2][Test only]Support path based table in TestCatalog

huan233usc:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

Support path based table in TestCatalog so that we could test the connector more easily before dsv2 connector support writes. (we could insert using existing delta-spark connector and read using new connector then)

Sample testing idea is in the early poc for reference https://github.com/delta-io/delta/pull/5105

## How was this patch tested?

Add unit test

No

### DIFF
--- a/spark-kernel-dsv2/src/test/java/io/delta/spark/dsv2/catalog/TestCatalog.java
+++ b/spark-kernel-dsv2/src/test/java/io/delta/spark/dsv2/catalog/TestCatalog.java
@@ -19,6 +19,7 @@ import io.delta.kernel.Operation;
 import io.delta.kernel.TableManager;
 import io.delta.kernel.defaults.engine.DefaultEngine;
 import io.delta.kernel.engine.Engine;
+import io.delta.kernel.exceptions.TableNotFoundException;
 import io.delta.kernel.internal.SnapshotImpl;
 import io.delta.kernel.utils.CloseableIterable;
 import io.delta.spark.dsv2.table.DeltaKernelTable;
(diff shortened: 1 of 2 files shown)


spark-kernel-dsv2/src/test/java/io/delta/spark/dsv2/catalog/TestCatalog.java
```diff
@@ -68,15 +69,24 @@ public Identifier[] listTables(String[] namespace) {
...
+      tablePath = tablePaths.get(tableKey);
+    }
     if (tablePath == null) {
       throw new NoSuchTableException(ident);
     }
     try {
       // Use TableManager.loadTable to load the table
       SnapshotImpl snapshot = (SnapshotImpl) TableManager.loadSnapshot(tablePath).build(engine);
       return new DeltaKernelTable(ident, snapshot);
+    } catch (TableNotFoundException kernelException) {
```
raveeram-db: is the `kernelException` name here intended?

raveeram-db (approved):
LGTM! Just a small nit

spark-kernel-dsv2/src/test/java/io/delta/spark/dsv2/catalog/TestCatalog.java
huan233usc: was named it as it is kernel side thrown error. But it seems that a generic naming is better

scottsand-db (approved):
LGTM!

Issue #5095, opened 2025-08-18
[BUG] [Kernel] Scan.transformPhysicalSchema does not get applied to nested fields


vpapavas:
### Describe the problem
The iterator in `Scan.transformPhysicalSchema`  https://github.com/delta-io/delta/blob/master/kernel/kernel-api/src/main/java/io/delta/kernel/Scan.java#L143 uses the physical schema to read but converts to logical before returning the rows to the user. The physical to logical transformation only happens for top-level fields but not nested structs. In my test, I saw that the schema in the ColumnarBatch has the correct logical names for the nested fields but the ColumnVector data type has the physical schema. 

#### Steps to reproduce

1. Write parquet files with a schema that has nested struct columns 
2. Read using 
```
final Table table = Table.forPath(engine, rootTablePath);
final Snapshot snapshot = table.getLatestSnapshot(engine);
final ScanBuilder scanBuilder = snapshot.getScanBuilder();
final StructType physicalReadSchema = ScanStateRow.getPhysicalDataReadSchema(engine, scanState);
final CloseableIterator<ColumnarBatch> physicalDataIter = engine
              .getParquetHandler()
              .readParquetFiles(singletonCloseableIterator(fileStatus), physicalReadSchema, Optional.empty())
              .map(res -> res.getData());
final CloseableIterator<FilteredColumnarBatch> transformedData = Scan.transformPhysicalData(
                engine,
                scanState,
                scanFileRow,
                physicalDataIter
              )
```
4. Check the resulting ColumnarBatch. Specifically, look at the schema of the vectors that still uses physical names.

#### Observed results

The schema of the `ColumnVector` uses the physical names for nested columns.

#### Expected results

Transform physical to logical for all fields, also nested.

### Environment information

* Delta Lake version: kernel-07-30-2025
* Spark version:
* Scala version:

### Willingness to contribute

The Delta Lake Community encourages bug fix contributions. Would you or another member of your organization be willing to contribute a fix for this bug to the Delta Lake code base?

- [ ] Yes. I can contribute a fix for this bug independently.
- [ ] Yes. I would be willing to contribute a fix for this bug with guidance from the Delta Lake community.
- [ ] No. I cannot contribute a bug fix at this time.

PR #5112, opened 2025-08-19
[Kernel][Spark]Implement PartitionReader for kernel based dsv2

huan233usc:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This is the row based PartitionReader, columnar based will be added later.

### DIFF
--- a/spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/scan/batch/KernelSparkPartitionReader.java
+++ b/spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/scan/batch/KernelSparkPartitionReader.java
@@ -88,9 +88,7 @@ public class KernelSparkPartitionReader implements PartitionReader<InternalRow> 
     return false;
   }
 
-  /**
-   * Returns the current row. Call only after {@link #next()} returns true.
-   */
+  /** Returns the current row. Call only after {@link #next()} returns true. */
   @Override
   public InternalRow get() {
     if (!initialized) {
@@ -107,17 +105,16 @@ public class KernelSparkPartitionReader implements PartitionReader<InternalRow> 
       return;
     }
 
-    try {
-      CloseableIterator<ColumnarBatch> physicalDataIter =
-          engine
-              .getParquetHandler()
-              .readParquetFiles(
-                  Utils.singletonCloseableIterator(
-                      InternalScanFileUtils.getAddFileStatus(scanFileRow)),
-                  ScanStateRow.getPhysicalDataReadSchema(scanState),
-                  // TODO: [delta-io/delta#5119] push down filter to parquet.
-                  Optional.empty())
-              .map(FileReadResult::getData);
+    try (CloseableIterator<ColumnarBatch> physicalDataIter =
+        engine
+            .getParquetHandler()
+            .readParquetFiles(
+                Utils.singletonCloseableIterator(
+                    InternalScanFileUtils.getAddFileStatus(scanFileRow)),
+                ScanStateRow.getPhysicalDataReadSchema(scanState),
+                // TODO: [delta-io/delta#5119] push down filter to parquet.
+                Optional.empty())
+            .map(FileReadResult::getData)) {
       CloseableIterator<FilteredColumnarBatch> logicalDataIter =
           Scan.transformPhysicalData(engine, scanState, scanFileRow, physicalDataIter);
       dataIterator = new RowIteratorAdapter(logicalDataIter);


spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/scan/batch/KernelSparkPartitionReader.java
```diff
@@ -0,0 +1,172 @@
...
+    @Override
+    public boolean hasNext() {
+      if (currentBatchRows != null && currentBatchRows.hasNext()) {
+        return true;
+      }
+
+      // Try to get next batch
+      while (batchIterator.hasNext()) {
+        FilteredColumnarBatch batch = batchIterator.next();
+        currentBatchRows = batch.getRows();
```
raveeram-db, line 162: should `currentBatchRows` be closed when moving on to the next FilteredColumnarBatch?

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/scan/batch/KernelSparkPartitionReader.java
```diff
@@ -0,0 +1,172 @@
...
+   */
+  @Override
+  public InternalRow get() {
+    if (!initialized || dataIterator == null) {
+      throw new IllegalStateException("Reader not initialized. Call next() first.");
+    }
+    if (!dataIterator.hasNext()) {
+      throw new IllegalStateException("No more data available.");
+    }
+    return dataIterator.next();
```
raveeram-db: Does this behavior match the comment above? Looks like get() is expected to return the same row until next is called? Should we be advancing the iterator in next() instead?

huan233usc: good catch, the semantic is different from the kernel's iterator. Updated according to spark's api

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/scan/batch/KernelSparkPartitionReader.java
huan233usc: Yes, we should. added

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/scan/batch/KernelSparkPartitionReader.java
```diff
@@ -0,0 +1,189 @@
...
+          engine
+              .getParquetHandler()
+              .readParquetFiles(
+                  Utils.singletonCloseableIterator(
+                      InternalScanFileUtils.getAddFileStatus(scanFileRow)),
+                  ScanStateRow.getPhysicalDataReadSchema(scanState),
+                  // TODO: [delta-io/delta#5119] push down filter to parquet.
+                  Optional.empty())
+              .map(FileReadResult::getData);
+      CloseableIterator<FilteredColumnarBatch> logicalDataIter =
```
raveeram-db, line 118: I guess if there's an exception in the transformPhysicalData, we still need to close the above `physicalDataIter`

huan233usc: update the code to try-with-resource so it will close for exception

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/scan/batch/KernelSparkPartitionReader.java
```diff
@@ -0,0 +1,186 @@
...
+            .getParquetHandler()
+            .readParquetFiles(
+                Utils.singletonCloseableIterator(
+                    InternalScanFileUtils.getAddFileStatus(scanFileRow)),
+                ScanStateRow.getPhysicalDataReadSchema(scanState),
+                // TODO: [delta-io/delta#5119] push down filter to parquet.
+                Optional.empty())
+            .map(FileReadResult::getData)) {
+      CloseableIterator<FilteredColumnarBatch> logicalDataIter =
+          Scan.transformPhysicalData(engine, scanState, scanFileRow, physicalDataIter);
```
allisonport-db, line 119: Spark InternalRow doesn't have schema right? So I think we shouldn't have any issues with this Kernel bug? https://github.com/delta-io/delta/issues/5095

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/scan/batch/KernelSparkPartitionReader.java
allisonport-db: Hmm, I'm a little confused here because I think we cannot close physicalDataIter until we are done with logicalDataIter. I'm not sure if this is maybe an undesirable/unintended dependency?

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/scan/batch/KernelSparkPartitionReader.java
```diff
@@ -0,0 +1,186 @@
...
+
+  @Override
+  public void close() throws IOException {
+    if (dataIterator != null) {
+      dataIterator.close();
+    }
+  }
+
+  /** Adapter that converts FilteredColumnarBatch Iterator to InternalRow Iterator. */
+  private static class RowIteratorAdapter implements CloseableIterator<InternalRow> {
```
allisonport-db, line 135: This is basically just a flatMap over `CloseableIterator<FilteredColumnarBatch>` right? Do you think this is a pattern that we'll see again -- maybe it's worth having this just as a separate util?

I'm having crazy deja-vu but can't seem to find this utility in Kernel at the moment.

allisonport-db: Aha I found it! Kernel has `Utils.intoRows` -- can we use this and just map the `new KernelToSparkRowAdapter(currentBatchRows.next())` op? Seems like it will simplify this code a lot and is safer since we'll be reusing tested code.

Issue #5118, opened 2025-08-20
[Kernel] [UC] When/if applicable, update UCCatalogManagedClient codes to handle UC ratifying v0 commit


scottsand-db:
Today, UC's catalogManaged commit API is not involved in table creation, as table creating uses unique staging tables and so commits are performed via filesystem writes directly to 000.json.

This causes a few interesting `if` statements / special cases in UCCatalogManagedClient code paths, as the `UC::getCommits` API will return `maxRatifiedVersion=-1` for a table at version 0 (that has just been created).

If this is ever updated or changed, let's be sure to address all the relevant code paths.

timothyw553:
This has been partially addressed. 

https://github.com/unitycatalog/unitycatalog/blob/main/server/src/main/java/io/unitycatalog/server/persist/DeltaCommitRepository.java#L179

```diff
    if (commitCount == 0) {
      // Table is validated as a managed Delta table. UC is the source of truth for
      // managed tables, so a newly created table with no commits is at version 0.
-      return new DeltaGetCommitsResponse().latestTableVersion(-1L);
+      return new DeltaGetCommitsResponse().latestTableVersion(0L);
    }
```

Next, we will have kernel `createImpl` call `commitToUC(...)`

Currently `createImpl` looks like:

```java
  private CommitResponse createImpl(...) {
    // writes 000.json
    final FileStatus kernelPublishedDeltaFileStatus =
        writeDeltaFile(engine, finalizedActions, commitMetadata.getPublishedDeltaFilePath(), ...);
    // returns. That's it. No ucClient call.
    return new CommitResponse(
        ParsedPublishedDeltaData.forFileStatus(kernelPublishedDeltaFileStatus));
  }
```

we want to make it look like:

```java
  private CommitResponse createImpl(...) {
    // writes 000.json
    final FileStatus kernelPublishedDeltaFileStatus =
        writeDeltaFile(engine, finalizedActions, commitMetadata.getPublishedDeltaFilePath(), ...);
    
    // NEW: commit to uc
    commitToUC(commitMetadata, ....);

    return new CommitResponse(
        ParsedPublishedDeltaData.forFileStatus(kernelPublishedDeltaFileStatus));
  }
```

scottsand-db:
```java
// NEW: commit to uc
commitToUC(commitMetadata, ....);
```

Err, this isn't really a "commit" in that we are not invoking the "commit" API for UC. We are creating the table.

[...]

PR #5121, opened 2025-08-20
[Kernel] [UC] UCatalogManagedCommitter CREATE support

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/5121/files) to review incremental changes.
- [**stack/kernel_uc_catalog_managed_committer_create**](https://github.com/delta-io/delta/pull/5121) [[Files changed](https://github.com/delta-io/delta/pull/5121/files)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Enable the UCCatalogManagedCommiter to be invoked for v0 at to thus CREATE the delta table. Note that UC still needs to be contacted about 00.json in a post-commit RPC hook. This is outside the purview of Kernel.

## How was this patch tested?

Simple UTs. More complex integration tests will come soon.

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
@@ -345,6 +345,13 @@ public class TransactionImpl implements Transaction {
           return doCommit(
               engine, commitAsVersion, attemptCommitInfo, dataActions, transactionMetrics);
         } catch (CommitFailedException cfe) {
+          // TODO: When catalog-managed CREATE conflicts on 00.json, yielding CFE(retryable=false,
+          //       conflict=true), we can't retry with 001.json since CREATE must be version 0.
+          //       However, we might be conflicting with our own previous attempt that succeeded but
+          //       the client didn't learn about it. Similar to Case 4 below, we need detection to
+          //       distinguish between genuine conflict and conflict with our own previous commit
+          //       attempt. For now, we just fail.
+
           if (!cfe.isRetryable()) {
             // Case 1: Non-retryable exception. We must throw this. We don't expect connectors to
             //         be able to recover from this.
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
@@ -345,13 +345,6 @@ public class TransactionImpl implements Transaction {
           return doCommit(
               engine, commitAsVersion, attemptCommitInfo, dataActions, transactionMetrics);
         } catch (CommitFailedException cfe) {
-          // TODO: When catalog-managed CREATE conflicts on 00.json, yielding CFE(retryable=false,
-          //       conflict=true), we can't retry with 001.json since CREATE must be version 0.
-          //       However, we might be conflicting with our own previous attempt that succeeded but
-          //       the client didn't learn about it. Similar to Case 4 below, we need detection to
-          //       distinguish between genuine conflict and conflict with our own previous commit
-          //       attempt. For now, we just fail.
-
           if (!cfe.isRetryable()) {
             // Case 1: Non-retryable exception. We must throw this. We don't expect connectors to
             //         be able to recover from this.
--- a/unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
+++ b/unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
@@ -35,6 +35,7 @@ import io.delta.storage.commit.Commit;
 import io.delta.storage.commit.uccommitcoordinator.UCClient;
 import io.delta.storage.commit.uccommitcoordinator.UCCommitCoordinatorException;
 import java.io.IOException;
+import java.nio.file.FileAlreadyExistsException;
 import java.util.Optional;
 import org.apache.hadoop.fs.Path;
 import org.slf4j.Logger;
@@ -75,11 +76,33 @@ public class UCCatalogManagedCommitter implements Committer {
 
     final CommitMetadata.CommitType commitType = commitMetadata.getCommitType();
 
-    if (commitType != CommitMetadata.CommitType.CATALOG_WRITE) {
-      // TODO: Support CATALOG_CREATE in the future, too
-      throw new UnsupportedOperationException("Unsupported commit type: " + commitType);
+    if (commitType == CommitMetadata.CommitType.CATALOG_CREATE) {
+      return createImpl(engine, finalizedActions, commitMetadata);
+    }
+    if (commitType == CommitMetadata.CommitType.CATALOG_WRITE) {
+      return writeImpl(engine, finalizedActions, commitMetadata);
     }
 
+    throw new UnsupportedOperationException("Unsupported commit type: " + commitType);
+  }
+
+  /** Handles CATALOG_CREATE by writing the published delta file for version 0. */
+  private CommitResponse createImpl(
+      Engine engine, CloseableIterator<Row> finalizedActions, CommitMetadata commitMetadata)
+      throws CommitFailedException {
+    final FileStatus kernelPublishedDeltaFileStatus =
+        writePublishedDeltaFileForVersion0(engine, finalizedActions, commitMetadata);
+
+    return new CommitResponse(ParsedLogData.forFileStatus(kernelPublishedDeltaFileStatus));
+  }
+
+  /**
+   * Handles CATALOG_WRITE by writing the staged commit file and then committing (e.g. REST or RPC
+   * call) to UC server.
+   */
+  private CommitResponse writeImpl(
+      Engine engine, CloseableIterator<Row> finalizedActions, CommitMetadata commitMetadata)
+      throws CommitFailedException {
     if (commitMetadata.getNewProtocolOpt().isPresent()) {
       // TODO: support this
       throw new UnsupportedOperationException("Protocol change is not yet implemented");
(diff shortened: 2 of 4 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
```diff
@@ -345,6 +345,13 @@ private long commitWithRetry(
           return doCommit(
               engine, commitAsVersion, attemptCommitInfo, dataActions, transactionMetrics);
         } catch (CommitFailedException cfe) {
+          // TODO: When catalog-managed CREATE conflicts on 00.json, yielding CFE(retryable=false,
```
raveeram-db: A naive question - if we throw a CFE(retryable=false, conflict=true) along with information on the path (00.json),  we should leave it up to the Committer/Catalog specific client to make a call on how to handle the conflict (even if it's with 00.json, e.g they retry with a new path)

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -75,11 +76,33 @@ public CommitResponse commit(
...
     }
+    if (commitType == CommitMetadata.CommitType.CATALOG_WRITE) {
+      return writeImpl(engine, finalizedActions, commitMetadata);
+    }
+
+    throw new UnsupportedOperationException("Unsupported commit type: " + commitType);
+  }
 
+  /** Handles CATALOG_CREATE by writing the published delta file for version 0. */
+  private CommitResponse createImpl(
```
scottsand-db, line 96: elaborate here on how CREATE works with UC. e.g. there must be a post-commit hook by the engine to inform UC of this commit.

also add a `// TODO: [delta-io/delta#5118] If UC changes CREATE semantics, update logic here.`

unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
```diff
@@ -428,4 +423,100 @@ class UCCatalogManagedCommitterSuite
...
+          committer.commit(throwingEngine, emptyActionsIterator, commitMetadata)
+        }
+
+        // ===== THEN =====
+        assert(ex.isRetryable == testCase.expectedRetryable)
+        assert(ex.isConflict == testCase.expectedConflict)
+        assert(ex.getMessage.contains(testCase.expectedMessageContains))
+      }
+    }
+  }
```
scottsand-db, line 526: more complex integration tests will come in a subsequent PR

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -120,34 +143,70 @@ private FileStatus writeStagedCommitFile(
       throws CommitFailedException {
     checkArgument(
         commitMetadata.getVersion() > 0, "Can only write staged commit files for versions > 0");
-    final String stagedCommitFilePath = commitMetadata.generateNewStagedCommitFilePath();
+    try {
+      // Do not use Put-If-Absent for staged commit files since we assume that UUID-based
+      // commit files are globally unique, and so we will never have concurrent writers
+      // attempting to write the same commit file.
+      return writeDeltaFile(
```
raveeram-db, line 156: Just for my information, are we breaking down a previous PR on this into smaller ones?

scottsand-db: Nope -- is there something off / confusing here that makes you think that?

Previously there was just `writeStagedCommitFile`.

Now there is:
- `createImpl --> writePublishedDeltaFileForVersion0 --> writeDeltaFile`
- `writeImpl --> writeStagedCommitFile --> writeDeltaFile`

raveeram-db: Ahh got it, i think i just got confused by that. Thanks for clarifying!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
xzhseh: > we should leave it up to the Committer/Catalog specific client to make a call on how to handle the conflict (even if it's with 00.json, e.g they retry with a new path)

I agree - Given that this is CREATE TABLE command, IMO it might not be worth checking whether the conflicting `0.json` is made by our previous _successful-but-not-acknowledged_ (e.g., network issue) commit attempt or not. We can defer the decision to the upper client or simply fail.

^ cc @sumeet-db @LukasRupprecht.

scottsand-db: I agree. I'm realizing that Kernel's Txn cannot just see some CFE and conclude "don't worry, this is just conflicting with a previous commit attemp --> we can assume it was successful". Kernel Txn **cannot** do that. Because just because a file was written does not mean that the "commit", as per the semantics of the Committer, was successful. (And to be clear -- that is what we are doing today. I am just realizing that "we need detection to distinguish between genuine conflict and conflict with our own previous commit attempt" is insufficient)

I'll think more deeply on this. In the short term, Im inclined to ship code quickly that is correct and never yields false positives (cases that we _think_ succeeded but didn't actually). i.e. I'll err on the side of throwing an exceptoin, and we can iterate from there and make this more robust.

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
xzhseh: > there must be a post-commit hook by the engine to inform UC of this commit.

This seems like engine-specific bit. Have we mentioned this in CCv2 RFC?

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -120,34 +143,70 @@ private FileStatus writeStagedCommitFile(
...
+        commitMetadata.getVersion());
+    try {
+      return writeDeltaFile(
+          engine,
+          finalizedActions,
+          commitMetadata.getPublishedDeltaFilePath(),
+          false /* overwrite */);
+    } catch (FileAlreadyExistsException ex) {
+      // We are either conflicting with (a) another writer concurrently writing to the same staging
+      // table location (which should be impossible) or (b) a previous commit attempt from this
```
xzhseh: If two writers start to CREATE the same table by name at almost the same time - the actual table location would be different but one of the writers would get conflict indicating that the table already exists, depending on the concrete timing. Hence, `FileAlreadyExistsException` should exactly map to case (b) here?

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -120,34 +143,70 @@ private FileStatus writeStagedCommitFile(
...
-          } catch (IOException ex) {
-            throw new CommitFailedException(
-                true /* retryable */,
-                false /* conflict */,
-                "Failed to write staged commit file due to: " + ex.getMessage(),
-                ex);
-          }
+          // Note: the engine is responsible for closing the actions iterator once it has been
+          //       fully consumed.
+          engine.getJsonHandler().writeJsonFileAtomically(filePath, finalizedActions, overwrite);
```
xzhseh, line 214: Is there a way to enforce engine to do so at interface level? Or the best we can do is to rely on a verbal contract with the engine APIs?

unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
```diff
@@ -428,4 +423,100 @@ class UCCatalogManagedCommitterSuite
...
+      val committer = new UCCatalogManagedCommitter(ucClient, "ucTableId", tablePath)
+
+      val commitMetadata = createCommitMetadata(
+        version = 0,
+        logPath = logPath,
+        newProtocolOpt = Optional.of(protocolWithCatalogManagedSupport),
+        newMetadataOpt = Optional.of(basicPartitionedMetadata))
+
+      // ===== WHEN =====
+      val response = committer.commit(defaultEngine, actionsIterator, commitMetadata)
```
xzhseh, line 446: Please add a comment here explaining 0th commit will go through FS rather than UC for now.

xzhseh (approved):
LGTM. Mostly nits + clarification questions.

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
scottsand-db: > This seems like engine-specific bit. Have we mentioned this in CCv2 RFC?

Yup! We leave it up to the engine and the catalog to work together on this. 

https://github.com/delta-io/delta/blob/master/protocol_rfcs/catalog-managed.md#creating-and-dropping-catalog-managed-tables

IMO it's fine to have a comment like this? Just to give context to a reader? This is part of the "catalog" clide-side implementation, after all.

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
scottsand-db: > Hence, FileAlreadyExistsException should exactly map to case (b) here?

 I agree! I state that case (a) should be impossible. I can try and update this comment to make this language clearer and firmer.

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
scottsand-db: There's no way in Java -- you can define a method signature and what should be received, but you cannot then statically specify what the invoked method must then do with those parameters

unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
scottsand-db: yup! will do 👍

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TransactionImpl.java
raveeram-db: SGTM!

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
scottsand-db: @xzhseh -- actually, in this case: if our previous commit attempt did succeed, and that is what we are now conflicting with in **this** commit attempt, can't we just return success ??

scottsand-db: perhaps we should add the validation that the `00.json` files have the same content? but ... concurrent writes here should be **impossible**, right?

scottsand-db:
```java
// This can happen if a previous commit attempt from this writer succeeded in writing 00.json,
// but the client was not successfully notified of that write (e.g. due to network
// issues).
//
// Since we know we are conflicting with our previous write, we can safely return a successful
// response.
return FileStatus.of(commitMetadata.getPublishedDeltaFilePath());
```

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -75,11 +76,33 @@ public CommitResponse commit(
...
+  }
 
+  /** Handles CATALOG_CREATE by writing the published delta file for version 0. */
+  private CommitResponse createImpl(
+      Engine engine, CloseableIterator<Row> finalizedActions, CommitMetadata commitMetadata)
+      throws CommitFailedException {
+    final FileStatus kernelPublishedDeltaFileStatus =
+        writePublishedDeltaFileForVersion0(engine, finalizedActions, commitMetadata);
+
+    return new CommitResponse(ParsedLogData.forFileStatus(kernelPublishedDeltaFileStatus));
```
allisonport-db, line 102: > Note that UC still needs to be contacted about 00.json in a post-commit RPC hook. This is outside the purview of Kernel.

Where does this happen / who is responsible for doing this? This is not immediately clear to me

xzhseh: The post commit hook of each transaction is responsible to update UC (or other catalogs) asynchronously. This logic exists in Delta-Spark as of today: https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala#L413.
IMO this is more engine-specific - they choose when to notify the catalog regarding the table creation, etc.

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
xzhseh: > @xzhseh -- actually, in this case: if our previous commit attempt did succeed, and that is what we are now conflicting with in this commit attempt, can't we just return success ??

If the catalog is UC - then yes, we could return success as the possibility of race condition during CREATE is almost certain 0%.

But, for any arbitrary catalog (and engine) there is no guarantee of their table creation logic. The assumption we are making here only applies to UC and should not be _generalized_ unless we enforce this in the CCv2 spec.

The current implementation LGTM since we are returning the successful response only in `UCCatalogManagedCommitter`.

cc @scovich @LukasRupprecht @sumeet-db for visibility.

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
xzhseh: Then let's at least add the comment regarding `finalizedActions` to this function.

scottsand-db: Agreed!

PR #5123, opened 2025-08-21
[Spark] Improve backfill termination conditions

andreaschat-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Currently the backfill executor could hang in the presence of (infinite) competing third-party transactions that keep producing data eligible for backfill. This PR improves this behaviour as follows:

The executor will now terminate if the last batch is less than the maximum batch size. This is going to protect again fast small competing txns.
The executor will now terminate after processed a total of `totalNumberOfFilesInTable` * `factor`. This is going to protect against large competing txns. In practice, it should be quite hard for competing txns to be fast enough in order to pass (1). This is because backfill batches are fast due to only committing log files while business transactions write data files. Live-locking would be more likely to occur from a competing txn than only writes log files as well.

## How was this patch tested?

N/A.

No.


tomvanbussel (approved):
LGTM

PR #5130, opened 2025-08-25
[kernel] icebergWriterCompatV3 should support all features in icebergWriterCompatV1

KaiqiJinWow:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

Update the code to support multiple Iceberg compat features. Note that only one Iceberg compat version can be enabled at a time.

## How was this patch tested?

Unit tests

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/icebergcompat/IcebergWriterCompatMetadataValidatorAndUpdater.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/icebergcompat/IcebergWriterCompatMetadataValidatorAndUpdater.java
@@ -72,13 +72,13 @@ abstract class IcebergWriterCompatMetadataValidatorAndUpdater
   }
 
   /**
-   * Allowed table features for IcebergWriterCompatV1. This includes the incompatible legacy
-   * features (invariants, changeDataFeed, checkConstraints, identityColumns, generatedColumns)
-   * because they may be present in the table protocol even when they are not in use. In later
-   * checks we validate that these incompatible features are inactive in the table. See the protocol
-   * spec for more details.
+   * Common set of allowed table features shared across all Iceberg writer compatibility versions.
+   * This includes the incompatible legacy features (invariants, changeDataFeed, checkConstraints,
+   * identityColumns, generatedColumns) because they may be present in the table protocol even when
+   * they are not in use. In later checks we validate that these incompatible features are inactive
+   * in the table. See the protocol spec for more details.
    */
-  protected static final Set<TableFeature> V1_ALLOWED_FEATURES =
+  protected static final Set<TableFeature> COMMON_ALLOWED_FEATURES =
       Stream.of(
               // Incompatible, but not active, legacy table features
               INVARIANTS_W_FEATURE,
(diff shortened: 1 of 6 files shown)


lzlfred (approved):

murali-db (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/icebergcompat/IcebergWriterCompatMetadataValidatorAndUpdater.java
```diff
@@ -96,7 +96,9 @@ protected static IcebergCompatRequiredTablePropertyEnforcer<Boolean> createIcebe
               CLUSTERING_W_FEATURE,
               TIMESTAMP_NTZ_RW_FEATURE,
               TYPE_WIDENING_RW_FEATURE,
-              TYPE_WIDENING_RW_PREVIEW_FEATURE)
+              TYPE_WIDENING_RW_PREVIEW_FEATURE,
+              ICEBERG_COMPAT_V2_W_FEATURE,
```
vkorukanti: please add a comment on why this is a common featurue. Also how do we handle (common features) when icebergCompatWriter4 is introduced?

vkorukanti (changes requested):
Can we just remove the icebergCompatV2 and icebergCompatWriterV1 when icebergcompatWriterV3 is enabled? Does it simplify the check.

Also missing an integration tests

Should the same logic applies to icebergCompat3? cc @murali-db

kernel/kernel-api/src/main/java/io/delta/kernel/internal/icebergcompat/IcebergWriterCompatMetadataValidatorAndUpdater.java
KaiqiJinWow: comment updated, move all allowed features in this base class. If later icebergWriterCompatV4 is introduced, it should extends from V3_ALLOWED_FEATURES

KaiqiJinWow:
> Can we just remove the icebergCompatV2 and icebergCompatWriterV1 when icebergcompatWriterV3 is enabled? Does it simplify the check.

Dropping the table feature requires more effort (update protocol, rewrite  data file if needed) compared to just allow it in icebergWriterCompatV3. 

> Should the same logic applies to icebergCompat3

IcebergCompatV3 only check metadata and here in IcebergWriterCompat we check allowed table features

kernel/kernel-api/src/main/java/io/delta/kernel/internal/icebergcompat/IcebergWriterCompatMetadataValidatorAndUpdater.java
```diff
@@ -72,13 +72,13 @@ protected static IcebergCompatRequiredTablePropertyEnforcer<Boolean> createIcebe
...
-   * they are not in use. In later checks we validate that these incompatible features are inactive
-   * in the table. See the protocol spec for more details.
+   * Allowed table features for IcebergWriterCompatV1. This includes the incompatible legacy
+   * features (invariants, changeDataFeed, checkConstraints, identityColumns, generatedColumns)
+   * because they may be present in the table protocol even when they are not in use. In later
+   * checks we validate that these incompatible features are inactive in the table. See the protocol
+   * spec for more details.
    */
-  protected static final Set<TableFeature> COMMON_ALLOWED_FEATURES =
+  protected static final Set<TableFeature> V1_ALLOWED_FEATURES =
```
vkorukanti: Should this be moved to `IcebergWriterCompatV1MetadataValidatorAndUpdater` as it no longer the common allowed features.

vkorukanti: IMO just keep this list as is and add the `icebergWriterCompatV1` and `icebergCompatV2` to  `V3_ALLOWED_FEATURES`.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/icebergcompat/IcebergWriterCompatV3MetadataValidatorAndUpdaterSuite.scala
```diff
@@ -382,7 +382,9 @@ class IcebergWriterCompatV3MetadataValidatorAndUpdaterSuite
       "rowTracking",
       "variantType",
       "variantType-preview",
-      "variantShredding-preview")
+      "variantShredding-preview",
+      "icebergCompatV2",
+      "icebergWriterCompatV1")
```
vkorukanti, line 387: where is the integration test that tests this scenario? I can think of at least one test.

1) table has icebergCompatWriterV1 enabled
2) Use Spark to disable the compat writer v1 (may be use the Kernel here)
3) Use kernel to enable the icebergCompatWriterv3. Ensure the table is in a sane state and can still read it.

vkorukanti (changes requested):
cc @lzlfred I think this needs protocol update. So far we don't have the `icebergCompatWriterV3` yet in protocol. Also we are adding additional requirement, which should be documented in the protocol. I will leave it up to you on how to handle the protocol updates.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/icebergcompat/IcebergWriterCompatV3MetadataValidatorAndUpdaterSuite.scala
KaiqiJinWow: We don't support disable writerCompatV1 so we cannot add this integration test. The only goal for this PR is to support both v2 and writerCompatV1 table feature is compatible with writerCompatV3, which is verified in this test case

vkorukanti (approved):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/icebergcompat/IcebergWriterCompatV3MetadataValidatorAndUpdater.java
```diff
@@ -107,8 +107,11 @@ public static Optional<Metadata> validateAndUpdateIcebergWriterCompatV3Metadata(
                   VARIANT_RW_FEATURE,
                   VARIANT_SHREDDING_PREVIEW_RW_FEATURE,
                   VARIANT_RW_PREVIEW_FEATURE,
-                  ROW_TRACKING_W_FEATURE))
-          .collect(toSet());
+                  ROW_TRACKING_W_FEATURE,
+                  // also allow writerV1 features for backward compatibility
```
allisonport-db: Can you add a comment that we already check that these must not be enabled due to XX check in YY?

KaiqiJinWow: added

PR #5136, opened 2025-08-27
[Kernel][CommitRange] Initial interfaces for CommitRange and CommitRangeBuilder

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Part of https://github.com/delta-io/delta/issues/5138

Adds the initial interfaces for CommitRange and its builder, which will be used to query actions over a commit range for CCV2 tables.

## How was this patch tested?

Just interfaces.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/CommitRangeBuilder.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/CommitRangeBuilder.java
@@ -136,23 +136,19 @@ public interface CommitRangeBuilder {
       this.latestSnapshot = latestSnapshot;
     }
 
-    /**
-     * @return {@code true} if this is a version-based boundary, {@code false} otherwise
-     */
+    /** @return {@code true} if this is a version-based boundary, {@code false} otherwise */
     public boolean isVersion() {
       return isVersion;
     }
 
-    /**
-     * @return {@code true} if this is a timestamp-based boundary, {@code false} otherwise
-     */
+    /** @return {@code true} if this is a timestamp-based boundary, {@code false} otherwise */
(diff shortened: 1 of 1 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/CommitRangeBuilder.java
```diff
@@ -0,0 +1,206 @@
...
+      checkArgument(isVersion || latestSnapshot.isPresent());
+      this.isVersion = isVersion;
+      this.value = value;
+      this.latestSnapshot = latestSnapshot;
+    }
+
+    /**
+     * Returns {@code true} if this boundary is based on a version number.
+     *
+     * @return {@code true} if this is a version-based boundary, {@code false} otherwise
```
scottsand-db: nit: my personal preference is reducing the rebosity of simple getters like this. e.g. just the `@return` is fine

kernel/kernel-api/src/main/java/io/delta/kernel/CommitRangeBuilder.java
```diff
@@ -0,0 +1,206 @@
...
+      return !isVersion;
+    }
+
+    /**
+     * Returns the version number for version-based boundaries.
+     *
+     * @return the version number
+     * @throws IllegalStateException if this boundary is timestamp-based
+     */
+    public long getVersion() {
```
scottsand-db, line 156: nit: for these `get` methods, since they do not return optional, probably worth mentioning that callers should first check `isX` method returns true before accessing this method

PR #5137, opened 2025-08-27
[Kernel][CommitRange] Pre-factor to current getChanges internal code, namely TableImpl and DeltaLogActionUtils

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Refactors internal code used by the current getChanges API, in preparation for the CommitRange implementation which will use this code as well (at least initially).

## How was this patch tested?

Updated tests for the 1 small logic change, other than that, just refactor so existing tests should suffice.

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
@@ -18,33 +18,26 @@ package io.delta.kernel.internal;
 import static io.delta.kernel.internal.DeltaErrors.wrapEngineExceptionThrowsIO;
 
 import io.delta.kernel.*;
-import io.delta.kernel.data.ColumnVector;
 import io.delta.kernel.data.ColumnarBatch;
 import io.delta.kernel.engine.Engine;
 import io.delta.kernel.exceptions.CheckpointAlreadyExistsException;
 import io.delta.kernel.exceptions.KernelException;
 import io.delta.kernel.exceptions.TableNotFoundException;
-import io.delta.kernel.internal.actions.Protocol;
 import io.delta.kernel.internal.checkpoints.Checkpointer;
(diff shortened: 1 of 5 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
```diff
@@ -191,41 +184,15 @@ public CloseableIterator<ColumnarBatch> getChanges(
...
-                batchToReturn = batchToReturn.withDeletedColumnAt(protocolIdx);
-              }
-              int commitInfoIdx = batchToReturn.getSchema().indexOf("commitInfo");
-              if (shouldDropCommitInfoColumn) {
-                batchToReturn = batchToReturn.withDeletedColumnAt(commitInfoIdx);
-              }
-              return batchToReturn;
-            });
+    logger.info(
+        "{}: Getting the commit files for versions [{}, {}]", tablePath, startVersion, endVersion);
```
scottsand-db: Can't this `logger.info` go into `DeltaLogActionUtils.getCommitFilesForVersionRange`

scottsand-db: i.e. I'm a fan of pushing logs like this into the underlying called methods.

Instead of 

```
// call site A
doLog
callMethod

// call site B
doLog
callMethod
```

we can just do

```
// call site A
callMethod

// call site B
callMethod

and then the underlying method does the log
```

scottsand-db (approved):
LGTM!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableImpl.java
allisonport-db: I agree -- just lost in the refactor. Updated.

PR #5139, opened 2025-08-27
[Kernel] Specify benchmarking workloads for automated execution

OussamaSaoudi:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description
This PR introduces the core infrastructure for running benchmarking workloads using Java Microbenchmark Harness. The benchmarking infrastructure currently supports metadata read workloads on local tables. 

The main classes are as follows:
* **WorkloadSpec**: Abstract class that represents all workload specifications.
* **ReadSpec**: A WorkloadSpec that defines a metadata read benchmark. Contains fields for snapshot version and table root.
* **WorkloadRunner**: Interface for all the Workload runners 
* **ReadMetadataRunner**: WorkloadRunner for the ReadMetadata workload. Given a workload spec, this runner performs all the setup in its constructor and can be executed as a benchmark.
* **WorkloadBenchmark**: Executes WorkloadRunners as benchmarks, measuring their time using the Java Microbenchmarking Harness
* **BenchmarkingUtils**: Utilities to read workload specs.
* **AbstractBenchmarkState**: JMH benchmark state that sets up workloads and invokes them. This is extended to provide different Engine implementations.
* **KernelMetricsProfiler**: Provides a BenchmarkingEngine implementation that collects metrics reports to a JMH profiler.
* **WorkloadOutputFormat**: Generates a json-formatted report of the benchmark results

## How was this patch tested?

This was run locally using
`build/sbt "kernelDefaults/test:runMain io.delta.kernel.defaults.benchmarks.WorkloadBenchmark"`

This correctly produces a `benchmark_report.json`.

No


kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/ReadMetadata.java
```diff
@@ -0,0 +1,65 @@
...
+package io.delta.kernel.defaults.benchmarks;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.nio.file.Path;
+
+/** Workload specification for read_metadata benchmarks. */
+public class ReadMetadata extends WorkloadSpec {
+
+  /**
+   * @return The path to the root of the table. If it is a URI, this must be treated as an absolute
```
scottsand-db: Can you help me understand: Why not always use absolute paths? Why make consumers have to think and handle different cases? I'm sure there's a good reason for doing this, I'd just like to get brought up to speed :)

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/BenchmarkUtils.java
```diff
@@ -0,0 +1,49 @@
...
+  public static final Path WORKLOAD_SPECS_DIR = RESOURCES_DIR.resolve("workload_specs");
+
+  /**
+   * Scans the workloads directory and loads all JSON workload specifications.
+   *
+   * @return List of loaded workload specifications
+   * @throws RuntimeException if the workloads directory doesn't exist
+   */
+  public static List<String> loadAllWorkloads(Path specDirPath) {
+
```
scottsand-db, line 82: nit: superfluous blank line

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/ReadMetadata.java
```diff
@@ -0,0 +1,65 @@
...
+ * limitations under the License.
+ */
+
+package io.delta.kernel.defaults.benchmarks;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.nio.file.Path;
+
+/** Workload specification for read_metadata benchmarks. */
+public class ReadMetadata extends WorkloadSpec {
```
scottsand-db: `ReadMetadataSpec` -- this is, afterall, a specific type of WorkloadSpec

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/models/WorkloadSpec.java
```diff
@@ -0,0 +1,43 @@
...
+package io.delta.kernel.defaults.benchmarks;
+
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import java.nio.file.Path;
+
+/** Base class for all workload specifications. */
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type")
+@JsonSubTypes({@JsonSubTypes.Type(value = ReadMetadata.class, name = "read_metadata")})
+public abstract class WorkloadSpec {
```
scottsand-db, line 37: any reason you went for abstract class rather than interface?

edit: ah, probably cuz of the jackson stuff with the type

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/ReadMetadata.java
```diff
@@ -0,0 +1,65 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.delta.kernel.defaults.benchmarks;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.nio.file.Path;
+
+/** Workload specification for read_metadata benchmarks. */
```
scottsand-db: I think you should also explain to the reader what exactly a "read metadata" spec is -- what does it test for? what is an example of it?

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/ReadMetadataRunner.java
```diff
@@ -0,0 +1,122 @@
...
+import io.delta.kernel.data.FilteredColumnarBatch;
+import io.delta.kernel.defaults.engine.DefaultEngine;
+import io.delta.kernel.engine.Engine;
+import io.delta.kernel.utils.CloseableIterator;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import org.apache.hadoop.conf.Configuration;
+import org.openjdk.jmh.infra.Blackhole;
+
+public class ReadMetadataRunner implements WorkloadRunner {
```
scottsand-db: nit: classdocs. How does this runner interelate to ReadMetadataSpec and WorkloadRunner?

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/ReadMetadataRunner.java
```diff
@@ -0,0 +1,122 @@
...
+
+  /**
+   * Resolves the table root path based on whether it's a URI or a relative path. If it is a
+   * relative path, it is resolved based on the workload base directory.
+   *
+   * @param baseWorkloadDirPath The base directory containing workload tables.
+   * @param workloadTableRoot The table root path from the workload spec.
+   * @return The resolved table root path.
+   */
+  private static String getResolvedTableRoot(Path baseWorkloadDirPath, String workloadTableRoot) {
```
scottsand-db: why private static? can we put these private helper methods at the bottom of the class? below any public methods?

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/ReadMetadataRunner.java
```diff
@@ -0,0 +1,122 @@
...
+   * Constructs the Scan from the workload specification for this runner.
+   *
+   * @param engine The kernel engine to use for table operations.
+   * @param workload The read_metadata workload specification.
+   * @param resolvedTableRoot The resolved table root path.
+   * @return The Scan object used to execute the workload.
+   */
+  private static Scan getScanFromWorkloadSpec(
+      Engine engine, ReadMetadata workload, String resolvedTableRoot) {
+    Table table = Table.forPath(engine, resolvedTableRoot);
```
scottsand-db: Can we please use the new `TableManager` APIs?

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/ReadMetadataRunner.java
```diff
@@ -0,0 +1,122 @@
...
+    // Get snapshot (use specified version or latest)
+    Snapshot snapshot;
+    if (workload.getSnapshotVersion() != null) {
+      snapshot = table.getSnapshotAsOfVersion(engine, workload.getSnapshotVersion());
+    } else {
+      snapshot = table.getLatestSnapshot(engine);
+    }
+    // Build scan
```
scottsand-db: nit: superfluous comment --> this doesn't give us any more context than the line of code below `ScanBuilder scanBuilder = snapshot.getScanBuilder();` :) we can delete this

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/ReadMetadataRunner.java
```diff
@@ -0,0 +1,122 @@
...
+    ScanBuilder scanBuilder = snapshot.getScanBuilder();
+    // in the read_metadata workload spec
```
scottsand-db: nit: is this an incomplete sentence? sorry -- not sure the takeaway here

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/WorkloadRunner.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.delta.kernel.defaults.benchmarks;
+
+import org.openjdk.jmh.infra.Blackhole;
+
+public interface WorkloadRunner {
```
scottsand-db: nit: classdocs and method docs on a public interface like this

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/ReadMetadataRunner.java
```diff
@@ -0,0 +1,122 @@
...
+    return scanBuilder.build();
+  }
+
+  /**
+   * Executes the read_metadata workload, returning an iterator over the results. This must be fully
+   * consumed by the caller to ensure the workload is fully executed.
+   *
+   * @return Iterator of results from the read_metadata workload.
+   */
+  public CloseableIterator<FilteredColumnarBatch> execute() {
```
scottsand-db: So this is public but it is not part of the WorkloadRunner benchmark, right? so would there be some callers that are sub-casing on the type of workload runner? if if that runner is a ReadMetadata then they can call this to get back the rows?

should there be a top-level `ReadWorkloadRunner` that reads _something_ and has this method as a public API?

Just curious about the usage here

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/ReadMetadataRunner.java
```diff
@@ -0,0 +1,122 @@
...
+   * Constructs the ReadMetadataRunner from the workload spec and the base workload directory
+   * containing test tables. This initializes any state necessary to execute this workload as a
+   * benchmark.
+   *
+   * @param baseWorkloadDirPath The base directory containing workload tables.
+   * @param workload The read_metadata workload specification.
+   */
+  public ReadMetadataRunner(Path baseWorkloadDirPath, ReadMetadata workload) {
+    String resolvedTableRoot = getResolvedTableRoot(baseWorkloadDirPath, workload.getTableRoot());
+    engine = DefaultEngine.create(new Configuration());
```
scottsand-db: High-level question: Why is this benchmark stuff in kernel-defaults? Do we want it to be hard-coded to using the KernelEngine? I'd imagine we want it to be extensible to be run with any engine, right?

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/WorkloadBenchmark.java
```diff
@@ -0,0 +1,120 @@
...
+    // Convert paths into a String array for JMH. JMH requires that parameters be of type String[].
+    String[] paths = workloadSpecPaths.toArray(new String[0]);
+
+    // Create and run the benchmark
+    Options opt =
+        new OptionsBuilder()
+            .include(WorkloadBenchmark.class.getSimpleName())
+            .param("workloadPath", paths)
+            .forks(1)
+            .warmupIterations(3)
```
scottsand-db: should all of these configurations eventually be configurable? // doesn't need to be done in this PR, just curious. if so, can we add this as a TODO like so: `TODO: [delta-io/delta#12345] Allow users to configure XXX` and can we also track this future work in a tracking github issue?

kernel/kernel-defaults/src/test/resources/workload_specs/basic_with_checkpoint_latest.json
```diff
@@ -0,0 +1,4 @@
+{
+  "type": "read_metadata",
```
scottsand-db: can we also include a name? or do you want to just infer the name from the table path?

I do think a name string could be longer and more descriptive?

scottsand-db (changes requested):
Looks great! Lots of tiny questions -- thanks for helping to take the time and teach me / bring me up to speed on this!

kernel/kernel-defaults/src/test/resources/workload_specs/basic_with_checkpoint_v10.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "read_metadata",
```
scottsand-db: I also think we should err on the side of being **very** descriptive with our naming. I wonder if we should be including the protocol and the schema data types? and the partioning and clustering?

e.g. what about this workload spec is "basic"? // and I know you didn't create the `basic_with_checkpoint` table against which this is testing :) not your original problem

This is just a starter point: `Table schema=(ID LONG), protocol=(1,2), dataLayoutSpec=None, checkpoints=True`

But I think things like this help us see and identify our independent variables per each table?

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/ReadMetadata.java
scottsand-db: I also wonder -- would we ever want a ReadMetadataSpec (for some table, at some version) to be extended to also be a ReadDataSpec?

i.e. would that class extend this class? like ... is a workload spec in JSON really just "read metadata" or is it really just "read" and then different runners can test reading metadata or reading data, too ?

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/WorkloadBenchmark.java
```diff
@@ -0,0 +1,120 @@
...
+    public WorkloadRunner getRunner() {
+      return runner;
+    }
+  }
+
+  /**
+   * Benchmark method that executes the workload specified in the state.
+   *
+   * @param state The benchmark state containing the workload specification and runner.
+   * @param blackhole The Blackhole to consume results and prevent dead code elimination.
```
scottsand-db: Can you help me to understand: Where does this Blackhole come from? Basically, I never saw `val x = new Blackhole` in this code -- so I think something is being injected magically? would love to understand this more

kernel/kernel-defaults/src/test/resources/workload_specs/basic_with_checkpoint_v10.json
OussamaSaoudi: I think a description field like this would be very useful. Unsure how structured we want it to be. This can be completely unstructured through manual input to fully generated.


I'm leaning towards starting off with manual and evolving to generated in the future. This can be under a "description" json field.

kernel/kernel-defaults/src/test/resources/workload_specs/basic_with_checkpoint_latest.json
OussamaSaoudi: I think a name is a good addition 👍  Description field (mentioned below) would also be nice.

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/ReadMetadata.java
OussamaSaoudi: the problem I found with absolute paths is that the location of the data can be different depending on where you clone your repo.

Example: suppose the data is under `~/delta/resources/my_benchmark_table`.

If someone else clones the repository to a different path: `~/projects/delta/`, then the absolute path would no longer be correct, because it's `~/projects/delta/resources/my_benchmark_table`, not `~/delta/resources/my_benchmark_table`

scottsand-db: Makes sense! And then can you give an example of when/why it _would_ be absolute? Would that just be paths in like S3?

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/ReadMetadataRunner.java
OussamaSaoudi: hmm yeah perhaps this can be kept private for now.

The intended usage I had were the following:
```
WorkloadRunner runner = spec.getRunner();
runner.executeAsTest() 
runner.executeAsBenchmare(Blackhole) 
```

In both cases execute is used behind the scenes, but we don't need to expose it 👍

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/WorkloadBenchmark.java
OussamaSaoudi: This is part of the Java Microbenchmark Harness. To my understanding, the `@Benchmark` annotation tells JMH that this is a benchmark, and that it should take a black hole (that JMH provides).

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/WorkloadBenchmark.java
OussamaSaoudi: Good point. In fact there's a lot of mental TODO's that I haven't formally put into Issues/comments. I'll add that.

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/models/WorkloadSpec.java
OussamaSaoudi: > edit: ah, probably cuz of the jackson stuff with the type

Correct :) 

Should I promote the comments below to be class docs?
```
// The type of workload (e.g., "read_metadata") This is used by Jackson's polymorphic
  // deserialization to automatically instantiate the correct subclass based on the "type" field in
  // JSON.
```

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/ReadMetadata.java
OussamaSaoudi: Hmm that's an interesting point. I'd  initially intended for readMetadata and readData to be separate workload spec and separate runners. 

This property actually becomes very handy with how the polymorphism works. It all becomes a simple flow like this:
```
WorkloadSpec spec = objectMapper.read(json_file)
WorkloadRunner runner = spec.getRunner(...)
runner.executeAsBenchmark()
```

We could introduce a flag in the WorkloadSpec itself. That way, getRunner(..) constructs the right one.


I think introducing this shared notion at the spec is interesting. Given that we're not implementing the data phase just yet, I'd advocate to punt on this design for now. Renaming is easy since this is private after all :)

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/WorkloadBenchmark.java
scottsand-db: Thanks! And yeah it need not necessarily even be an Issue -- Just so long as you have one tracking issue with tables that track the current/future stuff, similar to https://github.com/delta-io/delta/issues/4573 would be great

kernel/kernel-defaults/src/test/resources/workload_specs/basic_with_checkpoint_latest.json
scottsand-db: Clarifying question: What's the differencce between a name and a description here?

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/ReadMetadata.java
OussamaSaoudi: yep, any object store path like `s3://my-bucket/...`, or one can create workload specs for their own local test tables using `file:///...`

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/ReadMetadataRunner.java
OussamaSaoudi: > Do we want it to be hard-coded to using the KernelEngine? 

I've changed the metadata workload runner to take the engine. It's currently hardcoded in BenchmarkState, but that's where we can add conditional logic in the future to handle these cases.

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/WorkloadBenchmark.java
OussamaSaoudi: Still needs to be filled out fully, but here it is: https://github.com/delta-io/delta/pull/5139

kernel/kernel-defaults/src/test/resources/workload_specs/basic_with_checkpoint_latest.json
OussamaSaoudi: I'd consider a description to be something like what you proposed below: `Table schema=(ID LONG), protocol=(1,2), dataLayoutSpec=None, checkpoints=True`. Describes the workload in greater detail.

A name can give a high level description of the workload just like how a test name describes the high level of the test.

kernel/kernel-defaults/src/test/resources/workload_specs/basic_with_checkpoint_v10.json
OussamaSaoudi: Update: instead, I'm planning on omitting the description `Table schema=(ID LONG), protocol=(1,2), dataLayoutSpec=None, checkpoints=True` for now until there's an easy way to generate nice, well structured descriptions.

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/ReadMetadata.java
OussamaSaoudi: Updated the docs :) PTAL

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/ReadMetadataRunner.java
OussamaSaoudi: updated. PTAL

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/WorkloadBenchmark.java
```diff
@@ -0,0 +1,142 @@
...
+    private WorkloadRunner runner;
+
+    /**
+     * Setup method that runs before each benchmark invocation. Reads the workload specification
+     * from the given path and initializes the corresponding workload runner.
+     *
+     * @throws Exception If any error occurs during setup.
+     */
+    @Setup(Level.Invocation)
+    public void setup() throws Exception {
```
scottsand-db: Is it possible for WorkloadBenchmark to be abstract? and it has a single abstract method `createEngine`?

Or perhaps this benchmarking framework has a better solution to dependency injection?

So someone can easily have:

```java
class WorkloadBenchmarkWithDefaultEngine extends AbstractWorkloadBenchmark {
  @Override
  public Engine createEngine() {
    return DefaultEngine....
  }
}

class WorkloadBenchmarkWithDeltaSparkDSv2Engine extends AbstractWorkloadBenchmark {
  @Override
  public Engine createEngine() {
    return CustomDeltaSparkDSv2Engine..
  }
}
```

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/WorkloadBenchmark.java
```diff
@@ -0,0 +1,115 @@
...
+  public static class DefaultBenchmarkState extends AbstractBenchmarkState {
+    @Override
+    protected Engine getEngine(String engineName) {
+      if (engineName.equals("default")) {
+        return DefaultEngine.create(new Configuration());
+      } else {
+        throw new IllegalArgumentException("Unsupported engine: " + engineName);
+      }
+    }
+  }
```
OussamaSaoudi, line 58: @scottsand-db Someone could extend this benchmark simply by modifying the `getEngine` to also support their own engine.

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/models/WorkloadSpec.java
```diff
@@ -0,0 +1,145 @@
...
+    spec.setTableInfo(tableInfo);
+    spec.setCaseName(caseName);
+    return spec;
+  }
+
+  /** @return Returns a list of benchmark variants for this workload specification. */
+  @JsonIgnore
+  public List<WorkloadSpec> getBenchmarkVariants() {
+    return Collections.singletonList(this);
+  }
```
OussamaSaoudi: @scottsand-db  this is how get all variants for a benchmark. By default a workloadSpec is its own only variant. But the `read` spec overrides this and generates a couple other variants.

One alternative is to have a notion of`UnspecializedWorkloadSpec` and `SpecializedWorkloadSpec`, would love your thoughts.

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/KernelMetricsProfiler.java
```diff
@@ -0,0 +1,137 @@
...
+    return out.collect(Collectors.toList());
+  }
+
+  /**
+   * An {@link Engine} implementation that wraps an existing engine and delegates all engine tasks.
+   * The BenchmarkingEngine sends all metrics reports to the KernelMetricsProfiler for collection
+   * during benchmarks. The metrics reports can then be extracted and reported by the
+   * KernelMetricsProfiler.
+   */
+  public static class BenchmarkingEngine implements Engine {
```
OussamaSaoudi, line 152: @scottsand-db  this is my solution to collecting Scan metrics using any provided engine. The idea is to intercept the MetricsReports and send them all to the JMH metrics collector `KernelMetricsProfiler`

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/BenchmarkUtils.java
```diff
@@ -0,0 +1,164 @@
...
+import io.delta.kernel.defaults.benchmarks.models.WorkloadSpec;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/** Useful utilities and values for benchmarks. */
```
OussamaSaoudi, line 28: This file contains all the logic for handling the DAT-style directory structure and reading specs.

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/BenchmarkUtils.java
```diff
@@ -0,0 +1,164 @@
...
+public class BenchmarkUtils {
+
+  public static final Path RESOURCES_DIR =
+      Paths.get(System.getProperty("user.dir") + "/src/test/resources");
```
scottsand-db: can we throw a nice exception if `user.dir` is not set?

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/BenchmarkUtils.java
```diff
@@ -0,0 +1,164 @@
...
+  private static List<WorkloadSpec> loadSpecsFromTable(Path tableDir) {
+    validateTableStructure(tableDir);
+
+    Path tableInfoPath = tableDir.resolve(TABLE_INFO_FILE_NAME);
+    Path deltaDir = tableDir.resolve(DELTA_DIR_NAME);
+    Path specsDir = tableDir.resolve(SPECS_DIR_NAME);
+
+    TableInfo tableInfo =
+        TableInfo.fromJsonPath(tableInfoPath.toString(), deltaDir.toAbsolutePath().toString());
+    System.out.println("Loading specs for table: " + tableInfo);
```
scottsand-db: nit: We should use a more formal logger framework than Sys.out.println

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/KernelMetricsProfiler.java
scottsand-db: LGTM

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/KernelMetricsProfiler.java
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  private static ScalarResult generateAvgScalarCount(String name, double value) {
+    return new ScalarResult(name, value, "count", AggregationPolicy.AVG);
+  }
+
+  private static SampleTimeResult generateTimeSample(String name, long value, TimeUnit unit) {
+    SampleBuffer buf = new SampleBuffer();
+    buf.add(value);
+    return new SampleTimeResult(ResultRole.SECONDARY, name, buf, unit);
```
scottsand-db, line 101: Javadocs -- what does this do? As a reviewer, how should I know that SECONDARY is right and PRIMARY would have been wrong?

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/WorkloadBenchmark.java
scottsand-db: Mostly LGTM! One question: why take in an engine name? WorkloadSpecs should not define which engine they should run against, right?

IMO if we wanted different engines to run, we would just do

`class DefaultEngineBenchmark ...`

`class SuperFancyEngineBenchmark ...`

right?

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/WorkloadBenchmark.java
```diff
@@ -0,0 +1,115 @@
...
+    // Convert paths into a String array for JMH. JMH requires that parameters be of type String[].
+    String[] workloadSpecsArray =
+        filteredSpecs.stream().map(WorkloadSpec::toJsonString).toArray(String[]::new);
+
+    // Configure and run JMH benchmark with the loaded workload specs
+    Options opt =
+        new OptionsBuilder()
+            .include(WorkloadBenchmark.class.getSimpleName())
+            .param("workloadSpecJson", workloadSpecsArray)
+            // TODO: In the future, this can be extended to support multiple engines.
```
scottsand-db, line 102: would you run multiple engines all at once? like one command runs ALL specs using 3 different engines?

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/models/ReadSpec.java
```diff
@@ -0,0 +1,116 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.delta.kernel.defaults.benchmarks.workloadRunners.ReadMetadataRunner;
+import io.delta.kernel.defaults.benchmarks.workloadRunners.WorkloadRunner;
+import io.delta.kernel.engine.Engine;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Workload specification for read_metadata benchmarks. This workload reads the metadata of the
```
scottsand-db: I think out of date? since read is for both metadata and data?

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/models/WorkloadSpec.java
scottsand-db: Question: So when I run my benchmark, will it all of the read_metadata and all of the read_data? or will I configure it?

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/models/ReadSpec.java
```diff
@@ -0,0 +1,116 @@
...
+ * "name": "read-basic-checkpoint-at-latest", "table_root":
+ * "s3:///my-bucket/delta-tables/basic-checkpoint-table" }
+ */
+public class ReadSpec extends WorkloadSpec {
+
+  /** The snapshot version to read. If null, the latest version must be read. */
+  @JsonProperty("version")
+  private Long version;
+
+  @JsonProperty("expected_data")
```
scottsand-db: is this for both read_data and read_metadata?

Basically: we only have read_metadata so far as the child variant -- yet I'm seeing `expected_data.parquet` ? Can you help me to understand?

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/WorkloadBenchmark.java
OussamaSaoudi: >  WorkloadSpecs should not define which engine they should run against, right?

Yep, WorkloadSpec doesn't contain the engine name. engineName is provided through jmh. That way, if you implement a benchmark with multiple engines, jmh will run each variant {engineA, engineB} x {testA, testB, testC} ...

This can be exposed as a command line argument in the future too

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/WorkloadBenchmark.java
OussamaSaoudi: If you have multiple engines (ex: DefaultEngine vs MyFancyCustomEngine), you may want to compare the two.

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/models/ReadSpec.java
OussamaSaoudi: ah good point. I'll comb through the docs and update as we converge to a final design. Designs are still somewhat in flux as the end-to-end is built out

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/models/ReadSpec.java
OussamaSaoudi: I haven't generated any `expected_metadata.parquet` yet, so it's not reflected in the WorkloadSpec. The intent is for the `operationType` to determine whether this is a metadata or a data operation.

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/models/WorkloadSpec.java
OussamaSaoudi: By default it will convert all reads into `read_metadata` and `read_data`. In the future, I plan to allow us to filter workloads. See below:
```
    for (WorkloadSpec spec : workloadSpecs) {
      // TODO: In the future, we can filter specific workloads using command line args here.
      filteredSpecs.addAll(spec.getBenchmarkVariants());
    }
```

kernel/kernel-defaults/src/test/resources/workload_specs/basic_append/table_info.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "basic_append",
```
scottsand-db, line 2: Can you also include: which engine created this table?
Can you also include: how did you create this table? e.g. let's say I fix a bug and now I want to re-create this table -- how did you create it in the first place?

We should at the very least include the SQL or pyspark etc code here, too. Just as JSON but it's a start.

This is the same problem DAT had :)

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/KernelMetricsProfiler.java
OussamaSaoudi: Updated the docs. This function is meant to generate a SecondaryMetric and is called as part of the (newly renamed) `extractSecondaryScanMetrics`

kernel/kernel-defaults/src/test/resources/workload_specs/basic_append/table_info.json
OussamaSaoudi: Added the engineInfo. Regarding repro, I'd like to think some more on that as I generate tables for the benchmark. I have a feeling there's a better way to represent reproductions.

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/models/ReadSpec.java
```diff
@@ -0,0 +1,188 @@
...
+ *
+ * <p>When getWorkloadVariants() is called, this spec generates variants:
+ *
+ * <pre>{@code
+ * ReadSpec(type="read", version=5)
+ *   → ReadSpec(type="read", version=5, operationType="read_metadata")
+ *   → ReadSpec(type="read", version=5, operationType="read_data")  // future
+ * }</pre>
+ *
+ * <h2>Fields</h2>
```
scottsand-db: nit: not sure if we need this level of detail in classdocs? this seems like it could get out of date/sync with the code very easily (i.e. a different developer changes the code and forgets to change here)

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/BenchmarkUtils.java
OussamaSaoudi: Removed those logging statements. I left the sysout one for the benchmark report since that's handled by JMH

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/models/TableInfo.java
```diff
@@ -0,0 +1,134 @@
...
+
+  /**
+   * Information about the engine used to create this table (e.g., "Apache-Spark/3.5.1
+   * Delta-Lake/3.1.0"). Optional field to track which engine/version created the table data.
+   */
+  @JsonProperty("engineInfo")
+  public String engineInfo;
+
+  /** The root path where the Delta table is stored. */
+  @JsonProperty("table_root")
```
scottsand-db, line 64: nit: why `table_root` snake case, and then `engineInfo` camel case?

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/models/ReadSpec.java
scottsand-db: Hm, not sure if that answered my question. This is a parent, top-level workload spec, yes? Like will it have `expected_data` and `expected_metadata`? is it valid for it to only have one and not the other?

scottsand-db: would `expectedDataFile` be a better variable name?

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/models/ReadSpec.java
```diff
@@ -0,0 +1,188 @@
...
+  public Long getVersion() {
+    return version;
+  }
+
+  /**
+   * @return the full name of this workload, derived from table name, case name, and operation type.
+   */
+  @Override
+  public String getFullName() {
+    return this.tableInfo.name + "/" + this.caseName + "/read/" + this.operationType;
```
scottsand-db: nit: you don't need to keep specifying `this.`

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/models/ReadSpec.java
scottsand-db: im also a bit confused becayse we only have `read_metadata` child types but we have `expectedData`?

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/models/TableInfo.java
```diff
@@ -0,0 +1,134 @@
...
+  public String tableRoot;
+
+  /**
+   * Gets the absolute path to the root of the Delta table.
+   *
+   * @return the absolute path to the root of the table
```
scottsand-db: nit: we don't need these 2 lines that basically say the same thing. Just the `@return` is sufficient and succinct. You can apply this elsewhere, too, plz.

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/models/WorkloadSpec.java
```diff
@@ -0,0 +1,207 @@
...
+ * spec.json (on disk)
+ *     ↓ fromJsonPath()
+ * WorkloadSpec + tableInfo + caseName
+ *     ↓ getWorkloadVariants()
+ * List&lt;WorkloadSpec&gt; (with operation types set)
+ *     ↓ getRunner()
+ * WorkloadRunner instances (ready to execute)
+ * </pre>
+ *
+ * <h2>Fields</h2>
```
scottsand-db: Ditto IMO we don't need all this info here, wdyt?

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/workloadrunners/WorkloadRunner.java
```diff
@@ -0,0 +1,66 @@
...
+ * A runner that can execute a specific workload as a benchmark or test. A WorkloadRunner is created
+ * from a {@link WorkloadSpec} and is responsible for setting up any state necessary to execute the
+ * workload using {@link WorkloadRunner#setup()}, as well as executing the workload itself.
+ *
+ * <h2>Execution Modes</h2>
+ *
+ * <ul>
+ *   <li><b>Benchmark</b>: Execute via {@link #executeAsBenchmark(Blackhole)} for JMH performance
+ *       measurements
+ *   <li><b>Test</b>: Execute via executeAsTest() for correctness validation (future work)
```
scottsand-db, line 32: Just a thought: IMO we can avoid including class docs for things that don't really exist yet? Not a blocker, just food for thought. I'm fine if you keep it.

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/models/ReadSpec.java
OussamaSaoudi: To make things clear I will get rid of expectedData and expectedMetadata and add them in the future once we actually support tests.

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/models/WorkloadSpec.java
OussamaSaoudi: cool, trimmed it down 👍

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/workloadrunners/WorkloadRunner.java
OussamaSaoudi: makes sense 👍

kernel/kernel-defaults/src/test/java/io/delta/kernel/defaults/benchmarks/models/TableInfo.java
OussamaSaoudi: good catch! thank you

Issue #5034, opened 2025-08-06
[Kernel] Update `SnapshotBuilder` to take in a timestamp for time travel


scottsand-db:
Title.

scottsand-db:
Closed by https://github.com/delta-io/delta/pull/5145

PR #5145, opened 2025-08-29
[Kernel] Implement SnapshotBuilder::atTimestamp (doesn't support ratified commits)

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/5145/files) to review incremental changes.
- [**stack/kernel_snapshot_builder_at_timestamp**](https://github.com/delta-io/delta/pull/5145) [[Files changed](https://github.com/delta-io/delta/pull/5145/files)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Implement `SnapshotBuilder::atTimestamp` API for the new SnapshotBuilder Kernel interface.

For now, this only supports tables where you are NOT passing in any ratifiedCommits (we need to still build the timestamp->version parsing logic to handle ratified commits)

## How was this patch tested?

New UTs for the builder.

Un-ignore other test cases so now we e2e test time-travel-by-timestamp with the new SnapshotBuilder API too 👍 

New SnapshotBuilder::atTimestamp API

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/SnapshotQueryContext.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/SnapshotQueryContext.java
@@ -91,7 +91,7 @@ public class SnapshotQueryContext {
     return tablePath;
   }
 
-  public Optional<Long> getVersion() {
+  public Optional<Long> getResolvedVersion() {
     return resolvedVersion;
   }
 
@@ -121,13 +121,15 @@ public class SnapshotQueryContext {
   }
 
   /**
-   * Updates the {@code version} stored in this snapshot context. This version should be updated
-   * upon version resolution for non time-travel-by-version queries. For latest snapshot queries
-   * this is after log segment construction. For time-travel by timestamp queries this is after
-   * timestamp to version resolution.
+   * Set the resolved version that was actually loaded for this snapshot query.
+   *
+   * <p>For AS OF TIMESTAMP queries, this should be set upon timestamp-to-version resolution.
+   *
+   * <p>For AS OF LATEST queries, this should be set after log segment construction, when we learn
+   * what the latest version of the table really is.
    */
-  public void setVersion(long updatedVersion) {
-    resolvedVersion = Optional.of(updatedVersion);
+  public void setResolvedVersion(long resolvedVersion) {
+    this.resolvedVersion = Optional.of(resolvedVersion);
   }
 
   /** Updates the {@code checkpointVersion} stored in this snapshot context. */
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/SnapshotQueryContext.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/SnapshotQueryContext.java
@@ -52,10 +52,18 @@ public class SnapshotQueryContext {
   }
 
   private final String tablePath;
+
+  /** The version provided in a time-travel-by-version query, if any. */
+  private final Optional<Long> providedVersion;
+
+  /** The timestamp provided in a time-travel-by-timestamp query, if any. */
   private final Optional<Long> providedTimestamp;
+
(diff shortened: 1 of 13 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/SnapshotQueryContext.java
```diff
@@ -99,12 +108,15 @@ public SnapshotMetrics getSnapshotMetrics() {
   }
 
   public String getQueryDisplayStr() {
```
scottsand-db, line 110: Since even AS OF LATEST queryies eventually set the version, this was incorrectly returning AS OF VERSION X.

Now, it will return `LATEST SNAPSHOT (RESOLVED TO VERSION X)` which is much clearer.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/table/SnapshotBuilderImpl.java
```diff
@@ -99,22 +106,60 @@ public SnapshotBuilderImpl withProtocolAndMetadata(Protocol protocol, Metadata m
...
   ////////////////////////////
   // Private Helper Methods //
   ////////////////////////////
 
-  private void validateInputOnBuild() {
-    checkArgument(ctx.versionOpt.orElse(0L) >= 0, "version must be >= 0");
+  private void validateInputOnBuild(Engine engine) {
+    validateVersionNonNegative();
+    validateTimestampNonNegative();
+    validateTimestampNotGreaterThanLatestSnapshot(engine);
```
scottsand-db, line 120: oop. validate that logDatas is empty (not currently supported).

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/SnapshotQueryContext.java
```diff
@@ -115,7 +127,7 @@ public String getQueryDisplayStr() {
    * timestamp to version resolution.
    */
   public void setVersion(long updatedVersion) {
```
allisonport-db: nit - maybe rename to `setResolvedVersion` since we're changing the semantics of these variables?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/metrics/SnapshotQueryContext.java
```diff
@@ -73,7 +81,8 @@ private SnapshotQueryContext(
       Optional<Long> checkpointVersion,
       Optional<Long> providedTimestamp) {
     this.tablePath = tablePath;
-    this.version = providedVersion;
+    this.providedVersion = providedVersion;
+    this.resolvedVersion = providedVersion;
```
allisonport-db, line 85: wdyt - should we wait to set `resolvedVersion = providedVersion` till we have listed and can confirm such version exists and is reconstructable?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/table/SnapshotFactory.java
```diff
@@ -154,4 +164,40 @@ private Protocol getProtocol(LogReplay logReplay) {
   private Metadata getMetadata(LogReplay logReplay) {
     return ctx.protocolAndMetadataOpt.map(x -> x._2).orElseGet(logReplay::getMetadata);
   }
+
+  public static long resolveTimestampToSnapshotVersion(
```
allisonport-db: Super minor but since this is a static util used outside of this class can we maybe add simple docs? Should it be at the top or in a separate section?

scottsand-db: Will definitely add the docs. I originally put it here since this is really just a helper method (if no one else needed this, it would be private) ... but ... yeah, can move to top 👍

scottsand-db:
Closes #5034

PR #5146, opened 2025-08-30
[Unity Catalog] UC-CatalogManaged-Client create table API

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/5146/files) to review incremental changes.
- [**stack/uc_catalog_managed_client_create_table**](https://github.com/delta-io/delta/pull/5146) [[Files changed](https://github.com/delta-io/delta/pull/5146/files)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Exposes a `UCCatalogManagedClient::buildCreateTableTransaction` API that injects the custom UC committer and sets required table properties.

Also updates the Kernel CreateTableTransactionBuilder::withTableProperties API semantics to allow multiple invocations.

## How was this patch tested?

New simple UTs.

New `UCCatalogManagedClient::buildCreateTableTransaction` API

### DIFF
--- a/unity/src/test/scala/io/delta/unity/UCCatalogManagedClientSuite.scala
+++ b/unity/src/test/scala/io/delta/unity/UCCatalogManagedClientSuite.scala
@@ -212,7 +212,7 @@ class UCCatalogManagedClientSuite extends AnyFunSuite with UCCatalogManagedTestU
     assert(snapshot.getCommitter.isInstanceOf[UCCatalogManagedCommitter])
   }
 
-  test("buildCreateTableTransaction sets required properties") {
+  test("buildCreateTableTransaction sets required properties and uses UC committer") {
     // ===== GIVEN =====
     val ucClient = new InMemoryUCClient("ucMetastoreId")
     val ucCatalogManagedClient = new UCCatalogManagedClient(ucClient)
@@ -228,19 +228,7 @@ class UCCatalogManagedClientSuite extends AnyFunSuite with UCCatalogManagedTestU
     assert(builderTableProperties.get("delta.feature.catalogOwned-preview") == "supported")
     assert(builderTableProperties.get("ucTableId") == "ucTableId")
     assert(builderTableProperties.get("foo") == "bar")
-  }
 
-  test("buildCreateTableTransaction uses UC committer") {
-    // ===== GIVEN =====
-    val ucClient = new InMemoryUCClient("ucMetastoreId")
-    val ucCatalogManagedClient = new UCCatalogManagedClient(ucClient)
-
-    // ===== WHEN =====
-    val createTableTxnBuilder = ucCatalogManagedClient
-      .buildCreateTableTransaction("ucTableId", baseTestTablePath, testSchema, "test-engine")
-      .asInstanceOf[CreateTableTransactionBuilderImpl]
-
-    // ===== THEN =====
     val committerOpt = createTableTxnBuilder.getCommitterOpt
     assert(committerOpt.get().isInstanceOf[UCCatalogManagedCommitter])
   }


unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
```diff
@@ -163,6 +211,17 @@ private void validateLoadTableVersionExists(
     }
   }
 
+  private Map<String, String> getRequiredTablePropertiesForCreate(String ucTableId) {
+    final Map<String, String> requiredProperties = new HashMap<>();
+
+    requiredProperties.put(
+        CATALOG_MANAGED_ENABLEMENT_PROPERTY, TableFeatures.SET_TABLE_FEATURE_SUPPORTED_VALUE);
+
+    // TODO: use the ucTableId
```
allisonport-db: what will this be used for?

allisonport-db (commented):
overall makes sense to me; it's super simple which I think means we're doing something right :)

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
scottsand-db: Just something UC requires! `delta-spark` may sometimes poke at this.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/CreateTableTransactionBuilderImpl.java
```diff
@@ -54,10 +55,31 @@ public CreateTableTransactionBuilderImpl(String tablePath, StructType schema, St
...
+    if (!this.tableProperties.isPresent()) {
+      this.tableProperties = Optional.of(Collections.unmodifiableMap(normalizedNewProperties));
+      return this;
+    }
+
+    // Case 2: Properties have already been set; ensure no duplicates
+    final Map<String, String> existingProperties = this.tableProperties.get();
+    for (String key : normalizedNewProperties.keySet()) {
+      if (existingProperties.containsKey(key)) {
+        throw new IllegalArgumentException(
```
xzhseh: QQ: Where do we update/override the existing properties if we directly throw an exception for every duplicate here?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/TablePropertiesSuite.scala
```diff
@@ -27,7 +27,37 @@ import org.scalatest.funsuite.AnyFunSuite
 class TablePropertiesTransactionBuilderV1Suite extends TablePropertiesSuiteBase with WriteUtils {}
 
 class TablePropertiesTransactionBuilderV2Suite extends TablePropertiesSuiteBase
-    with WriteUtilsWithV2Builders {}
+    with WriteUtilsWithV2Builders {
+  test("create table (V2 only) - withTableProperties can be called multiple times") {
```
xzhseh, line 31: What does V2 here refer to?

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
```diff
@@ -99,6 +118,38 @@ public Snapshot loadSnapshot(
...
 
+  /**
+   * Builds a create table transaction for a Unity Catalog managed Delta table.
+   *
+   * <p>Configures the transaction with a {@link UCCatalogManagedCommitter} and required table
+   * properties for catalog-managed table enablement.
+   *
+   * <p>This assumes the table is being created in a staging location as per UC semantics. Once this
+   * transaction is built and committed, creating 000.json, you must call {@code
+   * TablesApi::createTable} to inform Unity Catalog of the successful table creation.
```
xzhseh, line 129: Can we add a callback function that will always be called as part of `buildCreateTableTransaction`? This way we ensure the callback will be invoked after finishing building and committing the transaction. This would complicate the interface, though.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/TablePropertiesSuite.scala
scottsand-db: There's old/legacy/v1 APIs, and then theres new/v2 APIs

unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
scottsand-db: Sorry, what would this callback function do? You want this to be the `createTable` UC RPC call?

That could just go into the `UCCatalogManagedCommitter` -- there's no need to pass in a callback function.

We've discussed in the past that for now it's fine for this `UCCatalogManagedCommitter` to just do a normal `000.json` wriite for CREATE, and then return back to the Connector, who is then responsible for the UC::CreateTable RPC call.

In the future, we can push that UC::CreateTable RPC call into the committer.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/CreateTableTransactionBuilderImpl.java
scottsand-db: These API semantics do not let you update or override properties that are already set on this CreateTableTransactionBuilder.

For example, the `UCCatalogManagedClient::buildCreateTableTransaction` will set the `ucTableId` and will set `delta.feature.catalogOwned-preview=supported`. We don't want the Connector to override these fields.

Do you agree/disagree with this semantic?

unity/src/test/scala/io/delta/unity/UCCatalogManagedClientSuite.scala
```diff
@@ -212,4 +212,22 @@ class UCCatalogManagedClientSuite extends AnyFunSuite with UCCatalogManagedTestU
     assert(snapshot.getCommitter.isInstanceOf[UCCatalogManagedCommitter])
   }
 
+  test("buildCreateTableTransaction creates transaction and sets required properties") {
```
allisonport-db: Does there need to be an E2E test with committing? Also should check the committer right?

allisonport-db (commented):
src code LGTM just a question about testing

unity/src/test/scala/io/delta/unity/UCCatalogManagedClientSuite.scala
scottsand-db: Great question. There already are! https://github.com/delta-io/delta/blob/74ef11b45cf875d24f9d58615910cd2e2ed681a9/unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala#L428

allisonport-db: (1) Isn't that not an E2E test using the full API? Wondering if we can check committing the transaction that we are creating using this API from building it to committing?

(2) Can we at least assert the committer is set to the UCCatalogManagedCommitter in the builder?

allisonport-db: Related slightly - should setting the committer twice fail? I wonder if we need/want to safeguard the committer like we now are for the properties.

My inclination would be that it's not necessary.

scottsand-db: For (1) yup fair observation, sorry I wasn't trying to be misleading.

I plan on doing extensive followup tests for CREATE + INSERT in followup PR(s).

What I meant from the above is that, we do have the individual E2E components tested:
a. We can get a CreateTableTransactionBuilder from the UCCatalogManagedClient
b. We know that Kernel can do all the snapshot & P+M & transaction work and invoke the committer
c. We know that he UCCatalogManagedCommitter can be invoked for CREATE (version 0)

> (2) Can we at least assert the committer is set to the UCCatalogManagedCommitter in the builder?

Sure will do

scottsand-db: > My inclination would be that it's not necessary.

Agreed -- I don't think things like this would come up in practice

kernel/kernel-api/src/main/java/io/delta/kernel/internal/CreateTableTransactionBuilderImpl.java
```diff
@@ -54,10 +55,31 @@ public CreateTableTransactionBuilderImpl(String tablePath, StructType schema, St
...
+        TableConfig.validateAndNormalizeDeltaProperties(properties);
+
+    // Case 1: First time properties are being set
+    if (!this.tableProperties.isPresent()) {
+      this.tableProperties = Optional.of(Collections.unmodifiableMap(normalizedNewProperties));
+      return this;
+    }
+
+    // Case 2: Properties have already been set; ensure no duplicates
+    final Map<String, String> existingProperties = this.tableProperties.get();
```
scottsand-db, line 69: actually -- only throw if the value is DIFFERENT

unity/src/test/scala/io/delta/unity/UCCatalogManagedClientSuite.scala
```diff
@@ -230,4 +230,19 @@ class UCCatalogManagedClientSuite extends AnyFunSuite with UCCatalogManagedTestU
...
+    val ucCatalogManagedClient = new UCCatalogManagedClient(ucClient)
+
+    // ===== WHEN =====
+    val createTableTxnBuilder = ucCatalogManagedClient
+      .buildCreateTableTransaction("ucTableId", baseTestTablePath, testSchema, "test-engine")
+      .asInstanceOf[CreateTableTransactionBuilderImpl]
+
+    // ===== THEN =====
+    val committerOpt = createTableTxnBuilder.getCommitterOpt
+    assert(committerOpt.get().isInstanceOf[UCCatalogManagedCommitter])
```
allisonport-db, line 233: nit but I think this assert could have just been added to the previous test instead of a whole new one

allisonport-db (approved):
LGTM

unity/src/test/scala/io/delta/unity/UCCatalogManagedClientSuite.scala
scottsand-db: will push

PR #5149, opened 2025-08-31
[Kernel] Fix Snapshot::getTimestamp to lookup the CommitInfo using Delta file from LogSegment

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/5149/files) to review incremental changes.
- [**stack/kernel_snapshot_get_commit_info_from_ratified_commit_b**](https://github.com/delta-io/delta/pull/5149) [[Files changed](https://github.com/delta-io/delta/pull/5149/files)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

On ICT tables, Snapshot::getTimestamp has to read the CommitInfo file.

Previously, this was blindly assuming and then trying to read `0000<version>.json`.

However on catalogManaged tables, there's no guarantee this version exists.

This PR fixes this so that we instead look up the right delta file using the existing LogSegment.

## How was this patch tested?

A few new unit tests and E2E tests.

No.


kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/CommitInfo.java
```diff
@@ -61,6 +61,35 @@
  */
 public class CommitInfo {
 
+  //////////////////////////////////
+  // Static variables and methods //
+  //////////////////////////////////
+
+  public static final StructType FULL_SCHEMA =
```
scottsand-db, line 68: I just fix the order of static / member methods

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/CommitInfo.java
```diff
@@ -86,27 +115,83 @@ public static CommitInfo fromColumnVector(ColumnVector vector, int rowId) {
...
+    return commitInfo.inCommitTimestamp.orElseThrow(
+        () -> DeltaErrors.tableWithIctMissingIct(dataPath.toString(), version));
+  }
 
-  private static final Logger logger = LoggerFactory.getLogger(CommitInfo.class);
+  /**
+   * Get the CommitInfo action (if available) from the delta file at the given logPath and version.
+   */
+  // TODO: [delta-io/delta#5147] Can't just use the logPath & version on catalogManaged tables.
+  public static Optional<CommitInfo> getCommitInfoOpt(Engine engine, Path logPath, long version) {
```
scottsand-db: Same signature as the old method, now pointing to the new method.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/CommitInfo.java
```diff
@@ -86,27 +115,83 @@ public static CommitInfo fromColumnVector(ColumnVector vector, int rowId) {
...
-          .add(
-              "operationMetrics",
-              new MapType(StringType.STRING, StringType.STRING, true /* nullable */));
+  /**
+   * Returns the `inCommitTimestamp` of delta file at the requested version. Throws an exception if
+   * the delta file does not exist or does not have a commitInfo action or if the commitInfo action
+   * contains an empty `inCommitTimestamp`.
+   */
+  // TODO: [delta-io/delta#5147] Can't just use the logPath & version on catalogManaged tables.
+  public static long getRequiredInCommitTimestampFromFile(
```
allisonport-db: Where is this used currently? Also maybe `..FromBackfilledDeltaFile` or something else of that specificity?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/CommitInfo.java
allisonport-db: is this used anywhere besides `getRequiredInCommitTimestampFromFile`? I would almost prefer we just get rid of this? or at least similarly re-name this to be clear this uses the log-path delta files only

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
```diff
@@ -273,6 +272,16 @@ public List<FileStatus> allFilesWithCompactionsReversed() {
...
   }
 
+  /**
+   * Returns the delta file for the given version, if it exists in this LogSegment.
+   *
+   * @param version the version to lookup
+   * @return Optional containing the delta file, or empty if not found
+   */
+  public Optional<FileStatus> getDeltaFileForVersion(long version) {
+    return Optional.ofNullable(versionToDeltaMap.get(version));
```
allisonport-db: is it worth validating that the version is in the logsegment start/end range? and throwing an exception then instead of just returning empty?

I guess another way of phrasing this is, when do we expect this to potentially return empty? If never -- should this validate the input version and return a non-optional?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/CommitInfo.java
scottsand-db: > Where is this used currently?

https://github.com/delta-io/delta/blob/c2e4df98cbc0ff084ed4333d40179ab365c5b710/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java#L207

https://github.com/delta-io/delta/blob/c2e4df98cbc0ff084ed4333d40179ab365c5b710/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java#L262

https://github.com/delta-io/delta/blob/c2e4df98cbc0ff084ed4333d40179ab365c5b710/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java#L266

> Also maybe ..FromBackfilledDeltaFile or something else of that specificity?

Great idea! Maybe: `unsafeGetRequiredInCommitTimestampFromPublishedDeltaFile`. e.g. `delta-spark` uses some `unsafeXXX` for methods that should be fixed.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/CommitInfo.java
scottsand-db: > is this used anywhere besides getRequiredInCommitTimestampFromFile?

Yes -- used in DeltaHistoryManager

Will rename as well!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
scottsand-db: > when do we expect this to potentially return empty?

Great question. For example. Suppose I have a Snapshot at version 13. Is there a 10.json? Maybe. Maybe we have json 0 to 13. Maybe we have 10.checkpoint and 11->13.json.

So, I think it's reasonable to return an optional for cases like this instead of throwing an exception.

If we are at version 13 and you ask for a delta at version 14, yup I can throw an exception 👍

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/CommitInfo.java
allisonport-db: SGTM. And it seems like this will be obsolete once I update DeltaHistoryManager for CCV2 :)

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
allisonport-db: What happens here if I have a snapshot with only the checkpoint in my log segment? Will this getTimestamp just throw an exception?

allisonport-db: To formalize it, we have the delta file for > checkpointVersion and <= logSegment.version right? I just wonder if this should be more transparent rather than just hidden behind this method returning optional. Also just trying to better understand the use-cases where we are calling this API and hitting the empty clause.

scottsand-db: > What happens here if I have a snapshot with only the checkpoint in my log segment? Will this getTimestamp just throw an exception?

Great catch! Yup I need to think through this.

> To formalize it, we have the delta file for > checkpointVersion and <= logSegment.version right? I just wonder if this should be more transparent rather than just hidden behind this method returning optional.

Sorry, I don't know what you mean by more transparent?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/CommitInfo.java
```diff
@@ -86,27 +115,84 @@ public static CommitInfo fromColumnVector(ColumnVector vector, int rowId) {
...
+        FileStatus.of(
+            FileNames.deltaFile(logPath, version), /* path */
+            0, /* size */
+            0 /* modification time */);
+
+    return tryReadCommitInfoFromDeltaFile(engine, file);
+  }
+
+  /** Read the CommitInfo action (if available) from the given delta file. */
+  public static Optional<CommitInfo> tryReadCommitInfoFromDeltaFile(
```
scottsand-db, line 164: just renamed; and take in the delta file instead of logPath + version

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
```diff
@@ -34,35 +34,16 @@
 
 public class LogSegment {
 
-  //////////////////////////////////
-  // Static variables and methods //
-  //////////////////////////////////
-
-  public static LogSegment empty(Path logPath) {
```
scottsand-db, line 41: LogSegment::empty is no longer useful

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
```diff
@@ -186,21 +165,6 @@ public LogSegment(
   // Public APIs //
   /////////////////
 
-  /**
-   * @return true if this LogSegment is complete and fully describes a Snapshot version. A partial
-   *     LogSegment is missing some information. We consider an empty LogSegment to be incomplete.
-   */
-  public boolean isComplete() {
```
scottsand-db, line 193: Was never used

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/snapshot/LogSegmentSuite.scala
```diff
@@ -39,132 +39,121 @@ class LogSegmentSuite extends AnyFunSuite with MockFileSystemClientUtils {
     FileStatus.of(s"${logPath.toString}/gibberish.checkpoint.parquet", 1, 1))
   private val logPath2 = new Path("/another/fake/path/to/table/", "_delta_log")
 
-  test("constructor -- valid case (empty)") {
-    LogSegment.empty(logPath)
-  }
+  private def createLogSegmentForTest(
```
scottsand-db, line 42: simplifies tests!

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/snapshot/LogSegmentSuite.scala
```diff
@@ -299,84 +283,20 @@ class LogSegmentSuite extends AnyFunSuite with MockFileSystemClientUtils {
...
-        1)
+      createLogSegmentForTest(
+        version = 12,
+        deltas = deltasFs11To12List,
+        checkpoints = checkpointsForDifferentTable)
     }
     assert(ex.getMessage.contains("doesn't belong in the transaction log"))
   }
 
-  test("isComplete") {
```
scottsand-db, line 314: deleted this method; no longer applicable

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/CommitInfo.java
```diff
@@ -86,27 +115,84 @@ public static CommitInfo fromColumnVector(ColumnVector vector, int rowId) {
...
+    return commitInfo.inCommitTimestamp.orElseThrow(
+        () -> DeltaErrors.tableWithIctMissingIct(dataPath.toString(), version));
+  }
 
-  private static final Logger logger = LoggerFactory.getLogger(CommitInfo.class);
+  /**
+   * Get the CommitInfo action (if available) from the delta file at the given logPath and version.
+   */
+  // TODO: [delta-io/delta#5147] Can't just use the logPath & version on catalogManaged tables.
+  public static Optional<CommitInfo> unsafeTryReadCommitInfoFromPublishedDeltaFile(
```
raveeram-db, line 152: Perhaps add a comment explaining context around the `unsafe` prefix?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
```diff
@@ -243,16 +207,33 @@ public Optional<FileStatus> getLastSeenChecksum() {
...
+   *
+   * <p>If this LogSegment has only checkpoints (i.e. 10.checkpoint only) then this is the
+   * filesystem timestamp of the delta at that checkpoint version.
+   *
+   * <p>By unadjusted, we mean that this commit timestamp may not necessarily be monotonically
+   * increasing for the commits within this segment.
    */
-  public List<FileStatus> allLogFilesUnsorted() {
-    return deltasAndCheckpoints;
+  public long getDeltaFileTimestampAtEndVersion() {
```
huan233usc: It seems to me a little bit overkill to have this method as for now the only use case is in snapshot. getTimestamp. 

I wonder if directly use "logSegment.getDeltaFileAtEndVersion().getModificationTime()" there is more readable -- as a reader, I could directly know it is the modification time of last delta file.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/CommitInfo.java
huan233usc: +1, it seems that the action of ReadCommitInfoFromPublishedDeltaFile is not unsafe, it is just the result not trusted for catalog managed table

kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java
scottsand-db: Sounds good! Will implement this feedback.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/actions/CommitInfo.java
scottsand-db: @huan233usc that's actually not correct. it is unsafe. there's no guarantee that `_delta_log/0000N.json` exists!

scottsand-db: Added the comment 👍

allisonport-db (approved):
LGTM

huan233usc (approved):
LGTM

PR #5160, opened 2025-09-05
[Kernel][Test] Test CommitRange API with ICT timestamp-resolution

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Adds test coverage for ICT case.

## How was this patch tested?

Test change.

No.

### DIFF
--- a/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/TableChangesSuite.scala
+++ b/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/TableChangesSuite.scala
@@ -173,11 +173,11 @@ class CommitRangeTableChangesSuite extends TableChangesSuite {
 
   test("ICT timestamp resolution") {
     withTempDirAndEngine { (tablePath, engine) =>
-      // Create a table with ICT enabled from the start
+      // Create a table with ICT enabled from the start and commits with specific custom-set ICTs
       val startTime = 1000L
       val clock = new ManualClock(startTime)
 
-      // Create initial table with ICT enabled
+      // Version 0 has ICT=1000L, but modificationTime=approx current time (1757368326512+)
       appendData(
         engine,
         tablePath,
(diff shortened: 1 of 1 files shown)


kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/TableChangesSuite.scala
```diff
@@ -171,7 +171,94 @@ class CommitRangeTableChangesSuite extends TableChangesSuite {
     }
   }
 
-  // TODO: test ICT-based timestamp-to-version resolution
+  test("ICT timestamp resolution") {
+    withTempDirAndEngine { (tablePath, engine) =>
+      // Create a table with ICT enabled from the start
```
scottsand-db: nit: this comment seems really similar to the one below?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/TableChangesSuite.scala
```diff
@@ -171,7 +171,94 @@ class CommitRangeTableChangesSuite extends TableChangesSuite {
     }
   }
 
-  // TODO: test ICT-based timestamp-to-version resolution
+  test("ICT timestamp resolution") {
```
scottsand-db, line 174: How does this test relate to the test above `Basic timestamp resolution` ?

Like, besides this test having the table feature ICT enabled, what is different? e.g. do we somewhere check that even though the file modification time is X, we are _really_ actually using the ICT time Y != X, and that's how we know ICT is working for this streaming API?

Or, can we combine this method with the other, similar test?

Perhaps it's the manualClock thingy. Maybe just add a comment explaining this sort of stuff to the reader? ie to me it's not obvious. ty!

allisonport-db: I'll add a comment. But yeah, basically the modificationTime should be based on system time, the ICT based on the manual clock. So given we can query using the manual clock timestamps, we must be using ICT for resolution.

scottsand-db: Thank you!

PR #5161, opened 2025-09-05
Implement DSV2 table reader for Java kernel

gengliangwang:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description
Implement DSV2 table reader for Java kernel.
High level design:
* Delta Kernel
	•	Resolves table state from Catalog and Delta logs.
	•	Performs file skipping to minimize unnecessary reads.
* Spark Engine (via DSV2)
	•	Pushes down static and dynamic predicates.
	•	Partitions files(default 128MB per split) for efficient scheduling.
	•	Executes Parquet file scans with Spark's existing `ParquetFileFormat`.

Delta-kernel-connector

This PR implements the basic batch read. The predicate pushdown will be done in a follow-up PR.
This is a team work with @huan233usc @yhuang-db.
## How was this patch tested?

Existing tests. 

No


spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/read/SparkPartitionReader.java
```diff
@@ -0,0 +1,57 @@
...
+    this.readFunc = readFunc;
+    this.partition = partition;
+    ;
+  }
+
+  @Override
+  public boolean next() {
+    while (!currentIterator.hasNext() && currentFileIndex < partition.files().length) {
+      PartitionedFile file = partition.files()[currentFileIndex];
+      currentIterator = (Iterator<T>) readFunc.apply(file);
```
huan233usc: Do we want to close the previous one by casting to the closable? 

(the actual impl did implements closable)

https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/RecordReaderIterator.scala#L32

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/read/SparkPartitionReader.java
```diff
@@ -0,0 +1,57 @@
...
+      currentFileIndex++;
+    }
+    return currentIterator.hasNext();
+  }
+
+  @Override
+  public T get() {
```
huan233usc, line 70: Do we want to follow the api document to implement the iterator?

https://github.com/apache/spark/blob/1043d53b62bd5d152aa4d685fa2c6880ce015b77/sql/catalyst/src/main/java/org/apache/spark/sql/connector/read/PartitionReader.java#L41-L51 i.e. next move forwards and get should return same value until `next` is called.

  /**
   * Proceed to next record, returns false if there is no more records.
   *
   * @throws IOException if failure happens during disk/network IO like reading files.
   */
  boolean next() throws IOException;

  /**
   * Return the current record. This method should return same value until `next` is called.
   */
  T get();

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/read/SparkScan.java
```diff
@@ -58,6 +59,7 @@ public class SparkScan implements Scan, SupportsReportStatistics {
 
   public SparkScan(
       String tableName,
```
raveeram-db: bit of a nit but does it makes sense to have a tableInfo object with these two bits of information? Might make it easier in other parts of the code and to make sure they're consistent (e.g. equality checks and such)

raveeram-db (approved):
LGTM! Just a small comment but fine to address it later.

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/read/SparkScan.java
gengliangwang: @raveeram-db actually we can remove table name here. The physical plan `BatchScanExec` has the table name already:
https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/v2/BatchScanExec.scala#L166

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/read/SparkPartitionReader.java
gengliangwang: Sure, updated.

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/read/SparkPartitionReader.java
gengliangwang: Added comment on line 74. The behavior is `RecordReaderIterator.next` is exactly the same as `PartitionReader.get`

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/read/SparkScanBuilder.java
```diff
@@ -0,0 +1,93 @@
...
+    this.requiredDataSchema =
+        new StructType(
+            Arrays.stream(requiredSchema.fields())
+                .filter(f -> !partitionColumnSet.contains(f.name().toLowerCase(Locale.ROOT)))
+                .toArray(StructField[]::new));
+  }
+
+  @Override
+  public org.apache.spark.sql.connector.read.Scan build() {
+    // TODO: Implement predicate pushdown by translating Spark Filters to Delta Kernel Predicates.
```
raveeram-db, line 82: could we please add an issue for this @gengliangwang ? Good first issue for someone to pick up

raveeram-db (approved):
Overall looks good! Thanks for the PR

PR #5162, opened 2025-09-05
[Spark] Some minor refactoring and more clarification to the comment about CommitInfo deserialization

ctring:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR does some minor refactoring and add more clarification to commit info deserialization

## How was this patch tested?

Existing unit tests

No

### DIFF
--- a/spark/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala
+++ b/spark/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala
@@ -1214,23 +1214,6 @@ case class CommitInfo(
   @JsonIgnore
   override def getVersion: Long = version.get
 
-  /**
-   * Returns the legacy value of operation parameters after deserialization.
-   * See [[JsonMapDeserializer]] for more details about how operation
-   * parameter deserialization was broken before. These legacy values are the
-   * same as the original broken version.
-   */
-  @JsonIgnore
-  def getLegacyPostDeserializationOperationParameters: Map[String, String] = {
-    // Use JsonMapSerializer to serialize the operation parameters
-    val serializedOperationParameters =
-      JsonUtils.toJson(CommitInfoOperationParametersOnly(operationParameters))
-    // Instead of using JsonMapDeserializer,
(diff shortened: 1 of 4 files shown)


spark/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala
```diff
@@ -1484,51 +1484,66 @@ class JsonMapSerializer extends JsonSerializer[Map[String, String]] {
...
+ *     strings.
+ *   - However, the deserialization process used a default deserializer that was not an inverse of
+ *     JsonMapSerializer. The default deserializer automatically decoded JSON strings into their
+ *     inferred types, and then cast the values to strings. This meant the output of the
+ *     deserialization process was a (String -> String(Any)).
+ *   - Note that String() and JsonEncodedString() are not the same. For example, if the original
+ *     value was a string "abc", then String("abc") is still "abc", but JsonEncodedString("abc") is
+ *     "\"abc\"". We want to the deserialization process to recover the input of the serialization
+ *     process, i.e. (String -> JsonEncodedString(Any)).
+ *
```
prakharjain09: The code comment related to previous state may get out of date very soon.

    Previously, the deserialization of operationParameters was broken.

Any reason why we want to add the old state?

ctring: We still have this legacy behavior in `DeltaHistory` (see `spark/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala` below), so I'm adding context for what the legacy behavior is.

prakharjain09: got it. In that case, should this comment be with `CommitInfoOperationParametersOnly` class? so it explains why old deserializer is retained in code?

ctring: I will move it there.

PR #5163, opened 2025-09-05
[Spark] Column mapping conflict free Enablement

andreaschat-db:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR allow the conflict free enablement of column mapping. This is based on the fact that when a table is first created without column mapping and then enabled later, we do not write any metadata to the file footers. Therefore, if a transaction conflicted with a column mapping enablement transaction, it does not need to write any metadata changes to adapt.

## How was this patch tested?

Added new suite.

No.


xupefei (approved):
Lgtm!

PR #5168, opened 2025-09-05
[Kernel] UCCatalogManagedCommitter uses `getFileStatus` Engine API; remove impossible exception handling cases

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/5168/files) to review incremental changes.
- [**stack/kernel_uc_committer_fixes**](https://github.com/delta-io/delta/pull/5168) [[Files changed](https://github.com/delta-io/delta/pull/5168/files)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

This PR updates the UCCatalogManagedCommitter to:
- for file writes, always use overwrite=true. We are always either writing to a UUID file or a staging location.
- use the Engine::getFileSystemClient::getFileStatus API in its return value

## How was this patch tested?

New simple UT.

No.

### DIFF
--- a/unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
+++ b/unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
@@ -234,6 +234,30 @@ class UCCatalogManagedCommitterSuite
       org.apache.hadoop.fs.permission.FsPermission.getFileDefault)
   }
 
+  test("writeDeltaFile returns real FileStatus") {
+    withTempTableAndLogPathAndStagedCommitFolderCreated { case (tablePath, logPath) =>
+      // ===== GIVEN =====
+      val ucClient = new InMemoryUCClient("ucMetastoreId")
+      val committer = new UCCatalogManagedCommitter(ucClient, "ucTableId", tablePath)
+      val testValue = "TEST_FILE_STATUS_DATA"
+      val actionsIterator = getSingleElementRowIter(testValue)
+
+      val commitMetadata = createCommitMetadata(
+        version = 0,
+        logPath = logPath,
+        newProtocolOpt = Optional.of(protocolWithCatalogManagedSupport),
+        newMetadataOpt = Optional.of(basicPartitionedMetadata))
+
+      // ===== WHEN =====
+      val response = committer.commit(defaultEngine, actionsIterator, commitMetadata)
+
+      // ===== THEN =====
+      val fileStatus = response.getCommitLogData.getFileStatus
+      assert(fileStatus.getSize > 0)
+      assert(fileStatus.getModificationTime > 0)
+    }
+  }
+
   // ===============================================================
   // ===================== CATALOG_WRITE Tests =====================
   // ===============================================================


unity/src/test/scala/io/delta/unity/UCCatalogManagedCommitterSuite.scala
```diff
@@ -466,35 +490,6 @@ class UCCatalogManagedCommitterSuite
     }
   }
 
-  test("CATALOG_CREATE: FileAlreadyExistsException returns success") {
```
scottsand-db, line 469: Add Test: if engine file system client throws IO exception then we throw a CFE

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -144,80 +151,45 @@ private void validateLogPathBelongsToThisUcTable(CommitMetadata cm) {
...
-          commitMetadata.getPublishedDeltaFilePath());
-      // This can happen if a previous commit attempt from this writer succeeded in writing 00.json,
-      // but the client was not successfully notified of that write (e.g. due to network
-      // issues).
-      //
-      // Since we know we are conflicting with our previous write, we can safely return a successful
-      // response.
-      return FileStatus.of(commitMetadata.getPublishedDeltaFilePath());
+          () -> {
+            // Note: the engine is responsible for closing the actions iterator once it has been
```
raveeram-db, line 173: A bit tangential, should we add this note to the `commit()` methods API docs too? https://github.com/delta-io/delta/blob/2368503757507a260f7281dd1dc1fe328e894f08/kernel/kernel-api/src/main/java/io/delta/kernel/commit/Committer.java#L63
Just so the engine is aware of this responsibility of closing the iterator

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -144,80 +151,45 @@ private void validateLogPathBelongsToThisUcTable(CommitMetadata cm) {
...
+            //       fully consumed.
+            engine
+                .getJsonHandler()
+                .writeJsonFileAtomically(filePath, finalizedActions, true /* overwrite */);
+
+            return engine.getFileSystemClient().getFileStatus(filePath);
```
huan233usc, line 179: On retrying due to commit failure(if we do so), will we generating a new uuid or use the existing one? 

I am thinking about the case that write json succeeds but get file status fails due to whatever reason. 

If we are generating a new one or not doing retry at all, it should be fine.

scottsand-db: Each retry attempt will use a new UUID.

Even if we didn't -- that's fine too! e.g. we do overwrite=true -- totally fine if we replace our previous write

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
allisonport-db: By the engine being responsible for closing the actions iterator, I interpreted this to mean the engine JsonHandler is responsible for closing it during the writeJsonFileAtomically. Is that correct or is it somewhere else?

allisonport-db: Can someone clarify for me how this relates to the Committer?

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
```diff
@@ -144,80 +151,45 @@ private void validateLogPathBelongsToThisUcTable(CommitMetadata cm) {
...
+          });
     } catch (IOException ex) {
```
allisonport-db, line 181: Is there any nuance or extra complexity to consider regarding if this originated from the write call vs the getFileStatus call?

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
scottsand-db: Great questions.

1. The `Committer::commit` is invoekd and passed in the finalizedActions. The Committer is the consumer here. The Committer is responsible for closing the actions (this is what Rakesh commented above)

2. Our UC-Committer implementaiton writes out the finalizedActions using the Engine::JsonHandler API. When we invoke that API, we pass on the responsibility to that API to close the finalizedActions iterator (this is an implementation detail of our UC Committer)

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
scottsand-db: Great question. I thought about that, too, for some time. The only thing it _could_ impact is the error we throw, i.e. we succeeded in writing the file but failed in getting the file status. I think the chances of this is quite slim.

In either case, if we fail in either, then we will (maybe) retry. And if we retry, we will write with overwrite=true, so it actually doesn't really matter where it fails.

Anyways, I was inclined not to go to that level of detail in the error message. Just felt like a bit overkill. If we start experiencing lots of user reports that his is a big deal, we can fix it later IMO.

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
allisonport-db: Cool thanks!

unity/src/main/java/io/delta/unity/UCCatalogManagedCommitter.java
allisonport-db: Thanks, that sounds reasonable to me!

PR #5172, opened 2025-09-08
[Spark] ConditionalIncrementMetric

juliuszsompolski:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Add `ConditionalIncrementMetric` expression that allows to increment a metric only if an additional conditions is passed.
In some cases, `IncrementMetric` is used inside a more complex expression, like `if (updated) IncrementMetric("numUpdatedRows", true) else IncrementMetric("numCopiedRows", false)`. But in some cases, Spark can optimize some of these out. `ConditionalIncrementMetric` will let us express this better.

## How was this patch tested?

Tests added.

No


juliuszsompolski:
@andreaschat-db

PR #5176, opened 2025-09-08
[Kernel] Update IcebergWriterCompat ALLOW_LIST to allow CatalogManaged table feature

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/5176/files) to review incremental changes.
- [**stack/kernel_iceberg_writer_compat_allow_catalog_managed_feature**](https://github.com/delta-io/delta/pull/5176) [[Files changed](https://github.com/delta-io/delta/pull/5176/files)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

This PR adds `catalogManaged` table feature to the IcebergWriterCompatV1/V3 allow list of features.

## How was this patch tested?

Update existing tests where applicable to support catalogManaged.

No.

### DIFF
--- a/kernel/kernel-api/src/test/scala/io/delta/kernel/internal/icebergcompat/IcebergCompatMetadataValidatorAndUpdaterSuiteBase.scala
+++ b/kernel/kernel-api/src/test/scala/io/delta/kernel/internal/icebergcompat/IcebergCompatMetadataValidatorAndUpdaterSuiteBase.scala
@@ -92,7 +92,7 @@ trait IcebergCompatMetadataValidatorAndUpdaterSuiteBase
       }
   }
 
-  SIMPLE_TYPES.toSet.diff(simpleTypesToSkip).foreach {
+  PRIMITIVE_TYPES.diff(simpleTypesToSkip).foreach {
     dataType: DataType =>
       Seq(true, false).foreach { isNewTable =>
         test(s"allowed partition column types: $dataType, new table = $isNewTable") {
--- a/kernel/kernel-api/src/test/scala/io/delta/kernel/test/TestFixtures.scala
+++ b/kernel/kernel-api/src/test/scala/io/delta/kernel/test/TestFixtures.scala
@@ -30,7 +30,7 @@ import io.delta.kernel.types.{ArrayType, BinaryType, BooleanType, ByteType, Data
 trait TestFixtures extends ActionUtils {
 
   /** All simple data type used in parameterized tests where type is one of the test dimensions. */
-  val SIMPLE_TYPES = Seq(
+  val PRIMITIVE_TYPES = Set(
     BooleanType.BOOLEAN,
     ByteType.BYTE,
     ShortType.SHORT,
@@ -45,13 +45,13 @@ trait TestFixtures extends ActionUtils {
     BinaryType.BINARY,
     new DecimalType(10, 5))
 
-  val COMPLEX_TYPES: Set[DataType] = Set(
+  val NESTED_TYPES: Set[DataType] = Set(
     new ArrayType(BooleanType.BOOLEAN, true),
     new MapType(IntegerType.INTEGER, LongType.LONG, true),
     new StructType().add("s1", BooleanType.BOOLEAN).add("s2", IntegerType.INTEGER))
 
   /** All types. Used in parameterized tests where type is one of the test dimensions. */
-  val ALL_TYPES = SIMPLE_TYPES ++ COMPLEX_TYPES
+  val ALL_TYPES: Set[DataType] = PRIMITIVE_TYPES ++ NESTED_TYPES
 
   def createCommitMetadata(
       version: Long,
--- a/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/IcebergWriterCompatV1Suite.scala
+++ b/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/IcebergWriterCompatV1Suite.scala
@@ -743,7 +743,7 @@ trait IcebergWriterCompatV1SuiteBase
   // We cannot test enabling icebergCompatV1 since it is not a table feature in Kernel; This is
   // tested in the unit tests in IcebergWriterCompatV1MetadataValidatorAndUpdaterSuite
 
-  (SIMPLE_TYPES ++ COMPLEX_TYPES)
+  (PRIMITIVE_TYPES ++ NESTED_TYPES)
     // filter out the types unsupported by icebergWriterCompatV1
     .filter(dataType => dataType != ByteType.BYTE && dataType != ShortType.SHORT)
     .foreach { dataType: DataType =>
--- a/kernel/kernel-api/src/test/scala/io/delta/kernel/internal/icebergcompat/IcebergCompatV2MetadataValidatorAndUpdaterSuite.scala
+++ b/kernel/kernel-api/src/test/scala/io/delta/kernel/internal/icebergcompat/IcebergCompatV2MetadataValidatorAndUpdaterSuite.scala
@@ -33,11 +33,11 @@ trait IcebergCompatV2MetadataValidatorAndUpdaterSuiteBase
 
   override def icebergCompatVersion: String = "V2"
 
-  override def supportedDataColumnTypes: Set[DataType] = ALL_TYPES.toSet
+  override def supportedDataColumnTypes: Set[DataType] = ALL_TYPES
 
   override def unsupportedDataColumnTypes: Set[DataType] = Set(VariantType.VARIANT)
 
-  override def unsupportedPartitionColumnTypes: Set[DataType] = COMPLEX_TYPES.toSet
+  override def unsupportedPartitionColumnTypes: Set[DataType] = NESTED_TYPES
 
   override def isDeletionVectorsSupported: Boolean = false
 
--- a/kernel/kernel-api/src/test/scala/io/delta/kernel/internal/icebergcompat/IcebergCompatV3MetadataValidatorAndUpdateSuite.scala
+++ b/kernel/kernel-api/src/test/scala/io/delta/kernel/internal/icebergcompat/IcebergCompatV3MetadataValidatorAndUpdateSuite.scala
@@ -36,11 +36,11 @@ trait IcebergCompatV3MetadataValidatorAndUpdaterSuiteBase
 
   override def icebergCompatVersion: String = "V3"
 
-  override def supportedDataColumnTypes: Set[DataType] = ALL_TYPES.toSet
+  override def supportedDataColumnTypes: Set[DataType] = ALL_TYPES
 
   override def unsupportedDataColumnTypes: Set[DataType] = Set(VariantType.VARIANT)
 
-  override def unsupportedPartitionColumnTypes: Set[DataType] = COMPLEX_TYPES.toSet
+  override def unsupportedPartitionColumnTypes: Set[DataType] = NESTED_TYPES
 
   override def isDeletionVectorsSupported: Boolean = true
 
--- a/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaIcebergCompatV2Suite.scala
+++ b/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaIcebergCompatV2Suite.scala
@@ -45,10 +45,9 @@ trait DeltaIcebergCompatV2SuiteBase extends DeltaIcebergCompatBaseSuite {
     TableFeatures.ICEBERG_COMPAT_V2_W_FEATURE,
     TableFeatures.COLUMN_MAPPING_RW_FEATURE)
 
-  override def supportedDataColumnTypes: Seq[DataType] =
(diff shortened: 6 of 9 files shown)


kernel/kernel-api/src/test/scala/io/delta/kernel/internal/icebergcompat/IcebergCompatMetadataValidatorAndUpdaterSuiteBase.scala
```diff
@@ -219,27 +222,3 @@ trait IcebergCompatMetadataValidatorAndUpdaterSuiteBase extends AnyFunSuite
     }
   }
 }
-
-object IcebergCompatMetadataValidatorAndUpdaterSuiteBase {
```
scottsand-db, line 223: This was just copied from TestUtils.

TestUtils is in kernel-defaults. This is in kernel-api. So we could not import that. I moved these types to TestFixtures.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/utils/WriteUtils.scala
```diff
@@ -68,9 +69,9 @@ trait AbstractWriteUtils extends TestUtils with TransactionBuilderSupport {
   val testEngineInfo = "test-engine"
 
   /** Test table schemas and test */
-  val testSchema = new StructType().add("id", INTEGER)
-  val dataBatches1 = generateData(testSchema, Seq.empty, Map.empty, 200, 3)
-  val dataBatches2 = generateData(testSchema, Seq.empty, Map.empty, 400, 5)
+  lazy val testSchema = new StructType().add("id", INTEGER)
```
scottsand-db: Required for some scala trait initialization order magic shenanigans

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/IcebergWriterCompatV1Suite.scala
```diff
@@ -243,11 +265,19 @@ class IcebergWriterCompatV1Suite extends AnyFunSuite with WriteUtils
                 .build())
         }
         val e = intercept[KernelException] {
-          updateTableMetadata(
```
scottsand-db: This was actually just wrong for createTable

kernel/kernel-api/src/test/scala/io/delta/kernel/test/TestFixtures.scala
```diff
@@ -21,12 +21,38 @@ import java.util.Optional
...
+import io.delta.kernel.types.{ArrayType, BinaryType, BooleanType, ByteType, DataType, DateType, DecimalType, DoubleType, FloatType, IntegerType, LongType, MapType, ShortType, StringType, StructType, TimestampNTZType, TimestampType}
 
 /**
  * Test fixtures including factory methods and constants for creating test objects with sensible
  * defaults.
  */
 trait TestFixtures extends ActionUtils {
+
+  /** All simple data type used in parameterized tests where type is one of the test dimensions. */
+  val SIMPLE_TYPES = Seq(
```
scottsand-db: Copied from TestUtils

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/utils/TransactionBuilderSupport.scala
```diff
@@ -117,8 +123,12 @@ trait TransactionBuilderV1Support extends TransactionBuilderSupport with TestUti
       clusteringColsOpt: Option[List[Column]] = None,
       logCompactionInterval: Int = 10,
       txnId: Option[(String, Long)] = None,
-      tablePropertiesRemoved: Set[String] = null): Transaction = {
+      tablePropertiesRemoved: Set[String] = null,
+      committer: Option[Committer] = None): Transaction = {
     // scalastyle:on argcount
+    committer.foreach(_ =>
```
scottsand-db: use `{`

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/IcebergWriterCompatV1Suite.scala
```diff
@@ -243,11 +265,19 @@ class IcebergWriterCompatV1Suite extends AnyFunSuite with WriteUtils
...
-            engine,
-            tablePath,
-            schema = schemaToCommit,
-            tableProperties = tblPropertiesIcebergWriterCompatV1Enabled)
+          if (isNewTable) {
```
scottsand-db: I think we can just do

```scala
          appendData(
            engine,
            tablePath,
            isNewTable = isNewTable,
            schema = schemaToCommit,
            data = Seq.empty,
            tableProperties = tblPropertiesIcebergWriterCompatV1Enabled)
        }
```

allisonport-db: Merged the other PR that does this

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/IcebergWriterCompatV1Suite.scala
```diff
@@ -580,26 +610,38 @@ class IcebergWriterCompatV1Suite extends AnyFunSuite with WriteUtils
...
     withTempDirAndEngine { (tablePath, engine) =>
       getCreateTxn(
         engine,
         tablePath,
         schema,
-        tableProperties = tblProperties,
-        withDomainMetadataSupported = true)
+        tableProperties = allTblProperties,
+        withDomainMetadataSupported = true,
+        committer = getCatalogManagedCommitter)
```
allisonport-db: Can you clarify to me why we need to set the committer here for this test? Do we just want to ensure that _we can_ or does this in some way not work if we don't?

allisonport-db: My thinking is, in general, it's simpler to put the tests for things like catalogManaged that don't make sense for the legacy APIs, in that concrete child suite, instead of having to factor out extra functionality in the base suite. WDYT?

allisonport-db: I guess along those lines, my question would be, instead of all the changes throughout this file, should we instead just add a simple test that you can use catalogManaged with icebergWriterCompatV1 for the CREATE case in `IcebergWriterCompatV1WithNewAPISuite`?

allisonport-db (commented):
Just a testing comment

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/IcebergWriterCompatV1Suite.scala
allisonport-db: Depending on your thoughts, I would also remove the committer from the shared fxns in `TransactionBuilderSupport`. Shouldn't suites that support both the V1 and V2 builders not be setting committers ever?

scottsand-db: Great questions.

> Can you clarify to me why we need to set the committer here for this test? 

Anytime we want to write into a table that has the catalogManaged table feature enabled on it, we must specify a committer. We cannot use the default-path-based-table-only-committer.

> Do we just want to ensure that we can or does this in some way not work if we don't?

It won't work if we do not provide our own committer. The default committer will identify it is trying to commit into a catalogManaged table and throw!

------------------------

Your other questions are basically: Why not put this into a new suite. Totally fair. I can do that

scottsand-db: @allisonport-db -- Actually, sorry. Your revent PR already has this suite run with both the old and new APIs, right?

So what value is there in putting this in its own suite? Like, we want to test that all of the expected table features are compatible with Iceberg writer compat v1. So that includes catalogManaged. And we want to use BOTH old and new APIs. So we will have to use `getCreateTxn`. And we will have to be able to inject the right committer.

Like, we will still need to pass through the committer to our transaction builder API test abstractions (for V2 only).

The only reason I could think of to move this to a new suite is that the 

```scala
  protected def supportsCatalogManaged: Boolean
  protected def getCatalogManagedCommitter: Option[Committer]
```

methods wouldn't be in the parent suite? but that doesn't seem so bad? wdyt?

allisonport-db: I thought we don't want to test catalogManaged + icebergWriterCompatV1 for the old APIs? Since a committer is required and that shouldn't be supported through the old APIs.

scottsand-db: Ah, I see the disconnect.

You want _just_ the catalogManaged and icebergWriterCompatV1 to be its own test. i.e. leave all the other table features here in this test.

Sorry maybe I just missed that. Or maybe I knew that yesterday and forgot. Ha. That makes sense. Will do.

allisonport-db: Yep! I also think we can add `createTxn(committer=...)` to the TransactionBuilderV2Support trait as a helper, but it should not be in the abstract parent class, since any suites that support both old and new APIs shouldn't be setting a committer.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/IcebergWriterCompatV1Suite.scala
```diff
@@ -66,6 +85,23 @@ trait IcebergWriterCompatV1SuiteBase extends AnyFunSuite with AbstractWriteUtils
...
+}
+
+trait IcebergWriterCompatV1SuiteBase
+    extends AnyFunSuite
+    with AbstractWriteUtils
+    with TestFixtures
+    with IcebergWriterCompatV1TestUtils
+    with ColumnMappingSuiteBase {
+
+  private val tblPropertiesIcebergWriterCompatV1Enabled = Map(
```
scottsand-db, line 96: these aren't added by my PR -- just showing up as added due to the way a new trait and suite were inserted above

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/IcebergWriterCompatV1Suite.scala
```diff
@@ -40,18 +40,36 @@ class IcebergWriterCompatV1TransactionBuilderV1Suite extends IcebergWriterCompat
...
-  private val tblPropertiesIcebergWriterCompatV1Enabled = Map(
-    TableConfig.ICEBERG_WRITER_COMPAT_V1_ENABLED.getKey -> "true")
+class CatalogManagedWithIcebergWriterCompatV1Suite
+    extends AnyFunSuite
+    with WriteUtilsWithV2Builders
+    with IcebergWriterCompatV1TestUtils {
 
-  private val tblPropertiesIcebergCompatV2Enabled = Map(
-    TableConfig.ICEBERG_COMPAT_V2_ENABLED.getKey -> "true")
+  test("can create a catalogManaged table with icebergWriterCompatV1") {
```
scottsand-db, line 48: Was asked in a previous PR review to move this to a separate test/class -- so as not to muddy the other tests/trait with having to plumb through the committer through all of our test utils 👍

kernel/kernel-api/src/test/scala/io/delta/kernel/test/TestFixtures.scala
huan233usc: How hard it is to rename to primitive types and nested type?

scottsand-db: SimpleTypes -> 8 usages. ComplexTypes --> 6 usages.

huan233usc: Thanks, we could do the renaming in a separate pr(as it is an orthogonal change), if it make senses.

nicklan: Why make this a `Seq` and then have to convert it to a `Set` above, rather than just make it a `Set`?

scottsand-db: I just copied + pasted the existing code from TestUtils, which used `Seq`.

Will implement this change (and I don't _think_ I will have to upadate many/any other code sites, because the `Set` should be interchangable with the `Seq` ...)

scottsand-db: https://github.com/delta-io/delta/pull/5176/commits/744af186de4a34c3e0867945df45b036c7296342

PR #5183, opened 2025-09-09
[Kernel][Test-only] Run more read test suites with both legacy Table Snapshot APIs and new TableManager APIs

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Runs additional test suites with both the legacy Table Snapshot APIs and the new TableManager APIs to increase test coverage. Previously these suites were run using only the legacy APIs.

## How was this patch tested?

Test-only change

No


kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/utils/TestUtils.scala
allisonport-db, line 1: Previously I had factored out these changes into this PR https://github.com/delta-io/delta/pull/5181 but i think it's small enough it is not necessary

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/utils/DefaultEngineOfBatchSize2.scala
```diff
@@ -0,0 +1,31 @@
...
+ */
+package io.delta.kernel.defaults.utils
+
+import io.delta.kernel.defaults.engine.DefaultEngine
+
+import org.apache.hadoop.conf.Configuration
+
+trait DefaultEngineOfBatchSize2 extends AbstractTestUtils {
+
+  override lazy val defaultEngine = DefaultEngine.create(new Configuration() {
```
scottsand-db: IMO this single value doesn't warrant a whole new trait. e.g. I don't think any other values would go in this trait, right? seems very single-purpose?

Could we not just define `val engineWithBatchSize2` inside of `AbstractTestUtils` ? and then explicitly use that inside of the required tests?

scottsand-db (commented):
Just the 1 comment/question!

scottsand-db (approved):

PR #5188, opened 2025-09-10
[Kernel][Test-only] Run DomainMetadataSuite with the new and legacy APIs for both read/write loads

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Refactors DomainMetadataSuite to run with both
- New APIs: load snapshots using TableManager and build transactions using V2 builders
- Legacy APIs: load snapshots using Table and build transactions with the V1 builder

This is done to increase test coverage.

## How was this patch tested?

Test change.

No.

### DIFF
--- a/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DomainMetadataSuite.scala
+++ b/kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DomainMetadataSuite.scala
@@ -91,17 +91,14 @@ trait AbstractDomainMetadataSuite extends AnyFunSuite with AbstractWriteUtils
       tablePath: String,
       domainMetadatas: Seq[DomainMetadata],
       expectedValue: Map[String, DomainMetadata],
-      useInternalApi: Boolean = false,
-      enableDomainMetadata: Boolean = false): Unit = {
+      useInternalApi: Boolean = false): Unit = {
     // Create the transaction with domain metadata and commit
 
-    // I think this can actually just not override enableDomainMetadata & remove the arg?...
     val txn = createTxnWithDomainMetadatas(
       engine,
       tablePath,
       domainMetadatas,
-      useInternalApi,
-      enableDomainMetadata = enableDomainMetadata)
+      useInternalApi)
     commitTransaction(txn, engine, emptyIterable())
 
     // Verify the final state includes the expected domain metadata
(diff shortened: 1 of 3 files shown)


kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/utils/WriteUtils.scala
```diff
@@ -50,11 +50,19 @@ import org.apache.commons.io.FileUtils
...
 
-/** Write utilities that use the V2 transaction builders to create transactions */
+/**
+ * DO NOT MODIFY this trait -- this is just syntactic sugar to clearly indicate we are extending the
+ * "default" WriteUtils which happens to use the legacy Kernel APIs
+ */
+trait WriteUtilsWithV1Builders extends WriteUtils
+
+/** Write utilities that use the V2 transaction builders to create transactions and TableManager snapshot APIs */
 trait WriteUtilsWithV2Builders extends AbstractWriteUtils with TransactionBuilderV2Support
```
allisonport-db, line 67: Note: this change does mean that all other existing write suites that extend `WriteUtilsWithV2Builders` will now run with `TestUtilsWithTableManagerAPIs`. However these suites have not been refactored to use the `getTableManagerAdapter` method so this does not mean they've been fully refactored to use the new read APIs (probably some mix of it, helpers within test utilities may use the new APIs, but the suites may also manually load snapshots)

I think this is a fine state for now!

raveeram-db (approved):
Awesome, thank you @allisonport-db and @scottsand-db for the extended test coverage here

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DomainMetadataSuite.scala
```diff
@@ -281,13 +300,12 @@ class DomainMetadataSuite extends AnyFunSuite with WriteUtils with ParquetSuiteB
       }
 
       // Checkpoint the table
-      val latestVersion = table.getLatestSnapshot(engine).getVersion()
-      table.checkpoint(engine, latestVersion)
+      val latestVersion = getTableManagerAdapter.getSnapshotAtLatest(engine, tablePath).getVersion()
+      Table.forPath(engine, tablePath).checkpoint(engine, latestVersion)
```
scottsand-db, line 304: Can we track this work that we need to add `Snapshot::checkpoint` and also eventually then upgrade our test abstractions, too?

scottsand-db (approved):

PR #5193, opened 2025-09-10
[Kernel][Test-only] Run InCommitTimestampSuite using both the legacy and new APIs for snapshot loading and txn building

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Increases test coverage

## How was this patch tested?

Test only change

No


kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/InCommitTimestampSuite.scala
```diff
@@ -42,8 +42,23 @@ import io.delta.kernel.utils.FileStatus
...
+ * Runs in-commit timestamp tests using the legacy Table snapshot APIs and V1 transaction builders
+ */
+class LegacyInCommitTimestampSuite extends AbstractInCommitTimestampSuite
+    with WriteUtilsWithV1Builders
+
+trait AbstractInCommitTimestampSuite extends AnyFunSuite {
+  self: AbstractWriteUtils =>
+
+  private def getLogPath(engine: Engine, tablePath: String): Path = {
+    new Path(Table.forPath(engine, tablePath).getPath(engine), "_delta_log")
```
scottsand-db: you can't just do `new Path(s"$tablePath/_delta_log")` ?

scottsand-db: if so -- can we delete this helper method and just do that inline?

and if we can't do it, that's probably because we need the path to be resolved by the engine? can we just use that engine API? instead of relying on a `Table.forPath` API that might get deprecated and deleted?

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/utils/TransactionBuilderSupport.scala
```diff
@@ -196,6 +197,7 @@ trait TransactionBuilderV2Support extends TransactionBuilderSupport with TestUti
       tablePath,
       schema,
       "test-engine")
+      .asInstanceOf[CreateTableTransactionBuilderImpl].withClock(clock)
```
scottsand-db: nit IMO `.withClock(clock)` being on a new line would more-so match the existing style

PR #5194, opened 2025-09-11
[Kernel][Test-only] Refactor CommitIcebergActionSuite and TransactionCommitLoopSuite to run with both the v1 and v2 txn builders

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

See title. Increases test coverage.

## How was this patch tested?
Test change

no


kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/TransactionCommitLoopSuite.scala
```diff
@@ -34,16 +34,18 @@ import io.delta.kernel.utils.CloseableIterator
 import org.apache.hadoop.conf.Configuration
 import org.scalatest.funsuite.AnyFunSuite
 
-class TransactionCommitLoopSuite extends AnyFunSuite with WriteUtils {
+class TransactionCommitLoopTransactionBuilderV1Suite extends AbstractTransactionCommitLoopSuite
+    with WriteUtilsWithV1Builders {}
 
-  // TODO: Refactor this test suite to use both Table.forPath().getLatestSnapshot() and
```
allisonport-db, line 39: This suite loads no snapshots :) so this is not needed

PR #5195, opened 2025-09-11
[Kernel] Refactor validation of schemas for new tables in TransactionMetadataFactory

amogh-jahagirdar:
…MetadataFactory

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This change is just a refactoring for validations that we perform on schemas for new tables (create or replace).
Currently we do additional validations on the user provided schema for create or replace but
down the line we want validations of schemas to incorporate any protocol upgrades. 

This refactoring to do validation post upgrades enables us to extract table features post upgrade and pass that through to validations.  An example of this would be to extract variant table features and pass that through to schema validations.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
@@ -53,7 +53,18 @@ public class SchemaUtils {
    */
   public static void validateSchema(StructType schema, boolean isColumnMappingEnabled) {
     checkArgument(schema.length() > 0, "Schema should contain at least one column");
+    validateColumnNames(schema, isColumnMappingEnabled);
+    validateSupportedType(schema);
+  }
 
+  /**
+   * Validate the column names in the schema (uniqueness, whether the column names can be
+   * represented in Parquet)
+   *
+   * @param schema
+   * @param isColumnMappingEnabled
+   */
+  public static void validateColumnNames(StructType schema, boolean isColumnMappingEnabled) {
     List<String> flattenColNames =
         new SchemaIterable(schema)
             .stream()
@@ -90,8 +101,6 @@ public class SchemaUtils {
             }
           });
     }
-
-    validateSupportedType(schema);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


allisonport-db:
@amogh-jahagirdar seems like some tests are failing related to the re-ordering; I think some other checks rely on the initial schema validation having happened earlier. Just took a glance at a failing job though haven't taken a closer look.

amogh-jahagirdar:
@allisonport-db Yeah I'm looking into it. Many of these  failing cases are expectations on specific error messages but now we're still failing at a later point with slightly different error messages so it mismatches in the test expectations, I'm going through case by case and seeing what makes sense...

allisonport-db:
@amogh-jahagirdar maybe it also makes sense to split up the check? like very basic schema validation as early as possible, and then more detailed validation later on

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -53,7 +53,18 @@ private SchemaUtils() {}
    */
   public static void validateSchema(StructType schema, boolean isColumnMappingEnabled) {
```
amogh-jahagirdar, line 54: I added a ToDo at the relevant call point, but we'd plumb through the set of table features in `validateSchema`. I refactored it to separate validating the column names and validating the supported types, where validating supported types will need to validate considering the table features.

This separation also allows us to just do column name validation on the initial metadata passed in to TransactionMetadataFactory which is what we care about pre-protocol upgrade. Post protocol upgrade we'd do another pass over the schema but also validate the supported types with any new features which are enabled.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/IcebergWriterCompatV1Suite.scala
```diff
@@ -479,8 +479,8 @@ trait IcebergWriterCompatV1SuiteBase extends AnyFunSuite with AbstractWriteUtils
       .add("c1", IntegerType.INTEGER)
       .add("c2", VariantType.VARIANT),
     testOnExistingTable = false, // we don't currently support schema updates
-    // We throw an error earlier for variant for some reason
```
amogh-jahagirdar, line 482: For this case based on the comment it looks like it was unexpected to throw the error that existed previously. The new error is at a different point in table feature validation but it effectively conveys the same failure to a client.

PR #5197, opened 2025-09-11
[Kernel][Test-only] Refactor TransactionReportSuite to run with both the V1 and V2 transaction builders

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

Increases test coverage.

## How was this patch tested?

Test-only change.

No.


kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/metrics/TransactionReportSuite.scala
```diff
@@ -121,9 +128,7 @@ class TransactionReportSuite extends AnyFunSuite with WriteUtils
       expectedTotalAddFilesSizeInBytes: Long = 0,
       expectedTotalRemoveFilesSizeInBytes: Long = 0,
       expectedFileSizeHistogramResult: Option[FileSizeHistogramResult] = None,
-      buildTransaction: (TransactionBuilder, Engine) => Transaction = (tb, e) => tb.build(e),
-      engineInfo: String = "test-engine-info",
```
allisonport-db, line 125: I got rid of this param because it didn't seem worth it to add it to all the methods in TransactionBuilderSupport and I think checking the engineInfo is set to this test value is sufficient

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/metrics/TransactionReportSuite.scala
```diff
@@ -625,21 +616,19 @@ class TransactionReportSuite extends AnyFunSuite with WriteUtils
           new Column(Array[String]("nested", "nestedName"))),
         expectedNumAddFiles = 0,
         expectedNumRemoveFiles = 0,
+        getTransaction = (engine) =>
```
raveeram-db, line 619: Very nice abstractions!

PR #5205, opened 2025-09-15
[Kernel] Support reading/writing variant stats in JSON

amogh-jahagirdar:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This change adds support for reading and writing variant stats in the Delta Log JSON.  
Details for the structure are in this protocol RFC: https://github.com/delta-io/delta/blob/master/protocol_rfcs/variant-shredding.md#statistics-for-variant-columns

These stats are themselves variants (metadata + values concatenated) and then z85 encoded into a String. 

-->

## How was this patch tested?
Added unit tests

No because the ability to even create tables with variant is not supported yet, these are the underlying primitives....

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/JsonUtils.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/JsonUtils.java
@@ -254,6 +254,13 @@ public class JsonUtils {
         long epochMicros = ChronoUnit.MICROS.between(EPOCH.toLocalDateTime(), localDateTime);
         return Literal.ofTimestampNtz(epochMicros);
 
+      } else if (dataType instanceof VariantType) {
+        if (!valueNode.isTextual()) {
+          throw new KernelException(
+              String.format("Expected variant as string value but got: %s", valueNode));
+        }
+        String textValue = valueNode.asText();
+        return Literal.ofString(textValue);
       } else {
         throw unsupportedStatsDataType(dataType);
       }
--- a/kernel/kernel-api/src/test/scala/io/delta/kernel/internal/util/DataFileStatisticsSuite.scala
+++ b/kernel/kernel-api/src/test/scala/io/delta/kernel/internal/util/DataFileStatisticsSuite.scala
@@ -46,6 +46,7 @@ class DataFileStatisticsSuite extends AnyFunSuite with Matchers {
     val nestedStructType = new StructType()
       .add("aa", StringType.STRING)
       .add("ac", new StructType().add("aca", IntegerType.INTEGER))
+      .add("nested_variant", VariantType.VARIANT)
 
     val schema = new StructType()
       .add("ByteType", ByteType.BYTE)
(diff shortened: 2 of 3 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/JsonUtils.java
```diff
@@ -254,6 +254,13 @@ public static Literal parseJsonValueToLiteral(JsonNode valueNode, DataType dataT
         long epochMicros = ChronoUnit.MICROS.between(EPOCH.toLocalDateTime(), localDateTime);
         return Literal.ofTimestampNtz(epochMicros);
 
+      } else if (dataType instanceof VariantType) {
+        if (!valueNode.isTextual()) {
+          throw new KernelException(
+              String.format("Expected variant as string value but got: %s", valueNode));
+        }
+        String textValue = valueNode.asText();
+        return Literal.ofString(textValue);
```
amogh-jahagirdar, line 263: This intentionally just reads the z85 encoding directly into the literal. We do have base85 decoding in the project but we don't really have the in-memory variant structures to contain the relevant buffers yet. Parquet is defining those in the new version and those are also defined in Iceberg as well. 

Even so, from a stats reading perspective, I don't think there's much that we get out of validating that the string is valid z85 and then attempting to read that into variant structures in a literal. When the stats for a variant are actually needed during filtering then we would lazily evaluate those and at that point we would fail since we wouldn't be able to create the variant.

I think this also aligns with the intent of this class since the comment says "with basic physical-type validation" which we are validating that it's just a string. Lastly, these are all just the in-memory representations, we can always change these if we don't like the API

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/util/DataFileStatisticsSuite.scala
```diff
@@ -76,7 +77,9 @@ class DataFileStatisticsSuite extends AnyFunSuite with Matchers {
       new Column("TimestampNTZType") -> Literal.ofTimestampNtz(1L),
       new Column("BinaryType") -> Literal.ofBinary("a".getBytes),
       new Column(Array("NestedStruct", "aa")) -> Literal.ofString("a"),
-      new Column(Array("NestedStruct", "ac", "aca")) -> Literal.ofInt(1)).asJava
+      new Column(Array("NestedStruct", "ac", "aca")) -> Literal.ofInt(1),
```
vkorukanti, line 81: case where variant is inside the nestedStructType. You can actually separate out the primitive schema and use it the topLevel (primitiveSchema), struct(primitiveSchema)

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/util/DataFileStatisticsSuite.scala
```diff
@@ -958,4 +974,24 @@ class DataFileStatisticsSuite extends AnyFunSuite with Matchers {
     assert(resultFromEmpty.getTightBounds.isPresent &&
       !resultFromEmpty.getTightBounds.get)
   }
+
+  test("deserializing invalid variant stats throws KernelException") {
```
vkorukanti, line 987: Is there an integration test that we could write to test this on the actual production table use case path?

vkorukanti (approved):
Looks great! couple of minor comments about the test.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/util/DataFileStatisticsSuite.scala
amogh-jahagirdar: No integration tests through the current OSS Spark + Delta integration can be written yet because the parquet write logic to write shredded values, and the logic to collect min/max stats, and plumb that through hasn't been implemented yet

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/util/DataFileStatisticsSuite.scala
amogh-jahagirdar: I went ahead and added a test to test variant inside the nestedStructType. I didn't do the additional suggested changes about separating the primitive schema since it ends up being a larger diff with changes to some of the existing test logic which I'd prefer to avoid just to keep this change small.

PR #5206, opened 2025-09-15
[Kernel connector]introduce expression convertion utils for dsv2 connector

huan233usc:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Introduce utils to convert spark's `Filter` to kernel's `Predicate`

## How was this patch tested?

unit test

No

### DIFF
--- a/kernel-spark/src/main/java/io/delta/kernel/spark/utils/ExpressionUtils.java
+++ b/kernel-spark/src/main/java/io/delta/kernel/spark/utils/ExpressionUtils.java
@@ -143,7 +143,8 @@ public final class ExpressionUtils {
       Not f = (Not) filter;
       // NOT disables partial pushdown for semantic correctness.
       // Example: NOT(A AND B) requires both A and B to be convertible.
-      // We cannot convert it to just NOT A if only A is convertible, because:
+      // We cannot convert it to just return NOT A if only A is convertible when B is not,
+      // because:
       //
       // Original: NOT(age < 30 AND name = "John")
       //
--- a/kernel-spark/src/main/java/io/delta/kernel/spark/utils/ExpressionUtils.java
+++ b/kernel-spark/src/main/java/io/delta/kernel/spark/utils/ExpressionUtils.java
@@ -58,7 +58,7 @@ public final class ExpressionUtils {
    *     supported
    */
   public static Optional<Predicate> convertSparkFilterToKernelPredicate(Filter filter) {
-    return convertSparkFilterToKernelPredicate(filter, true);
+    return convertSparkFilterToKernelPredicate(filter, true /*canPartialPushDown*/);
   }
 
   /**
@@ -73,44 +73,44 @@ public final class ExpressionUtils {
     if (filter instanceof EqualTo) {
       EqualTo f = (EqualTo) filter;
       return convertValueToKernelLiteral(f.value())
-          .map(l -> new Predicate("=", column(f.attribute()), l));
+          .map(l -> new Predicate("=", kernelColumn(f.attribute()), l));
     }
     if (filter instanceof EqualNullSafe) {
       EqualNullSafe f = (EqualNullSafe) filter;
       // EqualNullSafe with null value should be translated to IS_NULL
       // For non-null values, we use "=" operator.
       return f.value() == null
-          ? Optional.of(new Predicate("IS_NULL", column(f.attribute())))
+          ? Optional.of(new Predicate("IS_NULL", kernelColumn(f.attribute())))
           : convertValueToKernelLiteral(f.value())
-              .map(l -> new Predicate("=", column(f.attribute()), l));
+              .map(l -> new Predicate("=", kernelColumn(f.attribute()), l));
     }
     if (filter instanceof GreaterThan) {
       GreaterThan f = (GreaterThan) filter;
       return convertValueToKernelLiteral(f.value())
-          .map(l -> new Predicate(">", column(f.attribute()), l));
+          .map(l -> new Predicate(">", kernelColumn(f.attribute()), l));
     }
     if (filter instanceof GreaterThanOrEqual) {
       GreaterThanOrEqual f = (GreaterThanOrEqual) filter;
       return convertValueToKernelLiteral(f.value())
-          .map(l -> new Predicate(">=", column(f.attribute()), l));
+          .map(l -> new Predicate(">=", kernelColumn(f.attribute()), l));
     }
     if (filter instanceof LessThan) {
       LessThan f = (LessThan) filter;
       return convertValueToKernelLiteral(f.value())
-          .map(l -> new Predicate("<", column(f.attribute()), l));
+          .map(l -> new Predicate("<", kernelColumn(f.attribute()), l));
     }
     if (filter instanceof LessThanOrEqual) {
       LessThanOrEqual f = (LessThanOrEqual) filter;
       return convertValueToKernelLiteral(f.value())
-          .map(l -> new Predicate("<=", column(f.attribute()), l));
+          .map(l -> new Predicate("<=", kernelColumn(f.attribute()), l));
     }
     if (filter instanceof IsNull) {
       IsNull f = (IsNull) filter;
-      return Optional.of(new Predicate("IS_NULL", column(f.attribute())));
+      return Optional.of(new Predicate("IS_NULL", kernelColumn(f.attribute())));
     }
     if (filter instanceof IsNotNull) {
       IsNotNull f = (IsNotNull) filter;
-      return Optional.of(new Predicate("IS_NOT_NULL", column(f.attribute())));
+      return Optional.of(new Predicate("IS_NOT_NULL", kernelColumn(f.attribute())));
     }
     if (filter instanceof org.apache.spark.sql.sources.And) {
       org.apache.spark.sql.sources.And f = (org.apache.spark.sql.sources.And) filter;
@@ -141,9 +141,15 @@ public final class ExpressionUtils {
     }
     if (filter instanceof Not) {
       Not f = (Not) filter;
-      // NOT disables partial pushdown for semantic correctness
-      // NOT (A AND B) cannot become NOT A - they have different meanings
-      Optional<Predicate> child = convertSparkFilterToKernelPredicate(f.child(), false);
+      // NOT disables partial pushdown for semantic correctness.
+      // Example: NOT(A AND B) requires both A and B to be convertible.
+      // We cannot convert it to just NOT A if only A is convertible, because:
+      // - NOT(A AND B) = (NOT A) OR (NOT B) by De Morgan's law
+      // - If we only push down NOT A, we lose information about condition B
+      // - This could incorrectly filter data: some rows that should pass the full
+      //   NOT(A AND B) filter might be wrongly excluded based on NOT A alone
+      Optional<Predicate> child =
+          convertSparkFilterToKernelPredicate(f.child(), false /*canPartialPushDown*/);
       return child.map(c -> new Predicate("NOT", c));
     }
 
@@ -154,12 +160,20 @@ public final class ExpressionUtils {
    * Creates a Delta Kernel Column from a Spark SQL column attribute name.
    *
    * <p>This method handles nested column references (e.g., "user.profile.name") by parsing the
-   * dot-separated path into an array of field names.
+   * dot-separated path into an array of field names using Spark's column path parser.
+   *
+   * <p>If a column name contains literal dots that should not be treated as field separators, it
+   * must be properly quoted/escaped in the original Spark SQL. For example:
+   *
+   * <ul>
+   *   <li>{@code `my.column.with.dots`} - treats the entire string as a single column name
+   *   <li>{@code my.nested.field} - treats this as nested field access: my -> nested -> field
+   * </ul>
    *
    * @param attribute the column attribute name, potentially dot-separated for nested fields
-   * @return Delta Kernel Column object
+   * @return Delta Kernel Column object representing the parsed column path
    */
-  private static Column column(String attribute) {
+  private static Column kernelColumn(String attribute) {
     scala.collection.Seq<String> seq = parseColumnPath(attribute);
     String[] parts = CollectionConverters.asJavaCollection(seq).toArray(new String[0]);
     return new Column(parts);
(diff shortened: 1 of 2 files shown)


spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/utils/ExpressionUtils.java
```diff
@@ -0,0 +1,261 @@
...
+  public static Optional<Predicate> convertSparkFilterToKernelPredicate(Filter filter) {
+    if (filter instanceof EqualTo) {
+      EqualTo f = (EqualTo) filter;
+      return convertValueToKernelLiteral(f.value())
+          .map(l -> new Predicate("=", column(f.attribute()), l));
+    }
+    if (filter instanceof EqualNullSafe) {
+      EqualNullSafe f = (EqualNullSafe) filter;
+      // EqualNullSafe with null value should be translated to IS_NULL
+      // We cannot use "IS NOT DISTINCT FROM" with null because Delta Kernel
```
gengliangwang: This comment doesn't match the code

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/utils/ExpressionUtils.java
```diff
@@ -0,0 +1,261 @@
...
+    }
+    if (filter instanceof IsNotNull) {
+      IsNotNull f = (IsNotNull) filter;
+      return Optional.of(new Predicate("IS_NOT_NULL", column(f.attribute())));
+    }
+    if (filter instanceof org.apache.spark.sql.sources.And) {
+      org.apache.spark.sql.sources.And f = (org.apache.spark.sql.sources.And) filter;
+      Optional<Predicate> left = convertSparkFilterToKernelPredicate(f.left());
+      Optional<Predicate> right = convertSparkFilterToKernelPredicate(f.right());
+      return left.flatMap(l -> right.map(r -> new And(l, r)));
```
gengliangwang: if left is present and right is not, or right is present and left is not, we can still partially push down filter

gengliangwang: And let's add unit test for this.

gengliangwang: The logical should be like
```
  private def convertibleFiltersHelper(
      predicate: sources.Filter,
      canPartialPushDown: Boolean): Option[sources.Filter] = {
    predicate match {
      case sources.And(left, right) =>
        val leftResultOptional = convertibleFiltersHelper(left, canPartialPushDown)
        val rightResultOptional = convertibleFiltersHelper(right, canPartialPushDown)
        (leftResultOptional, rightResultOptional) match {
          case (Some(leftResult), Some(rightResult)) => Some(sources.And(leftResult, rightResult))
          case (Some(leftResult), None) if canPartialPushDown => Some(leftResult)
          case (None, Some(rightResult)) if canPartialPushDown => Some(rightResult)
          case _ => None
        }

      case sources.Or(left, right) =>
        val leftResultOptional = convertibleFiltersHelper(left, canPartialPushDown)
        val rightResultOptional = convertibleFiltersHelper(right, canPartialPushDown)
        if (leftResultOptional.isEmpty || rightResultOptional.isEmpty) {
          None
        } else {
          Some(sources.Or(leftResultOptional.get, rightResultOptional.get))
        }
      case sources.Not(pred) =>
        val resultOptional = convertibleFiltersHelper(pred, canPartialPushDown = false)
        resultOptional.map(sources.Not)

      case other =>
        if (createFilter(other).isDefined) {
          Some(other)
        } else {
          None
        }
    }
  }
```

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/utils/ExpressionUtils.java
```diff
@@ -0,0 +1,261 @@
...
+      return Optional.of(Literal.ofString(s));
+    }
+    if (value instanceof Character) {
+      Character c = (Character) value;
+      return Optional.of(Literal.ofString(c.toString()));
+    }
+    if (value instanceof char[]) {
+      // Convert char array to string
+      char[] charArray = (char[]) value;
+      return Optional.of(Literal.ofString(String.valueOf(charArray)));
```
gengliangwang: QQ: why there are 3 types of string value from Spark filters?

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/utils/ExpressionUtils.java
```diff
@@ -0,0 +1,261 @@
...
+    if (value instanceof LocalDate) {
+      // Convert java.time.LocalDate to days since epoch
+      LocalDate localDate = (LocalDate) value;
+      return Optional.of(Literal.ofDate((int) localDate.toEpochDay()));
+    }
+    if (value instanceof Instant) {
+      // Convert java.time.Instant to microseconds since epoch
+      Instant instant = (Instant) value;
+      long micros = ChronoUnit.MICROS.between(Instant.EPOCH, instant);
+      return Optional.of(Literal.ofTimestamp(micros));
```
gengliangwang: QQ: why there are 2 types of timestamp value from Spark filters?

spark-kernel-dsv2/src/test/java/io/delta/spark/dsv2/utils/ExpressionUtilsTest.java
```diff
@@ -0,0 +1,213 @@
...
+    Optional<Predicate> nonNullResult =
+        ExpressionUtils.convertSparkFilterToKernelPredicate(nonNullFilter);
+
+    assertTrue(nonNullResult.isPresent(), "EqualNullSafe with value should be converted");
+    assertEquals("=", nonNullResult.get().getName());
+    assertEquals(2, nonNullResult.get().getChildren().size());
+  }
+
+  @Test
+  public void testAndFilter() {
```
gengliangwang: QQ: shall we add test case that a filter can't be converted to kernel predicate? Or can only be partially converted to kernel predicate?

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/utils/ExpressionUtils.java
huan233usc: Follow the logic here ->  https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/literals.scala#L74-L76 
Spark could have different representation of string

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/utils/ExpressionUtils.java
huan233usc: Follow here -> https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/literals.scala#L86-L87

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/utils/ExpressionUtils.java
gengliangwang: The input here is the value from Spark's V1 Filter.
There should be exactly one value type for each data type.
See https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/CatalystTypeConverters.scala#L603 for details.

spark-kernel-dsv2/src/main/java/io/delta/spark/dsv2/utils/ExpressionUtils.java
gengliangwang: See comment in https://github.com/delta-io/delta/pull/5206#discussion_r2353363769

kernel-spark/src/main/java/io/delta/kernel/spark/utils/ExpressionUtils.java
```diff
@@ -0,0 +1,259 @@
...
+   *   <li>Null-safe comparison: EqualNullSafe
+   *   <li>Logical operators: And, Or, Not
+   * </ul>
+   *
+   * @param filter the Spark SQL filter to convert
+   * @return Optional containing the converted Kernel predicate, or empty if conversion is not
+   *     supported
+   */
+  public static Optional<Predicate> convertSparkFilterToKernelPredicate(Filter filter) {
+    return convertSparkFilterToKernelPredicate(filter, true);
```
scottsand-db: nit: named boolean param

kernel-spark/src/main/java/io/delta/kernel/spark/utils/ExpressionUtils.java
```diff
@@ -0,0 +1,259 @@
...
+      // OR requires both operands to be convertible for correctness
+      if (!left.isPresent() || !right.isPresent()) {
+        return Optional.empty();
+      }
+      return Optional.of(new Or(left.get(), right.get()));
+    }
+    if (filter instanceof Not) {
+      Not f = (Not) filter;
+      // NOT disables partial pushdown for semantic correctness
+      // NOT (A AND B) cannot become NOT A - they have different meanings
```
scottsand-db: should this say `NOT(A AND B) cannot become NOT A AND NOT B` ? Maybe just explain this to me here in this github comment?

kernel-spark/src/main/java/io/delta/kernel/spark/utils/ExpressionUtils.java
```diff
@@ -0,0 +1,259 @@
...
+  /**
+   * Creates a Delta Kernel Column from a Spark SQL column attribute name.
+   *
+   * <p>This method handles nested column references (e.g., "user.profile.name") by parsing the
+   * dot-separated path into an array of field names.
+   *
+   * @param attribute the column attribute name, potentially dot-separated for nested fields
+   * @return Delta Kernel Column object
+   */
+  private static Column column(String attribute) {
```
scottsand-db: would `kernelColumn` be clearer?

kernel-spark/src/main/java/io/delta/kernel/spark/utils/ExpressionUtils.java
```diff
@@ -0,0 +1,259 @@
...
+      return child.map(c -> new Predicate("NOT", c));
+    }
+
+    return Optional.empty();
+  }
+
+  /**
+   * Creates a Delta Kernel Column from a Spark SQL column attribute name.
+   *
+   * <p>This method handles nested column references (e.g., "user.profile.name") by parsing the
```
scottsand-db, line 173: what happens if a column has a dot inside of it?

kernel-spark/src/main/java/io/delta/kernel/spark/utils/ExpressionUtils.java
huan233usc: NOT(A AND B) is actually NOT A OR  NOT B, added the doc

kernel-spark/src/main/java/io/delta/kernel/spark/utils/ExpressionUtils.java
huan233usc: Add a comment, if column name has doc, it should be passed as "`user.profile.name`"

kernel-spark/src/main/java/io/delta/kernel/spark/utils/ExpressionUtils.java
scottsand-db: Oh yeah. Oop. Thanks

kernel-spark/src/main/java/io/delta/kernel/spark/utils/ExpressionUtils.java
```diff
@@ -0,0 +1,273 @@
...
+      if (!left.isPresent() || !right.isPresent()) {
+        return Optional.empty();
+      }
+      return Optional.of(new Or(left.get(), right.get()));
+    }
+    if (filter instanceof Not) {
+      Not f = (Not) filter;
+      // NOT disables partial pushdown for semantic correctness.
+      // Example: NOT(A AND B) requires both A and B to be convertible.
+      // We cannot convert it to just NOT A if only A is convertible, because:
```
scottsand-db: Nit: Sorry, this paragraph has still taken me > 5 mins to really read and understand with Claude.

`We cannot convert it to just NOT A ` --> I think you need to back up and answer: Why would someone _think_ we **could** convert it to NOT A in the first place?

For example, `NOT(A OR B)` we CAN push down `NOT(A)`.

However, `NOT(A AND B)` we CANNOT push down only `NOT(A)` because that could lead to some incorrect results.

Claude gave me this example, can we udpate these comments to explain this a bit more clearly to new and naive readers?

```
Original: NOT(age < 30 AND name = "John")

Row: age=25, name="John"
(age < 30 AND name = "John") = (true AND true) = true
NOT(true) = false → row should be EXCLUDED

But if we naively push down just NOT(age < 30):

NOT(age < 30) = NOT(true) = false → system excludes row
We lose the chance to evaluate the full condition properly at the Spark level
```

huan233usc: updated comments with example

scottsand-db: Thanks! `// We cannot convert it to just NOT A if only A is convertible` --> what if both A and B are convertible?

huan233usc: Updated, explicitly saids B is not for this case and state that "Pushing down NOT(A AND B) requires both A and B to be convertible."

gengliangwang: @scottsand-db so bacially the logic here is consistent with https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFilters.scala#L600


> what if both A and B are convertible?

In such a case, the predicate can be pushed down, regardless of the value of`canPartialPushDown`

kernel-spark/src/main/java/io/delta/kernel/spark/utils/ExpressionUtils.java
```diff
@@ -0,0 +1,284 @@
...
+      // OR requires both operands to be convertible for correctness
+      if (!left.isPresent() || !right.isPresent()) {
+        return Optional.empty();
+      }
+      return Optional.of(new Or(left.get(), right.get()));
+    }
+    if (filter instanceof Not) {
+      Not f = (Not) filter;
+      // NOT disables partial pushdown for semantic correctness.
+      // Example: Pushing down NOT(A AND B) requires both A and B to be convertible.
```
scottsand-db, line 145: Thank you! Very clear!

PR #5215, opened 2025-09-16
temp

ilicmarkodb:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/skipping/PartitionPredicate.java
```diff
@@ -0,0 +1,22 @@
...
+import io.delta.kernel.expressions.Predicate;
+import io.delta.kernel.types.CollationIdentifier;
+import java.util.List;
+
+/**
+ * A specialized {@link Predicate} used for partition pruning. This class does not add any new
+ * functionality beyond {@link Predicate}, but serves as a marker to indicate that the predicate is
+ * specifically intended for partition pruning operations.
+ */
+public class PartitionPredicate extends Predicate {
```
vkorukanti: why do we need to create a separate class for partition predicate?

kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/internal/expressions/DefaultExpressionEvaluator.java
```diff
@@ -346,6 +407,83 @@ ExpressionTransformResult visitStartsWith(Predicate startsWith) {
       return new ExpressionTransformResult(transformedExpression, BooleanType.BOOLEAN);
     }
 
+    @Override
+    ExpressionTransformResult visitDataSkippingPredicate(
+        DataSkippingPredicate dataSkippingPredicate) {
```
vkorukanti: this is defined as internal class. Can we look at the collation property on the expression?

PR #5201, opened 2025-09-12
Spark Kernel Connector: Fix Column Order and Propagate Scan Options

gengliangwang:
#### Which Delta project/connector is this regarding?

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

* Change the table schema of partitioned table. After changes, the column order is consistent with existing Spark behavior. See https://github.com/delta-io/delta/issues/5200 for more details.
* Propagate scan options to `SparkScanBuilder` and `SparkScan`
* Add unit test for validating `SparkTable`, `SparkScanBuilder` and `SparkScan`

## How was this patch tested?

New UTs
## Does this PR introduce _any_ user-facing changes?

No

PR #5218, opened 2025-09-17
[kernel-spark] Fix table schema column order; add golden table comparison test

gengliangwang:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Follow-up to https://github.com/delta-io/delta/pull/5201.  
Before that PR, the table schema column order was correct. After it, the schema aligned with Spark’s Parquet tables but diverged from Delta tables.  

This PR reverts the affected changes and adds tests to validate consistency between the delta-spark and kernel-spark connectors in both schema and table data.

## How was this patch tested?

Added new tests to compare schema and table data between delta-spark and kernel-spark connectors.

No, the kernel-spark connector is not released yet.


gengliangwang:
cc @huan233usc @raveeram

PR #5221, opened 2025-09-17
Update table feature to indicate writing variant is supported by kern…

amogh-jahagirdar:
…el, update SChemaUtils to consider table features and compare variant

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/tablefeatures/TableFeatures.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/tablefeatures/TableFeatures.java
@@ -220,7 +220,10 @@ public class TableFeatures {
 
     @Override
     public boolean hasKernelWriteSupport(Metadata metadata) {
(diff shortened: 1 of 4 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/tablefeatures/TableFeatures.java
```diff
@@ -220,7 +220,10 @@ private static class VariantTypeTableFeatureBase extends TableFeature.ReaderWrit
 
     @Override
     public boolean hasKernelWriteSupport(Metadata metadata) {
```
amogh-jahagirdar, line 222: I'll break away this change along with another change I had to enable variant when shredding is enabled; for some reason those are treated as independent features but shredding is a super set of  variant support and if it's enabled it implies that variant itself is enabled....

PR #5222, opened 2025-09-17
[Kernel] Enable variant and shredding RW support

amogh-jahagirdar:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This change enables variant and shredding RW support. We still preserve the preview for variant though probably at some point we should remove supporting that in Kernel. Another change that is made is that when shredding is enabled, variant is automatically enabled by making it a required dependent feature. 

Furthermore, since this change is about enabling metadata write support in Kernel we block physical writes when there's a variant in Transaction#transformLogicalData similar to how kernel blocks writes on table for column mapping but allows metadata writes.

## How was this patch tested?

Added unit tests

### DIFF
--- a/kernel/kernel-api/src/test/scala/io/delta/kernel/internal/actions/ProtocolSuite.scala
+++ b/kernel/kernel-api/src/test/scala/io/delta/kernel/internal/actions/ProtocolSuite.scala
@@ -284,6 +284,11 @@ class ProtocolSuite extends AnyFunSuite {
     // expect the dependency features also to be supported
     Set("icebergCompatV2") ->
       (2, 7, Set[String](), Set[String]("icebergCompatV2", "columnMapping")),
+    Set("variantShredding-preview") -> (
+      3,
+      7,
+      Set[String]("variantType", "variantShredding-preview"),
+      Set[String]("variantType", "variantShredding-preview")),
     Set("rowTracking") -> (
       1,
       7,


kernel/kernel-api/src/test/scala/io/delta/kernel/TransactionSuite.scala
```diff
@@ -230,6 +230,22 @@ class TransactionSuite extends AnyFunSuite with VectorTestUtils with MockEngineU
         "Writing into column mapping enabled table is not supported yet."))
     }
   }
+
+  test("transformLogicalData: Writing to tables with variant is blocked") {
+    val txnState = testTxnState(new StructType().add("variant", VariantType.VARIANT))
```
vkorukanti, line 235: also the case where variant type is inside a nested struct

kernel/kernel-api/src/main/java/io/delta/kernel/internal/tablefeatures/TableFeatures.java
```diff
@@ -262,8 +257,8 @@ public boolean metadataRequiresFeatureToBeEnabled(Protocol protocol, Metadata me
     }
 
     @Override
-    public boolean hasKernelWriteSupport(Metadata metadata) {
-      return false; // TODO: yet to be implemented in Kernel
+    public Set<TableFeature> requiredFeatures() {
+      return new HashSet<>(Arrays.asList(TableFeatures.VARIANT_RW_FEATURE));
```
vkorukanti, line 261: one issue (not sure if it is):
If the table already has the `varaintType-preview` and you enable shredding through metadata config. With this `requiredFeatures`, it will have both the `variantType` and `variantType-preview` in the table protocol.

vkorukanti (changes requested):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/tablefeatures/TableFeatures.java
amogh-jahagirdar: I don't think this is an issue because the feature sets are the same (so there's no logical source of truth issue) but I do wonder if we should (in a follow on) introduce an abstraction for features to have some ability to disable existing features. e.g. any time a "graduated" feature is enabled, it goes and removes the "preview" feature if it's set on the table.  This keeps the protocol cleaner

kernel/kernel-api/src/test/scala/io/delta/kernel/TransactionSuite.scala
amogh-jahagirdar: added!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/tablefeatures/TableFeatures.java
allisonport-db: Edit: ignore this comment
I think for type widening we did something specific for graduating features. Can we do the same thing here?

allisonport-db: I think this abstraction makes sense -- if you have a table with variantType-preview, and you somehow (either through implicit updates, or explicitly) enable variantType, can we remove variantType-preview?

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/tablefeatures/TableFeaturesSuite.scala
```diff
@@ -560,13 +563,13 @@ class TableFeaturesSuite extends AnyFunSuite {
...
   Seq("variantType", "variantType-preview", "variantShredding-preview").foreach { feature =>
-    checkWriteUnsupported(
+    checkWriteSupported(
       s"validateKernelCanWriteToTable: protocol 7 with $feature, " +
         s"metadata doesn't contains $feature",
       new Protocol(3, 7, singleton(feature), singleton(feature)),
       testMetadata())
 
-    checkWriteUnsupported(
+    checkWriteSupported(
```
allisonport-db, line 572: I think this test metadata needs to include variant right?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/tablefeatures/TableFeatures.java
vkorukanti: @richardc-db any concern if both `variant-preview` and `variant` are enabled on the table?

richardc-db: i don't think theres a concern for now if a small subset of tables have both (ideally most tables are on `variantType` and not `variantType-preview`). Agree it would be cleaner to somehow remove the `-preview` feature, but AFAIK we don't currently have that capability.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/tablefeatures/TableFeaturesSuite.scala
amogh-jahagirdar: This one already does (line 576), but it looks like the one above doesn't, I'll add that good catch!

amogh-jahagirdar: oh actually the intent of the test above was to validate in case the feature was not contained in the metadata, so i think it was good as is actually

allisonport-db: `includeInvariant` includes invariants in the metadata AFAIK, not _variants_ haha

allisonport-db: but yes the above one should not have variant in the metadata, and this one should

allisonport-db: should be `includeVariantTypeCol` it seems

amogh-jahagirdar: Ah hah, okay yeah when reading it was I confused myself between invariant and variant....cool I think you're right

kernel/kernel-api/src/main/java/io/delta/kernel/internal/tablefeatures/TableFeatures.java
```diff
@@ -262,8 +257,8 @@ public boolean metadataRequiresFeatureToBeEnabled(Protocol protocol, Metadata me
     }
 
     @Override
-    public boolean hasKernelWriteSupport(Metadata metadata) {
-      return false; // TODO: yet to be implemented in Kernel
+    public Set<TableFeature> requiredFeatures() {
```
allisonport-db, line 260: Should we test that we enable dependent features here? Not sure where we test this for other table features

amogh-jahagirdar: Sure, I updated ProtocolSuite which looks like it was doing some tests around dependent table features for a given feature.

felipepessoto:
@amogh-jahagirdar, @allisonport-db. The release notes claim variant read and write is supported by Kernel, but in the description here I noticed it is limited to metadata write.

What do we mean by "metadata write"?

PR #5223, opened 2025-09-17
[Kernel] Update SchemaUtils validation to  allow for variant

amogh-jahagirdar:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [x] Kernel
- [ ] Other (fill in here)

## Description

This change allows variant in SchemaUtils validateSchema/validateUpdatedSchema

## How was this patch tested?

Added unit tests to SchemaUtilsSuite

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
@@ -28,6 +28,8 @@ import io.delta.kernel.internal.DeltaErrors;
 import io.delta.kernel.internal.TableConfig;
 import io.delta.kernel.internal.actions.Metadata;
 import io.delta.kernel.internal.skipping.StatsSchemaHelper;
+import io.delta.kernel.internal.tablefeatures.TableFeature;
+import io.delta.kernel.internal.tablefeatures.TableFeatures;
 import io.delta.kernel.internal.types.TypeWideningChecker;
 import io.delta.kernel.types.*;
 import java.util.*;
@@ -785,11 +787,14 @@ public class SchemaUtils {
         || dataType instanceof BinaryType
         || dataType instanceof DateType
         || dataType instanceof TimestampType
-        || dataType instanceof TimestampNTZType) {
+        || dataType instanceof TimestampNTZType
+        || dataType instanceof VariantType) {
       // supported types
       return;
     } else if (dataType instanceof StructType) {
-      ((StructType) dataType).fields().forEach(field -> validateSupportedType(field.getDataType()));
+      ((StructType) dataType)
+          .fields()
+          .forEach(field -> validateSupportedType(field.getDataType()));
     } else if (dataType instanceof ArrayType) {
       validateSupportedType(((ArrayType) dataType).getElementType());
     } else if (dataType instanceof MapType) {
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
@@ -28,8 +28,6 @@ import io.delta.kernel.internal.DeltaErrors;
 import io.delta.kernel.internal.TableConfig;
 import io.delta.kernel.internal.actions.Metadata;
 import io.delta.kernel.internal.skipping.StatsSchemaHelper;
-import io.delta.kernel.internal.tablefeatures.TableFeature;
-import io.delta.kernel.internal.tablefeatures.TableFeatures;
 import io.delta.kernel.internal.types.TypeWideningChecker;
 import io.delta.kernel.types.*;
 import java.util.*;
@@ -792,9 +790,7 @@ public class SchemaUtils {
       // supported types
       return;
     } else if (dataType instanceof StructType) {
-      ((StructType) dataType)
-          .fields()
-          .forEach(field -> validateSupportedType(field.getDataType()));
+      ((StructType) dataType).fields().forEach(field -> validateSupportedType(field.getDataType()));
     } else if (dataType instanceof ArrayType) {
       validateSupportedType(((ArrayType) dataType).getElementType());
     } else if (dataType instanceof MapType) {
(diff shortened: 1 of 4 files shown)


kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -794,16 +814,19 @@ protected static void validateSupportedType(DataType dataType) {
         || dataType instanceof BinaryType
         || dataType instanceof DateType
         || dataType instanceof TimestampType
-        || dataType instanceof TimestampNTZType) {
+        || dataType instanceof TimestampNTZType
+        || dataType instanceof VariantType && features.contains(TableFeatures.VARIANT_RW_FEATURE)) {
```
vkorukanti: On qn: why do we need to check the table feature is enabled or not? Before coming here, we do enabled the protocol automatically (if not already exists) to include the variant type. Similar to the `TimestampNTZType`. What is special about the variant type?

vkorukanti (changes requested):

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
amogh-jahagirdar: I think you're definitely right that in when the auto upgrade is done in TransactionMetadataFactory and we go through auto enablement of features based on metadata, the variant feature will identify that there's a variant in the schema and add the feature. So in that sense there's nothing special w variant type, it's the same idea as timestamp ntz.

I made this change since I think `SchemaUtils.validateSchema` should be able to be invoked standalone at least in the current state of Kernel and that the utility API should take in a set of supported features (since as a user of this API I'd want to check that a given schema or changed schema is OK considering some set of features) . 

Considering that `TransactionMetadataFactory` already does a schema validation post-protocol upgrade, I thought it just made sense to pass in the new feature. But I think you're right that this change to SchemaUtils is not _strictly_ required, I just think it makes for a better validation API. cc @allisonport-db who was also thinking about validation.

allisonport-db: Not sure about variant, but for other future type related features, is it possible they won't be auto-enabled?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
```diff
@@ -52,9 +54,26 @@ private SchemaUtils() {}
...
+   * @param schema the schema to validate
+   * @param features the table features to validate the schema against
+   * @throws IllegalArgumentException if the schema is invalid
+   */
+  public static void validateSchema(StructType schema, Set<TableFeature> features) {
     checkArgument(schema.length() > 0, "Schema should contain at least one column");
-    validateColumnNames(schema, isColumnMappingEnabled);
-    validateSupportedType(schema);
+    checkArgument(features != null, "Table features should not be null");
+    validateColumnNames(schema, features.contains(TableFeatures.COLUMN_MAPPING_RW_FEATURE));
```
allisonport-db: The COLUMN_MAPPING_RW_FEATURE doesn't necessarily imply column mapping is enabled in theory, this value should really come from the metadata configuration.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/util/SchemaUtils.java
allisonport-db: Oop sorry didn't see the full thread for the above comment. But I think this validation is fine by me, and technically maybe a bit safer. I think we should do the same thing for TimestampNTZ though and check for the table feature.

allisonport-db: Or, if we think they'll always be auto-enabled, we can skip this and revert the earlier PR :) honestly not that picky here

vkorukanti: Hi @allisonport-db, this is an internal utility method that checks the whether Kernel supports the type or not. IMO we shouldn't mix the protocol details here. Protocol details are already taken care by the metadata update PR. I think lets revert this check (don't see adding any value here).

amogh-jahagirdar: I reverted this change so SchemaUtils just checks the type

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWritesSuite.scala
```diff
@@ -92,21 +92,6 @@ abstract class AbstractDeltaTableWritesSuite extends AnyFunSuite with AbstractWr
     }
   }
 
-  test("create table - provide unsupported column types - expect failure") {
```
allisonport-db, line 95: Minor: Could we maybe create a custom data type `FooType extends DataType` and keep this test?

allisonport-db (commented):
Just 1 request then LGTM

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/DeltaTableWritesSuite.scala
allisonport-db: To be specific, just create this custom data type just within this test here

amogh-jahagirdar: Sure that's reasonable just to make sure we have the right error message

amogh-jahagirdar: Discussed offline, will take this in a follow on right after this

allisonport-db: yeah, also with this change, I think we never hit that error message at all since all types are now supported!

PR #5229, opened 2025-09-18
[Spark] Support RANDOMIZE_FILE_PREFIXES

sigmod:

- [x] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

## Description

Add implementation for DeltaConfigs.RANDOM_PREFIX_LENGTH.

## How was this patch tested?

Added a unit test in DeltaTableCreationTest.


scottsand-db (changes requested):
Thanks for this, looks great! Can we add tests? And can you please update the PR description?

sigmod:
> Thanks for this, looks great! Can we add tests? And can you please update the PR description?

Done.

sigmod:
@scottsand-db @brkyvz

spark/src/test/scala/org/apache/spark/sql/delta/DeltaAlterTableTests.scala
```diff
@@ -222,6 +222,54 @@ trait DeltaAlterTableTests extends DeltaAlterTableTestBase {
     }
   }
 
+  ddlTest("SET TBLPROPERTIES - delta.randomizeFilePrefixes") {
```
brkyvz, line 225: Can you also unset the property and write more data to check?

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTableCreationTests.scala
```diff
@@ -1026,6 +1026,46 @@ trait DeltaTableCreationTests
...
 
+ test("create table with table properties - delta.randomizeFilePrefixes") {
+    withTable("delta_test") {
+      sql(s"""
+             |CREATE TABLE delta_test(a LONG, b String)
+             |USING delta
+             |TBLPROPERTIES(
+             |  'delta.randomizeFilePrefixes' = 'true',
+             |  'delta.randomPrefixLength' = '5'
+             |)
```
brkyvz, line 1037: Can you please test with a partitioned table as well?

spark/src/test/scala/org/apache/spark/sql/delta/DeltaAlterTableTests.scala
brkyvz: both on a partitioned and unpartitioned table

brkyvz (changes requested):
Can you please add a couple more tests? Otherwise LGTM!

spark/src/test/scala/org/apache/spark/sql/delta/DeltaAlterTableTests.scala
sigmod: done.

spark/src/test/scala/org/apache/spark/sql/delta/DeltaTableCreationTests.scala
sigmod: done.

brkyvz (approved):
LGTM!

scottsand-db (approved):
LGTM! Simple prod code and great tests!

felipepessoto:
@sigmod, @scottsand-db is this change working as expected? There isn't much info about it, but it is unexpected to me that once you enable column mapping the data files are now in a subfolder with random names.

cc @timothyw553

PR #5230, opened 2025-09-18
[PROTOTYPE] [EARLY FEEDBACK REQUESTED] Kernel accept opaque committer properties; UC committer extensibility

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta/pull/5230/files) to review incremental changes.
- [**stack/kernel_uc_committer_catalog_properties**](https://github.com/delta-io/delta/pull/5230) [[Files changed](https://github.com/delta-io/delta/pull/5230/files)]

---------

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [ ] Kernel
- [ ] Other (fill in here)

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/commit/CommitMetadata.java
@@ -59,28 +59,28 @@ public class CommitMetadata {
   private final long version;
   private final String logPath;
   private final CommitInfo commitInfo;
+  private final Supplier<Map<String, String>> committerProperties;
+
   private final Optional<Tuple2<Protocol, Metadata>> readPandMOpt;
   private final Optional<Protocol> newProtocolOpt;
   private final Optional<Metadata> newMetadataOpt;
-  private final Optional<Supplier<Map<String, String>>> committerPropertiesOpt;
 
   public CommitMetadata(
       long version,
       String logPath,
       CommitInfo commitInfo,
+      Supplier<Map<String, String>> committerProperties,
       Optional<Tuple2<Protocol, Metadata>> readPandMOpt,
       Optional<Protocol> newProtocolOpt,
-      Optional<Metadata> newMetadataOpt,
-      Optional<Supplier<Map<String, String>>> committerPropertiesOpt) {
+      Optional<Metadata> newMetadataOpt) {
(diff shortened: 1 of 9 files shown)


unity/src/main/java/io/delta/unity/UCCatalogManagedClient.java
```diff
@@ -142,10 +143,24 @@ public CreateTableTransactionBuilder buildCreateTableTransaction(
...
+  // Protected Methods for Extensibility //
+  /////////////////////////////////////////
+
+  /**
+   * Creates a UC committer instance for the specified table.
+   *
+   * <p>This method allows subclasses to provide custom committer implementations for specialized
+   * use cases.
+   */
+  protected Committer createUCCommitter(UCClient ucClient, String ucTableId, String tablePath) {
```
scottsand-db, line 160: Code smell here. I wonder if we should abandon the UCCatalogManagedClient::loadSnapshot API -- and instead just provide static utilities (like getting the commits, parsing the commits, etc.)

kernel/kernel-api/src/main/java/io/delta/kernel/Transaction.java
```diff
@@ -108,6 +109,12 @@ public interface Transaction {
   TransactionCommitResult commit(Engine engine, CloseableIterable<Row> dataActions)
       throws ConcurrentWriteException;
 
+  /**
+   * Adds custom properties that will be passed through to the committer. These properties allow
+   * connectors to inject catalog-specific metadata without Kernel inspection.
+   */
+  void withCommitterProperties(Supplier<Map<String, String>> committerProperties);
```
scottsand-db, line 116: Added here instead of on the builder so Connectors can inject any properies they'd like at any point.

e.g. maybe these properties depend on the data files written

kernel/kernel-api/src/main/java/io/delta/kernel/Transaction.java
```diff
@@ -108,6 +109,12 @@ public interface Transaction {
   TransactionCommitResult commit(Engine engine, CloseableIterable<Row> dataActions)
       throws ConcurrentWriteException;
 
+  /**
```
scottsand-db, line 112: I don't love how this is on the Txn and not the TxnBuilder ...

PR #5235, opened 2025-09-20
[Kernel][CCV2] Support DeltaHistoryManager. getActiveCommitAtTimestamp with catalog commits

allisonport-db:

- [ ] Spark
- [ ] Standalone
- [ ] Flink
- [X] Kernel
- [ ] Other (fill in here)

## Description

Updates getActiveCommitAtTimestamp to accept an optional list of parsedLogDelta. For now, these can only be ratified staged commits. When provided, these are used instead of published commits, for reading ICT timestamps and resolving timestamp to version.

## How was this patch tested?

Adds unit tests.

TODO: add E2E tests

No.

### DIFF
--- a/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java
+++ b/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java
@@ -18,6 +18,7 @@ package io.delta.kernel.internal;
 import static io.delta.kernel.internal.DeltaErrors.wrapEngineExceptionThrowsIO;
 import static io.delta.kernel.internal.TableConfig.*;
 import static io.delta.kernel.internal.fs.Path.getName;
+import static io.delta.kernel.internal.util.Preconditions.checkArgument;
 
 import io.delta.kernel.engine.Engine;
 import io.delta.kernel.exceptions.KernelException;
@@ -25,6 +26,7 @@ import io.delta.kernel.exceptions.TableNotFoundException;
 import io.delta.kernel.internal.actions.CommitInfo;
 import io.delta.kernel.internal.actions.Metadata;
 import io.delta.kernel.internal.checkpoints.CheckpointInstance;
+import io.delta.kernel.internal.files.ParsedLogData;
 import io.delta.kernel.internal.fs.Path;
 import io.delta.kernel.internal.util.FileNames;
 import io.delta.kernel.internal.util.InCommitTimestampUtils;
@@ -35,6 +37,7 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.util.*;
+import java.util.function.Function;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -62,6 +65,7 @@ public final class DeltaHistoryManager {
    * @return latest commit that happened at or before {@code timestamp}.
    * @throws KernelException if the timestamp is more than the timestamp of any committed version
    */
+  // TODO: update to accept parsedLogDatas
   public static long getVersionAtOrAfterTimestamp(
       Engine engine, Path logPath, long millisSinceEpochUTC, SnapshotImpl latestSnapshot) {
     DeltaHistoryManager.Commit commit =
@@ -76,7 +80,8 @@ public final class DeltaHistoryManager {
             false, /* canReturnLastCommit */
             // e.g. we give time T-1 and first commit has time T, then we DO want that earliest
             // commit
-            true /* canReturnEarliestCommit */);
+            true /* canReturnEarliestCommit */,
+            Optional.empty());
 
     if (commit.getTimestamp() >= millisSinceEpochUTC) {
       return commit.getVersion();
@@ -106,6 +111,7 @@ public final class DeltaHistoryManager {
    * @return latest commit that happened before or at {@code timestamp}.
    * @throws KernelException if the timestamp is less than the timestamp of any committed version
    */
+  // TODO: update to accept parsedLogDatas
   public static long getVersionBeforeOrAtTimestamp(
       Engine engine, Path logPath, long millisSinceEpochUTC, SnapshotImpl latestSnapshot) {
     return DeltaHistoryManager.getActiveCommitAtTimestamp(
@@ -118,7 +124,8 @@ public final class DeltaHistoryManager {
             true, /* canReturnLastCommit */
             // e.g. we give time T-1 and first commit has time T, then do NOT want that earliest
             // commit
-            false /* canReturnEarliestCommit */)
+            false /* canReturnEarliestCommit */,
+            Optional.empty())
         .getVersion();
   }
 
@@ -150,16 +157,86 @@ public final class DeltaHistoryManager {
       long timestamp,
       boolean mustBeRecreatable,
       boolean canReturnLastCommit,
-      boolean canReturnEarliestCommit)
+      boolean canReturnEarliestCommit) {
+    return getActiveCommitAtTimestamp(
+        engine,
+        latestSnapshot,
+        logPath,
+        timestamp,
+        mustBeRecreatable,
+        canReturnLastCommit,
+        canReturnEarliestCommit,
+        Optional.empty());
+  }
+
+  /**
+   * Returns the latest commit that happened at or before {@code timestamp}.
+   *
+   * <p>If the timestamp is outside the range of [earliestCommit, latestCommit] then use parameters
+   * {@code canReturnLastCommit} and {@code canReturnEarliestCommit} to control whether an exception
+   * is thrown or the corresponding earliest/latest commit is returned.
+   *
+   * @param engine instance of {@link Engine} to use
+   * @param logPath the _delta_log path of the table
+   * @param timestamp the timestamp find the version for in milliseconds since the unix epoch
+   * @param mustBeRecreatable whether the state at the returned commit should be recreatable
+   * @param canReturnLastCommit whether we can return the latest version of the table if the
+   *     provided timestamp is after the latest commit
+   * @param canReturnEarliestCommit whether we can return the earliest version of the table if the
+   *     provided timestamp is before the earliest commit
+   * @param parsedLogDatas parsed log Deltas to use
+   * @throws KernelException if the provided timestamp is before the earliest commit and
+   *     canReturnEarliestCommit is false
+   * @throws KernelException if the provided timestamp is after the latest commit and
+   *     canReturnLastCommit is false
+   * @throws TableNotFoundException when there is no Delta table at the given path
+   */
+  public static Commit getActiveCommitAtTimestamp(
+      Engine engine,
+      SnapshotImpl latestSnapshot,
+      Path logPath,
+      long timestamp,
+      boolean mustBeRecreatable,
+      boolean canReturnLastCommit,
+      boolean canReturnEarliestCommit,
+      // TODO: update this to be ParsedLogDelta after #4805 is merged
+      Optional<List<ParsedLogData>> parsedLogDatas)
       throws TableNotFoundException {
+
+    // For now, we only accept ratified staged commits
+    parsedLogDatas.ifPresent(
+        logDatas ->
+            checkArgument(
+                logDatas.stream()
+                    .allMatch(
+                        logData ->
+                            logData.isMaterialized()
+                                && FileNames.isStagedDeltaFile(logData.getFileStatus().getPath())),
+                "Currently getActiveCommitAtTimestamp only accepts ratified staged commits"));
+
+    // Create a mapper for delta version -> file status that takes into account ratified commits
+    Function<Long, FileStatus> versionToFileStatusFunction =
+        getVersionToFileStatusFunctionWithParsedLogDelta(
+            parsedLogDatas.orElse(Collections.emptyList()), logPath);
+    Optional<Long> earliestRatifiedCommitVersion =
+        parsedLogDatas.flatMap(
+            logDatas -> logDatas.stream().map(ParsedLogData::getVersion).min(Long::compare));
 
     long earliestVersion =
         (mustBeRecreatable)
-            ? getEarliestRecreatableCommit(engine, logPath)
-            : getEarliestDeltaFile(engine, logPath);
+            ? getEarliestRecreatableCommit(engine, logPath, earliestRatifiedCommitVersion)
+            : getEarliestDeltaFile(engine, logPath, earliestRatifiedCommitVersion);
 
     Commit placeholderEarliestCommit = new Commit(earliestVersion, -1L /* timestamp */);
     Commit ictEnablementCommit = getICTEnablementCommit(latestSnapshot, placeholderEarliestCommit);
+    // Validate our assumptions: ICT must be enabled for all catalog-provided commits
+    earliestRatifiedCommitVersion.ifPresent(
+        v ->
+            checkArgument(
+                ictEnablementCommit.version <= v,
+                "catalogManaged tables must have ICT enabled but given catalog-provided commit "
+                    + "with version < ictEnablementVersion"));
+
     Commit searchResult;
     if (ictEnablementCommit.getTimestamp() <= timestamp) {
       // The target commit is in the ICT range.
@@ -185,7 +262,8 @@ public final class DeltaHistoryManager {
                   searchWindowLowerBound,
                   latestSnapshot.getVersion(),
                   engine,
-                  latestSnapshot.getLogPath());
+                  latestSnapshot.getLogPath(),
+                  versionToFileStatusFunction);
         } catch (IOException e) {
           throw new RuntimeException(
               "There was an error while reading a historical commit while performing a timestamp-"
@@ -204,10 +282,15 @@ public final class DeltaHistoryManager {
         // Else, when `canReturnEarliestCommit` is `true`, the earliest commit
         // is the desired result.
         long ict =
-            CommitInfo.unsafeGetRequiredIctFromPublishedDeltaFile(
-                engine, logPath, placeholderEarliestCommit.getVersion());
+            CommitInfo.getRequiredIctFromDeltaFile(
+                engine,
+                logPath.getParent(),
+                versionToFileStatusFunction.apply(placeholderEarliestCommit.getVersion()),
+                placeholderEarliestCommit.getVersion());
         searchResult = new Commit(placeholderEarliestCommit.getVersion(), ict);
       } else {
+        // We know the table was not catalogManaged here since ICT was not enabled ==> we don't
+        // need to worry about catalogCommits
         // start non-ICT linear search over [earliestVersion, )
         List<Commit> commits = getCommits(engine, logPath, earliestVersion);
         searchResult =
@@ -250,7 +333,8 @@ public final class DeltaHistoryManager {
       long startCommitVersionInclusive,
       long endCommitVersionInclusive,
       Engine engine,
-      Path logPath)
+      Path logPath,
+      Function<Long, FileStatus> versionToFileStatusFunction)
       throws IOException {
     // Now we have a range of commits to search through. We can use binary search to find the
     // commit that is closest to the search timestamp.
@@ -260,12 +344,19 @@ public final class DeltaHistoryManager {
             startCommitVersionInclusive,
             endCommitVersionInclusive,
             version ->
-                CommitInfo.unsafeGetRequiredIctFromPublishedDeltaFile(engine, logPath, version));
+                CommitInfo.getRequiredIctFromDeltaFile(
+                    engine,
+                    logPath.getParent(),
+                    versionToFileStatusFunction.apply(version),
+                    version));
     // This indicates that the search timestamp is less than the earliest commit.
     if (!greatestLowerBoundOpt.isPresent()) {
       long startIct =
-          CommitInfo.unsafeGetRequiredIctFromPublishedDeltaFile(
-              engine, logPath, startCommitVersionInclusive);
+          CommitInfo.getRequiredIctFromDeltaFile(
+              engine,
+              logPath.getParent(),
+              versionToFileStatusFunction.apply(startCommitVersionInclusive),
+              startCommitVersionInclusive);
       return new Commit(startCommitVersionInclusive, startIct);
     }
     Tuple2<Long, Long> greatestLowerBound = greatestLowerBoundOpt.get();
@@ -317,8 +408,17 @@ public final class DeltaHistoryManager {
    * <p>We search for the earliest checkpoint we have, or whether we have the 0th delta file. This
    * method assumes that the commits are contiguous.
    */
-  public static long getEarliestRecreatableCommit(Engine engine, Path logPath)
+  public static long getEarliestRecreatableCommit(
+      Engine engine, Path logPath, Optional<Long> earliestRatifiedCommitVersion)
       throws TableNotFoundException {
+    // For a catalogManaged table, the only time no published commits exist is when v0 has not yet
+    // been published. Otherwise, since checkpoints must have a published delta file, and log clean
+    // up
+    // must always preserve a checkpoint, there must be published commits present on the file-system
+    if (earliestRatifiedCommitVersion.isPresent() && earliestRatifiedCommitVersion.get() == 0) {
+      return 0;
+    }
+    // Thus, if there is no v0 ratified commit, then there must be published commit
     try (CloseableIterator<FileStatus> files =
         listFrom(engine, logPath, 0)
             .filter(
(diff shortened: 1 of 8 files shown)


kernel/kernel-api/src/test/scala/io/delta/kernel/internal/DeltaHistoryManagerSuite.scala
```diff
@@ -23,25 +23,51 @@ import scala.collection.JavaConverters._
...
 import io.delta.kernel.test.MockSnapshotUtils.getMockSnapshot
 import io.delta.kernel.types.StringType
+import io.delta.kernel.types.StructType
 import io.delta.kernel.utils.FileStatus
 
 import org.scalatest.funsuite.AnyFunSuite
 
 class DeltaHistoryManagerSuite extends AnyFunSuite with MockFileSystemClientUtils {
 
+  // Helper function for non-catalog-managed tables (no staged commits)
```
allisonport-db, line 50: Since there are no default params in Java, but I didn't want to keep a second overloaded method def just for test-only usage, I had to add this test helper method. While I was at it, also refactored the boolean parameters to have defaults and only be defined when different. Sorry a lot of code change from this :(

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/DeltaHistoryManagerSuite.scala
```diff
@@ -1145,4 +1099,500 @@ class DeltaHistoryManagerSuite extends AnyFunSuite with MockFileSystemClientUtil
...
+    // Query the exact timestamp of V3
+    checkGetActiveCommitAtTimestampWithParsedLogData(
+      fileList,
+      parsedLogData,
+      versionToICT,
+      480L,
+      3)
+  }
+
+  test("getActiveCommitAtTimestamp with staged commits: v0 published and staged") {
```
allisonport-db: add "+ prefer staged"

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java
```diff
@@ -76,7 +81,8 @@ public static long getVersionAtOrAfterTimestamp(
             false, /* canReturnLastCommit */
             // e.g. we give time T-1 and first commit has time T, then we DO want that earliest
             // commit
-            true /* canReturnEarliestCommit */);
+            true /* canReturnEarliestCommit */,
+            Optional.empty());
```
huan233usc: nit: `/*parsedLogDatas*/`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java
```diff
@@ -150,16 +159,44 @@ public static Commit getActiveCommitAtTimestamp(
       long timestamp,
       boolean mustBeRecreatable,
       boolean canReturnLastCommit,
-      boolean canReturnEarliestCommit)
+      boolean canReturnEarliestCommit,
+      Optional<List<ParsedDeltaData>> parsedLogDatas)
```
scottsand-db: Does `None` represent anything different from `Some(empty_list)`? If not, can we just pass in an empty list?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java
```diff
@@ -150,16 +159,44 @@ public static Commit getActiveCommitAtTimestamp(
...
       throws TableNotFoundException {
 
+    // For now, we only accept ratified staged commits
+    parsedLogDatas.ifPresent(
```
scottsand-db: Ugh. Yeah. We should be able to represent this better (in parsedDeltaData) (this is on me). FYI the complexity is: can we have inline data of both a published delta file and a ratified commit?

Perhaps yes -- in which case, we can just add a simple boolean to the `ParsedDeltaData.forInlineData(..., isCatalogRatifiedCommit)`.

In the meantime -- want to just add `ParsedDeltaData::isRatifiedCommit`? For now, return true if (a) file and correct file name, or (b) is inline

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java
```diff
@@ -530,4 +600,22 @@ public int hashCode() {
       return Objects.hash(version, timestamp);
     }
   }
+
+  private static Function<Long, FileStatus> getVersionToFileStatusFunctionWithParsedLogDelta(
```
scottsand-db: nit: `WithParsedDeltaDelta(`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java
```diff
@@ -260,12 +304,19 @@ private static Commit getActiveCommitAtTimeFromICTRange(
             startCommitVersionInclusive,
             endCommitVersionInclusive,
             version ->
-                CommitInfo.unsafeGetRequiredIctFromPublishedDeltaFile(engine, logPath, version));
+                CommitInfo.getRequiredIctFromDeltaFile(
+                    engine,
+                    logPath.getParent(),
+                    versionToFileStatusFunction.apply(version),
```
scottsand-db, line 303: Isn't this still "unsafe"? Our version -> fileStatus made an assumption that if the version is _not_ in the ratified commits then it is published. But we never LISTed and saw that it was published?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java
```diff
@@ -530,4 +600,22 @@ public int hashCode() {
...
+    Map<Long, FileStatus> versionToFileStatusMap = new HashMap<>();
+    for (ParsedDeltaData parsedLogData : parsedLogDatas) {
+      versionToFileStatusMap.put(parsedLogData.getVersion(), parsedLogData.getFileStatus());
+    }
+    return version -> {
+      if (versionToFileStatusMap.containsKey(version)) {
+        return versionToFileStatusMap.get(version);
+      } else {
+        return FileStatus.of(
+            FileNames.deltaFile(logPath, version), /* path */
```
scottsand-db, line 622: When I see this -- I just ask: why aren't we just unioning the ratified commits with a result from the LIST of the file system to get an authoritative file status?

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java
```diff
@@ -317,8 +368,17 @@ private static Commit getICTEnablementCommit(SnapshotImpl snapshot, Commit earli
    * <p>We search for the earliest checkpoint we have, or whether we have the 0th delta file. This
    * method assumes that the commits are contiguous.
    */
-  public static long getEarliestRecreatableCommit(Engine engine, Path logPath)
+  public static long getEarliestRecreatableCommit(
+      Engine engine, Path logPath, Optional<Long> earliestRatifiedCommitVersion)
       throws TableNotFoundException {
+    // For a catalogManaged table, the only time no published commits exist is when v0 has not yet
+    // been published. Otherwise, since checkpoints must have a published delta file, and log clean
+    // up
```
scottsand-db: nit: fix comment structure

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/DeltaHistoryManagerSuite.scala
scottsand-db: Can this refactor be done in a prefactor PR?

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/DeltaHistoryManagerSuite.scala
```diff
@@ -1145,4 +1099,501 @@ class DeltaHistoryManagerSuite extends AnyFunSuite with MockFileSystemClientUtil
...
+      s"Expected version $expectedVersion but got ${activeCommit.getVersion} " +
+        s"for timestamp=$timestampToQuery")
+  }
+
+  test("getActiveCommitAtTimestamp with staged commits: empty log, 1 staged commit") {
+    // Published commits: _
+    // Ratified commits: V0
+    val stagedCommitFiles = Seq(stagedCommitFile(0L))
+    val parsedLogData = stagedCommitFiles.map(ParsedDeltaData.forFileStatus(_))
+    val versionToICT = Map(0L -> 180L)
```
scottsand-db, line 1156: Can you help me understand where 180 came from?

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/DeltaHistoryManagerSuite.scala
```diff
@@ -1145,4 +1099,501 @@ class DeltaHistoryManagerSuite extends AnyFunSuite with MockFileSystemClientUtil
...
+
+    // Query the exact timestamp
+    checkGetActiveCommitAtTimestampWithParsedLogData(
+      stagedCommitFiles,
+      parsedLogData,
+      versionToICT,
+      180L,
```
scottsand-db: can you do named params to make this easier to read?

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/DeltaHistoryManagerSuite.scala
```diff
@@ -1145,4 +1099,501 @@ class DeltaHistoryManagerSuite extends AnyFunSuite with MockFileSystemClientUtil
...
+    intercept[KernelException] {
+      checkGetActiveCommitAtTimestampWithParsedLogData(
+        stagedCommitFiles,
+        parsedLogData,
+        versionToICT,
+        170L,
+        0)
+    }
+
+    // With canReturnEarliestCommit passes
```
scottsand-db: Nit: Querying before with canReturnEarliestCommit passes

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/DeltaHistoryManagerSuite.scala
```diff
@@ -1145,4 +1099,501 @@ class DeltaHistoryManagerSuite extends AnyFunSuite with MockFileSystemClientUtil
...
+    intercept[KernelException] {
+      checkGetActiveCommitAtTimestampWithParsedLogData(
+        stagedCommitFiles,
+        parsedLogData,
+        versionToICT,
+        190L,
+        0)
+    }
+
+    // With canReturnEarliestCommit passes
```
scottsand-db: ditto

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/DeltaHistoryManagerSuite.scala
```diff
@@ -1145,4 +1099,501 @@ class DeltaHistoryManagerSuite extends AnyFunSuite with MockFileSystemClientUtil
...
+    // Query the exact timestamp of V3
+    checkGetActiveCommitAtTimestampWithParsedLogData(
+      fileList,
+      parsedLogData,
+      versionToICT,
+      480L,
+      3)
+  }
+
+  test("getActiveCommitAtTimestamp with staged commits: v0 published and staged + prefer staged") {
```
scottsand-db: to me the `+` (an "and", really) isn't part of the "input" or "start state" of the test -- but rather it's the assertion we want. so `v0 published and staged => prefer staged` seems clearer

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/DeltaHistoryManagerSuite.scala
```diff
@@ -1145,4 +1099,501 @@ class DeltaHistoryManagerSuite extends AnyFunSuite with MockFileSystemClientUtil
...
+  test("getActiveCommitAtTimestamp with staged commits: overlap + prefer staged") {
+    // Published commits: V10, V11
+    // Ratified commits: V11, V12
+    val stagedCommitFiles = Seq(stagedCommitFile(11), stagedCommitFile(12))
+    val parsedLogData = stagedCommitFiles.map(ParsedDeltaData.forFileStatus(_))
+    val fileList = Seq(
+      classicCheckpointFileStatus(10),
+      deltaFileStatus(10),
+      deltaFileStatus(11)) ++ stagedCommitFiles
+    val versionToICT = Map(10L -> 1000L, 11L -> 1100L, 12L -> 1200L)
```
scottsand-db, line 1326: so, is this vesionToICT only for published deltas? can we make that name clearer?

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/DeltaHistoryManagerSuite.scala
```diff
@@ -1145,4 +1099,501 @@ class DeltaHistoryManagerSuite extends AnyFunSuite with MockFileSystemClientUtil
...
+    // We have v10=1000, v11=1110 (if we use the correct file), v12=1210
```
scottsand-db: by the "correct file" you mean the ratified commit? can we just say that?

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/DeltaHistoryManagerSuite.scala
```diff
@@ -1145,4 +1099,501 @@ class DeltaHistoryManagerSuite extends AnyFunSuite with MockFileSystemClientUtil
...
+        fileList,
+        parsedLogData,
+        versionToICT,
+        200L,
+        0,
+        add10ToICTForStagedFiles = true)
+    }
+  }
+
+  test("getActiveCommitAtTimestamp with staged commits: overlap + prefer staged") {
```
scottsand-db: instead of "staged" can we say "ratified" --> multiple files can be staged for the same version, but only one of them will ever be ratified

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/catalogManaged/CatalogManagedE2EReadSuite.scala
```diff
@@ -86,4 +91,88 @@ class CatalogManagedE2EReadSuite extends AnyFunSuite with TestUtilsWithTableMana
...
+    // ===== GIVEN =====
+    val tablePath = getTestResourceFilePath("catalog-owned-preview")
+    val logPath = new Path(tablePath, "_delta_log")
+
+    // Note: We need to *resolve* each test resource file path, because the table root file path
+    //       will itself be resolved when we create the Snapshot. If we resolved some paths but
+    //       not others, we would get an error like `File <commit-file> doesn't belong in the
+    //       transaction log at <log-path>`.
+
+    val parsedDeltaData = Seq(
```
scottsand-db: I see this is dupicated from above? Let's move to a helper method?

We can keep it table-agnostic with just `def resolvedFileStatusesToParsedDeltaData(...)`

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java
allisonport-db: I think we briefly talked about this in a sync at some point. In the current implementation (before & after my changes) we never actually _fully list_ the _delta_log. We only list to find the earliest commit and exit early, so we never actually have that full listing to reference or combine with. We rely on the assumption that the commits after the earliest commit exist and are contiguous.

This change keeps it that way. ==> if you think it's worth the cost to do a full listing, then that's a larger change we could consider.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java
allisonport-db: No, I'm happy to do that instead. Just affects the semantics on the callers really.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java
allisonport-db: So I thought the naming of "unsafe" was in reference to the fact that we are not accounting for catalog commits (i.e. maybe the commit is ratified but not published, we shouldn't blindly only read the published file names).

I think this is somewhat related to the other convo. Currently, we confirm the earliest commit that exists on the FS, and then assume that commits after that up to the latest version exist (either on the FS if not provided by the catalog, or as a ratified commit if provided). This is what we currently do, and while it is "unsafe" in some ways, I don't think it's unsafe for the reason we changed the name of that method.

Also, it's kind of a blurry line. In theory, even if we list, commits could be cleaned up in the meantime and also not exist for that reason.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/DeltaHistoryManagerSuite.scala
allisonport-db: Done https://github.com/delta-io/delta/pull/5237

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/catalogManaged/CatalogManagedE2EReadSuite.scala
allisonport-db: I actually did update this in my next PR changes, I was going to add a test fx so that you can do
```
withCatalogOwnedPreviewTestTable { (tablePath, parsedLogData) =>
   ...
}
```
I'll move the changes up to this PR. Let me know if you don't like this method and prefer helper methods.

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java
```diff
@@ -530,4 +596,22 @@ public int hashCode() {
...
+      List<ParsedDeltaData> parsedDeltaDatas, Path logPath) {
+    Map<Long, FileStatus> versionToFileStatusMap = new HashMap<>();
+    for (ParsedDeltaData parsedDeltaData : parsedDeltaDatas) {
+      versionToFileStatusMap.put(parsedDeltaData.getVersion(), parsedDeltaData.getFileStatus());
+    }
+    return version -> {
+      if (versionToFileStatusMap.containsKey(version)) {
+        return versionToFileStatusMap.get(version);
+      } else {
+        return FileStatus.of(
```
huan233usc, line 621: Do we want to separate this logic and return Optional<FileStatus> with only taking care of parsedDeltaDatas(the method name give me an impression that it tries to find delta file from parsed log data but it does also return a fallback when the version is not found.)


getVersionToFileStatusFunctionWithParsedDeltaData().apply(version).orElse(FileNames.deltaFile(logPath, version), /* path */)

allisonport-db: I feel like that just increases the number of callers that need to be aware of this logic, and it would be duplicated in more than once place. Do you think we could just name this better to make it more clear?

`getVersionToFileStatusFunction`
`getVersionToFileStatusFunctionForPublishedAndRatifiedCommits`
??

allisonport-db: I think the total logic of resolving a version -> some file status should be abstracted away from the callers that actually read the files if possible.

huan233usc: getVersionToFileStatusFunction should be fine iiuc. (Most of my confusion came from `parsedDeltaDatas`)

allisonport-db: Sounds good! It's obvious that it's "withParsedDeltaData" considering that's passed in as an arg anyways xD

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java
allisonport-db: https://github.com/delta-io/delta/pull/5238

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/DeltaHistoryManagerSuite.scala
allisonport-db: Hmm, I agree mostly, but I also think it makes it seem like we're _only testing that we prefer staged_ whereas that's just component of what we are testing.

Updated it anyways let me know if you prefer differently.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/DeltaHistoryManagerSuite.scala
allisonport-db: I get your point. My confusion at least with the terminology for myself, is "ratified commit" makes me think only of a version, not necessarily _the staged file_.

As in
- `0010.json` that's been backfilled, is a ratified commit
- `_staged_commits/0010.<uuid>.json` provided by the catalog, is also a ratified commit

I'm happy to update all these instances. But please let me know if "ratified commit" = "ratified staged commit" (I was thinking this is not necessarily true).

allisonport-db: Updated terminology throughout the suite to follow this

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/DeltaHistoryManagerSuite.scala
allisonport-db: So we use this map for both published and staged files. We just have an optional "hack" `add10ToICTForStagedFiles` which means for a staged commit file at version v, the ICT will be `versionToICT(v) + 10`.

This is a lot less invasive than having separate maps for published and staged commit files.

allisonport-db: The line below this outlines what this looks like including the +10. Let me know how I can make it more clear.

kernel/kernel-api/src/test/scala/io/delta/kernel/internal/DeltaHistoryManagerSuite.scala
allisonport-db: Just randomly chose a non-zero number (and that isn't close to the default file-system mod times we populate when creating the FileStatus).

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java
scottsand-db: Ack -- Thanks for that explanation. This is some context that is worth adding to this method as a comment.

```
Historically, DeltaHistoryManager has never done a full LIST. Instead, it has done XXX.
With catalgManaged ratified commits, we must now YYY
```

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java
scottsand-db: Thanks for this context. SG -- this is still making an "assumption" -- but we were already doing that in this class. Now, we just add on to that assumption (really a version -> file status lookup) some ratified commit context.

kernel/kernel-defaults/src/test/scala/io/delta/kernel/defaults/catalogManaged/CatalogManagedE2EReadSuite.scala
```diff
@@ -60,30 +64,108 @@ class CatalogManagedE2EReadSuite extends AnyFunSuite with TestUtilsWithTableMana
...
+          logPath,
+          timestamp,
+          true, /* mustBeRecreatable */
+          canReturnLastCommit,
+          canReturnEarliestCommit,
+          parsedDeltaData.asJava)
+        assert(activeCommit.getVersion == expectedVersion)
+      }
+
+      // ===== WHEN =====
```
scottsand-db: nit: this actually isn't a "when" as nothing is happening. The "when" is for the operations that are performed that we "then" evaluate and check.

here, `checkGetActiveCommitAtTimestamp` is doing the WHEN and the THEN.

IMO for this test setup we can probably abandaone the given when then paradigm!

kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java
allisonport-db: Added a full method docs

